PR #704, opened 2014-11-26
Add getProjectByFile. 

rimmeraj:
This allows bndtools to pass in the path of the project it is adding to the workspace. This will allow mutiple git
repo's per eclipse works space if the cnf redirect is used in each repo
to point to the repo holding the cnf folder. If this accepted there are 4 spots in bndtools that need to be changed. Each location already has a file is is doing getName() on it. It just has to change to getProjectByFile() and drop the getName()

Signed-off-by: Dave Smith <dave.smith@candata.com>

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
@@ -77,7 +77,7 @@ public class Workspace extends Processor {
 		assert projectDir.isDirectory();
 
 		Workspace ws = getWorkspace(projectDir.getParentFile());
-		return ws.getProject(projectDir.getName());
+		return ws.getProjectByFile(projectDir);
 	}
 
 	static synchronized public Processor getDefaults() {
@@ -210,14 +210,29 @@ public class Workspace extends Processor {
 			if (project != null)
 				return project;
 
-			File projectDir = getFile(bsn);
-			project = new Project(this, projectDir);
-			if (!project.isValid())
-				return null;
-
-			models.put(bsn, project);
-			return project;
+			return addProject(getFile(bsn));
 		}
+	}
+	
+	public Project getProjectByFile(File projectDir) throws Exception
+	{
+		synchronized (models) {
+			Project project = models.get(projectDir.getName());
+			if (project != null)
+				return project;
+			
+			return addProject(projectDir);
+		}
+	}
+	
+	Project addProject(File projectDir) throws Exception
+	{
+		Project project = new Project(this, projectDir);
+		if (!project.isValid())
+			return null;
+
+		models.put(projectDir.getName(), project);
+		return project;
 	}
 
 	void removeProject(Project p) throws Exception {
@@ -712,6 +727,7 @@ public class Workspace extends Processor {
 	 * Report details of this workspace
 	 */
 
+	@Override
 	public void report(Map<String,Object> table) throws Exception {
 		super.report(table);
 		table.put("Workspace", toString());
@@ -760,6 +776,7 @@ public class Workspace extends Processor {
 		return null;
 	}
 
+	@Override
 	public void close() {
 		cache.remove(getPropertiesFile().getParentFile().getParentFile());
 	}


pkriens:
I am not sure I like this idea and do not want to let the workspace and the file system interact in this way. So unless you provide some really compelling argument I would like to close this PR

rimmeraj:
This fix allows multiple git repos in one eclipse workspace. With a few more lines of code I could also allow the project name != bundle symbolic name but I will leave that for now.

For us this a show stopper. I have read all of the posts on multiple eclipse workspaces per repo and it for a project of any size  (10+ repos and growing) it just does not work. In a perfect world eclipse would be better at handling git but at this point I think bnd/bndtools should just code around the limitations. 

I would disagree about the workspace and file system.  When you are working in eclipse (bndtools) you are always starting with a directory. Eclipse already knows where all the projects are in it's workspace. I would suggest that is where is disconnect is. A build workspace is different that an eclipse workspace and we should not force users down the path of making the two the same. A build system is infinitely more flexible than eclipse so if a user chooses to use this functionality then they will have to deal with the consequences when they try and write a builder for the projects.

bjhargrave:
You still have failed to update your commit with the required Signed-off-by message. Until that is fixed, we wouldn't accept the PR anyway.

This fix seems to want to decouple the project folder name from the bsn so that a project folder can have a different name than the bsn of the project. This seems a non-trivial change with perhaps not-well understood consequences since the bsn value defaults from the project folder name. I am also not sure how this will interact with the offline build (gradle, ant) which may also make the assumption that folder name=bsn. There are also assumptions that cnf dir and project dirs are peers.

rimmeraj:
I think I updated the commit message , my git is not the best. 

The change has nothing to do with decoupling the folder name from BSN. I agree that would open up a whole other set of problems thus, I left that change out.  This change makes it possible for the Eclipse Workspace=bnd Workspace.  This seems logical since eclipse knows the physical location of all of it's projects, and bnd should only be driven by that when used inside eclipse.

bjhargrave:
> I think I updated the commit message , my git is not the best.

The PR has 2 commits and a merge commit. This is bad. You need to reset back to the original commit (git reset --hard fbf984adba2fdba127225d6ade3a537737ecf844), amend the message (git commit --amend) and force push (git push --force) to "overwrite" the original commit on github.



> The change has nothing to do with decoupling the folder name from BSN.

This seems to be the net effect of the commit since you in fact decouple them. The commit looks up a project by folder name not bsn which is also used as a folder name. So the folder name can differ from the bsn.

rimmeraj:
No. All I have done is moved the BSN=folder name into one spot. Regardless of what you pass in as the file location the code always calls getName() and uses that as the BSN.

bjhargrave:
So `getProjectByFile(File projectDir)` could be implemented more simply as

    getProject(projectDir.getName())

So why all the duplicate code and the need for `addProject`?

I also don't see how this change helps anything then.

rimmeraj:
If you call getProject(projectDir.getName()) then the project folder location is determined by calling getFile(bsn). This figures out the location **relative** to the Workspace root (cnf/..) .  If you pass in the absolute location it skips that step because it knows where the folder is. The change simply splits the logic into two pieces. If you only have a BSN, figure out where we **think** the location is and then try and get/create the project otherwise based on the location passed in  get/create the project.

bjhargrave:
Your commit changes the static getProject(File) which is used by the ant build. But the gradle build uses the non-static getProject(String) in a number of places. That would have to changed too to maintain fidelity, no?

Also, the main issue seems to be that if cnf is a file pointing to the cnf folder, then this causes bnd to use the parent folder of the cnf folder as the workspace. Perhaps it is better to attack the issue there. I would think a cnf file should act as a symlink to a cnf folder and not cause the workspace to be "relocated".

rimmeraj:
If you look at static getProject(File) all I have done is moved the getName() down a level. The only case where this would not work is if the folder passed in from ant was not actually the folder you were trying to build. Thought that was a safe bet.. The non-static getProject(String) has not changed, it's logic has been split. Firstly it resolves the bsn folder on the relative path to cnf (as before) and then creates the project based on that location in addProject(File projectDir). 

If you want to support eclipse workspace=bnd workspace then this is the only way. With eclipse you can not make any assumptions about the underlying layout of the projects.

bjhargrave:
> If you want to support eclipse workspace=bnd workspace then this is the only way. With eclipse you can not make any assumptions about the underlying layout of the projects.

Well for offline builds, which are very important, you cannot use the full "flexibility" of eclipse. When building offline, Eclipse is not in the picture to stitch an eclipse workspace together from a random assortment of folders. There is just the project folders in the workspace folder. So the bnd workspace must be the focus for bnd to enable fidelity between offline builds and eclipse.

rimmeraj:
Why must my eclipse environment match my build environment? Bnd is a tool for creating bundles.  The user of the tool should be able to decide the layout their project not bnd. This change has **zero** impact on existing offline builds and gives people some control if they decide they do not like bnd's hard coded project layouts and need something different. 

Regardless off offline build's , it will allow users who want to have more than 1 git repo per eclipse workspace...

pkriens:
> On 13 jan. 2015, at 23:36, Dave Smith <notifications@github.com> wrote:
> Why must my eclipse environment match my build environment? Bnd is a tool for creating bundles. The user of the tool should be able to decide the layout their project not bnd. This change has zero impact on existing offline builds and gives people some control if they decide they do not like bnd’s hard coded project layouts and need something different.
> 
We had that discussion MANY times, please look it up. Yes, bnd does constrain your flexibility but in return it provides more predictability and much less error cases to handle. A bnd workspace is a module and part of that workspace module is that (just like Git) a workspace is a single directory that hosts a cnf directory and its projects. That is how bnd very consciously was designed since I like simplicity.

We should not sacrifice that simplicity and robustness for what I consider (arguably subjective) a small user benefit. Actually, many users later report that it tends to grow on them.

In many ways this discussion is like the eternal “friend” or fragments discussion in object oriented/OSGi lingo. You want to punch holes in a module because it is more convenient. Unfortunately I always find out later that the convenience came at an often steep price.

> Regardless off offline build's , it will allow users who want to have more than 1 git repo per eclipse workspace.
> 
I have not looked in detail what your patch does but if it allows projects outside the workspace directory then I fully support BJ. We now can ensure that offline builds work for any workspace because we define the disk structure. Once we open it up, we get lots of different variations that will not be a-priori compatible with the tools. Constraints are good, the resulting simplicity is highly undervalued.

With Git module support you can manage this decently. Alternatively you can also use Git modules to maintain the cnf directory as a separate repository.

	workspace
		.git
		cnf
			.git
			build.bnd
		project
			bnd.bnd

Again, this discussion has been done many times already. I feel the pain but I am afraid it will require my dead body so far :-)

Kind regards,

	Peter Kriens

rimmeraj:
I think the disconnect is that when using a service based model we get into the mode of having one service do one thing. (Which is the right way!)  When bnd was developed is was for offline builds and in a maven style where instead of specifying locations it makes assumptions about where things are for simplicity. Now along comes bndtools/eclipse which has it's own ideas about project layout and bnd is saying "My way or the highway", I only do one thing and do it well. I still think this patch will have zero impact since the rest of the code just uses the file path that is assigned to it on creation. I would also suggest that in the spirit of only doing one thing , that the motto "only do it in one place" should apply and that when the project is added that the location is set and everyone else uses it. 

I am not a violent man so since Peter's dead body is not an option, and unless he has a change of heart, I will just fork bnd and bndtools and build a version that supports this and make it publicly available. If people find it useful then maybe you will reconsider, otherwise maybe the process will make me see the error in my ways.

bjhargrave:
> Also, the main issue seems to be that if cnf is a file pointing to the cnf folder, then this causes bnd to use the parent folder of the cnf folder as the workspace. Perhaps it is better to attack the issue there. I would think a cnf file should act as a symlink to a cnf folder and not cause the workspace to be "relocated".

I spoke with Peter about this change and he agrees this would be a good fix and should also address your concern.

bjhargrave:
> I spoke with Peter about this change and he agrees this would be a good fix and should also address your concern.

I made a PR, https://github.com/bndtools/bnd/pull/781, with this solution. Please look at it to see if it addresses your needs instead of this PR.

rimmeraj:
Firstly thanks for spending cycles on this, I appreciate it.  Unfortunately this will not work in bndtools. For example

Repo1 
   proj1
   proj2
   cnf (file containing ../Repo2)
Repo2
   cnf

So in eclipse you would see
proj1
proj2
cnf

When bndtools starts up is asks eclipse for a list of projects, finds the cnf project and initializes the bnd workspace with a call to getWorkspace with the cnf directory. Then it goes through and gets the rest of the projects. Eclipse returns an IProject for each project. Bndtools then does getLocation().toFile().getName() and calls Workspace.getProject() with this value. Workspace then sees if it has already created the project, if not looks for that folder relative to the cnf directory. In the above example Repo1/proj1 gets turned into proj1 and getProject looks for it in Repo2/proj1 and can not find it. My patch allows bndtools to call Workspace.getProject with getLocation().toFile(). The important thing here is in this world bndtools knows where all the projects are and does not need resolution by bnd.  

The other changes I made were simply tidying the code so once we had a project file path bnd only initialized it in one spot, thus the split.

bjhargrave:
> if not looks for that folder relative to the cnf directory

I see this <https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java#L79>:

    Workspace ws = getWorkspace(projectDir.getParentFile());
    return ws.getProject(projectDir.getName());

But that is the static method `getProject(File)` perhaps you mean the instance method `getProject(String)`. (Your comment is not clear.)

If bndtools used the static method with the File argument, proj1 and proj2 will use the Workspace for Repo1 (parentFile of repo1 and repo2). cnf will use the Workspace for Repo2 (parentFile of cnf). This means that under the covers, there are two Workspace objects. You should try that.

PS. I fixed your comment to get the proper indentation (I hope) of the projects.

rimmeraj:
Comment looks good. There is only one workspace in bndtools, it is calling the instance method not the static.

Here is where the workspace is saved
https://github.com/bndtools/bndtools/blob/master/bndtools.core/src/bndtools/central/Central.java

I would change line 581. The refresh handler needs some work but that is for later....

bjhargrave:
Looking around, there seems to be a lot in bndtools that assumes a single workspace containing all the projects (which is the way bnd also assumes things). So even though Eclipse can cobble a workspace from multiple folders, bnd and bndtools are not really designed or prepared to do that. So this whole effort seems a folly without a significant effort to rearchitect around a new assumption that all projects are not in the same workspace.

rimmeraj:
I think it is a dead simple change with no side effects (where have you heard that before ;-) ) . I will fork bndtools and create a pull request with the required changes. I will run a bunch of tests in my env and report back with the required changes. Then you can decide.

rimmeraj:
I have created a bndtools pull request showing the changes for multiple git repos

https://github.com/bndtools/bndtools/pull/1008

I did add one more method to Workspace that is not in this pull request ,   I was not sure what I should do. The method looks like this 

public boolean isPresent(File fileName) {
                Project p = models.get(fileName.getName());
                return p != null && p.getBase().equals(fileName);
        }
I am not sure what triggers this call in bndtools (I was not able to see it being called running in the debugger) so some guidance would be appreciated.

pkriens:
How does this work in the offline build?

Kind regards,

	Peter Kriens

rimmeraj:
It uses the current feature where the cnf folder can be a file that re-directs to the actual cnf location. In the ant task, the basedir is passed to the static method Workspace.getProject() which now uses the absolute path of the project. If you are doing a build of all the repos I have a master layer that calls all of the repos , but it is all within the same Workspace so the repo's that depend on other repo's are already in the workspace. 

I do not have any experience with gradle so I can not comment on that.

pkriens:
Not sure I get this. In general, the gradle (and also the ant as I know it) builds the workspace from the directories in the workspace directory to build the workspace. I can see that any project can find the cnf directory, but I fail to see how I can find the projects from the workspace directory?

Kind regards,

	Peter Kriens

rimmeraj:
Is the build system asking the Workspace for all the projects and their build order once it has found the cnf folder OR is the build system adding the projects to the workspace and then getting the build order? From the settings.gradle in the bnd project, it would seem the latter? 

if (projectNames.remove('')) {
  rootDir.eachDir {
    def projectName = it.name
    if (!projectName.startsWith('.')) {
      projectNames += projectName
    }
  }
}
/* Add each project and its dependencies to the graph */
projectNames.each { projectName ->
  include projectName
  def project = getBndProject(workspace, projectName)
  project?.dependson.each {
    include it.name
  }
}
If so then projectName becomes projectDir and you should be all set. rootDir is not the cnf root but the root you are working in.

bjhargrave:
The gradle build, like bnd and bndtools, assumes all projects are peer folders in the same workspace folder. Yes, you can use the cnf-as-a-file "trick" to point to the real cnf elsewhere, but that does not change the assumption that all project folders are peers in the workspace folder. The gradle build looks for folders in the workspace folder and adds them as projects to gradle. The bnd Workspace object does the same and so they are in sync. The gradle build also asks the bnd Project object for the project dependencies to configure the same dependencies into the gradle build. All of this is predicated on bnd knowing where all the projects are: which is that they are peer folders in the workspace folder.

Furthermore, bndtools build is changing so the eclipse build order will be set from the bnd workspace build order which means that the bnd workspace has to know about all the projects in order to set the build order for eclipse. If the eclipse projects are not all in the same bnd workspace, then this, of course, will not work.

rimmeraj:
From my example above, if I go into Repo1 and build the project, the rootDir in gradle will be Repo1 and all of the projects added will be children of Repo1 with the correct path. Once a project is added with file path, bnd never has to find the project file path again. 

I would suggest if you are building multiple repos there is a dependency order between the repos so  there would be a master build that would call the build on each repo in a specific order. In ant I would launch the build task for each repo in the same environment so the projects added to the Workspace in previous Repo's  would be available.  This then looks the same as having multiple repos in eclipse.

As far as eclipse, the first thing bndtools does when it starts up is it asks eclipse for all of it's projects and adds them to the Workspace (With my change , the absolute path of the project) . Bnd will never have to go hunting for a project based on path because bndtools has given it all of it's projects.

bjhargrave:
> From my example above, if I go into Repo1 and build the project, the rootDir in gradle will be Repo1 and all of the projects added will be children of Repo1 with the correct path.

Yes but that example is pretty trivial. It is just moving cnf out of the workspace folder. A more complex example of having other projects in Repo2 where the projects in Repo1 and Repo2 having dependencies will not work.

> I would suggest if you are building multiple repos there is a dependency order between the repos so there would be a master build that would call the build on each repo in a specific order.

So suppose `proj1` depends upon `proj3` which is Repo2, how is the gradle build for the workspace Repo1 supposed to even know where proj3 is? Even if you build the Repo2 workspace first via some parent build.

> As far as eclipse, the first thing bndtools does when it starts up is it asks eclipse for all of it's projects and adds them to the Workspace (With my change , the absolute path of the project) . Bnd will never have to go hunting for a project based on path because bndtools has given it all of it's projects.

Only in Eclipse. In a offline build, Eclipse is not there to impart its knowledge of the "workspace" to bnd.

rimmeraj:
> Only in Eclipse. In a offline build, Eclipse is not there to impart its knowledge of the "workspace" to bnd.
I know. You were suggesting that this might break eclipse, I was pointing out that it would not.

> So suppose proj1 depends upon proj3 which is Repo2, how is the gradle build for the workspace Repo1 supposed to even know where proj3 is? Even if you build the Repo2 workspace first via some  parent build.

In my world if the build order is Repo1,Repo2 , Repo1 can not contain references to projects in Repo2. 

However if you wanted to support projects across repos you would just have to do the same thing that eclispe does in your master build. Go into each Repo, add all of the projects and then build as normal. Again, if the project is created with the absolute file path then all output/artifacts are relative to that project location not the workspace.

It would only be a few lines of build script in a master build to "init" each of the repos in the build workspace, then continue on as normal.

bjhargrave:
> However if you wanted to support projects across repos

Isn't that the whole point of your work? To have projects in multiple repos in the same bnd workspace? If they don't have any inter dependencies then why bother with any of this? You can just run multiple instances of Eclipse if you need to work on multiple bnd workspaces at the same time (like the rest of us :-).

> It would only be a few lines of build script in a master build to "init" each of the repos in the build workspace, then continue on as normal.

Perhaps, but that is highly custom logic that breaks whenever you change project dependencies (by editing bnd.bnd files). Hardly a robust build environment.

As I have said before, there is so much in bnd and bndtools which assumes all projects are peer folders in the workspace folder, that I don't see that there are any simple changes to alter this. It would be a major architectural change which is not appealing to the project leaders at this time. So I don't think this PR or the companion PR in bndtools (bndtools/bndtools#1008) will be merged.

rimmeraj:
It is obvious we both have different options and quite possibly different problems to solve. The change is so trivial I will fork the code and just maintain it. I will post the artifacts some docs on github on how to use it.  

If people find it useful then maybe you will re-consider, I appreciate the thoughtful discussion.

Dave

bjhargrave:
> I will fork the code and just maintain it

OK. Then we should close the PRs for now.

derheld42:
@rimmeraj- we use a forked version of bnd and bndtools that supports projects being in different levels in a workspace. There were a number of places in the code base that had to be adjusted to allow for this, but it's been working for us for a while now. I would guess it wouldn't be a large stretch to handle projects outside of the main cnf workspace. 

I could push the changes to my fork on github if you're interested. I've been meaning to refactor the changes into something that would be palletable to go into the baseline, but haven't found the motivation recently.

PR #749, opened 2014-12-23
Added transitive property on a project to be optionally used with eclipse

derheld42:
Added transitive property on a project to be optionally used with eclipse processing of dependencies.

This is used by the cooresponding bndtools change. Unused in bnd.

Signed-off-by: Carter Smithhart <carter.smithhart@gmail.com>

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
@@ -206,6 +206,7 @@ public interface Constants {
 	String							WAB											= "-wab";
 	String							WABLIB										= "-wablib";
 	String							REQUIRE_BND									= "-require-bnd";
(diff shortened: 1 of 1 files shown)


pkriens:
a) This property is specific for Eclipse so it should be in some -eclipse: property if we do it.
b) It feels like a hack? Are we really understand what goes on here? 

basically I am not sure why it should be an instruction? Can't we figure out what the right option is? After all the desired behavior is clear?

derheld42:
it's true that bnd doesn't use this property, so I suppose it doesn't need to be defined at all in bnd.

For some reason putting this in bnd made sense to me when I made the change...

So, I suppose this could be rejected and bndtools/bndtools#984 could be turned into a bndtools specific property (or whatever). I'm okay with that.

derheld42:
bndtools/bndtools#999 makes this obsolete.

PR #756, opened 2014-12-26
Generate correct caps for Felix SCR known bundle

seanbright:
Felix SCR 1.8+ supports DS 1.2.

If/when this is merged, the Bundle-Hub index.xml will need to be regenerated.

Signed-off-by: Sean Bright <sean@malleable.com>


bjhargrave:
Why did you close this? I think this change is fine for the next bnd release.

seanbright:
It didn't solve the whole issue (osgi.ds instead of osgi.component) and it would need to be revisited when that issue was resolved.

Issue #763, opened 2015-01-05
LocalIndexedRepo (R5) does not generate index file that is gzipped


gamerson:
While trying to create a local index file (index.xml.gz) using LocalIndexRepo I noticed that the output was getting correctly named index.xml.gz which is the default but the output was not gzipped, instead it was plain text.  I was not able to force the LocalIndexRepo to actually generate a gzipped output even when explicitly setting the pretty=false.

bjhargrave:
Fixed by #764

PR #764, opened 2015-01-05
Fix for #763

gamerson:


biz.aQute.repository/test/aQute/bnd/deployer/repository/TestLocalIndexGeneration.java
```diff
@@ -143,6 +144,30 @@ public static void testFailToGenerate() throws Exception {
...
+		PutResult r = repo.put(new BufferedInputStream(new FileInputStream(
+				"testdata/bundles/name.njbartlett.osgi.emf.minimal-2.6.1.jar")), new RepositoryPlugin.PutOptions());
+		File deployedFile = new File(r.artifact);
+
+		File indexFile = IO.getFile("generated/testoutput/index.xml.gz");
+		assertTrue(indexFile.exists());
+
+		GZIPInputStream gzip = new GZIPInputStream(new FileInputStream(indexFile));
+		assertTrue(gzip.read() > -1);
+	}
```
bjhargrave: The test case should probably close the file and delete the temp files. Windows hates left open files and Linux has limited file handles.

gamerson: Gotcha, pushed new commit.

gamerson:
@bjhargrave can this fix be considered for upcoming 2.4.1?

bjhargrave:
First we would need to take the PR into master. We can discuss this on Friday at the dev call.

bjhargrave:
Shouldn't we fix the problem at the source (LocalIndexRepo) rather than just working around it here? Then other users will not have to work around this.

gamerson:
So the problem as I understand it is this... in the repo indexer here: https://github.com/bndtools/bnd/blob/master/org.osgi.impl.bundle.repoindex.lib/src/org/osgi/service/indexer/impl/RepoIndex.java#L145

According to that table are only 3 ways to get a compressed index file.  
pretty=null, compresed=null
pretty=null, compresed=true
pretty=true, compressed=true

but the problem is that the API for repository provider here:  https://github.com/bndtools/bnd/blob/master/biz.aQute.repository/src/aQute/bnd/deployer/repository/api/IRepositoryContentProvider.java#L77

It only provides a pretty parameter and not a compressed parameter.

And of all the repositoryContent providers in this repo, only the R5ContentProvider  makes use of the pretty param.  But with just a single pretty boolean it seems hard to express all of the output behavior.  So it seems there are two options.

1) we add a new compressed parameter to generateIndex() interface breaking change
2) make the change in each of the providers but right now only the R5 provider seems to care about the pretty param.

I've pushed a new commit that doesn't touch generateIndex() sense that seems like a big change and just applies fix to R5 provider again (which is the only provider that seems to care about that pretty param)

bjhargrave:
> According to that table are only 3 ways to get a compressed index file.
> 
> pretty=null, compresed=null
> pretty=null, compresed=true
> pretty=true, compressed=true

pretty=false, compressed=true also will work. See https://github.com/bndtools/bnd/blob/master/org.osgi.impl.bundle.repoindex.lib/src/org/osgi/service/indexer/impl/RepoIndex.java#L160

> I've rewritten and pushed a new commit gamerson@ce21ab6 that doesn't touch generateIndex() sense that seems like a big change and just applies fix to R5 provider again (which is the only provider that seems to care about that pretty param)

Makes sense since https://github.com/bndtools/bnd/blob/master/biz.aQute.repository/src/aQute/bnd/deployer/repository/providers/R5RepoContentProvider.java#L54 implies pretty handles the choice between pretty=true/compressed=false and pretty=false/compressed=true. 

I'll merge.

Issue #767, opened 2015-01-07
Bindex can't handle version > 1.2.0 of DS


pkriens:
Bindex throws an error (a builder error in Eclipse!) when it encounters an unrecognized version. This is a bit harsh since the URI contains the required version.

bjhargrave:
Fixed

PR #768, opened 2015-01-07
755 + 767 + error if DS version required > package version

pkriens:


bjhargrave:
We should consider pulling this in to 2.4.1.

bjhargrave:
Fixes #767

PR #809, opened 2015-02-16
Issue 797: generating service permissions

marcdejonge:
So, I've started on an implementation that only parses the DS service definitions for now. Can you review if this is what you were expecting?

See discussion here: https://groups.google.com/forum/#!topic/bndtools-dev/Ra7GDT7x_4U

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
+++ b/biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
@@ -2807,5 +2807,4 @@ public class DSAnnotationTest extends BndTestCase {
 				errors.get(3));
 		assertEquals(0, b.getWarnings().size());
 	}
-
 }
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
@@ -1,0 +1,209 @@
+package aQute.bnd.osgi;
+
+import java.util.Collections;
+import java.util.EnumSet;
+import java.util.HashSet;
+import java.util.Map.Entry;
+import java.util.Set;
+import java.util.TreeSet;
+
+import aQute.bnd.header.Attrs;
+import aQute.bnd.osgi.Descriptors.PackageRef;
+import aQute.bnd.osgi.resource.FilterParser;
+import aQute.bnd.osgi.resource.FilterParser.Expression;
+import aQute.bnd.osgi.resource.FilterParser.ExpressionVisitor;
+import aQute.bnd.osgi.resource.FilterParser.Op;
+import aQute.bnd.osgi.resource.FilterParser.Or;
+import aQute.bnd.osgi.resource.FilterParser.PatternExpression;
+import aQute.bnd.osgi.resource.FilterParser.SimpleExpression;
+
+public class PermissionGenerator {
+	public static enum Parameter {
+		ADMIN {
+			@Override
+			public void generate(StringBuilder sb, Builder builder) {
+				sb.append("(org.osgi.framework.AdminPermission)\n");
+			}
+		},
+		CAPABILITIES {
+			@Override
+			public void generate(StringBuilder sb, Builder builder) {
+				for (String namespace : builder.getProvideCapability().keySet()) {
+					if (!Processor.isDuplicate(namespace)) {
+						sb.append("(org.osgi.framework.CapabilityPermission \"").append(namespace)
+								.append("\" \"provide\")\n");
+					}
+				}
+				for (String namespace : builder.getRequireCapability().keySet()) {
+					if (!Processor.isDuplicate(namespace)) {
+						sb.append("(org.osgi.framework.CapabilityPermission \"").append(namespace)
+								.append("\" \"require\")\n");
+					}
+				}
+			}
+		},
+		PACKAGES {
+			@Override
+			public void generate(StringBuilder sb, Builder builder) {
+				if (builder.getImports() != null) {
+					for (PackageRef imp : builder.getImports().keySet()) {
+						if (!imp.isJava()) {
+							sb.append("(org.osgi.framework.PackagePermission \"");
+							sb.append(imp);
+							sb.append("\" \"import\")\n");
+						}
+					}
+				}
+				if (builder.getExports() != null) {
+					for (PackageRef exp : builder.getExports().keySet()) {
+						sb.append("(org.osgi.framework.PackagePermission \"");
+						sb.append(exp);
+						sb.append("\" \"export\")\n");
+					}
+				}
+			}
+		},
+		SERVICES {
+			@Override
+			public void generate(StringBuilder sb, Builder builder) {
+				for (String declaredService : getDeclaredServices(builder)) {
+					sb.append("(org.osgi.framework.ServicePermission \"").append(declaredService)
+							.append("\" \"register\")\n");
+				}
+				for (String referencedService : getReferencedServices(builder)) {
+					sb.append("(org.osgi.framework.ServicePermission \"").append(referencedService)
+							.append("\" \"get\")\n");
+				}
+			}
+		};
+
+		public abstract void generate(StringBuilder sb, Builder builder);
+	}
+
+	public static final String	KEY					= "permissions";
+
+	public static Set<String> getDeclaredServices(Builder builder) {
+		Set<String> declaredServices = new TreeSet<>();
+		for (Entry<String,Attrs> entry : builder.getProvideCapability().entrySet()) {
+			if (Processor.removeDuplicateMarker(entry.getKey()).equals("osgi.service")) {
+				Attrs attrs = entry.getValue();
+				String ifaces = attrs.get("objectClass");
+				if (ifaces != null) {
+					for (String iface : ifaces.split(",")) {
+						declaredServices.add(iface.trim());
+					}
+				}
+			}
+		}
+		return declaredServices;
+	}
+
+	private static final String	MATCH_ALL	= "*";
+	private static final String	VALID_WILDCARD	= ".*";
+
+	static final class FindReferencedServices extends ExpressionVisitor<Set<String>> {
+
+		public FindReferencedServices() {
+			super(Collections.<String> emptySet());
+		}
+
+		@Override
+		public Set<String> visit(SimpleExpression expr) {
+			if ("objectClass".equals(expr.getKey())) {
+				if (expr.getOp() == Op.EQUAL) {
+					String v = expr.getValue();
+
+					if (!v.contains(MATCH_ALL) || v.equals(MATCH_ALL)) {
+						return Collections.singleton(v);
+					} else if (v.endsWith(VALID_WILDCARD) && !v.substring(0, v.length() - 2).contains(MATCH_ALL)) {
+						return Collections.singleton(v);
+					} else {
+						// All other matchings have no meaning to us, the user
+						// should generate them by hand
+						return Collections.emptySet();
+					}
+				} else {
+					// All other matchings have no meaning, so match nothing
+					return Collections.emptySet();
+				}
+			}
+			return Collections.emptySet();
+		}
+
+		@Override
+		public Set<String> visit(PatternExpression expr) {
+			// Simply use the default visit method for simple expressions, since
+			// the patterns are also supported in the permissions
+			return visit((SimpleExpression) expr);
+		}
+
+		@Override
+		public Set<String> visit(Or expr) {
+			Set<String> matches = new HashSet<>();
+			for (Expression expression : expr.getExpressions()) {
+				matches.addAll(expression.visit(this));
+			}
+			return matches;
+		}
+
+		@Override
+		public Set<String> visitTrue() {
+			return Collections.singleton(MATCH_ALL);
+		}
+	}
+
+	public static Set<String> getReferencedServices(Builder builder) {
+		Set<String> referencedServices = new TreeSet<>();
+		for (Entry<String,Attrs> entry : builder.getRequireCapability().entrySet()) {
+			if (Processor.removeDuplicateMarker(entry.getKey()).equals("osgi.service")) {
+				Attrs attrs = entry.getValue();
+				String filter = attrs.get("filter:");
+				if (filter != null && !filter.isEmpty()) {
+					FilterParser filterParser = new FilterParser();
+					Expression expr = filterParser.parse(filter);
+					referencedServices.addAll(expr.visit(new FindReferencedServices()));
+				}
+			}
+		}
+
+		if (referencedServices.contains(MATCH_ALL)) {
+			return Collections.singleton(MATCH_ALL);
+		} else {
+			return referencedServices;
+		}
+	}
+
+	private static EnumSet<Parameter> parseParams(Builder builder, final String... args) {
+		EnumSet<Parameter> parameters = EnumSet.noneOf(Parameter.class);
+		// Skip the key name, so start at index 1
+		for (int ix = 1; ix < args.length; ix++) {
+			String name = args[ix].toUpperCase();
+			try {
+				parameters.add(Parameter.valueOf(name));
+			}
+			catch (IllegalArgumentException ex) {
+				builder.error("Could not parse argument for ${permissions}: %s", args[ix]);
+			}
+		}
+		return parameters;
+	}
+
+	private final Builder		builder;
+	private final Set<Parameter>	parameters;
+
+	public PermissionGenerator(Builder builder, final String... args) {
+		assert args.length > 0 && KEY.equals(args[0]);
+		this.builder = builder;
+		this.parameters = parseParams(builder, args);
+	}
+
+	public String generate() {
+		StringBuilder sb = new StringBuilder();
+
+		for (Parameter param : parameters) {
+			param.generate(sb, builder);
+		}
+
+		return sb.toString();
+	}
+}
--- a/biz.aQute.bndlib.tests/src/test/component/PermissionGeneratorTest.java
+++ b/biz.aQute.bndlib.tests/src/test/component/PermissionGeneratorTest.java
@@ -1,0 +1,248 @@
+package test.component;
+
+import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileWriter;
+import java.io.InputStreamReader;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import aQute.bnd.osgi.Builder;
+import aQute.bnd.osgi.Constants;
+import aQute.bnd.osgi.Jar;
+import aQute.bnd.osgi.PermissionGenerator;
+import aQute.bnd.osgi.Resource;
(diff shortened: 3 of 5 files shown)


bjhargrave:
Have you seen <https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/indexer/analyzers/SCRAnalyzer.java>?

Also, you should probably develop bnd with the latest bndtools builds. 2.4.0.REL is a bit old now :-)

marcdejonge:
Hmm, and I thought I had looked for something like that. But of course I didn't look for SCR...

And for bndtools build you need a snapshot of the latest version?

bjhargrave:
You can install the latest bndtools builds from this update site: https://bndtools.ci.cloudbees.com/job/bndtools.master/lastSuccessfulBuild/artifact/build/generated/p2/

djencks:
That is totally not the approach I thought we were talking about.  I'm going to have the DS stuff write provides and requires service capabilities into the manifest.  Your permission code should read these (not yet generated) service capabilities and write permissions based on them.  Until they are generated by e.g. ds you can test by just writing a manifest with service provides/requires capabilities by hand.

pkriens:
@djencks: I think that is the future. Until that time that we have those, reading the XML makes sense. However, if you think you can quickly provide this 'capability' then it might be worth waiting?

marcdejonge:
It seems that this is working well. Can you check and merge it?

bjhargrave:
I think per 

> Your permission code should read these (not yet generated) service capabilities and write permissions based on them.

you were to redo your changes to use the generated `osgi.service` capabilities and requirements as the source material to generate the permissions info rather then the parsing the DS xml and processing the imports and exports.

https://github.com/bndtools/bnd/pull/822 is now in bnd to generate the `osgi.service` capabilities and requirements for the DS annotations.

marcdejonge:
What exactly is the advantage of parsing the capabilities in stead of the
source XML? I'd rather use the source and this is already working.

There could be an advantage to have an internal list of services that both
these outputs would be based on, but I don't see that in the implementation.

bjhargrave:
> What exactly is the advantage of parsing the capabilities in stead of the source XML? I'd rather use the source and this is already working. There could be an advantage to have an internal list of services that both these outputs would be based on, but I don't see that in the implementation.

The advantage is that if new features are added to the DS XML, or if the bundle programmer doesn't use DS but writes (or has written) her own osgi.service capabilities and requirements, then by processing the osgi.service capabilities and requirements, this features would work for them. It is future proofing.

djencks:
After thinking about it a bit, we decided that the service capabilities/requirements _were_ an appropriate universal store for this information, and any other data structure would be redundant and more likely to cause problems than provide a simplification.

thanks
david jencks

marcdejonge:
You have a good point and it does seem reasonable to use the capabilities as a generic storage.

Is there then also some intermediate representation in the builder for the generated capabilities? Or should I just parse the properties?

bjhargrave:
You need to remove the merge commits. Your PR must not include any merges, just plain commits.

marcdejonge:
Ok, just rebased and added another feature. Hopefully this will built correctly.

marcdejonge:
And should I also update the documentation about this? Can I just edit this file:

https://github.com/bndtools/bnd.manual/blob/gh-pages/_macros/permissions.md

???

bjhargrave:
> And should I also update the documentation about this? Can I just edit this file:
> 
> https://github.com/bndtools/bnd.manual/blob/gh-pages/_macros/permissions.md
> 
> ???

Once the PR is merged, then the page on the gh-pages branch of the bnd.manual repo is the proper place to update the documentation.

marcdejonge:
I've tested it and it seems to work perfectly. Can you look at it and merge it?

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+	public static final String	PARAM_SERVICES		= "services";
+
+	private final Builder		builder;
+	private final List<String>	args;
+	private final StringBuilder	sb;
+
+	public PermissionGenerator(Builder builder, final String... args) {
+		this.builder = builder;
+
+		if (args.length == 2 && PARAM_ALL.equals(args[1])) {
```
bjhargrave: I think we sholud get rid of "all". It is not documented at http://bnd.bndtools.org/macros/permissions.html and was broken in the prior implementation anyways (It was basically an equivalent to "packages"). And this implementation only work if "all" is the sole argument.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+		for (String arg : args) {
+			if (PARAM_ADMIN.equals(arg)) {
+				generateAdmin();
+			} else if (PARAM_CAPABILITY.equals(arg)) {
+				generateCapability();
+			} else if (PARAM_PACKAGES.equals(arg)) {
+				generatePackages();
+			} else if (PARAM_SERVICES.equals(arg)) {
+				generateServices();
+			} else if (KEY.equals(arg))
```
bjhargrave: I am not sure what the point of `${permissions; permissions}` since it appears to be basically a noop. I know it was this way before this PR so I am just calling into question the need to continue to support it.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+				;
+			else
+				builder.error("Invalid option in ${permissions}: %s", arg);
+		}
+
+		return sb.toString();
+	}
+
+	private void generateAdmin() {
+		sb.append("(org.osgi.framework.AdminPermission)");
```
bjhargrave: End with `\n`

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	private void generateCapability() {
+		for (String namespace : builder.getProvideCapability().keySet()) {
+			if (!Processor.isDuplicate(namespace)) {
+				sb.append("(org.osgi.framework.CapabilityPermission \"").append(namespace)
+						.append("\" \"provide\")\r\n");
```
bjhargrave: End with `\n`. We shouldn't use `\r\n` and we want a consistent output regardless of whether the bundle is built on *ix or windows.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+			}
+		}
+		for (String namespace : builder.getRequireCapability().keySet()) {
+			if (!Processor.isDuplicate(namespace)) {
+				sb.append("(org.osgi.framework.CapabilityPermission \"").append(namespace)
+						.append("\" \"require\")\r\n");
```
bjhargrave: End with `\n`. We shouldn't use `\r\n` and we want a consistent output regardless of whether the bundle is built on *ix or windows.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+
+	private void generatePackages() {
+		if (builder.getImports() != null) {
+			for (PackageRef imp : builder.getImports().keySet()) {
+				if (!imp.isJava()) {
+					sb.append("(org.osgi.framework.PackagePermission \"");
+					sb.append(imp);
+					sb.append("\" \"import\")\r\n");
```
bjhargrave: End with `\n`. We shouldn't use `\r\n` and we want a consistent output regardless of whether the bundle is built on *ix or windows.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+				}
+			}
+		}
+		if (builder.getExports() != null) {
+			for (PackageRef exp : builder.getExports().keySet()) {
+				sb.append("(org.osgi.framework.PackagePermission \"");
+				sb.append(exp);
+				sb.append("\" \"export\")\r\n");
```
bjhargrave: End with `\n`. We shouldn't use `\r\n` and we want a consistent output regardless of whether the bundle is built on *ix or windows.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+			}
+		}
+	}
+
+	private void generateServices() {
+		for (String declaredService : getDeclaredServices()) {
+			sb.append("(org.osgi.framework.ServicePermission \"").append(declaredService)
+					.append("\" \"register\")\r\n");
```
bjhargrave: End with `\n`. We shouldn't use `\r\n` and we want a consistent output regardless of whether the bundle is built on *ix or windows.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+		}
+		for (String referencedService : getReferencedServices()) {
+			sb.append("(org.osgi.framework.ServicePermission \"").append(referencedService).append("\" \"get\")\r\n");
```
bjhargrave: End with `\n`. We shouldn't use `\r\n` and we want a consistent output regardless of whether the bundle is built on *ix or windows.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	public Set<String> getDeclaredServices() {
+		Set<String> declaredServices = new TreeSet<>();
+		for (Entry<String,Attrs> entry : builder.getProvideCapability().entrySet()) {
+			if (Processor.removeDuplicateMarker(entry.getKey()).equals("osgi.service")) {
+				Attrs attrs = entry.getValue();
+				String ifaces = attrs.get("objectClass");
+				if (ifaces != null) {
+					for (String iface : ifaces.split(",")) {
```
bjhargrave: Perhaps split with "\\s*,\\s*" to handle whitespace around comma? I wonder is the user specified this in the manifest with whitespace around the commas if bnd would strip that out? Handling it here means we don't care.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+		for (Entry<String,Attrs> entry : builder.getRequireCapability().entrySet()) {
+			if (Processor.removeDuplicateMarker(entry.getKey()).equals("osgi.service")) {
+				Attrs attrs = entry.getValue();
+					String filter = attrs.get("filter:");
+					if (filter != null && !filter.isEmpty()) {
+					Matcher matcher = FILTER_PATTERN.matcher(filter);
+					if (matcher.matches()) {
+						referencedServices.add(matcher.group(1));
+					} else {
+						builder.warning("Unexpected osgi.service requirement filter: {}", filter);
```
bjhargrave: A user could specify a valid filter directive without an objectClass. So we should not warn here. Just ignore.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
marcdejonge: What I did see while debugging is that for the macro `${permissions;services}` I get 2 parameters, `args[0] = "permissions"` and `args[1] = "services"`.

bjhargrave: Ah. The macro processor must put the macro name in arg[0]. Since the code does a foreach loop on the args array, it needs to ignore element 0 rather then using a for loop with an index starting at 1.

bjhargrave:
Also, please squash your commits together and force push it so there is just the single commit. Thanks.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
marcdejonge: I think this code is equivalent to the previous version, that also only worked if "all" was the only parameter.

But I agree that maybe it should be dropped or well documented.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
marcdejonge: I forgot, will fix all of these...

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
bjhargrave: In the original code `all` was equivalent to `packages` since each iteration over args would match the first if in the if/else chain. `${permissions; admin; all}` was the same as `${permissions; admin; packages}`.

biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
```diff
@@ -2774,4 +2809,105 @@ public static void testMixedStandardBnd() throws Exception {
...
 
+	public static void testPermissions() throws Exception {
+		Builder b = new Builder();
+		b.setProperty(Constants.DSANNOTATIONS, "test.component.*_basic");
+		b.setProperty("Private-Package", "test.component");
+		b.setProperty("Export-Package", "test.api");
+
+		File tmpFile = File.createTempFile("bndtest", "permissions.perm");
+		FileWriter fileWriter = new FileWriter(tmpFile);
+		fileWriter.write("${permissions;packages}${permissions;services}${permissions;capability}");
```
bjhargrave: Should this be `${permissions;packages;services;capabilities}`?

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
```diff
@@ -0,0 +1,145 @@
...
+
+import aQute.bnd.header.Attrs;
+import aQute.bnd.osgi.Descriptors.PackageRef;
+
+public class PermissionGenerator {
+	public static final String	KEY					= "permissions";
+
+	public static final String	PARAM_ALL			= "all";
+	public static final String	PARAM_ADMIN			= "admin";
+	public static final String	PARAM_CAPABILITY	= "capability";
```
bjhargrave: This should be plural like services and packages. So `capabilities`.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
bjhargrave: Since the constructor puts the array in a List, it could remove the first element "permissions" and thus the need for the test in the for loop.

biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
bjhargrave: Should the test not also cover `${permissions;admin}`?

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
marcdejonge: Furthermore, I'd often would want to use the capability, packages and services, but not the admin. The admin permission is not often used, but the others are. Maybe create a special parameter for that?

bjhargrave: Lets just remove `all`. It is not helpful since the definition is subject to change as we add new support like now. It is not hard to say `${permissions;packages;services;capabilities}`.

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
marcdejonge: I agree that the warning is not needed here. But what should we do if someone adds a wildcard filter?

marcdejonge:
Ok, I've fixed of lot of these comments and have extended the tests quite a bit. Since there were this many tests now, I've also put that into a separate test class.

Still not sure how we should handle the "all" parameters. Leave it out for now?

bjhargrave:
> Still not sure how we should handle the "all" parameters. Leave it out for now?

Yes. Remove the all option entirely.

marcdejonge:
Ok, removed in the last amend :smile:

biz.aQute.bndlib/src/aQute/bnd/osgi/PermissionGenerator.java
```diff
@@ -0,0 +1,147 @@
...
+
+	public static Set<String> getReferencedServices(Builder builder) {
+		Set<String> referencedServices = new TreeSet<>();
+		for (Entry<String,Attrs> entry : builder.getRequireCapability().entrySet()) {
+			if (Processor.removeDuplicateMarker(entry.getKey()).equals("osgi.service")) {
+				Attrs attrs = entry.getValue();
+					String filter = attrs.get("filter:");
+					if (filter != null && !filter.isEmpty()) {
+					Matcher matcher = FILTER_PATTERN.matcher(filter);
+					if (matcher.matches()) {
```
bjhargrave: Since the filter expression can  be arbitrarily complex, we can only try to handle a reasonable subset. Right now you handle the single expression `(objectClass=com.foo.Service)`. We could also expand to handle trailing wildcards in the service name: `(objectClass=com.foo.*)`. A more complex thing to support would be the objectClass expression in an AND list: `(&(X)(objectClass=com.foo.*)(Y))` where X and Y are other expressions. This latter thing would be a fairly complex regex so we can leave out for now.

pkriens: There is a FilterParser in aQute.bnd.osgi.resource that parses and optimizes filters. You could try to use this class since it handles a lot of these cases much better than a simple Regex

marcdejonge:
Ok, I've added a more advanced parsing mechanism for the capability filter, although it is still somewhat limited. It can now also parse filters such as:

```
(objectClass=*)
(objectClass=test.*)
(!(objectClass=test.*))
(|(objectClass=class1)(objectClass=class2))
```

And probably some more. But there are some limitations (e.g. what to do with the & operation? Now it generates nothing).

To do this, I needed to add a way to the FilterParser expressions such that it is much easier to do something based on the type of expression. For this I've added the visitor pattern to the expressions.

marcdejonge:
btw, why is the appveyor build failing? It is in some part I haven't touched as far as I know.

bjhargrave:
> btw, why is the appveyor build failing? It is in some part I haven't touched as far as I know.

We just added a CI build for Windows using Appveyor and there are some windows test failures we need to resolve. The test failures don't appear to be related to your changes.

marcdejonge:
Thanks for all your feedback @bjhargrave, you were right that in all 'exotic filters' we simple should not generate anything. Fixed and rebased.

Issue #823, opened 2015-02-22
[DS] Designate factory should encourage "require"


djencks:
112.13.3.10 Designate annotation with factory=true should result in default configuration-policy=require.

djencks:
Implementation currently on top of 792, would be nice to get that in first.

djencks:
Fixed with merge request #824

PR #824, opened 2015-02-22
#823 designate implies required 1

djencks:
Fix for #823


bjhargrave:
The commit comment for the the last commit has 832 instead of 823.

djencks:
Wrong branch name, replaced by pull request #837, which also has only relevant changes in it.

PR #825, opened 2015-02-23
Special treatment for artifacts with version numbers consisting only of ...

AdrianoKF:
I have been using the Aether repository plugin with success so far, but it a small bump in the road today when trying to fetch the Google Guava library from Maven Central. When adding a build dependency to "com.google.guava:guava;version=18.0;strategy=exact" the bndtools build fails with an Exception in aQute.bnd.version.Version.parseVersion(), since the version number "r03" gets passed there without proper handling.

This pull request suggests that version numbers following consisting of just a raw revision number (as "r03") be treated as if they had major and minor numbers 0 and a micro number corresponding to the build number.

### DIFF
--- a/biz.aQute.repository.aether/src/aQute/bnd/deployer/repository/aether/MvnVersion.java
+++ b/biz.aQute.repository.aether/src/aQute/bnd/deployer/repository/aether/MvnVersion.java
@@ -7,6 +7,7 @@ import aQute.bnd.version.*;
 public class MvnVersion implements Comparable<MvnVersion> {
 	
 	private static final Pattern QUALIFIER = Pattern.compile("[-.]?([^0-9.].*)$");
+	private static final Pattern RELEASE_ONLY_QUALIFIER = Pattern.compile("^[rR](0*[0-9]+)$");
 	
 	private static final String	QUALIFIER_SNAPSHOT	= "SNAPSHOT";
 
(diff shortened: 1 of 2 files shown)


bjhargrave:
How is "r03" a valid maven version string?

AdrianoKF:
Indeed a valid question - these artifacts should've never made it to the repository in the first place. That's why I suggested the workaround - at least that way the artifact can be specified as a build dependency without the exception during the indexing of the repository (especially since I can't be the only one having tried to include Guava via the Aether plugin).

I do agree though, that the actual burden of following the proper versioning scheme should be on the artifact maintainers.

bjhargrave:
Seems to me the better solution is for Guava to be re-released with a proper version number and everyone ignore the "r03" version. Hacking around someone else's mistake in bnd seems wrong. How many such hacks would we add over the years?

bjhargrave:
Where does "r03" come from? https://repo1.maven.org/maven2/com/google/guava/guava/18.0/guava-18.0.pom shows the version is, of course, 18.0. In the [jar](http://search.maven.org/remotecontent?filepath=com/google/guava/guava/18.0/guava-18.0.jar), it has `Bundle-Version: 18.0.0`.

Something else is wrong with your setup. Also there was a recent fix to MvnVersion to better handle versions from maven.

PR #828, opened 2015-02-24
Fix default JPM4J URL

njbartlett:


bjhargrave:
Your commit message is missing the Signed-off-by.

PR #834, opened 2015-02-27
dogfood: Use biz.aQute.repository from latest build

bjhargrave:
Replace the old (2.4.0), fixed version of biz.aQute.repository checked
into the repo with latest version downloaded from ${bnd_repourl}.


bjhargrave:
This is an alternate idea to #835. In this PR, you need to be connected to the network to download the repository jar into cnf/cache. In #835 the repository jar is part of the embedded repo which is part of bndtools and the bnd gradle plugin.

bjhargrave:
The repository code is now included in the gradle plugin (see #970). Closing this PR.

PR #835, opened 2015-02-27
embedded-repo: Add biz.aQute.repository

bjhargrave:
This will make sure we can use the matching version of repository
with the bnd version in use. We init the embedded repo early, so we
can refer to the repository jar from -pluginpath.


bjhargrave:
This is an alternate idea to #834. This PR would be followed by a change to use the repository jar from cnf/cache.

bjhargrave:
The repository code is now included in the gradle plugin (see #970). Closing this PR.

PR #837, opened 2015-03-01
#823 Implement and test effect of Designate annotation on configuration-p...

djencks:
...olicy


bjhargrave:
LGTM

Issue #690, opened 2014-11-17
custom annotation support to generate custom-namespace xml for ds and metatype


djencks:
Provide a way to specify custom annotation processors to add custom namespace attributes to generated DS and metatype xml resources.  Provide "generic" bnd annotations to directly specify the annotations as an example.

djencks:
Please review merge request #838

PR #838, opened 2015-03-01
#690 custom anno 3

djencks:
First commit extends bnd annotation processing so that nested annotations are automatically processed.

Second commit introduces interfaces for ds and metatype extension annotation processors.  To demonstrate that the extension mechanisms work, as an example, and as a last resort for those who don't want to write more specific annotation processors, I've included generic bnd extension annotations and processors.  I expect that Apache Felix DS and the proprietary project I work on will write and maintain project specific extension processors that are considerably easier to use than the generic ones.

In particular I would like review of whether I have used bnd plugins appropriately and whether the @BndPlugin annotation is used appropriately.


bjhargrave:
@pkriens needs to review this.

djencks:
>> why not field?  
Metatype can't have fields.  I haven't been able to think of any plausible reason to annotate methods or fields for DS components; only the class level annotations are supported.  I thought waiting for someone to come up with a reason to add attributes to <reference> elements was plausible before putting in generality that seemed useless to me.
>> prefix collision
The bnd annotations let you suggest a prefix.  If you make contradictory suggestions, for two different namespaces to use the same prefix, the processor alters at least one so there is no collision.

djencks:
Peter had a much better idea, closing this request.

PR #842, opened 2015-03-03
Lost main code in bnd & better location support for errors.

pkriens:


biz.aQute.bndlib.tests/src/test/LocationTest.java
```diff
@@ -22,6 +23,52 @@ public void tearDown() throws Exception {
 		IO.delete(new File("tmp"));
```
bjhargrave: tearDown still deletes "tmp" not "generated/tmp".

Issue #846, opened 2015-03-05
Baselining sees a MAJOR change when changing a return type compatibly


pkriens:
I got a major change from changing getExecutor in Processor. I am surprised since the ScheduledExecutorService extends the Executor through ExecutorService. So I guess we should not remove the Executor in this case. Vaguely recall I removed code because we reasoned that I did not need to analyze the hierarchy.

     MAJOR      METHOD     getExecutor()
       REMOVED    RETURN     java.util.concurrent.Executor
       ADDED      RETURN     java.util.concurrent.ScheduledExecutorService

bjhargrave:
According to <http://docs.oracle.com/javase/specs/jls/se7/html/jls-13.html#jls-13.4.15> changing the return type is a major change. So baselining is correct. You need to revert back to the original return type. You can add another method, e.g. getScheduledExecutor(), to return the subtype.

pkriens:
Yes, I am convinced now. I got it mixed up with covariant returns.

PR #847, opened 2015-03-05
[basleine] Change was not compatible enough

pkriens:
Filed #846 because I think this is a compatible change.


timothyjward:
I think this needs to be reverted as per BJ's comment in the linked bug.

PR #854, opened 2015-03-09
Corrected parsing of Maven versions with numerical qualifiers

rkrzewski:
This is a follow up to https://github.com/bndtools/bnd/pull/637

After upgrading to Bndtoools 3.0.0.DEV, I've discovered that Maven versions with numerical qualifiers were parsed incorrectly "3.2.2-2" for example was parsed as Version(3, 2, 2, "-2"). I'm a bit puzzled that I haven't noticed that before because I'm dealing with precisely this type of versions in a toy project of mine. Apparently there was an even number of bugs canceling each other out, and now this symmetry got broken :)

Anyway, I have a fix that involves an additional pattern matching and overwriting qualifier that was already parsed. I've tinkered with it a bit, I wasn't able to devise a single pattern that would handle all the edge cases.

The patch contains whitespace changes introduced by Eclipse save action (enabled by https://github.com/bndtools/bnd/commit/603e4b769b62e95cde7486fd28fb609c27f9ff73)

### DIFF
--- a/biz.aQute.repository.aether/src/aQute/bnd/deployer/repository/aether/MvnVersion.java
+++ b/biz.aQute.repository.aether/src/aQute/bnd/deployer/repository/aether/MvnVersion.java
@@ -5,47 +5,50 @@ import java.util.regex.*;
 import aQute.bnd.version.*;
 
 public class MvnVersion implements Comparable<MvnVersion> {
-	
-	private static final Pattern QUALIFIER = Pattern.compile("[-.]?([^0-9.].*)$");
-	
-	private static final String	QUALIFIER_SNAPSHOT	= "SNAPSHOT";
 
-	private final Version osgiVersion;
+	private static final Pattern	QUALIFIER			= Pattern.compile("[-.]?([^0-9.].*)$");
+
+	private static final Pattern	NUM_QUALIFIER		= Pattern.compile("-[0-9]+");
+
+	private static final String		QUALIFIER_SNAPSHOT	= "SNAPSHOT";
+
+	private final Version			osgiVersion;
 
 	public MvnVersion(Version osgiVersion) {
 		this.osgiVersion = osgiVersion;
 	}
-	
+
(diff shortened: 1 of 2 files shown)


bjhargrave:
> I wasn't able to devise a single pattern that would handle all the edge cases.

Why not use a modified version of the pattern used for OSGi versions?
<https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/version/Version.java#L10>

`"(\\d{1,9})(\\.(\\d{1,9})(\\.(\\d{1,9}))?)?([-\\.]?([-_\\.\\da-zA-Z]+))?"`

I tested the above pattern out and it it works fine for you example above and all the tests in MvnVersionTest.

group(1) is the major
group(3) is the minor (may be null)
group(5) is the micro (may be null)
group(7) is the qualifier (may be null)

Then the code does not need to try and peel off the qualifier and then parse the rest. It can just parse the whole maven version string into its parts.

rkrzewski:
Great! I'll update the PR tonight.

rkrzewski:
@bjhargrave I had to tweak the expression a bit to cover all the cases and implement some additional logic. I am not sure if the result is an improvement WRT readability ;).
If you like this version better, I'll squash both commits. Otherwise, I'll revert the latter.

bjhargrave:
The change you made to the regex I created does not work on some examples that mine did. `1rc1` does not parse. What example was not parseable by my regex?

You can test out the regex at <http://www.regexplanet.com/advanced/java/index.html>.

rkrzewski:
It didn't work for about half of the examples in MvnVersionTest, which makes me think I must have misunderstood you completely. I can't work on it right now and I'll be out of town for the weekend, so I'll get back to you on Monday.

bjhargrave:
It was easier for me to just author the fix which I released in #864 :-)

Hopefully that addresses your original concern.

rkrzewski:
#864 works great, thanks.

PR #855, opened 2015-03-09
bnd-maven-plugin new contribution

njbartlett:
This is a rewrite of the bnd-maven-plugin that can be used to build Maven projects with processing via bnd.

Development of this plugin was sponsored by Salesforce Inc. and is being contributed to the bnd project under the terms of the Apache License V2.


rotty3000:
:+1:

bnd-maven-plugin-parent/.settings/org.eclipse.core.runtime.prefs
```diff
@@ -0,0 +1,2 @@
+eclipse.preferences.version=1
```
bjhargrave: need a resource.prefs for UTF-8

bnd-maven-plugin-parent/LICENSE
```diff
@@ -0,0 +1,177 @@
+
```
bjhargrave: We already have a LICENSE file in the root of the REPO. Do we need another one here?

bnd-maven-plugin-parent/bnd-maven-plugin-dependencies/pom.xml
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>biz.aQute.bnd</groupId>
+        <artifactId>bnd-maven-plugin-parent</artifactId>
+        <version>2.5.0-SNAPSHOT</version>
```
bjhargrave: This is a PR for master branch which is 3.0.0. So this should be 3.0.0 here. I know you also need to build a version for bndlib 2.4.1 but that needs to be built outside of the master branch. Either in next (where 2.4.1 is built) or in some temporary branch.

bnd-maven-plugin-parent/bnd-maven-plugin-dependencies/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+            <plugin>
+                <artifactId>maven-antrun-plugin</artifactId>
+                <version>1.7</version>
+                <executions>
+                    <execution>
+                        <phase>initialize</phase>
+                        <configuration>
+                            <target>
+                                <!-- Install the bnd.jar built locally into the local maven repo under the version defined in ${bnd.version} -->
+                                <property name="bnd.jar" value="${basedir}/../../biz.aQute.bnd/generated/biz.aQute.bnd-3.0.0.jar"/>
```
bjhargrave: the 3.0.0 version in the file name is awkward. I think you should change the value to 

    value="${basedir}/../../dist/bundles/biz.aQute.bnd/biz.aQute.bnd-latest.jar"

This means that the `gradle :dist:releaseNeeded` target (or the default gradle target) need to be build first.

I am also not sure why this is bnd and not bndlib since we want to build the maven plugin against the bndlib we just built with gradle.

bnd-maven-plugin-parent/bnd-maven-plugin/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -0,0 +1,5 @@
+eclipse.preferences.version=1
```
bjhargrave: Need prefs for newline and UTF-8 (like other bnd projects).

bnd-maven-plugin-parent/bnd-maven-plugin/pom.xml
```diff
@@ -0,0 +1,75 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>biz.aQute.bnd</groupId>
+		<artifactId>bnd-maven-plugin-parent</artifactId>
+		<version>2.5.0-SNAPSHOT</version>
```
bjhargrave: This needs to be 3.0.0 as discussed above.

bnd-maven-plugin-parent/bnd-maven-plugin/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+
+		<dependency>
+			<groupId>org.apache.maven</groupId>
+			<artifactId>maven-project</artifactId>
+			<version>2.2.1</version>
+		</dependency>
+	
+		<dependency>
+			<groupId>biz.aQute.bnd</groupId>
+			<artifactId>bndlib</artifactId>
```
bjhargrave: artifactId is now biz.aQute.bndlib (bsn)

bnd-maven-plugin-parent/bnd-maven-plugin/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+			<version>2.4.0</version>
```
bjhargrave: This needs to be 3.0.0-SNAPSHOT which will have been installed in the local repo from the gradle build by the dependencies pom. That is, this needs to use the bndlib just built not one from maven central.

bnd-maven-plugin-parent/bnd-maven-plugin/pom.xml
```diff
@@ -0,0 +1,75 @@
...
+		</dependency>
+
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
+			<version>3.8.1</version>
+			<scope>test</scope>
```
bjhargrave: Somehow this junit dependency shows upon the built jar?

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/BndMavenPlugin.java
```diff
@@ -0,0 +1,165 @@
...
+	private MavenProject project;
+	
+	@Parameter(property = "bnd.trace", defaultValue = "false", readonly = true)
+	private boolean trace;
+
+	public void execute() throws MojoExecutionException {
+		File bndFile = new File(project.getBasedir(), Project.BNDFILE);
+		
+		Builder builder = new Builder();
+		builder.setTrace(true);
```
bjhargrave: Should the trace setting not be hard coded to true?

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/BndMavenPlugin.java
```diff
@@ -0,0 +1,165 @@
...
+				if (artifactFile != null)
+					classpath.add(artifactFile);
+			}
+			classpath.add(classesDir);
+			builder.setClasspath(classpath.toArray(new File[classpath.size()]));
+
+			// Include local project packages automatically
+			String includes = builder.getProperty(Constants.INCLUDERESOURCE);
+			if (includes == null || includes.trim().length() == 0)
+				builder.setProperty(Constants.INCLUDERESOURCE, classesDir.getPath());
```
bjhargrave: Shouldn't you add to an existing -includeresources?

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/BndMavenPlugin.java
```diff
@@ -0,0 +1,165 @@
...
+			builder.setProperty(Constants.BUNDLE_VERSION, mvnVersion.toBndVersion());
+
+			// Build bnd Jar (in memory)
+			Jar bndJar = builder.build();
+			
+			// Generate errors and warnings
+			List<String> errors = builder.getErrors();
+			if (errors != null) for (String error : errors) {
+				getLog().error(error);
+			}
```
bjhargrave: Shouldn't the build be failed if there are errors?

bnd-maven-plugin-parent/example-projects/example-api-bundle/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+			<version>6.0.0</version>
+		</dependency>
+	</dependencies>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>biz.aQute.bnd</groupId>
+				<artifactId>bnd-maven-plugin</artifactId>
+				<version>2.5.0-SNAPSHOT</version>
```
bjhargrave: 3.0.0-SNAPSHOT. I think we need to figure out how to pass the version as an arg to the maven command so we don't have to edit these after each release. (sigh)

bnd-maven-plugin-parent/example-projects/example-impl-bundle/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+			<version>0.0.1-SNAPSHOT</version>
+		</dependency>
+	</dependencies>
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>biz.aQute.bnd</groupId>
+				<artifactId>bnd-maven-plugin</artifactId>
+				<version>2.5.0-SNAPSHOT</version>
```
bjhargrave: 3.0.0-SNAPSHOT

bnd-maven-plugin-parent/pom.xml
```diff
@@ -0,0 +1,92 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    
+    <groupId>biz.aQute.bnd</groupId>
+    <artifactId>bnd-maven-plugin-parent</artifactId>
+    <version>2.5.0-SNAPSHOT</version>
```
bjhargrave: 3.0.0-SNAPSHOT

bnd-maven-plugin-parent/pom.xml
```diff
@@ -0,0 +1,92 @@
...
+    <developers>
+        <developer>
+            <id>njbartlett</id>
+            <name>Neil Bartlett</name>
+            <organization>Paremus Ltd</organization>
+        </developer>    
+    </developers>
+
+    <properties>
+        <bnd.version>2.5.0-SNAPSHOT</bnd.version>
```
bjhargrave: Ah. Here is a variable with the bnd maven plugin version. We need to use the var in all the other pom files.

It should also be 3.0.0-SNAPSHOT.

bjhargrave:
I added comments to the PR.

bjhargrave:
> :+1:

@rotty3000 Does this maven plugin do what you would want a workspace-less gradle plugin to do? That is, arrange `target/classes` as a pre-jar step?

rotty3000:
I think that would be ideal as one would not have to replace the default
jar task of the java gradle plugin.

bnd-maven-plugin-parent/bnd-maven-plugin-dependencies/pom.xml
njbartlett: @bjhargrave I don't have cycles to work on the 3.0 version right now and need to get 2.4 released. Not sure I understand why everything in the branch has to have a synchronized version. Doesn't sound very modular to me. Anyway if that's the rule then I will back out this PR and prepare a new one against `next`.

bnd-maven-plugin-parent/bnd-maven-plugin-dependencies/pom.xml
njbartlett: @bjhargrave It was bnd rather than bndlib in the code I restored from the old plugin. I don't think `-latest.jar` will work because no such JAR is generated during the build.

bnd-maven-plugin-parent/bnd-maven-plugin-dependencies/pom.xml
bjhargrave: Well the master branch is about building the upcoming release which is 3.0. This would mean that a bnd-maven-plugin built in this branch would build against the bndlib also built in this branch which is bndlib 3.0. We need to have this in the master branch to build the 3.0 version of the bnd-maven-plugin for 3.0. But you also have an additional requirement which is to build a version for bndlib 2.4.1 and that should be built out of another branch. That branch does not have to be in this repo  but it certainly can be. If it is in this repo, then the bnd-maven-plugin based upon bndlib 2.4.1 should be built in next or a new branch off next (which is probably safest).

bnd-maven-plugin-parent/bnd-maven-plugin-dependencies/pom.xml
bjhargrave: I am not sure why the cold code used bnd.jar at all. Why that is even there should be figured out. And a -latest is in dist/bundles.

bnd-maven-plugin-parent/bnd-maven-plugin-dependencies/pom.xml
bjhargrave: I am happy to help with the 3.0 version of this in the master branch if you need help.

njbartlett: Fine, I will prepare a new PR based on next. For that branch I can omit the dependency pulling stuff since it will build against Maven Central.

bjhargrave: Make sure to use bndlib 2.4.1 (rather than 2.4.0) in your new PR.

njbartlett: Thanks. I will close this PR and make your other suggested fixes in the new PR.

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/BndMavenPlugin.java
njbartlett: Not sure. I chose to allow the user to override the implicit `-includeresource` by setting it in bnd.bnd.

Perhaps if we prepend our implicit setting to the existing `-includeresource` then the user can use a negative pattern like `!target/classes/blah`?

bjhargrave: That sounds good. I just thought is was wrong to not add `target/classes` since then it is harder for the user to add something else without also having to themselves specify `target/classes`.

Also what about `target/resources`? I guess there is no way for bnd to process resources in there?

njbartlett: Yes probably `target/resources` is a good idea. I can't think of anything *currently* that bnd might process other than classfiles but that might change in the future.

bjhargrave: bnd can process resources for macro replacement when you do `-includeresource: {foo}`.

njbartlett: Turns out there is no `target/resources`. Maven copies the contents of `src/main/resources` into `target/classes` so we only need to care about that one folder.

PR #856, opened 2015-03-10
[gradle] The gradle plugin access a package private instance variable in...

pkriens:
...stead of the accessor method.


bjhargrave:
Good catch. I have a larger fix I will release instead in #857.

PR #865, opened 2015-03-14
Gradle builder plugin: Add gradle support for non-Workspace environments

bjhargrave:
A new Bundle task type is defined which extends the Jar task type and
uses the bnd Builder class to make the jar into a bundle. An
biz.aQute.bnd.builder plugin is defined which, when applied to a project
modifies the jar task to use the bnd Builder class to make the jar into
a bundle.


bjhargrave:
@rotty3000 Check this out! I need to make a wiki to document it.

bjhargrave:
Also, thanks to @njbartlett for the insight provided by the proposed bnd-maven-plugin from https://github.com/bndtools/bnd/pull/855.

PR #868, opened 2015-03-18
ant: Add Ant task to deploy run bundles like the Gradle version

### LINKED ISSUE #866: bnddistribute Ant task does not work when -runbundles uses properties
For example:

```
build.bundles: \
    xyz.core.api,\
    xyz.other.bundle
    
-runbundles: \
    ${build.bundles}
```

Results in:

[bnddistribute] ${build.bundles} could not be found in any repository

I'm using bnd 2.4.0

dhumeniuk:
Fixes #866

Signed-off-by: Humeniuk, Dave <david.humeniuk@udri.udayton.edu>

### DIFF
--- a/biz.aQute.bnd/src/aQute/bnd/ant/RunBundlesTask.java
+++ b/biz.aQute.bnd/src/aQute/bnd/ant/RunBundlesTask.java
@@ -1,0 +1,67 @@
+package aQute.bnd.ant;
+
+import java.io.*;
+
+import org.apache.tools.ant.*;
+
+import aQute.bnd.build.*;
+import aQute.bnd.build.Project;
+import aQute.bnd.build.model.*;
+
+public class RunBundlesTask extends Task {
+
+	private File			rootDir;
+	private File			buildProject;
+	private String			outputDir;
+	private File			bndFile;
+
+	@Override
+	public void execute() throws BuildException {
+		try {
+			createReleaseDir();
+			BndEditModel model = new BndEditModel();
+			model.loadFrom(bndFile);
+			Project bndProject = new Project(new Workspace(rootDir), buildProject, bndFile);
+			
+			bndProject.exportRunbundles(bndFile.getName(), new File(outputDir));
+
+			bndProject.close();
+		}
+		catch (Exception e) {
+			e.printStackTrace();
+		}
+	}
+
+	private File createReleaseDir() {
+		File releaseDir = new File(outputDir);
+		boolean deleted = releaseDir.delete();
+		if (deleted) {
+			log("Deleted directory " + outputDir);
+		}
+
+		boolean created = releaseDir.mkdir();
+		if (created) {
+			log("Created directory " + outputDir);
+		} else {
+			throw new BuildException("Output directory '" + outputDir + "' could not be created");
+		}
+
+		return releaseDir;
+	}
+
+	public void setRootDir(File rootDir) {
+		this.rootDir = rootDir;
+	}
+
+	public void setBuildProject(File buildProject) {
+		this.buildProject = buildProject;
+	}
+
+	public void setOutputDir(String outputDir) {
+		this.outputDir = outputDir;
+	}
+
+	public void setBndFile(File bndFile) {
+		this.bndFile = bndFile;
+	}
+}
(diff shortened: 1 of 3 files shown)


biz.aQute.bnd/src/aQute/bnd/ant/RunBundlesTask.java
```diff
@@ -0,0 +1,67 @@
...
+	private File			buildProject;
+	private String			outputDir;
+	private File			bndFile;
+
+	@Override
+	public void execute() throws BuildException {
+		try {
+			createReleaseDir();
+			BndEditModel model = new BndEditModel();
+			model.loadFrom(bndFile);
```
bjhargrave: Why do you create a BndEditModel since it is not used anywhere?

biz.aQute.bnd/src/aQute/bnd/ant/RunBundlesTask.java
```diff
@@ -0,0 +1,67 @@
...
+
+import java.io.*;
+
+import org.apache.tools.ant.*;
+
+import aQute.bnd.build.*;
+import aQute.bnd.build.Project;
+import aQute.bnd.build.model.*;
+
+public class RunBundlesTask extends Task {
```
bjhargrave: Shouldn't this extend BaseTask?

biz.aQute.bnd/src/aQute/bnd/ant/RunBundlesTask.java
```diff
@@ -0,0 +1,67 @@
...
+		return releaseDir;
+	}
+
+	public void setRootDir(File rootDir) {
+		this.rootDir = rootDir;
+	}
+
+	public void setBuildProject(File buildProject) {
+		this.buildProject = buildProject;
+	}
```
bjhargrave: Can this be defaulted to the current dir?

biz.aQute.bnd/src/aQute/bnd/ant/RunBundlesTask.java
```diff
@@ -0,0 +1,67 @@
...
+
+	public void setOutputDir(String outputDir) {
+		this.outputDir = outputDir;
+	}
+
+	public void setBndFile(File bndFile) {
+		this.bndFile = bndFile;
```
bjhargrave: Shouldn't this be defaulted to bnd.bnd in the BuildProject dir?

dhumeniuk:
This was copied from the RunconfigToDistributionTask. Maybe that wasn't the best example to use?

biz.aQute.bnd/src/aQute/bnd/ant/RunBundlesTask.java
dhumeniuk: copied from other task, will remove

biz.aQute.bnd/src/aQute/bnd/ant/RunBundlesTask.java
dhumeniuk: I don't know, just going by the task I copied from

biz.aQute.bnd/src/aQute/bnd/ant/RunBundlesTask.java
dhumeniuk: is there an example of how I could do this, first time writing an Ant task, I'm not seeing other Ant tasks that default to bnd.bnd, but I only sampled a couple

biz.aQute.bnd/src/aQute/bnd/ant/RunBundlesTask.java
dhumeniuk: how would I get that from this file?

bjhargrave: I would assume the base ant Task class would have access.

biz.aQute.bnd/src/aQute/bnd/ant/RunBundlesTask.java
bjhargrave: I am not an ant programmer either. But if you can get the default for the BuildProject dir, then this is just `new File(buildProject, "bnd.bnd")`.

bjhargrave:
> This was copied from the RunconfigToDistributionTask. Maybe that wasn't the best example to use?

I don't know :-) I am not a big ant task programmer myself.

dhumeniuk:
still need to address other comments

PR #841, opened 2015-03-03
fixes 632 - BaselineProcessor test

rotty3000:
this adds adds simple workspace-less baseline processor, and workspace-less deployer

PR #873, opened 2015-03-20
fixes 632 - BaselineProcessor test

rotty3000:


rotty3000:
Comments are here: https://github.com/bndtools/bnd/pull/841

bjhargrave:
@pkriens Please investigate for including a refactor like this in 3.0.

pkriens:
I understand that the purpose of this exercise is that you can use the BaselineProcessor independent from Workspace. However, you instantiate it with an Analyzer to provide you with the context of the build (e.g. the repositories).

As we discussed, I am uncomfortable with this. This the role of the workspace. I think we need to change the implementation of the workspace. In your case, you would provide a workspace that gets the properties etc. from Ant. In the Maven case we would get the properties from the parent pom.

Though I like Project getting smaller by refactoring, I think hat the way it is done now is trying too much to make it independent of aQute.bnd.build but not really. 

So we have 2 prongs:

* Create a workspace constructor that does not require cnf. I am ok to make Workspace an abstract class and then extend it with a CnfWorkspace. In your build you can then create an AntWorkspace.
* I think the new BaselineProcessor is useful but it should use workspace/project for the appropriate information of its environment. However, this is just cosmetics

rotty3000:
Is this something you would want me to do?

pkriens:
I won't have time for 3.0 but if Ray wants to make an attempt then I can of course assist him

rotty3000:
The introduction of a flexible workspace model would greatly simplify our use case.

Wishing this to get into 3.0

bjhargrave:
Closing per discussion in bndtools call.

PR #875, opened 2015-03-20
bndlib: Correct isValid check for workspace to look for build.bnd in the...

### LINKED ISSUE #874: Workspace.isValid check is incorrect
isValid looks for build.bnd to be at the root of the workspace instead of inside the build directory (cnf).

dhumeniuk:
... cnf folder.

Also, update rest of class to use constant instead of build.bnd string literal

Fixes #874

Signed-off-by: Humeniuk, Dave <david.humeniuk@udri.udayton.edu>

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
@@ -746,9 +746,9 @@ public class Workspace extends Processor {
 		if (!buildDir.isDirectory())
 			error("No directory for cnf %s", buildDir);
 		else {
-			File build = IO.getFile(buildDir, "build.bnd");
+			File build = IO.getFile(buildDir, BUILDFILE);
 			if (build.isFile()) {
-				error("No build.bnd file in %s", buildDir);
+				error("No %s file in %s", BUILDFILE, buildDir);
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -758,7 +758,7 @@ public File getBuildDir() {
 	}
 
 	public boolean isValid() {
-		return getFile(BUILDFILE).isFile();
+		return IO.getFile(buildDir, BUILDFILE).isFile();
```
dhumeniuk: getFile used the base workspace diretory

dhumeniuk:
If there's a good place to create unit/integration tests, point me in the direction and I will try that. Just wasn't sure how all this work when the code accesses files directly. As in, is there a framework to mock out filesystem stuff for unit tests?

bjhargrave:
> If there's a good place to create unit/integration tests, point me in the direction

Did you look at https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib.tests/src/test/WorkspaceTest.java? Most of the unit tests for bndlib are in the separate bndlib.tests project.

dhumeniuk:
I'm unable to build the biz.aQute.bndlib.tests project. Says it is missing biz.aQute.bnd.test which is clearly there, as is the test folder.

image

PR #877, opened 2015-03-21
#690 custom anno 4

djencks:
Implements Peter's idea for an XMLAttribute meta-annotation.


biz.aQute.bndlib/src/aQute/bnd/annotation/xml/XMLAttribute.java
```diff
@@ -0,0 +1,54 @@
...
+@Target({
+	ElementType.ANNOTATION_TYPE
+})
+public @interface XMLAttribute {
+
+	String namespace();
+
+	String prefix() default "";
+
+	String[] embedIn();
```
bjhargrave: Could this default to "*"?

bjhargrave:
I don't really have much expertise here, but I don't have any objections to this.

djencks:
I improved the defaults processing per BJ's comment: not specifying the namespace now actually works and embedIn defaults to {"*"}

djencks:
Changed how classes are represented in the Annotation backing map to be TypeRefs and added a test that now Class valued annotation elements are processed correctly.

PR #880, opened 2015-03-24
Import bnd-maven-plugin contribution from Salesforce.com

njbartlett:
This code was developed by Paremus Ltd on behalf of a customer. That customer has instructed us to contribute the code to the bnd project under the terms of the Apache License V2.

The customer is a large, well known company and we hope to confirm their identity in due course.


bnd-maven-plugin-parent/bnd-maven-plugin/pom.xml
```diff
@@ -0,0 +1,73 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+
+	<modelVersion>4.0.0</modelVersion>
+	<parent>
+		<groupId>biz.aQute.bnd</groupId>
+		<artifactId>bnd-maven-plugin-parent</artifactId>
+		<version>2.5.0</version>
```
bjhargrave: Curious why you call it 2.5.0 instead of 2.4.1 to match the version of bndlib it uses?

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/BndMavenPlugin.java
```diff
@@ -0,0 +1,176 @@
...
+	private MavenProject project;
+	
+	@Parameter(property = "bnd.trace", defaultValue = "false", readonly = true)
+	private boolean trace;
+
+	public void execute() throws MojoExecutionException {
+		File bndFile = new File(project.getBasedir(), Project.BNDFILE);
+		
+		Builder builder = new Builder();
+		builder.setTrace(trace);
```
bjhargrave: In the gradle plugin, I set the trace value based upon the current logging level. See <https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BundleTaskConvention.groovy#L102>. Perhaps the same could be done for this to avoid the need to a separate option?

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/BndMavenPlugin.java
```diff
@@ -0,0 +1,176 @@
...
+		try {
+			if (bndFile.isFile())
+				builder.setProperties(bndFile);
```
bjhargrave: If you don't set the properties, I think you still need to set the base of the build to the project. See <https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BundleTaskConvention.groovy#L108>.

bnd-maven-plugin-parent/bnd-maven-plugin/pom.xml
njbartlett: The current version on Maven Central is 2.4.0. That was the experimental plugin that we abandoned.

This is a complete rewrite...

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/BndMavenPlugin.java
```diff
@@ -0,0 +1,176 @@
...
+				File artifactFile = artifact.getFile();
+				if (artifactFile != null)
+					classpath.add(artifactFile);
+			}
+			classpath.add(classesDir);
+			builder.setClasspath(classpath.toArray(new File[classpath.size()]));
+
+			// Include local project packages automatically
+			String includes = builder.getProperty(Constants.INCLUDERESOURCE);
+			StringBuilder newIncludes = new StringBuilder().append(classesDir.getPath());
```
bjhargrave: To be safe, you need to quote the classesDir path. See <https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BundleTaskConvention.groovy#L147>.

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/BndMavenPlugin.java
```diff
@@ -0,0 +1,176 @@
...
+			@SuppressWarnings("unchecked")
+			Set<Artifact> artifacts = project.getArtifacts();
+			for (Artifact artifact : artifacts) {
+				File artifactFile = artifact.getFile();
+				if (artifactFile != null)
+					classpath.add(artifactFile);
+			}
+			classpath.add(classesDir);
+			builder.setClasspath(classpath.toArray(new File[classpath.size()]));
+
```
bjhargrave: Can you set the sourcepath if -sources=true? See <https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BundleTaskConvention.groovy#L138>.

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/BndMavenPlugin.java
```diff
@@ -0,0 +1,176 @@
...
+			}
+
+			// Output manifest to <classes>/META-INF/MANIFEST.MF
+			Files.createDirectories(manifestPath.toPath().getParent());
+			FileOutputStream manifestOut = new FileOutputStream(manifestPath);
+			try {
+				bndJar.writeManifest(manifestOut);
+			} finally {
+				manifestOut.close();
+			}
```
bjhargrave: Not sure if you need to check to errors again after writing the manifest. @pkriens told me to check again after writing the bundle since more errors could be detected. (For example: ${system} macro failure.). See <https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BundleTaskConvention.groovy#L190>.

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/BndMavenPlugin.java
```diff
@@ -0,0 +1,176 @@
...
+
+			// Expand Jar into target/classes
+			expandJar(bndJar, classesDir);
+
```
bjhargrave: Not sure if you need to check for errors again after writing the manifest. @pkriens told me to check again after writing the bundle since more errors could be detected. (For example: ${system} macro failure.). See <https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BundleTaskConvention.groovy#L190>.

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/MavenVersion.java
```diff
@@ -0,0 +1,63 @@
...
+
+	private static final String SNAPSHOT = "SNAPSHOT";
+	private static final String TSTAMP = "${tstamp}";
+
+	private int major;
+	private int minor;
+	private int micro;
+	private String qualifier;
+	
+	public MavenVersion(String versionString) throws IllegalArgumentException {
```
bjhargrave: You may want to use the new maven version class: <https://github.com/bndtools/bnd/blob/master/biz.aQute.repository.aether/src/aQute/bnd/deployer/repository/aether/MvnVersion.java>. It handles mapping bazaar maven versions into OSGi versions quite well.  See the tests <https://github.com/bndtools/bnd/blob/master/biz.aQute.repository.aether/test/aQute/bnd/deployer/repository/aether/MvnVersionTest.java>.

njbartlett: Perhaps but it lacks the feature I wanted which was to translate SNAPSHOT into ${tstamp}

bjhargrave: OK, so copy it and add that feature. Otherwise this plugin may fail on a number of versions encountered.

In the gradle plugin, I took the position that if they are deliberately building an OSGi bundle, then the jar version better be a valid OSGi version.

bosschaert:
Great to see that this is being contributed! Thank you Salesforce and @njbartlett !

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/BndMavenPlugin.java
njbartlett: What will this actually do when bnd isn't ultimately creating the JAR?

bjhargrave: Well it will put the sources in the Jar object which you will later expand into the target/classes folder. Then the maven jar task will put them in the jar just like everything else you wrote into target/classes.

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/BndMavenPlugin.java
```diff
@@ -0,0 +1,186 @@
...
+			checkErrors(builder);
+
+		} catch (Exception e) {
+			throw new MojoExecutionException("bnd error", e);
+		} finally {
+			builder.close();
+		}
+	}
+
+	private void checkErrors(Builder builder) throws MojoExecutionException {
```
bjhargrave: You will report duplicate warnings. If there are warnings (and no errors), you call this method twice and will display the warnings twice.

njbartlett: Yes, silly me. So why don't we just check once at the end?

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/MavenVersion.java
njbartlett: It's not so easy to copy because I can't preserve "${tstamp}" inside an actual OSGi Version object.

I like the rule that the version must be parseable as an OSGi version, and I think my code achieves that.

bjhargrave: ok

bnd-maven-plugin-parent/bnd-maven-plugin/src/main/java/org/bndtools/maven/BndMavenPlugin.java
bjhargrave: That's probably fine. Checking after build() but before writing just saves some time before you fail the build.

donald-w:
Looks like the identity of the company is in the pull request title!

PR #884, opened 2015-03-25
bnd-maven-plugin: Initial drop for master branch; version 3.0.0

bjhargrave:


bjhargrave:
This is based upon the submission in https://github.com/bndtools/bnd/pull/883.

The maven plugin is built against the bndlib built in the same build.

PR #903, opened 2015-04-03
Questions

gamerson:
@pkriens  here are some questions I had while reviewing launcher code.  I was able to answer my questions after I kept reading them, but you said you wanted questions "as they came" so here you go!


pkriens:
Answered in the comments, thanks

PR #909, opened 2015-04-07
[remote] deploy agent over JMX

gamerson:


pkriens:
This looks pretty cool. Just one remark, why use "jmx" as a property? You should be able to use it from the DTO?

gamerson:
No good reason, simply didn't notice the dto already had what I needed :)
 I pushed a new commit.

On Fri, Apr 10, 2015 at 9:59 PM, Peter Kriens <notifications@github.com>
wrote:

> This looks pretty cool. Just one remark, why use "jmx" as a property? You
> should be able to use it from the DTO?
>
> —
> <https://github.com/bndtools/bnd/pull/909#issuecomment-91567548>.
>



-- 
Greg Amerson
Liferay Developer Tools
Liferay, Inc. www.liferay.com

gamerson:
Hey @pkriens 

I just added a real jmx deploy testcase, but in order to do this I had to copy into the testdata folder several bundles that I used to launch a framework and get Apache Aries up and running.  Not sure if there is a better way to do that.

One another note... when I ran the build headless it always skips the remote tests.

```
:biz.aQute.remote:testClasses
:biz.aQute.remote:test SKIPPED
:biz.aQute.remote:check SKIPPED
```

My bnd+gradle knowledge fails me here.

bjhargrave:
See https://github.com/bndtools/bnd/blob/master/biz.aQute.remote/bnd.bnd#L31 which says no junit tests in the project.

gamerson:
Thanks @bjhargrave I enabled the tests locally and find a failing test when everything is run together and have pushed a new commit.

PR #910, opened 2015-04-09
[bnd-maven-plugin] add m2e plugin lifecycle mapping support

gamerson:
cc @njbartlett

### DIFF
--- a/bnd-maven-plugin/.gitignore
+++ b/bnd-maven-plugin/.gitignore
@@ -1,1 +1,4 @@
+/.classpath
+/.project
+/.settings/
 /target/
--- a/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
+++ b/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
@@ -17,6 +17,13 @@ package aQute.bnd.maven.plugin;
  */
 
 import static aQute.lib.io.IO.getFile;
+import aQute.bnd.build.Project;
+import aQute.bnd.osgi.Builder;
+import aQute.bnd.osgi.Constants;
+import aQute.bnd.osgi.FileResource;
+import aQute.bnd.osgi.Jar;
+import aQute.bnd.osgi.Resource;
+import aQute.lib.io.IO;
 
 import java.io.File;
 import java.io.FileOutputStream;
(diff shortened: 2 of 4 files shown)


bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -57,13 +58,16 @@
...
 	@Parameter(defaultValue = "${project.build.outputDirectory}/META-INF/MANIFEST.MF", readonly = true)
 	private File manifestPath;
-	
+
 	@Parameter(defaultValue = "${project}", required = true, readonly = true)
 	private MavenProject project;
-	
+
+	@Component
+	private BuildContext buildContext;
```
bjhargrave: What is the point of this since it is not used anywhere.

gamerson: So according to this https://wiki.eclipse.org/M2E_compatible_maven_plugins plugins must use buildContext API in order to get incremental workspace build support from m2e.  I assumed that meant the API needed to get injected.  However, let me try without it and if it works I'll remove it.

bnd-maven-plugin/.gitignore
```diff
@@ -1 +1,4 @@
+/.classpath
+/.project
+/.settings/
```
bjhargrave: How is this related to the m2e change? perhaps we should put eclipse metadata under source control like other bnd projects.

bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
gamerson: Another question... is there any time that you can think of that we wouldn't want to rerun the manifest generation based on an Eclipse workspace build?  If so we can use this API.  I figured that we want to rebuild the Manifest for every single Eclipse build just to keep things simple.

bnd-maven-plugin/.gitignore
gamerson: So in order to work on the "bnd-maven-plugin" project I imported into my eclipse using m2e plugin, which fires off the project configurators and generates those files.  So some developers will ignore these files since you can get them back whenever you "import" the project using m2e.

some people want these files checked in.  Both ways work I guess, what is bndtools pref and I can follow that way.

bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
gamerson: OK I reread that wiki page and the MUST only applies to cases where plugins want to change workspace resources (we don't) or they want to skip some incremental builds based on uninteresting changes (we don't), so I'll remove that.

bnd-maven-plugin/.gitignore
gamerson: I'll add my metadata (generated by m2e) and see if that is to everyone's liking.

njbartlett: As I understand, the convention in maven projects is to NOT check in those files. This does differ from other bnd projects.   

--  
Neil Bartlett

gamerson: Correct that is my understanding as well, let all eclipse metadata come
downstream of pom.xml.  however, some Eclipse metadata can't come
downstream of pom.xml like java code style.  For now I'll pull my commit of
that metadata until you guys decide.

On Thu, Apr 9, 2015 at 5:01 PM, Neil Bartlett <notifications@github.com>
wrote:

> In bnd-maven-plugin/.gitignore
> <https://github.com/bndtools/bnd/pull/910#discussion_r28045646>:
>
> > @@ -1 +1,4 @@
> > +/.classpath
> > +/.project
> > +/.settings/
>
> As I understand, the convention in maven projects is to NOT check in those
> files. This does differ from other bnd projects.
> … <#14c9d6a0e082818b_>
> -- Neil Bartlett Sent from a phone
> On Thursday, 9 April 2015 at 09:34, Gregory Amerson wrote: In
> bnd-maven-plugin/.gitignore (
> https://github.com/bndtools/bnd/pull/910#discussion_r28043828): > @@ -1
> +1,4 @@ > +/.classpath > +/.project > +/.settings/ So in order to work on
> the "bnd-maven-plugin" project I imported into my eclipse using m2e plugin,
> which fires off the project configurators and generates those files. So
> some developers will ignore these files since you can get them back
> whenever you "import" the project using m2e. some people want these files
> checked in. Both ways work I guess, what is bndtools pref and I can follow
> that way. — Reply to this email directly or view it on GitHub (
> https://github.com/bndtools/bnd/pull/910/files#r28043828).
>
> —
> <https://github.com/bndtools/bnd/pull/910/files#r28045646>.
>



-- 
Greg Amerson
Liferay Developer Tools
Liferay, Inc. www.liferay.com

PR #918, opened 2015-04-16
Fix #917

### LINKED ISSUE #917: Invalid file name check is missing escape character
The default regular expression for -invalidfilenames contains "CLOCK$". This is because "CLOCK$" (case-insensitive) is apparently an illegal name on Windows... who knew?

Anyway, the $ needs to be escaped because it can match a file that simply ends with CLOCK.

njbartlett:

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Verifier.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Verifier.java
@@ -84,7 +84,7 @@ public class Verifier extends Processor {
 
 	public final static Pattern	ReservedFileNames				= Pattern
 																		.compile(
-																				"CON(\\..+)?|PRN(\\..+)?|AUX(\\..+)?|CLOCK$|NUL(\\..+)?|COM[1-9](\\..+)?|LPT[1-9](\\..+)?|"
+																				"CON(\\..+)?|PRN(\\..+)?|AUX(\\..+)?|CLOCK\\\\$|NUL(\\..+)?|COM[1-9](\\..+)?|LPT[1-9](\\..+)?|"
 																						+ "\\$Mft|\\$MftMirr|\\$LogFile|\\$Volume|\\$AttrDef|\\$Bitmap|\\$Boot|\\$BadClus|\\$Secure|"
 																						+ "\\$Upcase|\\$Extend|\\$Quota|\\$ObjId|\\$Reparse",
 																				Pattern.CASE_INSENSITIVE);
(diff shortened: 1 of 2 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Verifier.java
```diff
@@ -84,7 +84,7 @@ public String toString() {
 
 	public final static Pattern	ReservedFileNames				= Pattern
 																		.compile(
-																				"CON(\\..+)?|PRN(\\..+)?|AUX(\\..+)?|CLOCK$|NUL(\\..+)?|COM[1-9](\\..+)?|LPT[1-9](\\..+)?|"
+																				"CON(\\..+)?|PRN(\\..+)?|AUX(\\..+)?|CLOCK\\\\$|NUL(\\..+)?|COM[1-9](\\..+)?|LPT[1-9](\\..+)?|"
 																						+ "\\$Mft|\\$MftMirr|\\$LogFile|\\$Volume|\\$AttrDef|\\$Bitmap|\\$Boot|\\$BadClus|\\$Secure|"
```
bjhargrave: Looking at your fix, does this mean the other expressions with $ are also wrong? They only have \\$ while your CLOCK$ fix has \\\\$.

bjhargrave: I tested `"CLOCK\\$"` at http://www.regexplanet.com/advanced/java/index.html which seems to be the correct value. `"CLOCK\\\\$"` did not match `CLOCK$` test value

njbartlett: The quadruple-backslash was necessary because the string gets passed through the macro parser... it's some nasty logic involved with the expansion of `${@}` in Verifier.

Whereas the other expressions I think are intended to be macros? I really don't know.

bjhargrave: I guess we need input from @pkriens since there should be consistency here. `CLOCK$` is  file name just like `$Mft`.

The issue may be in the way the test case is written.

njbartlett: Actually I think the logic is pretty broken. On line 85 of Verifier, the literal string is compiled to a Pattern... then we take the string representation of the pattern, macro-interpolate it, then recompile the result as a Pattern again. I think we should skip the first compile at least.

njbartlett: +1 for waiting for clarity from @pkriens

pkriens:
BJ is right. We must only use 1 \\ because the Pattern class must treat the $ as a literal. Escaping it with two times \\ will make the backslash a literal, so the invalid name would be CLOCK\ and not CLOCK$
I've created a new pull request for this and merged, so this pull request will be closed.

Issue #925, opened 2015-04-30
Automatically Create target/classes directory if it does not exist


skahmann:
If you do not have a src directory then the build fails because there will be no target/classes directory. See: https://groups.google.com/forum/#!topic/bndtools-users/ishII0W0rMY

PR #926, opened 2015-04-30
bnd-maven-plugin: Handle projects with no source code

bjhargrave:
The project might just want to wrap some other jars. This fix will test
for the target/classes dir and if it is not a directory, it will not
add it to bnd's classpath or the -includeresource directive.

Also added a new test project to the integration-test to test building
a project with no source code.

Fixes https://github.com/bndtools/bnd/issues/925

### DIFF
--- a/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
+++ b/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
@@ -94,21 +94,25 @@ public class BndMavenPlugin extends AbstractMojo {
 				if (artifactFile != null)
 					classpath.add(artifactFile);
 			}
-			classpath.add(classesDir);
+			if (classesDir.isDirectory()) {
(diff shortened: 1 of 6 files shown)


skahmann:
The classes we need (which we are wrapping) are being stored into the target/classes directory. Will they still be bundled into the jar?

bjhargrave:
If the target/classes directory exists when the bnd-maven-plugin executes, everything in the target/classes directory will be included in the jar (via the -includeresource directive).

Please make sure to test these changes to see if they work for you.

bjhargrave:
> The classes we need (which we are wrapping) are being stored into the target/classes directory.

Also, this statement seems to contradict #925 where you state that target/classes does NOT exist.

skahmann:
It seems to be creating the target/classes directory for the included dependencies after the check.

bjhargrave:
I am not sure what you mean. Is it working ok? Or do you still have a problem? If there is still a problem, please open a new issue with a detailed description and instructions on how to reproduce.

Issue #927, opened 2015-05-01
Failed bndlib tests on windows


marcdejonge:
There are a couple of failed tests when I run junit 3 on bndlib.tests.

There are 2 tests in `BuilderTest` that try to use the user path as an `URI` by doing:
```
new URI(pwd).toString()
```
This fails on a windows machine because a path such as `C:\Users\...` is not a valid URI. I will propose a quick fix for this

The second problem is a bit harder, but in involved the `CommandResource`. On [this line](https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/osgi/CommandResource.java#L25) the command `sh` in invoked, which of course will never work on a normal windows machine (without cygwin or someting similar). Because of this there are 3 tests in `ResourceTest` that fail.

Should the `CommandResource` ever work on a Windows machine? Or should we just ignore these? In that case I do propose that we give a clear warning when the `CommandResource` is used on a windows machine that this won't work.

pkriens:
Disable windows tests

bjhargrave:
We now have bnd building and testing on Windows using the AppVeyor CI system: <https://ci.appveyor.com/project/bjhargrave/bnd>. This does not exhibit any failures in CommandResource. So perhaps this issue can be closed?

bjhargrave:
Why do we need to disable these tests? They seem to be passing in the appveyor CI builds.

pkriens:
AppVeyor has this unix shell installed …

I do not think it is worth it to fine-tune this …

Kind regards,

	Peter Kriens

PR #928, opened 2015-05-01
Fixed 2 testcases in the BuilderTest that failed on Windows

marcdejonge:
These failures were because a Windows path is not a valid URI.
Using the new File(...).toUri() does work in all cases.

Signed-off-by: Marc de Jonge <marcdejonge@gmail.com>


marcdejonge:
Fixes the first 2 problems mentioned in #927

Issue #938, opened 2015-05-07
Feature request: parameterise RequireCapability filter using annotation attributes


njbartlett:
I would like to define an annotation as follows:

    @RequireCapability(ns = "foobar", filter = "(foobar=@value)")
    public @interface MyAnnotation { String value(); }

Such that using the annotation with a value attribute like this:

    @MyAnnotation("hello")

will result in a generated requirement like this:

    Require-Capability: foobar; filter:="(foobar=hello)"

The syntax for the placeholder is not important to me. We could reuse macro syntax but it may not be sensible to use the full power of macros here.

bjhargrave:
We discussed this basic topic at OSGi this week during a discussion about [RFP 167](https://github.com/osgi/design/blob/master/rfps/rfp-0167-Manifest-Annotations.pdf). So I expect something like this will be in the RFC.

njbartlett:
Fixed in a54a1b7

PR #940, opened 2015-05-08
Fix #937: @RequireCapability with 'value' attribute generates spuriou…

### LINKED ISSUE #937: @RequireCapability with 'value' attribute generates spurious entry
Define an annotation as follows:

    @RequireCapability(ns = "nsy", effective = "active", filter = "(foo=bar)")
    public @interface AnnotationWithValue {
	    String value();
    }

And use it on a component as follows:

    @AnnotationWithValue("hello")

The resulting `Require-Capability` header is:

    Require-Capability:
        nsy; value=hello; filter:="(foo=bar)"; effective:=active,
        hello; value=hello; filter:="(foo=bar)"; effective:=active

The first line with the `nsy` namespace is expected but the `hello` namespace is incorrect.

njbartlett:
@pkriens Please review, especially the removal of lines 308-310 of `AnnotationHeaders.java`, which were the direct cause of the bug. I don't understand why those lines were there in the first place, maybe you can recall the reason you wrote them.

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/annotationheaders/AnnotationHeadersTest.java
+++ b/biz.aQute.bndlib.tests/src/test/annotationheaders/AnnotationHeadersTest.java
@@ -52,14 +52,19 @@ public class AnnotationHeadersTest extends TestCase {
 			// namespaces must be "osgi.ee", "nsx" and "nsy" ONLY
 			assertNotNull(p.get("nsx"));
 			assertNotNull(p.get("nsy"));
+			assertNotNull(p.get("nsz"));
 			assertNotNull(p.get("osgi.ee"));
-			assertEquals("spurious Require-Capability namespaces", 3, p.size());
+			assertEquals("spurious Require-Capability namespaces", 4, p.size());
 
 			attrs = p.get("nsx");
 			assertEquals(Arrays.asList("abc","def"), attrs.getTyped("foo"));
 
 			attrs = p.get("nsy");
 			assertEquals("hello", attrs.get("value"));
+
+			attrs = p.get("nsz");
+			assertEquals("(nsz=*)", attrs.get("filter:"));
+			assertEquals("world", attrs.get("hello"));
 		}
 		finally {
 			b.close();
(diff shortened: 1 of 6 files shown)


pkriens:
The purpose of the value attribute was to allow the Require/Provide capability to just set the remainder of the clause, unchecked. I.e. this would allow directives and other tricks that the standard syntax does not allow. I guess this is ok for the Require/Provide Capability itself but should not be used for the meta annotations ...

Not sure if we should kill this feature for the Require/Provide Capability annotation itself?

njbartlett:
Okay I'm not sure I understand the difference between putting this on the user-defined annotation or the meta-annotation. However, if the feature exists at all it should at least work, i.e. not create random unsatisfiable requirements in the bundle...

pkriens:
Why would it create random unsatisfiable requirements? The value was just suffixed after the namespace so the author could do whatever he wanted, e.g. if you wanted to use a directive?

pkriens:
But I am ok to move that behavior to another named attribute though

njbartlett:
"Why would it create random unsatisfiable requirements?"

Umm... because there's a bug!

pkriens:
Why does RequireCapability(ns="foo", value="bar:=12") generate a bug?

You have not indicated (at least I do not see it) what the bug is that you're trying to solve? I assume(d) the bug is that if the meta annotation has 'value' you get shit. So I proposed to ONLY use the 'value' method from the RequireCapability annotation for this purpose. The annotated annotation's 'value' method should not leak upwards.

njbartlett:
Okay seems we've been talking at cross purposes. Yes I'm fine with that solution, as long as it's compatible with the use-case I want to implement in #938

pkriens:
So why not change the name of the value property used for the concatenation and then resubmit? The value property is not right anyway since you cannot really use it standalone anyway

njbartlett:
`value` can be used standalone... it's a "magic" name in the Java language:

    @MyAnnotation("hello")

is equivalent to

    @MyAnnotation(value = "hello")

pkriens:
Yes, so I suggested to rename the 'value' property in the  RequireCapability annotation to something like 'extra' so that it does not clash with this use.

njbartlett:
Ah okay. Yes that sounds good.

PR #945, opened 2015-05-19
[remote] wait until flush() is called to send output across the wire

### LINKED ISSUE #933: [remote] launcher CommandSession/GogoRedirector output can be slow
When using -runremote and specifying the shell to use "CommandSession/GogoRedirector" (specify the shell=-1) if I send a command in the eclipse console like "lb" it will return the output line by line, where it takes maybe 15-20 seconds to output a standard "lb" command with maybe 100 bundles.

gamerson:
Fixes #933

### DIFF
--- a/biz.aQute.remote/src/aQute/remote/agent/RedirectOutput.java
+++ b/biz.aQute.remote/src/aQute/remote/agent/RedirectOutput.java
@@ -68,7 +68,8 @@ public class RedirectOutput extends PrintStream {
 	}
 
 	public void flush() {
-		super.flush();
+		final String output = sb.toString();
+		sb = new StringBuilder();
 
 		for (AgentServer agent : agents) {
 			if (agent.quit)
@@ -76,9 +77,9 @@ public class RedirectOutput extends PrintStream {
 
 			try {
 				if (err)
-					agent.getSupervisor().stderr(sb.toString());
+					agent.getSupervisor().stderr(output);
 				else
-					agent.getSupervisor().stdout(sb.toString());
+					agent.getSupervisor().stdout(output);
 			}
 			catch (InterruptedException ie) {
 				return;
@@ -93,7 +94,7 @@ public class RedirectOutput extends PrintStream {
 			}
 		}
 
-		sb = new StringBuilder();
+		super.flush();
 	}
 
 	public void close() {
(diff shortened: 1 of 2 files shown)


biz.aQute.remote/src/aQute/remote/agent/Shell.java
```diff
@@ -49,6 +49,7 @@ public synchronized void add(String s) throws IOException {
 			running = true;
 			try {
 				Object result = session.execute(s);
+				this.out.flush();
```
bjhargrave: What about flush after the println and printStackTrace below? Perhaps the flush call should be at the bottom of the loop before the call to prompt?

biz.aQute.remote/src/aQute/remote/agent/RedirectOutput.java
```diff
@@ -85,6 +69,31 @@ public void write(byte b[], int off, int len) {
...
+				try {
+					agent.close();
+				}
+				catch (IOException e) {
+					// e.printStackTrace();
+				}
+			}
+		}
+
+		sb = new StringBuilder();
```
bjhargrave: Perhaps you should convert the StringBuffer to a local String (and use in the loop) and reset sb before the for loop?

biz.aQute.remote/src/aQute/remote/agent/RedirectOutput.java
```diff
@@ -85,6 +69,31 @@ public void write(byte b[], int off, int len) {
 
 	public void flush() {
 		super.flush();
```
bjhargrave: I am not sure of the proper order here. Should the new code be executed before super.flush()?

gamerson: now that you mention it, I think before super.flush() makes more sense.

biz.aQute.remote/src/aQute/remote/agent/RedirectOutput.java
gamerson: Done

biz.aQute.remote/src/aQute/remote/agent/Shell.java
gamerson: agree

PR #950, opened 2015-05-26
[remote][agent] move package to private

gamerson:


bjhargrave:
See mail thread at https://groups.google.com/forum/#!topic/bndtools-users/EUG7t7wQm4g for more background.

PR #953, opened 2015-05-26
Parameterise @RequireCapability filter using annotation attributes

njbartlett:
Example:

    @RequireCapability(ns="foo", filter="(foo=${value})")
    public @interface Foo { String value(); }

    @Foo("hello")
    public class MyComponent { ... }

    ==> Require-Capability: foo; filter:="(foo=hello)"

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
@@ -5,8 +5,11 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import aQute.bnd.annotation.headers.BundleCategory;
 import aQute.bnd.annotation.headers.BundleContributors;
@@ -57,6 +60,9 @@ import aQute.lib.strings.Strings;
  * {@link About} provides some more information.
  */
 class AnnotationHeaders extends ClassDataCollector implements Closeable {
+
+	static final Pattern SIMPLE_PARAM_PATTERN = Pattern
+			.compile("\\$\\{(\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*)\\}");
 
 	final Analyzer					analyzer;
 	final Set<TypeRef>				interesting	= new HashSet<TypeRef>();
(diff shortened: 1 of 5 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
```diff
@@ -315,6 +320,31 @@ private void doRequireCapability(Annotation a) throws Exception {
 		Attrs attrs = getAttributes(a, "ns");
 		directivesAndVersion(attrs, "filter", "effective", "resolution");
 		
+		for (Entry<String,String> entry : attrs.entrySet()) {
+			String value = entry.getValue();
+
+			Matcher matcher = SIMPLE_PARAM_PATTERN.matcher(value);
+			while (matcher.find()) {
```
bjhargrave: I think you want to use a loop like:

    StringBuffer result = new StringBuffer();
     while (m.find()) {
        m.appendReplacement(result, attrs.get(m.group()));
    }
    entry.setValue(m.appendTail(result).toString());

It is more efficient.

biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
```diff
@@ -58,6 +61,8 @@
  */
 class AnnotationHeaders extends ClassDataCollector implements Closeable {
 
+	static final Pattern SIMPLE_PARAM_PATTERN = Pattern.compile("\\$\\{(\\w*)\\}");
```
bjhargrave: Shouldn't this be: `\\$\\{(\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*)\\}` since it (1) requires at least one character and (2) has an alphabet which matches the alphabet of the names of annotation elements?

njbartlett:
Amended

biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
```diff
@@ -315,6 +321,27 @@ private void doRequireCapability(Annotation a) throws Exception {
...
+				modified = true;
+				String key = matcher.group(1);
+				String substitution = attrs.get(key);
+
+				if (SIMPLE_PARAM_PATTERN.matcher(substitution).find())
+					throw new IllegalArgumentException("nested substitutions not permitted");
+
+				matcher.appendReplacement(sb, substitution);
+			}
+			matcher.appendTail(sb);
```
bjhargrave: You could move the `appendTail` call inside the `if (modified)` test since there is no need to do it otherwise.

Issue #141, opened 2012-03-04
Support start levels and eager activation in launcher


njbartlett:
Unfortunately we sometimes need support for:

a) Specifying start levels;
b) Eagerly starting bundles even when they are marked as `Bundle-ActivationPolicy: lazy`.

Proposal: add directives to the -runbundles instruction. Also required is a single property to set the target start level of the framework. For example:

```
-runbundles: org.example.foo;version=latest;start:=eager;level:=3
-runlevel: 4
```

kriegfrj:
> I agree, let's do this [force-starting bundles] with a bundle.

@njbartlett, I know that this issue is pretty old, but I have hit up against this issue of needing to force-start bundles, and I have been searching in vain for a solution for this in bnd, and I simply keep coming back to this page.

Is this bundle that force starts bundles something that you developed in-house, or is it publicly available? If the latter would you be so kind as to put a post here with pointing readers like me (who have followed the trail this far) in the right direction? I know it shouldn't be too difficult to develop something like this from scratch, but I'd rather not re-invent the wheel.

njbartlett:
The "caffeine" bundle from this repo is probably what you are looking for: https://github.com/bndtools/bndtools-rt

I haven't touched the code for a while so I don't know how easy it will be to build. Then again, the code of the bundle itself is trivial.

kriegfrj:
Thanks @njbartlett, that will certainly get me started (and hopefully anyone else who follows this post after me!)

laeubi:
I found this thread while searching for the meaning of the property `launch.activation.eager` I see in the debug output of the launcher, was something like this added in the meantime?
If not, equinox-launcher behaves like this:

1. nothing specified = just behave as the default
2. `start=true` always start
3. `start=false` never start

probably BND could do something similar, having an extra bundle for this seems a bit cumbersome.

PR #963, opened 2015-06-01
Keep startup order from bndrun when launching remote

tverbele:
In order to make sure that the startup order of the bundles specified in the bndrun configuration is preserved when launching remote, a LinkedHashMap should be used in the Agent API. Peter already changed the implementation in commit b7285748eda20bcdbab0a632d7b4ef3465c99d05 , but this is not sufficient as the Map is serialized and deserialized by the converter in the remote call.

For the standard launcher startup order is preserved (also see issue #141) so it would be nice to have the same behaviour for the remote launcher.

### DIFF
--- a/biz.aQute.remote/src/aQute/remote/agent/AgentServer.java
+++ b/biz.aQute.remote/src/aQute/remote/agent/AgentServer.java
@@ -150,19 +150,19 @@ public class AgentServer implements Agent, Closeable, FrameworkListener {
 	}
 
 	@Override
-	public String update(Map<String,String> bundles) throws InterruptedException {
+	public String update(LinkedHashMap<String,String> bundles) throws InterruptedException {
 
 		refresh.await();
 
 		Formatter out = new Formatter();
 		if (bundles == null) {
-			bundles = Collections.emptyMap();
+			bundles = new LinkedHashMap<String,String>();
 		}
 
 		Set<String> toBeDeleted = new HashSet<String>(installed.keySet());
 		toBeDeleted.removeAll(bundles.keySet());
 
-		Set<String> toBeInstalled = new HashSet<String>(bundles.keySet());
+		Set<String> toBeInstalled = new LinkedHashSet<String>(bundles.keySet());
 		toBeInstalled.removeAll(installed.keySet());
 
 		Map<String,String> changed = new HashMap<String,String>(bundles);
--- a/biz.aQute.remote/src/aQute/remote/api/Agent.java
+++ b/biz.aQute.remote/src/aQute/remote/api/Agent.java
@@ -108,12 +108,13 @@ public interface Agent {
 	 * differences are reflected in the installed bundles. That is, a change in
 	 * the SHA will update, a new entry will install, and a removed entry will
 	 * uninstall. This is the preferred way to keep the remote framework
-	 * synchronized since it is idempotent.
+	 * synchronized since it is idempotent. Bundles are passed as a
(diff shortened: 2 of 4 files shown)


biz.aQute.remote/src/aQute/remote/agent/AgentServer.java
```diff
@@ -150,19 +150,19 @@ public String uninstall(long... ids) {
 	}
 
 	@Override
-	public String update(Map<String,String> bundles) throws InterruptedException {
+	public String update(LinkedHashMap<String,String> bundles) throws InterruptedException {
```
bjhargrave: It seems wrong to put an impl specific map type in the API. I understand the implementation will need to use a specific map impl, but the API should not use a specific impl map type.

biz.aQute.remote/src/aQute/remote/api/Agent.java
```diff
@@ -108,12 +108,13 @@
...
 	 * uninstall. This is the preferred way to keep the remote framework
-	 * synchronized since it is idempotent.
+	 * synchronized since it is idempotent. Bundles are passed as a
+	 * LinkedHashMap to preserve the startup order.
 	 * 
 	 * @param bundles
 	 *            the bundles to update
 	 */
-	String update(Map<String,String> bundles) throws Exception;
+	String update(LinkedHashMap<String,String> bundles) throws Exception;
```
bjhargrave: It seems wrong to put an impl specific map type in the API. I understand the implementation will need to use a specific map impl, but the API should not use a specific impl map type.

tverbele:
The problem is that the underlying converter will guess the type from the JSON, and thus will just instatiate a HashMap instead of the (in this case) whished LinkedHashMap. Another option would be to send the concrete type in exchanged JSON messages (thus patch JSONCodec)?

bjhargrave:
> thus will just instatiate a HashMap instead of the (in this case) whished LinkedHashMap

Perhaps the converter should always use LinkedHashMap as the default Map implementation to preserve the order in the JSON data?

pkriens:
I agree with BJ. The implementation of the Agent Supervisor now already uses a LinkedHashmap (I made this change) and the receiver of the communication code always creates linked hash maps. So the order should already be ok?

I have changed aQute.remote.plugin.RunSessionImpl.getBundles(Collection<String>, String) to use LinkedHashMap. Theoretically this should give ordering.

Can we close this?

pkriens:
The changes I made last week might not have made it until today ...

bjhargrave:
> The changes I made last week might not have made it until today ...

I assume you mean https://github.com/bndtools/bnd/commit/b7285748eda20bcdbab0a632d7b4ef3465c99d05

pkriens:
Yup

tverbele:
>and the receiver of the communication code always creates linked hash maps. So the order should already be ok?

Is this so? In the `Link` class the command parameters are parsed as follows:

`parameters[i] = codec.dec().from(args.get(i)).get(m.getGenericParameterTypes()[i]);`

If the `update` method in Agent has a generic Map type specified, this type will be explicitly given to the Decoder which will not return a LinkedHashMap but a HashMap implementation.

pkriens:
You’re right, but that is a bug in the JSON Codec. The default should be LinkedHashMap and it is now HashMap.

I’ve changed the JSON Codec.

Kind regards,

	Peter Kriens

tverbele:
Ok thanks! Then this can be closed.

Issue #961, opened 2015-06-01
Allow generation of multiple manifest headers from a single annotation


pkriens:
Currently you can only annotate a type once with a given annotation. This makes it impossible to generate multiple Require-Capability headers from a single annotation. We might want to look at solving this.

I think Java 8 already supports multiple annotations per target?

bjhargrave:
OSGi's design for bundle annotations will support Repeatable annotations. See <https://github.com/osgi/design/tree/master/rfcs/rfc0220>.

PR #968, opened 2015-06-02
make ProvideCapability and RequireCapability annotations repeatable

tverbele:
Use the `@Repeatable` annotation allow multiple ProvideCapability and RequireCapability annotations (#961). Probably this also requires changes to the annotation processing to also take into account the container annotation types?


tverbele:
Only works for Java 8 though...

bjhargrave:
So I guess we close this PR?

tverbele:
Yes, I guess there is no way then for now to include two require-capabilities with a single annotation...

PR #986, opened 2015-06-22
Support UUID class in Converter and JSONCodec

tverbele:
I regularly use UUIDs as identifiers, and when using for example the REST provider, I also want to use UUIDs there as arguments or attributes from my returning (DTO) class. To support this, I added an UUID handler to the JSONCodec to convert UUID from/to String, and also add a conversion in the Converter class when resultType is UUID.class.


bjhargrave:
LGTM.

bjhargrave:
FYI. I updated https://github.com/osgi/design/blob/master/rfps/rfp-0169-Object-Conversion.pdf to include UUIDs.

PR #995, opened 2015-07-10
Fix problem parsing OSGi headers with an explicit empty attribute at …

gnodet:
…the end.

For example:
  foo;a=\”\”
This is a valid header, but rejected by bndlib.

### DIFF
--- a/aQute.libg/src/aQute/libg/qtokens/QuotedTokenizer.java
+++ b/aQute.libg/src/aQute/libg/qtokens/QuotedTokenizer.java
@@ -80,7 +80,7 @@ public class QuotedTokenizer {
 		if (!hadstring)
 			result = result.trim();
 
-		if (result.length() == 0 && index == string.length())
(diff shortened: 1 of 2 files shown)


gnodet:
The original issue is the following: https://issues.apache.org/jira/browse/FELIX-4691
While it could be fixed by removing the attribute when the Provide-Capability header is generated, I don't see why we should go that way given the output is a valid OSGi header afaik.

pkriens:
Why are you using backslashes?

gnodet:
Sorry, this is a copy/paste from the one of my test, so it's a java string, hence the backslashes to escape the quotes inside the string.  There's no backslashes in the actual string.
The point is that if the header looks like
  foo;a=
it's an invalid header and the token should be null, but if the header looks like
  foo;a=""
or
  foo;a=''
then the header has a valid attribute a which is an empty string and the token should be the empty string, not null.

Issue #521, opened 2014-05-31
Make aQute.junit run as plain bundle


pkriens:
Currently aQute.junit runs as an "embedded bundle" and is not a plain bundle. This was likely done not to influence the set of bundles in a test. However, since it is not a bundle but on the classpath, it must export JUnit. This is causing constraints, i.e. aQute.junit must be updated to have the latest version of JUnit. 

aQute.junit must get a variation that runs as a bundle and that does not carry JUnit but only imports it.

PR #1015, opened 2015-07-31
New tester

pkriens:
The biz.aQute.junit testing placed itself on the class path so it was no bundle. This required it to contain JUnit which was awkward, especially since Hamcrest was old and split.

So this patch creates a new tester (and made it the default) that runs as a bundle.


bjhargrave:
Fixes https://github.com/bndtools/bnd/issues/521

PR #1033, opened 2015-08-17
Major cleanup of the resolver code

pkriens:


pkriens:
Went over this with Tim Ward

PR #1034, opened 2015-08-18
Ensure that the resolver bundle binds to OSGi version 4.3, even thoug…

timothyjward:
…h it uses some version 6.0 types

As can be seen on lines 22 and 23 of the original file, this code was not supposed to force us to use a newer OSGi version. The missing ,\ on line 29 didn't help...


bjhargrave:
I think we can close this since Peter removed the Core R6 dependency in https://github.com/bndtools/bnd/commit/7003d9cc0220b1e415a7ab343c95874bc3f5d64e.

PR #1044, opened 2015-08-22
Multiple repos in repo filter

njbartlett:
Support multiple repo patterns in the repo pattern, e.g.:

    -buildpath: *; repos=A|B


bjhargrave:
Why don't you just do `-buildpath: *;repo=A,*;repo=B`?

Also `|` is not the usual separator. Normally a `,` is used: `-buildpath: *;repo="A,B"`.

njbartlett:
Well I wanted to avoid the quoting if comma is used. But I like the idea of specifying `repo=` multiple times.

njbartlett:
...except that won't work because the parsing in OSGiHeader doesn't support repeated attribute names. Fine, I'll use the comma.

PR #1047, opened 2015-08-23
Match repo resources using SHA-256 hash

njbartlett:
Feedback on this patch is requested! 

Resources in repos can be referenced by hash rather than bsn and version range. For example:

    -buildpath: 7fe83bfd5999fa4ef9cec40282d5d67dd0ff3303bac6b8c7b0e8be80a821441c; version=hash

Currently it works only with repos implementing the OSGi Repository Service spec, and is implemented as a `findProviders` call in the `osgi.content` namespace.

Additionally the user can specify the *expected* bsn of the resource. An error is raised if the actual identity of the resource with the specified hash does not match what was expected:

    -buildpath: 7fe83bfd5999fa4ef9cec40282d5d67dd0ff3303bac6b8c7b0e8be80a821441c;\
        version=hash;\
        bsn=org.apache.felix.configadmin


njbartlett:
Modified so that the bsn is in the normal place, and the hash is specified in a separate attribute. For example:

    -buildpath: slf4j.api; version=hash; hash=7fe83bfd5999fa4ef9cec...

The hash algorithm can be specified as a prefix of the hash string:

    -buildpath: slf4j.api; version=hash; hash=SHA-256:7fe83bfd5999fa4ef9cec...

The default is SHA-256 and in fact only SHA-256 will currently work since OSGi Repository indexes always use it. However the syntax is supported for future use.

PR #1053, opened 2015-08-25
Bug fixes

pkriens:
See commits


biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReqBuilder.java
```diff
@@ -464,6 +489,7 @@ private Object toVersions(Object value) {
 
 		if (value instanceof String)
 			try {
+				System.out.println("Got " + value);
```
bjhargrave: @pkriens I don't think you mean to leave this println in the final code.

PR #1054, opened 2015-08-25
native_capability support for Windows 10

seanbright:
Signed-off-by: Sean Bright <sean@malleable.com>


bjhargrave:
cloudbees is acting weird. The builds cannot access cloudbees to download the gradle plugin. We'll have to restart the test builds later when cloudbees is happy again.

seanbright:
I've prodded it a few times and it still looks wonky. Let me know when it's fixed and I will push again.

bjhargrave:
You don't need to repush. We can restart the builds from the travis and appveyor websites. I just restarted them and they both were able to download the gradle plugin. So the issue with cloudbees seems to be resolved.

seanbright:
Even better. Thanks.

PR #1087, opened 2015-09-11
Allow the indexer maven plugin to limit scopes

timothyjward:
The indexer should only index compile and runtime dependencies by default. Other scopes may be added using configuration if desired.


timothyjward:
Github screwed up the order of commits in this PR - closing to recreate it. Hopefully that will fix it!

PR #1089, opened 2015-09-12
Allow the indexer maven plugin to limit scopes

timothyjward:
* The indexer should only index compile and runtime dependencies by default. Other scopes may be added using configuration if desired
* The output index should be attached to the project so that it can be installed and deployed
* The output index should also include a `.gz` version for more efficient file serving from remote repositories

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,9 +15,9 @@ install:
 
 script:
   - ./gradlew --continue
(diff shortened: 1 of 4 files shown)


timothyjward:
This is still messed up - I'll delete and try with a different branch name :(

PR #1099, opened 2015-09-13
Copy the bnd-indexer-maven-plugin to next for release 3.0.1

timothyjward:


timothyjward:
Note that this contains all plugin changes up to, but not including, c25ea500689a255cbb4e42ea264a795099be9f77. If we do want to do a 3.0.1 release of the plugins then we will need to pull in some more commits for bug fixes

PR #1117, opened 2015-09-16
Fix bndtools/bndtools#1177: error resolving EE caps

njbartlett:
The system packages from the EE and `-runsystempackages` omitted any version, so they could not resolve against a filter such as `(&(osgi.wiring.package=javax.xml.parsers)(version>=0.0.0))`.

If this patch works I would like it to be a candidate for inclusion in 3.0.1.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java
@@ -308,6 +308,9 @@ public class ResourceBuilder {
 	public void addExportPackage(String packageName, Attrs attrs) throws Exception {
 		CapReqBuilder capb = new CapReqBuilder(resource, PackageNamespace.PACKAGE_NAMESPACE);
 		capb.addAttributesOrDirectives(attrs);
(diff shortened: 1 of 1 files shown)


bjhargrave:
I would say the issue is on the indexing side. Why does the requirement filter get generated with a version test when there is no version (or version = 0)? That should not happen.

njbartlett:
Perhaps. But if somebody explicitly imports as version>=0.0.0 then that should surely match an export that does not specify its version.   

That is to say, an unspecified version on an export is equivalent to version 0.

njbartlett:
I have checked with the 2.4 source code. Prior to Peter's big refactoring, we solved this problem in exactly the same way as the patch. I.e., we always created the EE package capabilities with version 0.

pkriens:
A workaround is relatively easy. I've prepared two files for 1.7 and 1.8 (and if needed can make one for 1.6) with the patched packages:

Java 1.8 https://gist.github.com/pkriens/a0b91406b7e1edb89900
Java 1.7 https://gist.github.com/pkriens/c6e883e144fd2465157b

You can place these files in cnf/ext, this will patch all resolves. You can also include this in a bnd file:

   -include: packages-1.7.bnd

PR #1127, opened 2015-09-22
Added size limit to base64 macro

njbartlett:
Turned out this macro was pretty dangerous unless you had a way to limit the size...


pkriens:
Could you do the check on the byte[] instead of just reading the file? If you have a really large file it might cause Out Of Memory errors?

pkriens:
Neil??

pkriens:
Been superseded by #1242

PR #1136, opened 2015-09-29
mappings for member names in XMLAttribute

djencks:


biz.aQute.bndlib/src/aQute/bnd/annotation/xml/XMLAttribute.java
```diff
@@ -32,12 +35,38 @@
...
 	};
 
+	/**
+	 * Array of Strings of the form key=value where the key is the name of a
+	 * member and the value is the xml attribute name to use for that member's
+	 * value. For instance, "value=simple" causes @Simple("foo") to emit
+	 * ns:simple="foo".
+	 * 
+	 * @return Array of member-name to xml-attribute-name mappings.
+	 */
```
bjhargrave: @since 1.1

bjhargrave:
Also need Signed-off-by in commit message.

bjhargrave:
LGTM

djencks:
This adds 
{code}
	/**
	 * Array of Strings of the form key=value where the key is the name of a
	 * member and the value is the xml attribute name to use for that member's
	 * value. For instance, "value=simple" causes @Simple("foo") to emit
	 * ns:simple="foo".
	 * 
	 * @return Array of member-name to xml-attribute-name mappings.
	 */
	String[] mapping() default {};
{code}
to XMLAttribute, along with the supporting code.

PR #1151, opened 2015-10-12
Add capability to override spec attributes with extensions.

djencks:
For DS, make use of an extension attribute force the 1.1 version to provide a namespace-aware xml document.

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
+++ b/biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
@@ -2860,6 +2860,49 @@ public class DSAnnotationTest extends BndTestCase {
 
 	}
 
+	@XMLAttribute(namespace = "override", prefix = "", embedIn = "http://www.osgi.org/xmlns/scr/*")
+	@Retention(RetentionPolicy.CLASS)
+	@Target({
+		ElementType.TYPE
+	})
+	@interface TestSpecReplace {
+		boolean immediate() default false;
+
+		boolean enabled() default true;
+	}
+
+	@TestSpecReplace
+	@Component(immediate = true, enabled = false)
+	public static class ReplacedAttributes implements Serializable, Runnable {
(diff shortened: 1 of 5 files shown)


biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
```diff
@@ -2860,6 +2860,49 @@ public void testPrefixCollisionExtraAttributesDefaultPrefix() throws Exception {
...
+		ElementType.TYPE
+	})
+	@interface TestSpecReplace {
+		boolean immediate() default false;
+
+		boolean enabled() default true;
+	}
+
+	@TestSpecReplace
+	@Component(immediate = true, enabled = false)
```
bjhargrave: I can see wanting to maybe override the defaults but it seems wrong to override specified values. This seems very surprising. I specified immediate=true but get false because of some other annotation specified elsewhere. I don't like that.

djencks: Step back for a second.  You can already specify spec attributes directly through the spec annotations.  The only possible purpose I can think of for overriding them is if you want an easy way to  fill them in with a constant string.  If you want some other value, you will just use the spec annotations directly.  However I don’t think you can specify “final” on an annotation to indicate that it’s intended to be a constant.

Given this extremely limited use, I don’t see the point in the extreme code complexification involved in trying to make sure that explicitly specified values aren’t overwritten.

thanks
david jencks

bjhargrave: I don't see the point in adding a feature which will have surprising results. I say immediate=true right there in the source code, yet I don't get true because some other annotation overruled me.

I can certainly see the case where some other annotation can alter defaults but not override my explicit declaration.

bjhargrave:
Discussed in bndtools call Agreed to close.

PR #1152, opened 2015-10-12
Preserve bundle order from bndrun file when launching remote

tverbele:
In order to preserve the bundle ordering when launching remote a LinkedHashMap implementation is given to the update() method of the Agent.

However, internally the ordering is lost again by using HashSets for toBeStarted and toBeInstalled. Also changing these to LinkedHashSets saves the bundle ordering. 

I added a junit test to verify this.

And yes, I *know* that one shouldn't depend on bundle ordering :-)


pkriens:
Hmm, ordering??????

PR #1155, opened 2015-10-14
Allow the file pattern used in LocalIndexRepo to be customised

eli-jordan:
Some bundles don't use the standard naming convention for osgi bundles, which prevents them from being indexed by the LocalIndexedRepo. This pull request allows the pattern to be customised.

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
+++ b/biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
@@ -1,46 +1,82 @@
 package aQute.bnd.deployer.repository;
 
-import java.io.*;
-import java.net.*;
-import java.security.*;
-import java.util.*;
-import java.util.regex.*;
+import java.io.File;
+import java.io.FileFilter;
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URI;
+import java.security.DigestInputStream;
+import java.security.MessageDigest;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
-import org.osgi.service.coordinator.*;
-import org.osgi.service.log.*;
+import org.osgi.service.coordinator.Coordination;
+import org.osgi.service.coordinator.Coordinator;
+import org.osgi.service.coordinator.Participant;
+import org.osgi.service.log.LogService;
 
-import aQute.bnd.deployer.repository.api.*;
-import aQute.bnd.filerepo.*;
-import aQute.bnd.osgi.*;
+import aQute.bnd.deployer.repository.api.IRepositoryContentProvider;
+import aQute.bnd.filerepo.FileRepo;
+import aQute.bnd.osgi.Jar;
 import aQute.bnd.osgi.Verifier;
-import aQute.bnd.service.*;
+import aQute.bnd.service.Actionable;
+import aQute.bnd.service.Refreshable;
+import aQute.bnd.service.RepositoryListenerPlugin;
+import aQute.bnd.service.ResourceHandle;
 import aQute.bnd.service.ResourceHandle.Location;
-import aQute.bnd.version.*;
-import aQute.lib.hex.*;
-import aQute.lib.io.*;
-import aQute.libg.cryptography.*;
+import aQute.bnd.version.Version;
+import aQute.bnd.version.VersionRange;
+import aQute.lib.hex.Hex;
+import aQute.lib.io.IO;
+import aQute.libg.cryptography.SHA1;
+import aQute.libg.cryptography.SHA256;
 
 public class LocalIndexedRepo extends FixedIndexedRepo implements Refreshable, Participant, Actionable {
+	private final String		UPWARDS_ARROW				= " \u2191";
+	private final String		DOWNWARDS_ARROW				= " \u2193";
+	private static final String	CACHE_PATH					= ".cache";
+	public static final String	PROP_LOCAL_DIR				= "local";
+	public static final String	PROP_READONLY				= "readonly";
+	public static final String	PROP_PRETTY					= "pretty";
+	public static final String	PROP_OVERWRITE				= "overwrite";
+	public static final String	PROP_ONLYDIRS				= "onlydirs";
+	public static final String	PROP_FILE_INCLUDE_POLICY	= "file.include.policy";
 
-	private final String							UPWARDS_ARROW					= " \u2191";
-	private final String							DOWNWARDS_ARROW					= " \u2193";
-	Pattern	REPO_FILE	= Pattern.compile("([-a-zA-z0-9_\\.]+)-([0-9\\.]+)(-[-a-zA-z0-9_]+)?\\.(jar|lib)");
-	private static final String			CACHE_PATH				= ".cache";
-	public static final String			PROP_LOCAL_DIR			= "local";
-	public static final String			PROP_READONLY			= "readonly";
-	public static final String			PROP_PRETTY				= "pretty";
-	public static final String			PROP_OVERWRITE			= "overwrite";
-	public static final String			PROP_ONLYDIRS			= "onlydirs";
-	
+	static final String	VERSIONED_BUNDLE_PATTERN	= "([-a-zA-z0-9_\\.]+)-([0-9\\.]+)(-[-a-zA-z0-9_]+)?\\.(jar|lib)";
+	static final String	ALL_JARS_PATTERN			= ".*jar$";
+	static final String	ALL_JARS_AND_LIBS_PATTERN	= ".*(jar|lib)$";
+
+	public enum FileIncludePolicy {
+		VERSIONED_BUNDLE(VERSIONED_BUNDLE_PATTERN), ALL_JARS(ALL_JARS_PATTERN), ALL_JARS_AND_LIBS(
+				ALL_JARS_AND_LIBS_PATTERN);
+
+		private Pattern pattern;
+
+		FileIncludePolicy(String p_regexp) {
+			this.pattern = Pattern.compile(p_regexp);
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
```diff
@@ -1,46 +1,82 @@
...
+	public static final String	PROP_OVERWRITE				= "overwrite";
+	public static final String	PROP_ONLYDIRS				= "onlydirs";
+	public static final String	PROP_FILE_INCLUDE_POLICY	= "file.include.policy";
+
+	static final String	VERSIONED_BUNDLE_PATTERN	= "([-a-zA-z0-9_\\.]+)-([0-9\\.]+)(-[-a-zA-z0-9_]+)?\\.(jar|lib)";
+	static final String	ALL_JARS_PATTERN			= ".*jar$";
+	static final String	ALL_JARS_AND_LIBS_PATTERN	= ".*(jar|lib)$";
+
+	public enum FileIncludePolicy {
+		VERSIONED_BUNDLE(VERSIONED_BUNDLE_PATTERN), ALL_JARS(ALL_JARS_PATTERN), ALL_JARS_AND_LIBS(
```
bjhargrave: These capitalized enum names are not very nice to use in a bnd file. I think better names are needed.

`...;file.include-policy=ALL_JARS` looks awkward.

biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
```diff
@@ -1,46 +1,82 @@
...
 public class LocalIndexedRepo extends FixedIndexedRepo implements Refreshable, Participant, Actionable {
+	private final String		UPWARDS_ARROW				= " \u2191";
+	private final String		DOWNWARDS_ARROW				= " \u2193";
+	private static final String	CACHE_PATH					= ".cache";
+	public static final String	PROP_LOCAL_DIR				= "local";
+	public static final String	PROP_READONLY				= "readonly";
+	public static final String	PROP_PRETTY					= "pretty";
+	public static final String	PROP_OVERWRITE				= "overwrite";
+	public static final String	PROP_ONLYDIRS				= "onlydirs";
+	public static final String	PROP_FILE_INCLUDE_POLICY	= "file.include.policy";
```
bjhargrave: The prop names should be shorter and no dots like the other props. Perhaps just `files`?

biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
```diff
@@ -69,6 +105,30 @@ public synchronized void setProperties(Map<String,String> map) {
 			throw new IllegalArgumentException(String.format(
 					"Cannot create repository cache: '%s' already exists but is not directory.",
 					cacheDir.getAbsolutePath()));
+
+		setIncludePolicy(map);
```
bjhargrave: Why pass the whole map when only the files prop is needed? Just pass the prop value.

biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
```diff
@@ -69,6 +105,30 @@ public synchronized void setProperties(Map<String,String> map) {
...
+		String policy = props.get(PROP_FILE_INCLUDE_POLICY);
+		if (policy != null && !policy.isEmpty()) {
+			try {
+				FileIncludePolicy spec = FileIncludePolicy.valueOf(policy);
+				includePolicy = spec.pattern;
+			}
+			// this means the policy was not recognized, so try to interpret it
+			// as a regex
+			catch (IllegalArgumentException e) {
+				includePolicy = Pattern.compile(policy);
```
bjhargrave: You must handle the `PatternSyntaxException` here. You can't just let it bubble up. An error must be reported as the statement is malformed.

biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
```diff
@@ -100,8 +161,14 @@ public synchronized void setProperties(Map<String,String> map) {
 	}
 
 	/**
-	 * @param contentProvider the repository content provider
-	 * @return the filename of the index on local storage
+	 * <<<<<<< HEAD
```
bjhargrave: This is left overs from a merge error.

biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
```diff
@@ -168,7 +236,7 @@ private synchronized void generateIndex(File indexFile, IRepositoryContentProvid
 	}
 
 	@SuppressWarnings("deprecation")
-	private void gatherFiles(Set<File> allFiles) throws Exception {
+	void gatherFiles(Set<File> allFiles) throws Exception {
```
bjhargrave: Why did this method loose `private`? I did not notice `gatherFiles` used anywhere in this PR.

bjhargrave:
Overall, I am not sure of the value of this change. I made some specific comments on the code but I am not convinced this is a good idea. A repo should have some consistency in its layout and this includes file naming. Allowing any arbitrary file name rules defeats this value. So I don't think this is a great idea. But others should also provide their input.

biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
eli-jordan: That's fair, I didn't put a ton of thought into the names, I'll try to come up with some better names.

biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
eli-jordan: Yeah I thought about that, but the error appeared to be reported in bnd tools when I just let it bubble up. I can easily add exception handling here if that is best

biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
eli-jordan: No particular reason, easily changed.

biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
eli-jordan: Oops... My bad. Not super familiar with how marges work in egit

biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
eli-jordan: It's used in the tests I added, I call it directly to test that the files gathered respect the policy

eli-jordan:
I think that's fair, but not all jar follow the strict naming convention required by the existing convention. I think there needs to be some more flexibility here. Perhaps allowing an arbitrary regular expression is making it too flexible. 

I look forward to further feedback!

eli-jordan:
Ok, so I have come up with some better names for the properties etc.

The property key would be 'include' that takes a comma separated list of 
predefined patterns, and arbitrary regex is not allowed.

The predefined values would be;

'versionedJars' - include jars with the versioned bundle naming convention (the current implementation, and default if the 'include' prop is not set)
'jars' - include all jars i.e. regex .*jar$
'libs' - include all libs i.e. regex .*lib$
'bundle' - include any files that have a .jar suffix, where a manifest exists and has the header Bundle-MaifestVersion: 2
Note: I'm not convinced something like this is required, but would welcome feedback.
e.g.

...;include='jars,libs'

would include all files with the .jar suffix or the .lib suffix

Thoughts?

eli-jordan:
I added what I described above without the ```bundle``` file inclusion policy

pkriens:
It is interesting to pursue this a bit different.

First this should be a different repo, people might have assumptions about the fact this is is an old FileRepo. We could use Java 7 watching to pick changes as they go.

eli-jordan:
I was very careful to ensure that the new behaviour is opt in. i.e. if the ```include``` parameter is not specified then the behaviour of this repo is exactly the same as it was prior to this change.

Using a java 7 file walker + watcher would definitely be a more performant way, and would not require manual refreshing. However, that is significantly more work than this simple change.

pkriens:
Even if it is opt-in, I am worried that this plugin is a File Repo. I really think this should be a completely new repo.

Anyway, we're discussing some heavy changes to the repo model so I want to postpone this anyway until we know what we're going to do there.

eli-jordan:
Ok fair enough. My need for this change was about indexing the bundles in a WAS Liberty Profile installation for use during development. The current LocalIndexedRepo doesn't index the bundles because they do not meet the naming criteria.

Would a much more modest change of the regex from

``` ([-a-zA-z0-9_\\.]+)-([0-9\\.]+)(-[-a-zA-z0-9_]+)?\\.(jar|lib) ```
to
``` ([-a-zA-z0-9_\\.]+)(-|_)([0-9\\.]+)(-[-a-zA-z0-9_]+)?\\.(jar|lib) ```

be accepted? 

This would address my use case.

bjhargrave:
So this change is just to allow underscore as well as hyphen between the
name and version?  If so, then I am fine with that change.

On Mon, Nov 2, 2015, 12:36 eli-jordan <notifications@github.com> wrote:

> Ok fair enough. My need for this change was about indexing the bundles in
> a WAS Liberty Profile installation for use during development. The current
> LocalIndexedRepo doesn't index the bundles because they do not meet the
> naming criteria.
>
> Would a much more modest change of the regex from
>
> ([-a-zA-z0-9_\\.]+)-([0-9\\.]+)(-[-a-zA-z0-9_]+)?\\.(jar|lib)
>
> to
> ([-a-zA-z0-9_\\.]+)(-|_)([0-9\\.]+)(-[-a-zA-z0-9_]+)?\\.(jar|lib)
>
> be accepted?
>
> This would address my use case.
>
> —
> <https://github.com/bndtools/bnd/pull/1155#issuecomment-152994892>.
>
-- 
-- 
BJ

bjhargrave:
So we should close this PR since #1178 has been merged.

PR #1164, opened 2015-10-22
[build] Multiple sub bundles did not show up in Workspace Repository

pkriens:
It turned out that each sub bundle build was clearing the map with the other sub bundles. This caused only the last sub bundle to remain in that map. I.e. the workspace repository only showed the last sub bundle.

The project now clears the map before the overall build. Sub bundle builds remove and add their name + version to  this map.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
@@ -535,7 +535,7 @@ public class ProjectBuilder extends Builder {
 	 * version.
 	 */
 	protected void startBuild(Builder builder) {
-		project.versionMap.clear();
+		project.versionMap.remove(builder.getBsn());
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
```diff
@@ -535,7 +535,7 @@ public String toString() {
 	 * version.
 	 */
 	protected void startBuild(Builder builder) {
-		project.versionMap.clear();
+		project.versionMap.remove(builder.getBsn());
```
bjhargrave: Are these uses of versionMap safe? In some places Project, versionMap is synchronized before being mutated.

pkriens:
bnd is NOT concurrent and making it concurrent is wrong. We should protect at the side of the caller, in our case bndtools

bjhargrave:
OK, then why is there any synchronization in bnd then?

pkriens:
We should probably remote it. There are too many parts not synchronized

PR #1177, opened 2015-11-01
Resolve 'Bundle-ClassPath' header when adding bundles to the bnd build path

eli-jordan:
I noticed that the bnd build path doesn't expand ```Bundle-ClassPath``` entries. I opened this issue https://github.com/bndtools/bndtools/issues/1198 but it was closed as 'won't fix' and suggested that I could put together a pull request.

This pull request is by no means complete, I just figured this is the best way to get some feedback on the approach I am taking, before building upon this core. Please advise if there is a better channel to do such things.

Basically I am extracting the entries, and caching them using an indexed FileRepo, and performing the expansion from a jar to the jar + embedded entries in Container.getMembers()

I look forward to some feedback about whether this a good approach to take.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/BundleClasspathCache.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/BundleClasspathCache.java
@@ -1,0 +1,92 @@
+package aQute.bnd.build;
+
+import java.io.File;
+import java.io.InputStream;
+import java.util.HashMap;
+import java.util.Map;
+
+import aQute.bnd.osgi.Resource;
+import aQute.bnd.service.RepositoryPlugin.PutOptions;
+import aQute.bnd.service.RepositoryPlugin.PutResult;
+import aQute.bnd.service.repository.SearchableRepository.ResourceDescriptor;
+import aQute.bnd.version.Version;
+import aQute.lib.deployer.FileRepo;
+import aQute.libg.cryptography.SHA1;
+
+/**
+ * Caches the entries specified in the Bundle-ClassPath header in a file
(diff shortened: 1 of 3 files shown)


pkriens:
Thanks for the PR. However, this only works for the File Repo. We are rather leery of the Bundle-Classpath header. However, any solution should work for all repository types.

eli-jordan:
@pkriens Thanks for the feedback. I was just reading the doc for RepositoryPlugin, it seems that the reason the current code will only work for a FileRepo is that the File object returned by a repo may not yet exist. 

Is this what you were referring to?

If so, this seems to be easily rectified by calling Container.getFile() rather than referencing the field directly, because this calls DownloadBlocker.getReason() which blocks until the file exists of fails to be retrieved.

Other than this, is the general implementation strategy valid?

rotty3000:
I noticed that related to this there are two paths by which bnd "learns" about packages. In on instance it deeply expands the bundle classpath (when it's analying the "jar" it's building) and so it has all the logic one might expect. 

However, when analysing jars from the classpath it does not do the same deep analysis. I think a re-working of this logic to use the same in both cases would solve the issue here.

eli-jordan:
Thanks @rotty3000 I'll look more closely at the bundle class path handing when building the jar. But I think the key difference here is that for compile time, the inner entries need to be extracted and cached somewhere so they can be added the javac class path

bjhargrave:
Fixed by https://github.com/bndtools/bnd/commit/d8c177a055bf051f83acb6e0355d5f280e80d1bb

gamerson:
:+1:

Issue #1172, opened 2015-10-27
Spurious split package warnings from issues amongst jars inside the bundle classpath


benson-basis:
Please have a look at https://issues.apache.org/jira/browse/FELIX-5069. It seems to me that this has to be inside of bnd. What is happening here is that a package is split between jars that are all within the Bundle-Classpath, where the concept of a split package doesn't actually mean anything.

bjhargrave:
This is referring to where bnd is finding the package contents. It is split across multiple input jars some of which you may put on the `Bundle-Classpath`. Please just use the `-split-package` directive as suggested by the message.

benson-basis:
With all due respect, not 'some of which', but rather 'all of which'.

The problem here is that I've educated my local users to be very, very,
very, averse to split packages, so I wish they didn't get a warning in this
case.

Is there some reason that it's particularly hard to imagine changing this?
I would be willing to look at making a patch since I'm the one who is
whining.

bjhargrave:
bnd is complaining about the input. It is not making a statement about the resulting bundle. Since you present multiple inputs to bnd, it sees the package split across the inputs when they are analyzed.

bjhargrave:
And looking at your [example](https://github.com/bimargulies/felix-unwanted-split-package-messages-tc), you do have 2 projects contributing to the package. Since your combiner knows the packages are being combined, it can either use the `-split-package` directive to merge the packages or `-fixupmessages` to elide the message.

pkriens:
It is hard to see what is happening on the bnd side with the Maven plugin. Could you make a bnd test case that gives the error you want to get rid of? In principle I am ok to not have warnings when JARs come from the Bundle-ClassPath but I am not sure I see how you can get them yet.

benson-basis:
Can you help me a bit? I've never used bnd _except_ in maven. I started to try to see how to make a .bnd file and repro from command line, but I'm mildly stumped. The job description: Here's a tree of .class files and two jar files. I want to end up with a Bundle-Classpath of .,jar1,jar2.

bjhargrave:
You may want to ask for help in the felix mail list since they maintain the maven-bundle-plugin. This project has a different maven plugin, bnd-maven-plugin, which is fairly simple compared to the implicit behavior of maven-bundle-plugin.

benson-basis:
[...]

PR #1179, opened 2015-11-03
Test case for #1172.

benson-basis:
As requested, a test case for issue #1172.


biz.aQute.bndlib.tests/src/test/BuilderTest.java
```diff
@@ -41,6 +41,26 @@
...
+    /*
+     * #1172 a warning being generated when a package is split between two
+     * jar files that are both in the bundle classpath, and so part of the
+     * same bundle.
+     */
+    public void test1172SplitPackageWarningsInBundleClasspath() throws Exception {
+		Builder A = new Builder();
+        try {
+            A.addClasspath(new Jar(new File("jar/sp-tc-jar1.jar")));
+            A.addClasspath(new Jar(new File("jar/sp-tc-jar1.jar")));
```
kwin: is this really meant? Probably it should be A.addClasspath(new Jar(new File("jar/sp-tc-jar2.jar")));

pkriens:
Ok, bnd is ok. So what's happening?

You put the JARs on the class path AND on the Bundle-Classpath (BCP). The Export-Package now instructs bnd to put that package in the JAR in the root. During verification bnd gets confused with the packages on the class path and the BCP (it does not even see the packages in the JAR).

The solution is to use -exportcontents. 

	 /*
     * #1172 a warning being generated when a package is split between two
     * jar files that are both in the bundle classpath, and so part of the
     * same bundle.
     */
    public void test1172SplitPackageWarningsInBundleClasspath() throws Exception {
		Builder A = new Builder();
        try {
            A.setBundleClasspath("sp-tc-jar1.jar,sp-tc-jar2.jar");
			A.setProperty("-exportcontents", "org.apache.felix.tc");
            A.setIncludeResource("sp-tc-jar1.jar=jar/sp-tc-jar1.jar,sp-tc-jar2.jar=jar/sp-tc-jar2.jar");
            A.build();
            assertTrue(A.check());
        } finally {
            A.close();
        }
    }

PR #1184, opened 2015-11-06
Updated Felix resolver to 1.8.

jest:
This brings some bugfixes for resoler OOM errors, e.g.

* https://issues.apache.org/jira/browse/FELIX-4942
* https://issues.apache.org/jira/browse/FELIX-4926

Also removed old versions of the library.

Signed-off-by: Przemek Wesołek <jest@go.art.pl>


bjhargrave:
@tjwatson Any opinion on moving to the 1.8 resolver?

tjwatson:
Be aware that by default the felix resolver implementation will now create/use/close an executor for each resolve operation.  If you scenarios that involve doing many resolve operations then you will want to use the ResolverImpl constructor that takes an Executor that you can manage the lifecycle for yourself. See https://issues.apache.org/jira/browse/FELIX-4988

Also be aware that the List that is passed to org.osgi.service.resolver.ResolveContext.insertHostedCapability(List<Capability>, HostedCapability) does not appear to implement many of the methods I would expect.  For example, add is not supported, see some fall out because of that in https://issues.apache.org/jira/browse/ARIES-1439

Other than these two gotchas I am not aware of reasons not to move up.

bjhargrave:
> If you scenarios that involve doing many resolve operations then you will want to use the ResolverImpl constructor that takes an Executor that you can manage the lifecycle for yourself.

Hmmm, this may be an issue for bndtools where the user can call the resolver many times. So I think we need to put some more thought into updating the resolver to make sure it does create new issues.

tjwatson:
Note that if you want the resolver implementation to remain single threaded you can just call the constructor ResolverImpl(Logger logger, Executor executor) with a null executor.

Here is a link to how I plan to manage the executor for Equinox.
http://git.eclipse.org/c/equinox/rt.equinox.framework.git/tree/bundles/org.eclipse.osgi/container/src/org/eclipse/osgi/container/ModuleResolver.java?h=twatson/newResolver#n86 

The meat of the solution involves an executor that is constructed like this:

		// Always want to go to zero threads when idle
		int coreThreads = 0;
		// use the number of processors - 1 because we use the current thread when rejected
		int maxThreads = Math.max(Runtime.getRuntime().availableProcessors() - 1, 1);
		// idle timeout; make it short to get rid of threads quickly after resolve
		int idleTimeout = 5;
		// use sync queue to force thread creation
		BlockingQueue<Runnable> queue = new SynchronousQueue<Runnable>();
		// try to name the threads with useful name
		ThreadFactory threadFactory = new ThreadFactory() {
			@Override
			public Thread newThread(Runnable r) {
				Thread t = new Thread(r, "Resolver thread - " + adaptor.toString()); //$NON-NLS-1$
				t.setDaemon(true);
				return t;
			}
		};
		// use a rejection policy that simply runs the task in the current thread once the max threads is reached
		RejectedExecutionHandler rejectHandler = new RejectedExecutionHandler() {
			@Override
			public void rejectedExecution(Runnable r, ThreadPoolExecutor exe) {
				r.run();
			}
		};
		return new ThreadPoolExecutor(coreThreads, maxThreads, idleTimeout, TimeUnit.SECONDS, queue, threadFactory, rejectHandler);

Issue #1187, opened 2015-11-09
NPE in Clazz.parseClassFile() on analyzing self-referencing annotations


trespasserw:
```
java.lang.NullPointerException
  at aQute.bnd.osgi.Clazz.parseClassFile(Clazz.java:747)
  at aQute.bnd.osgi.Clazz.parseClassFile(Clazz.java:512)
  at aQute.bnd.osgi.Clazz.parseClassFileWithCollector(Clazz.java:501)
  at aQute.bnd.component.AnnotationReader.getDef(AnnotationReader.java:131)
  at aQute.bnd.component.AnnotationReader.getDefinition(AnnotationReader.java:127)
  at aQute.bnd.component.DSAnnotations.analyzeJar(DSAnnotations.java:79)
  at aQute.bnd.osgi.Analyzer.doPlugins(Analyzer.java:658)
  at aQute.bnd.osgi.Analyzer.analyze(Analyzer.java:213)
  at aQute.bnd.osgi.Builder.analyze(Builder.java:379)
  at aQute.bnd.osgi.Analyzer.calcManifest(Analyzer.java:688)
```

Happen on processing an annotation class containing a method annotated by the same annotation (e.g. [org.springframework.core.annotation.AliasFor](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/core/annotation/AliasFor.java) from spring-core) by DSAnnotations or MetatypeAnnotations plugins.

A test case for ClazzTest (requires DSAnnotations.Options to be public):
```java
	public @interface RecursiveAnno {
		@RecursiveAnno("recursive")
		String value() default "";
	}

	public void testRecursiveAnnotation() throws Exception {
		File file = IO.getFile("bin/test/ClazzTest$RecursiveAnno.class");
		Analyzer analyzer = new Analyzer();
		Clazz clazz = new Clazz(analyzer, file.getPath(), new FileResource(file));
		clazz.parseClassFile();
		analyzer.getClassspace().put(clazz.getClassName(), clazz);
		AnnotationReader.getDefinition(clazz, analyzer, 
			EnumSet.noneOf(DSAnnotations.Options.class),
			new XMLAttributeFinder(analyzer));
	}
```

bndlib 3.0.0.201509101326
Git-Descriptor: 3.0.0.REL
Git-SHA: 0ac3b76c9bd573dfc30cbeaa53700c13ee5187d0

trespasserw:
[...]

PR #1189, opened 2015-11-10
Fixes NPE in Clazz.parseClassFile() on analyzing self-referencing annotations #1187

trespasserw:


bjhargrave:
Thanks. LGTM.

@djencks, @pkriens should also review.

pkriens:
LGTM

PR #1198, opened 2015-11-13
prefer metatype namespace 1.2, using antique "Char" instead of "Chara…

### LINKED ISSUE #1183: Support generating 1.2 metatype files
Currently only 1.3 metatype files are being written by bnd. As some people cannot easily upgrade the metatype implementation (and therefore can only use at most 1.2 metatype files) and 1.3 did not add much new features (so most of the metatype files are automatically 1.2 compliant) bnd should by default use the 1.2 namespace if only 1.2 features are used. 
A more lengthy discussion about that can be found in https://github.com/bndtools/bnd/issues/675.

djencks:
…cter".  !.3 namespace only used for empty OCDs

fixes #1183
Signed-off-by David Jencks <david_jencks@yahoo.com


bjhargrave:
LGTM

PR #1200, opened 2015-11-14
Check declared service and class are compatible.

### LINKED ISSUE #894: RFE: improve DS: detect published but unimplemented interfaces
bnd should detect that interface C is published but not implemented and generate an error for the example below

```
@Component(designate = Config.class, provide = {
  B.class, C.class
})
public class ComponentImpl implements A, B {
```

cloned from https://github.com/bndtools/bndtools/issues/1084

djencks:
fixes #894


biz.aQute.bndlib/src/aQute/bnd/component/AnnotationReader.java
```diff
@@ -940,7 +940,15 @@ protected void doComponent(Component comp, Annotation annotation) throws Excepti
 			// We have explicit interfaces set
 			component.service = new TypeRef[x.length];
 			for (int i = 0; i < x.length; i++) {
-				component.service[i] = (TypeRef) x[i];
+				TypeRef typeRef = (TypeRef) x[i];
+				Clazz service = analyzer.findClass(typeRef);
+				if (assignable(clazz, service) == Boolean.FALSE) {
+					analyzer.error("Class %s is not assignable to specified service %s", clazz.getFQN(),
```
bjhargrave: Perhaps this should just be a warning and not an error. It is also possible that the service type is deeply inherited and not visible to bnd.

djencks: The entire inheritance hierarchy is checked, and if any class is not available (including the alleged service class) we assume everything is fine.  The error is raised only if all classes that could be involved in the check are visible to bnd.

bjhargrave: OK, then I am good.

PR #1201, opened 2015-11-17
ant: Add File.exists() checks for classpath elements

gmarcy:
signed-off-by Glenn Marcy <gmarcy@us.ibm.com>


bjhargrave:
I don't use the ant support, but this change seems reasonable to me.

PR #1202, opened 2015-11-17
Break out of loop when extended class is missing.

djencks:


biz.aQute.bndlib/src/aQute/bnd/component/AnnotationReader.java
```diff
@@ -143,6 +143,7 @@ private ComponentDef getDef() throws Exception {
 							clazz.getClassName())
 							.details(new DeclarativeServicesAnnotationError(className.getFQN(), null, null,
 									ErrorType.UNABLE_TO_LOCATE_SUPER_CLASS));
+					break;
```
bjhargrave: While this break is necessary, the test added here, <https://github.com/bndtools/bnd/commit/801dd138e6ccfa19b7368eceea12875e2ccce845#diff-f590bbd3071ea67fdfee0576b5f5e20cR3321>, is bogus in that it does not trigger the problem that the above check is supposed to catch. 

Furthermore, the above check only runs when the non-standard inherit option is set.

bjhargrave:
LGTM

PR #1204, opened 2015-11-18
Fix service implementation checks.

### LINKED ISSUE #1203: Potential issue with 32ad027 and Servlet?
Before 32ad027 I had code that looked something like this:

    @Component(service = Servlet.class)
    public class FooBar extends HttpServlet

That worked fine. After updating to the latest bndtools from cloudbees, I am getting the following error:

    Class com.seanbright.FooBar is not assignable to specified service javax.servlet.Servlet

Changing the class declaration to the following silences the error:

    @Component(service = Servlet.class)
    public class FooBar extends HttpServlet implements Servlet

But this seems a bit inelegant. Is this a bug in bnd or a user error?

djencks:
fixes #1203
Fix assignable check by looking at superclass interfaces, and add more tests.

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
+++ b/biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
@@ -1,6 +1,7 @@
 package test.component;
 
 import java.io.File;
+import java.io.IOException;
 import java.io.Serializable;
 import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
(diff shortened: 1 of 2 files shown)


bjhargrave:
LGTM

djencks:
Can you think of any more cases to test?  I don't see how to test the "we don't have one of the classes in the hierarchy" case.

bjhargrave:
> Can you think of any more cases to test? I don't see how to test the "we don't have one of the classes in the hierarchy" case.

Nope. I would go ahead and merge.

PR #1205, opened 2015-11-19
Fix NPE when manifest has a single stray '@'

djencks:


djencks:
I have a test if you like.

bjhargrave:
Would be good to add a test. Otherwise, LGTM.

bjhargrave:
@djencks Do you plan to add a test case?

PR #1209, opened 2015-11-20
Baseline improvements

timothyjward:
Tests for the bnd-baseline-maven-plugin, and a fix for the Baseline class when it is used to baseline two bundles with different symbolic names

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/baseline/BaselineTest.java
+++ b/biz.aQute.bndlib.tests/src/test/baseline/BaselineTest.java
@@ -1,25 +1,39 @@
 package test.baseline;
 
-import static org.mockito.Matchers.*;
-import static org.mockito.Mockito.*;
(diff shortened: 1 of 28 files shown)


timothyjward:
Just checking that @pkriens @bjhargrave @njbartlett are ok with the baselining change in b6bb46c1dff5b0841b0fb79c38deb662ba866ff8 - It is possible (and even useful) to baseline against a different bundle which offers the same API (for example a project rename). In this case the bundle version cannot be suggested because the symbolic names have changed.

The fix was very small, and easy to test, but as we're preparing for release 3.1 I thought some extra eyes would help.

rotty3000:
+1 I think this is a nice feature for restructuring efforts.

pkriens:
Except for the AppVeyor fail (which seems transient?) I am ok

timothyjward:
The windows failure turned out to be an actual problem when handling line-endings in the validation script. It's all fixed now

PR #1211, opened 2015-11-23
fixes #1210

### LINKED ISSUE #1210: add a PostAnalyzerPlugin lifecycle
it would be really nice if there was a PostAnalyzerPlugin lifecycle which would support analysis after manifest generation.

Currently the process is as follows (roughly):
```
- collect bundle classpath
- calculate ees
- analyze classpath exports & contracts
- Handle the bundle activator
- execute AnalyzerPlugins
- Conditional packages
- EXPORTS
- IMPORTS
- USES
- privates
- Checks
```
I'd like to have the following:
```
- collect bundle classpath
- calculate ees
- analyze classpath exports & contracts
- Handle the bundle activator
- execute AnalyzerPlugins
- Conditional packages
- EXPORTS
- IMPORTS
- USES
- privates
- execute PostAnalyzerPlugins   <----- NEW
- Checks
```

rotty3000:


biz.aQute.bndlib.tests/src/test/VerifierPluginTest.java
```diff
@@ -0,0 +1,33 @@
...
+public class VerifierPluginTest extends TestCase {
+
+	public static void testVerifierPluginExecution() throws Exception {
+		final AtomicBoolean executedCheck = new AtomicBoolean(false);
+
+		VerifierPlugin verifier = new VerifierPlugin() {
+
+			@Override
+			public void verify(Analyzer analyzer) throws Exception {
+				executedCheck.set(true);
```
bjhargrave: I think the test should validate that verifier is run after the analyzers, no? For example, check that manifest looks a certain way and the included resources are present in the jar.

This test just makes sure the plugin is run but not in relation to the normal analyzers.

bjhargrave:
Also, your commit message is missing the required Signed-off-by. And the commit message should be more useful rather than simply "fixes bug". Thanks.

rotty3000:
Addressing both remarks...

biz.aQute.bndlib/src/aQute/bnd/osgi/Verifier.java
```diff
@@ -531,6 +532,22 @@ public void verify() throws Exception {
...
+
+		doVerifierPlugins();
+	}
+
+	private void doVerifierPlugins() {
+		for (VerifierPlugin plugin : getPlugins(VerifierPlugin.class)) {
+			try {
+				Processor previous = beginHandleErrors(plugin.toString());
+				plugin.verify(analyzer);
+				endHandleErrors(previous);
```
bjhargrave: Perhaps the endHandleErrors should be in a finally block since plugin.verify can throw an exception meaning endHandleErrors will not be called for that plugin. Analyzer.doPlugins (which you copied this loop from) has this same flaw.

```java
Processor previous = beginHandleErrors(plugin.toString());
try {
   plugin.verify(analyzer);
} finally {
   endHandleErrors(previous);
}
```

rotty3000: Shall I fix this one only for the sake of this PR?
Or shall I fix both?
Or note it and fix the issue as a follow-on?

bjhargrave: You can fix both in this PR. I would put the fix for the Analyzer into a separate commit.

rotty3000: Will do!

biz.aQute.bndlib/src/aQute/bnd/osgi/Verifier.java
```diff
@@ -531,6 +532,24 @@ public void verify() throws Exception {
...
+			Processor previous = beginHandleErrors(plugin.toString());
+			try {
+				plugin.verify(analyzer);
+			}
+			catch (Exception e) {
+				e.printStackTrace(System.err);
+				error("Verifier Plugin %s failed %s", plugin, e);
+			}
+			finally {
+				endHandleErrors(previous);
```
bjhargrave: This is wrong too. The error call in the exception block will be be in the wrong Processor. You need to wrap the try/finally just around the call to plugin.verify.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -659,6 +658,9 @@ void doPlugins() {
 				e.printStackTrace(System.err);
 				error("Analyzer Plugin %s failed %s", plugin, e);
 			}
+			finally {
+				endHandleErrors(previous);
```
bjhargrave: This is wrong. Now the error in the exception block will be put in the wrong Processor. You need to wrap the try/finally just around the call to plugin.analyzeJar.

rotty3000: Ok, will fix both commits

rotty3000: Like so?
```java
	private void doVerifierPlugins() {
		for (VerifierPlugin plugin : getPlugins(VerifierPlugin.class)) {
			try {
				Processor previous = beginHandleErrors(plugin.toString());
				try {
					plugin.verify(analyzer);
				}
				finally {
					endHandleErrors(previous);
				}
			}
			catch (Exception e) {
				e.printStackTrace(System.err);
				error("Verifier Plugin %s failed %s", plugin, e);
			}
		}
	}
```

bjhargrave: Yes!

rotty3000:
updated!

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -648,11 +648,15 @@ void doPlugins() {
...
 			try {
 				Processor previous = beginHandleErrors(plugin.toString());
-				boolean reanalyze = plugin.analyzeJar(this);
-				endHandleErrors(previous);
-				if (reanalyze) {
-					classspace.clear();
-					analyzeBundleClasspath();
+				try {
+					boolean reanalyze = plugin.analyzeJar(this);
+					if (reanalyze) {
```
bjhargrave: This is still a behavior change. Before endHandleErrors was called before reanalyzing. Your change calls it after. I think the finally must be right after the plugin.analyzeJar call to avoid unintended behavioral changes.

rotty3000: gah... fixing

rotty3000: like so?
```java
	void doPlugins() {
		for (AnalyzerPlugin plugin : getPlugins(AnalyzerPlugin.class)) {
			try {
				boolean reanalyze = false;
				Processor previous = beginHandleErrors(plugin.toString());
				try {
					reanalyze = plugin.analyzeJar(this);
				}
				finally {
					endHandleErrors(previous);
				}
				if (reanalyze) {
					classspace.clear();
					analyzeBundleClasspath();
				}
			}
			catch (Exception e) {
				e.printStackTrace(System.err);
				error("Analyzer Plugin %s failed %s", plugin, e);
			}
		}
	}
```

bjhargrave: Yes! (not sure why you need to set reanalyze to false, perhaps the compiler whines...)

rotty3000: old bad habbits...

rotty3000:
BJ is this ok now?

bjhargrave:
LGTM

PR #1214, opened 2015-11-23
fixes #1213

### LINKED ISSUE #1213: gradle non-workspace plugin does not process includes from manifest configuration
Given the following:
```gradle
jar {
    manifest {
        attributes('-include': 'other.bnd')
    }
}
```
where `other.bnd` exists "at the correct place", the directives contained within do not get processed.

rotty3000:


bjhargrave:
This is not the right solution. There are also some larger issues here. 

The manifest is not a bnd file. It should be treated as a "prepend" to the bnd file. That is, anything in the bnd file will overwrite what is in the jar task manifest. This means that if the bnd file has a -include property, it must override the -include property in the jar task manifest. Your solution does not do this and will create a very different world than other bnd users have.

I am trying to think up a better solution.

rotty3000:
I agree with the "prepend to the bnd file" idea. Didn't know how else to achieve it.

rotty3000:
perhaps the setProperties methods in bnd should pull in any already set properties and then call the doInclude.

bjhargrave:
Replaced by https://github.com/bndtools/bnd/pull/1215

bjhargrave:
> perhaps the setProperties methods in Processor should pull in any already set properties (merging it's own passed properties last) and then call the doInclude().

This wouldn't work since you only want to process -include once you have all the properties loaded since later properties can override -include. And if setProperties can be called multiple times, you never know when the last call has been made to process -include. So the solution is to load the jar task manifest entries into a Properties object, add the bnd file entries over that (to allow overrides) and then setProperties on the final Properties object. This is the solution in #1215.

rotty3000:
Thx!

bjhargrave:
> Thx!

And thanks for finding the bugs!

PR #1216, opened 2015-11-24
Fix assignable checking where one class both extends and implements, …

djencks:
…case found by Ray Auge.


bjhargrave:
LGTM

PR #1222, opened 2015-11-26
Fix: java.net.BindException: Can't assign requested address

psoreide:
InetAddress.getLocalHost() may return ip's from wrong local interfaces,
use 'localhost' instead.

### DIFF
--- a/biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
+++ b/biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
@@ -81,7 +81,7 @@ public class ProjectLauncherImpl extends ProjectLauncher {
 		LauncherConstants lc = getConstants(getRunBundles(), false);
 
 		Method mainMethod = main.getMethod("main", args.getClass(), Properties.class);
-		Object o = mainMethod.invoke(null, (Object) args, lc.getProperties(new UTF8Properties()));
+		Object o = mainMethod.invoke(null, args, lc.getProperties(new UTF8Properties()));
 		if (o == null)
 			return 0;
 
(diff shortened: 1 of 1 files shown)


biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
```diff
@@ -161,7 +160,7 @@ private LauncherConstants getConstants(Collection<String> runbundles, boolean ex
 
 		if (!exported && !getNotificationListeners().isEmpty()) {
 			if (listenerComms == null) {
-				listenerComms = new DatagramSocket(new InetSocketAddress(InetAddress.getLocalHost(), 0));
+				listenerComms = new DatagramSocket(new InetSocketAddress("localhost", 0));
```
bjhargrave: Wouldn't 

```java
listenerComms = new DatagramSocket(new InetSocketAddress(InetAddress.getLoopbackAddress(), 0));
```

be better? Then no name lookup is even required?

psoreide: Absolutely! I'll fix!

psoreide: InetAddress.getLoopbackAddress() is since JavaSE-7 whilst the launcher is pointing to ee.j2se-1.6.0.jar. Is it supposed to do that?

bjhargrave: Oh yeah, forgot that the launcher is supporting older Java than the main build. 

So this will work on Java 6

```java
listenerComms = new DatagramSocket(new InetSocketAddress(InetAddress.getByName(null), 0));
```

Using `null` for the name represents the loopback address.

bjhargrave:
> Is the removal of the cast to Object of any significance? The formatting at save did that.

I don't think it matters since the invoke method takes an Object arg.

Could you squash all the commits together?

PR #1223, opened 2015-11-27
Fix type inference regression with added tests, and try to make error…

djencks:
… messages include the component name more often.


bjhargrave:
LGTM

PR #1225, opened 2015-12-01
[remote] added new agent methods for working with bundles

gamerson:
I'd like to request this additional API in the remote Agent so that we can call the agent from some of our dev tools at Liferay, namely the Liferay IDE eclipse plugin.

### DIFF
--- a/biz.aQute.remote/src/aQute/remote/agent/AgentServer.java
+++ b/biz.aQute.remote/src/aQute/remote/agent/AgentServer.java
@@ -5,6 +5,7 @@ import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -124,6 +125,14 @@ public class AgentServer implements Agent, Closeable, FrameworkListener {
 
 		Bundle b = context.installBundle(location, in);
 		installed.put(b.getLocation(), sha);
+		return toDTO(b);
+	}
+
+	@Override
+	public BundleDTO installFromURL(String location, String url) throws Exception {
+		InputStream is = new URL(url).openStream();
+		Bundle b = context.installBundle(location, is);
+		installed.put(b.getLocation(), url);
 		return toDTO(b);
 	}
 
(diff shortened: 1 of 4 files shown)


gamerson:
This commit adds new API to the Agent, so I guess that means the aQute.remote.api package needs to go from 1.0.0 -> 1.1.0 ? but I wasn't sure if 1) this api would get the nod and 2) how the team handles updating exported api versions.  Thanks for the feedback in advance.

pkriens:
I am not sure I like the idea of updating/installing from URL, it seems to bypass the whole idea of the agent? The controller of the agent can easily provide this capability itself by creating the association between the SHA and the URL? The controller can also handle local URLs like mvn: or file: that are impossible for the agent to handle. Also, the agent might not even be connected to the same network if an SSH tunnel is used to talk to the agent (or serial line). It also does not handle caching well. So it seems to introduce a lot of potential error cases and no functionality that is already possible?

I am ok with the new getBundles though I wonder if it is worth it ...

rotty3000:
I think one of the issues @gamerson found was that with Liferay the DTO
request for all the bundles was causing huge delays like 5-6 seconds per
operation due to the sheer number of bundles which is why he was trying to
cut down the scope.

pkriens:
Ok, then it seems worth it. Still surprises me that you can generate that many DTOs that you get a noticeable delay on a local network?

gamerson:
hey @pkriens regarding the URL... one of the use-cases I have is WABs.  So for Liferay 7 we support devs developing wars and then deploying them to equinox as a war using "webbundle:file:///..." style URI.  Then on the Liferay 7 side (the same jvm process that holds the agent bundle) there is a URL handler that handles the "webbundle:*" URI.  So we can't do this client side call "url.openStream()" that needs to be done in the same process as the the Liferay 7 WAB extender bundle.   So the URL in our cases is always local.  

I'm open to thoughts on how we can delay the calling of URL#openStream() until it can run in the same JVM process as the agent bundle.

pkriens:
But in that case you seem to be using the agent just like a remote management agent, unrelated to the debugging. This overlaps with the OSGi REST agent I guess. It seems to conflate the purpose of this debug agent a bit with an OSGi management agent. 

However, I am not strongly against it as long as the Javadoc clearly indicates that this will use out-of-band communication and make assumptions about the target.

gamerson:
Squashed commits and updated javadocs.

biz.aQute.remote/src/aQute/remote/agent/AgentServer.java
```diff
@@ -128,6 +129,15 @@ public BundleDTO install(String location, String sha) throws Exception {
 	}
 
 	@Override
+	public BundleDTO installFromURL(String location, String url) throws Exception {
+		InputStream is = new URL(url).openStream();
+		Bundle b = context.installBundle(location, is);
+		installed.put(b.getLocation(), url);
+		is.close();
```
bjhargrave: installBundle will always close the InputStream

gamerson: Updated.

gamerson:
Can this be considered for inclusion in 3.1?

pkriens:
I am ok to include this in 3.1

PR #1227, opened 2015-12-02
fixes #1226 allow donotcopy to match on path as well as on name

### LINKED ISSUE #1226: -donotcopy only handles matching on file names
It'd be really much more useful if `-donotcopy` matched any part of the path. Even if that path were relativized from the source base(s).

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
```diff
@@ -1458,6 +1458,10 @@ public boolean doNotCopy(String v) {
 		return getDoNotCopy().matcher(v).matches();
 	}
 
+	public boolean doNotCopy(File from) {
+		return doNotCopy(from.getName()) || doNotCopy(from.getAbsolutePath());
```
bjhargrave: You will need to do a lot more work to match full paths on Windows. I suspect this is why the windows build failed.

rotty3000: Right, will normalize the path by convert to a URI first.

bjhargrave: There must be places in bnd where file names must handle windows. Perhaps @pkriens can point you to some utility classes to help here.

rotty3000:
@bjhargrave @pkriens any thoughts on this? the tests pass, and I think this is fairly innocuous.
FWIW It does helps me plenty!

rotty3000:
oh, the paths are now relative to the bnd file of the builder.

pkriens:
You should probably put the getBase().getURI() in an instance variable. You now call it twice in the method and that method has the potential to be called quite often if you're making a large build.

I am not sure why you resolve the URI? The File is in bnd always absolute (IO.getFile() ensures this). So I think you only need to relativize it?

        URI baseURI = getBase().toURI();

        boolean doNotCopy( File f ) {
            if ( doNotCopy(f.getName())
                return true;

            URI uri = baseURI.relativize(f.toURI());
            if ( uri.isAbsolute() )
               return false; // there is a case to make this
                                   // true, should we allow copying outside the workspace?
                                   // guess it will break people but it is a security threat
           return doNotCopy( uri.toString() );
       }

bjhargrave:
> return doNotCopy( uri.toString() );

This should be 
```java
return doNotCopy( uri.getPath() );
```
since we don't want `file:` in the glob comparison.

rotty3000:
@pkriens I don't think there is any security threat here since we are not doing any action other than matching the regex. 

As far as I know this isn't a protection mechanism, if the developer did: `-includeresource: /etc/passwd` isn't that their own fault? Do not copy can't really assert any kind of security distinction between that and this: `-includeresource: /some/path/to.jar` which I should think is perfectly legit and well used.

pkriens:
Agree, we're not the security threat, but being able to copy /etc/passwd is worrying a bit. Then again, we allow them to call system functions so the only safety people have is to run their build in a sandbox.

biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
```diff
@@ -1458,6 +1459,17 @@ public boolean doNotCopy(String v) {
 		return getDoNotCopy().matcher(v).matches();
 	}
 
+	public boolean doNotCopy(File from) {
+		if (doNotCopy(from.getName())) {
+			return true;
+		}
+
+		URI baseURI = getBase().toURI();
```
bjhargrave: @pkriens suggested to put this in an instance var so you are not creating the baseURI many times. You could even add `getBaseURI()` to Processor (next to `getBase()`) to push the function to the superclass since others can then use.

rotty3000: Got it!

pkriens:
Be ready for lots of baseline errors if you add a method to Processor. It adds a method to a number of packages that extend Processor

rotty3000:
All good to merge?

pkriens:
Yup. Go and merge!

bjhargrave:
Hold off a second

bjhargrave:
If you had `-donotcopy: foo.*bar`, this change will match `/xxx/foo/yyy/bar/something` as well as `/xxx/foobaxbar`. So there is a risk of the change not copying things it used to.

So @pkriens suggests wrap the new function in `if (since(3.1) { ...}` (not exact code syntax).

bjhargrave:
LGTM

PR #1232, opened 2015-12-02
fixes #1231

### LINKED ISSUE #1231: bnd cannot "automatically" resolve exported packages properly from WABs
When attempting to use the auto calculation of exported packages; i.e.

`-exportcontents: ${packages;ANNOTATED;org.osgi.annotation.versioning.Version}`

I discovered that bnd does not properly *learn* packages which are not rooted in the based of the jar. This is because the *prefix* resolved for classes rooted in a non-`.` Bundle-ClassPath is ignored completely.

rotty3000:


rotty3000:
ok, I mixed the resources for two test cases together.. that was a bad idea. Fixing.

rotty3000:
hmm still issues apparently... investigating.

rotty3000:
@bjhargrave @pkriens thoughts?

bjhargrave:
I need some time to get to this... trying to finish some other things...

pkriens:
Looks ok to me ... nice work actually, this really was a bug.

Some hints:
a) Every Analyzer implements Domain so you can use setIncludeResource("...") instead of using  the ugly constants. Most OSGi headers and many bnd instructions are methods in there. Feel free to add missing methods/headers.
b) After you've done the build(), you can call `b.getExports().get("test.refer_versioned").getVersion()`. This is a lot smaller and cleaner then parsing the manifest

rotty3000:
If you want I'll cleanup the test then merge?

pkriens:
Yup, go ahead

PR #1236, opened 2015-12-03
Minor new functionality + JUnit big 1224

pkriens:


pkriens:
The 1224 bug is in bndtools ...

PR #1240, opened 2015-12-05
gradle: Add Baseline task

bjhargrave:
Add a Baseline task type to the gradle plugin.

Based upon Ray's poc:
https://github.com/rotty3000/phidias/blob/baseline-poc/build.gradle#L40-L94


bjhargrave:
Please review and try this out @rotty3000.

Issue #1078, opened 2015-09-05
badly terminated extension lines in .bnd files produce odd errors


djencks:
```
Include-Resource: \ 
  WEB-INF=resources/WEB-INF,\
```
produced the error

      [bnd]  Input file does not exist: b

I don't have an example handy but whitespace after a \ prevents it from being a line continuation and tends to produce incomprehensible errors, especially when this is in the middle of a long Service-Component header defining DS components.

Both of these were accepted and interpreted leniently by old versions of bnd (I don't know how old).

pkriens:
Check if the input likely has this case by checking for strange property keys.

pkriens:
Should be fixed in https://github.com/bndtools/bnd/pull/1241

PR #1241, opened 2015-12-05
#1078 Catching errors in bnd files

pkriens:
This commit is a bit large because I added a PropertiesParser. I think I also made it a bit faster. This affected a few places.

Sneaked in one commit addressing a problem in bnd. We added an include for the JDT properties via the cnf directory. Added a "-init" option, this will be macro evaluated. Hope to use this to safe this indirection but might require more work.


biz.aQute.bndlib.tests/src/test/AnnotationsTest.java
```diff
@@ -26,6 +26,9 @@
 import aQute.lib.io.IO;
 import junit.framework.TestCase;
 
+@SuppressWarnings({
+		"deprecation", "restriction"
+})
```
bjhargrave: I had already released commits which fixed these (and the other) warnings. I guess you did not pull from github before you started this. https://github.com/bndtools/bnd/commit/58832868cdc5841bdbdb3be0ecb5e474f2974d72

pkriens:
I did rebase but that gave me conflicts that I then merged. Another case of great minds? :-)

PR #1248, opened 2015-12-08
Assure that references have a name.

djencks:


djencks:
I'd like this in the rc1 if at all possible

timothyjward:
Is there a corresponding change in bndtools to add a new message for the new error constant?

https://github.com/djencks/bnd/commit/25401c87d6db06a80ac92f2d6c575cdffd8e5bf5#diff-7beabe7477d7bd687d8d83c335d78106R6

djencks:
I have no idea how bndtools uses these constants, I didn't add a message for the last error constant I added.

timothyjward:
There are error processors in bndtools which handle the bnd errors. It may be that nothing needs updating for the new constants, but the handling code is here:

https://github.com/bndtools/bndtools/blob/948fff97560022aa505fbf8e7eb87774901eb656/bndtools.builder/src/org/bndtools/builder/handlers/component/DSAnnotationErrorHandler.java#L19

If the error marker highlighting and error messages are working properly then there's nothing extra to do.

djencks:
I don't have a bndtools project I can test it on, but from looking at the code I think it's going to put the marker on the class, which is appropriate.  Again by inspection I think the previously added error type INCOMPATIBLE_SERVICE will also put the marker on the class as appropriate.

bjhargrave:
LGTM

The DSAnnotationErrorHandler does not seem to care about the errorType at the moment anyway.

PR #1267, opened 2015-12-15
Resolve various locked file issues under Windows

seanbright:
There are a couple resolver tests that were leaving behind a 'tmp' directory even though the tests were passing. There were a few issues contributing to this, but mostly we are just not closing things (PersistentMaps and Resources) after we are done using them.

This also changes Workspace.close() to call it's parent's (Processor) close() method as well for additional cleanup.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
@@ -816,6 +816,13 @@ public class Workspace extends Processor {
 
 	public void close() {
 		cache.remove(getPropertiesFile().getParentFile().getParentFile());
+
+		try {
+			super.close();
+		}
+		catch (IOException e) {
+			/* For backwards compatibility, we ignore the exception */
+		}
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


bjhargrave:
All of the commits are missing the Signed-off-by in their commit messages. A Signed-off-by in the PR comment is not useful since it is not part of the git repo.

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -816,6 +816,13 @@ public RepositoryPlugin getRepository(String repo) {
 
 	public void close() {
 		cache.remove(getPropertiesFile().getParentFile().getParentFile());
+
+		try {
+			super.close();
+		}
+		catch (IOException e) {
+			/* Swallow exception for BC */
```
bjhargrave: This seems rather dangerous. What does the comment mean?

bjhargrave:
Lets hold this out til 3.2. Or do you need it in 3.1?

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
seanbright: BC = Backwards Compatibility

It means that `Workspace.close()` did not throw an exception before but Processor.close() is declared to throw an `IOException`. I'm open to suggestions on how to make this better.

seanbright:
It can wait

PR #1280, opened 2015-12-21
support ~ prefix to not overwrite when including resources

erwint:
If included resources always overwrite you cannot make a bundle that
wraps a thirdparty dependendcy with some custom  modifications (e.g.
handpicked fixes from upstream) by including the thirdparty depedency
and just adding the fixed classes as source to the bundle.

The new prefix '~' will deactivate the overwrite mode, e.g.
-includeresource: ~@lib/thirdparty.jar


erwint:
any feedback?

rotty3000:
writing is completely controlled by order.

Just add your classes after the third party bundle!!!

- Ray

On Wed, Jan 6, 2016 at 8:34 AM, Erwin Tratar <notifications@github.com>
wrote:

> any feedback?
>
> —
> <https://github.com/bndtools/bnd/pull/1280#issuecomment-169323612>.
>



-- 
*Raymond Augé* <http://www.liferay.com/web/raymond.auge/profile>
 (@rotty3000)
Senior Software Architect *Liferay, Inc.* <http://www.liferay.com>
 (@Liferay)
Board Member & EEG Co-Chair, OSGi Alliance <http://osgi.org> (@OSGiAlliance)

erwint:
Probably I was unclear, or you got the wrong expression from the testcase, for which I used a simpler setup. 

I want to compile the "patches" in my wrapper project, so the compile output are the classes which I do not want to be overwritten. AFAICT there is no ordering that can be applied here, as the inclusion comes after compilation and would overwrite any compilation result.

biz.aQute.bndlib.tests/src/test/BuilderTest.java
```diff
@@ -2960,4 +2960,24 @@ public static void report(String title, Analyzer builder, Jar jar) {
...
+		Builder bmaker = new Builder();
+		bmaker.setClasspath(new File[] {
+				new File("bin"), new File("src")
+		});
+		bmaker.setProperty("Export-Package", "test.includeresource");
+		bmaker.setProperty("Include-Resource", "@jar/includeresource.jar");
+		Jar jar = bmaker.build();
+		assertTrue(jar.getResource("test/includeresource/SomeClass.class").size() == 294);
+		
+		// now again without overwrite
```
bjhargrave: I would break this in to two test methods.

biz.aQute.bndlib.tests/src/test/BuilderTest.java
```diff
@@ -2960,4 +2960,24 @@ public static void report(String title, Analyzer builder, Jar jar) {
 		System.err.println("Imports     " + builder.getImports());
 	}
 
+	public void testOverwriteIncludeResources() throws Exception {
+		Builder bmaker = new Builder();
+		bmaker.setClasspath(new File[] {
+				new File("bin"), new File("src")
+		});
+		bmaker.setProperty("Export-Package", "test.includeresource");
+		bmaker.setProperty("Include-Resource", "@jar/includeresource.jar");
```
bjhargrave: Why not use the more modern `-includeresources`?

biz.aQute.bndlib.tests/src/test/BuilderTest.java
```diff
@@ -2960,4 +2960,24 @@ public static void report(String title, Analyzer builder, Jar jar) {
...
+		Jar jar = bmaker.build();
+		assertTrue(jar.getResource("test/includeresource/SomeClass.class").size() == 294);
+		
+		// now again without overwrite
+		bmaker = new Builder();
+		bmaker.setClasspath(new File[] {
+				new File("bin"), new File("src")
+		});
+		bmaker.setProperty("Export-Package", "test.includeresource");
+		bmaker.setProperty("Include-Resource", "~@jar/includeresource.jar");
```
bjhargrave: Why not use the more modern `-includeresources`?

biz.aQute.bndlib.tests/src/test/BuilderTest.java
```diff
@@ -2960,4 +2960,24 @@ public static void report(String title, Analyzer builder, Jar jar) {
...
+		jar = bmaker.build();
+		assertTrue(jar.getResource("test/includeresource/SomeClass.class").size() > 294);
```
bjhargrave: This test seems a bit fragile since it greatly depends upon how the java compiler generates the class file. Why does the newer file always have to be larger than 294 bytes?

erwint: the testcase includes the already compiled "patched" file, so I know the exact size.

bjhargrave:
I think your use case is a bit awkward. You are effectively trying to do a split package where some of the package comes from the third party jar and the rest comes from your project. The combination can be messed up since your replacement class(es) can have very different inner classes that the class you are replacing. So the resulting jar can have left over inner classes from the original class.

biz.aQute.bndlib.tests/src/test/BuilderTest.java
bjhargrave: But you do not know the size of the class file compiled by the source code in this PR. You assume it will be > 294 but you do not know that positively.

erwint: you're right. I misremembered how the testcase works. I'll modify this.

erwint:
It's not exactly a split package, as the original third-party Jar is not even in my repository (and mostly not even an OSGi jar). Basically, we got the idea last year in a workshop with Neil.

The point with inner classes is valid, so you need to be careful. You could try to exclude some files explicitly with a pattern (e.g. `@library.jar!/!com/whatever/MySpecialClass*`) in that case, but that does not scale.

pkriens:
You could also do this by 
    -includeresource: @somejar.jar, ${bin}
    -exportcontents: Foo;version=1, Bar;version=2

erwint:
I don't think that using `${bin}` will work with the `bnd-maven-plugin` which we're using for the build. The included resources are expanded into the target/classes folder in this case.

erwint:
If you think it's preferable we could try to modify the handling in the `bnd-maven-plugin` to either never overwrite, or just when you have a `${bin}` in the `-includeresource`

bjhargrave:
For `bnd-maven-plugin` use `${project.output}/classes`.

bjhargrave:
Well, that you are using `bnd-maven-plugin` is an important detail you didn't mention initially.

It sets the target/class folder as the initial input. So it is effectively always added first to the bundle.

erwint:
Thanks for the suggestion, I wasn't aware of that. Indeed, it works like this:
```
-includeresource: ...,\
 ${project.output}/classes/,\
...
 OSGI-OPT/src=${project.sourcepath}
```

Then - even better - no change at all is needed.

Issue #1293, opened 2016-01-25
Maven plugin should support alternative bnd.bnd locations


paulbakker:
I'm helping the Vert.x team to osgi-ify their modules using the bnd maven plugin. They *really* want the bnd.bnd file in an alternative location though, so the plugin should support this. PR on the way.

bjhargrave:
Is this issue still relevant now?

paulbakker:
I'm ok with closing it, you convinced me this is too much effort.

PR #1294, opened 2016-01-25
#1293 Added configuration option for setting an alternative bnd.bnd location

paulbakker:
See issue #1293

Usage: 

```
<configuration>
     <bndLocation>src/main/osgi/bnd.bnd</bndLocation>
</configuration>
```

Signed-off-by: Paul Bakker <paul.bakker@luminis.eu>

### DIFF
--- a/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
+++ b/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
@@ -81,6 +81,9 @@ public class BndMavenPlugin extends AbstractMojo {
 	@Parameter(defaultValue = "${settings}", readonly = true)
 	private Settings			settings;
 
+	@Parameter(readonly = true)
+	private String				bndLocation;
+
 	@Component
 	private BuildContext		buildContext;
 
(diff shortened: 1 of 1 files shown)


maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -185,7 +188,13 @@ private void loadProjectProperties(Builder builder, MavenProject project) throws
 
 		// Merge in current project properties
 		File baseDir = project.getBasedir();
-		File bndFile = new File(baseDir, Project.BNDFILE);
+		File bndFile;
+		if(bndLocation != null) {
+			bndFile = new File(bndLocation);
```
bjhargrave: This won't work. `loadProjectProperties` is recursively called to load the bnd file of the parent project. So the configuration of the location of the bnd file in _this_ project may have no relation to the location of the bnd file in all the parent projects.

Note: It is also possible that a parent project may have a bnd file yet itself not be configured for the bnd-maven-plugin. In this way, they can provide common bnd settings for child projects.

rotty3000: Can't they just use an -include directive to achieve their goal of putting
the file where they like?

paulbakker: Ok, I'm clearly missing a lot of details about the plugin. Any hints on how this could be implemented correctly?

paulbakker:
Closing this, after more discussion it turned out to be much more complex to do this.

PR #1297, opened 2016-01-26
Added support for ${packages;versioned}

paulbakker:
As discussed with @pkriens, this adds support for `${packages;exported}`. Useful when wrapping other bundles, and the exported packages of that bundle should be exported by the wrapper as well.

Signed-off-by: Paul Bakker <paul.bakker@luminis.eu>

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/MacroTest.java
+++ b/biz.aQute.bndlib.tests/src/test/MacroTest.java
@@ -1121,21 +1121,23 @@ public class MacroTest extends TestCase {
 				new Jar(IO.getFile("bin"))
 		});
 		b.setProperty("Private-Package",
-				"test.packageinfo.annotated,test.packageinfo.notannotated,test.packageinfo.nopackageinfo");
-
+				"test.packageinfo.annotated,test.packageinfo.notannotated,test.packageinfo.nopackageinfo,test.activator");
 		b.setProperty("All-Packages", "${packages}");
 		b.setProperty("Annotated", "${packages;annotated;test.packageinfo.annotated.BlahAnnotation}");
 		b.setProperty("Named", "${packages;named;*.notannotated}");
 		b.setProperty("Negated", "${packages;named;!*.no*}");
+		b.setProperty("Versioned", "${packages;versioned}");
 		b.build();
 
 		assertEquals(0, b.getErrors().size());
 
-		assertEquals("test.packageinfo.annotated,test.packageinfo.notannotated,test.packageinfo.nopackageinfo",
+		assertEquals(
+				"test.packageinfo.annotated,test.packageinfo.notannotated,test.packageinfo.nopackageinfo,test.activator",
 				b.getProperty("All-Packages"));
 		assertEquals("test.packageinfo.annotated", b.getProperty("Annotated"));
 		assertEquals("test.packageinfo.notannotated", b.getProperty("Named"));
-		assertEquals("test.packageinfo.annotated", b.getProperty("Negated"));
+		assertEquals("test.packageinfo.annotated,test.activator", b.getProperty("Negated"));
+		assertEquals("test.packageinfo.annotated,test.packageinfo.notannotated", b.getProperty("Versioned"));
 	}
 
 	public void testBase64() {
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -2562,6 +2562,9 @@ public class Analyzer extends Processor {
 							throw new IllegalArgumentException("Not enough arguments in ${packages} macro");
 						accept = pkgInfo != null && pkgInfo.is(Clazz.QUERY.ANNOTATED, instr, this);
 						break;
+					case VERSIONED :
+						accept = entry.getValue().getVersion() != null;
+						break;
 				}
 			} else {
 				accept = true;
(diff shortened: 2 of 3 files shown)


biz.aQute.bndlib.tests/src/test/MacroTest.java
```diff
@@ -1121,21 +1121,23 @@ public static void testPackagesMacro() throws Exception {
...
 
-		assertEquals("test.packageinfo.annotated,test.packageinfo.notannotated,test.packageinfo.nopackageinfo",
+		assertEquals(
+				"test.packageinfo.annotated,test.packageinfo.notannotated,test.packageinfo.nopackageinfo,test.activator",
 				b.getProperty("All-Packages"));
 		assertEquals("test.packageinfo.annotated", b.getProperty("Annotated"));
 		assertEquals("test.packageinfo.notannotated", b.getProperty("Named"));
-		assertEquals("test.packageinfo.annotated", b.getProperty("Negated"));
+		assertEquals("test.packageinfo.annotated,test.activator", b.getProperty("Negated"));
+		assertFalse(b.getProperty("Exported").contains("test.activator"));
```
bjhargrave: This should be an assertEquals which will provide a useful message when there is a failure. assertFalse doesn't provide any useful message.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -2562,6 +2562,9 @@ public String _packages(String... args) throws Exception {
 							throw new IllegalArgumentException("Not enough arguments in ${packages} macro");
 						accept = pkgInfo != null && pkgInfo.is(Clazz.QUERY.ANNOTATED, instr, this);
 						break;
+					case EXPORTED :
+						accept = entry.getValue().getVersion() != null;
```
bjhargrave: Is this really the way to know the package is exported? Couldn't a non-exported package have a version? Perhaps we need a different macro which is invoked on the exported package set. This macro is invoked on the contained package set.

bjhargrave:
Doesn't the existing [`exports`](http://bnd.bndtools.org/macros/exports.html) macro do what you want?

paulbakker:
@bjhargrave I wouldn't know how to do this with the exports macro, but I could be missing something.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
paulbakker: It could, but it's very unlikely. Got this hint from @pkriens, which probably means there's not currently a better way to know if a package is exported.

bjhargrave:
`${exports}` will produce a comma separated list of the exported packages.

paulbakker:
Something like `Export-Package: my.custom.package, ${exports}` doesn't seem to work, the whole macro doesn't get replaced (but it's in the source for a long time already according to git).

pkriens:
@bj: I think the exported macro is what is exported for THIS bundle. The packages macro checks if they are exported by the buildpath

bjhargrave:
> @bj: I think the exported macro is what is exported for THIS bundle. The packages macro checks if they are exported by the buildpath

The macro as written in this PR is the list of _contained_ packages in this bundle which have a non-null version.  I guess the use case is unclear but the current proposal seems rather wrong.

bjhargrave:
> Something like Export-Package: my.custom.package, ${exports} doesn't seem to work, the whole macro doesn't get replaced (but it's in the source for a long time already according to git).

This won't work since the Export-Package header defines the exported package which are returned by the exports macro :-)

bjhargrave:
> These package are not exported. They just have versions. So calling the macro: ${packages;exported} is rather misleading. At best it should be called: ${packages;versioned}.

Then you could do: `-exportcontents: ${packages;versioned}`.

paulbakker:
@bjhargrave, I agree with the name change. I would rather have an implementation that *really* looks for exported packages though, so if you have any other suggestions...
If not, I'll change the name at least.

bjhargrave:
> I would rather have an implementation that really looks for exported packages though, so if you have any other suggestions...
 
There is a bit of a cycle here. The `Export-Package` or `-exportcontents` instructions set the packages exported by the bundle you are making. It sound like you want a macro that will return the packaged exported by  a specified jar: `${exported;somejar.jar}`.

pkriens:
@BJ: the code from Paul takes the contained packages and checks if they are exported by a JAR on the -buildpath. Taken the version for this seems reasonable because If they were not exported, they would not have a version. So I think EXPORTED is ok imho.

bjhargrave:
> So I think EXPORTED is ok imho.

It is not because his test case is wrong. It passes if the package has a version. It does not care if the package was exported by a jar on the -buildpath.

paulbakker:
Let me explain the use case better: For convenience I'm creating a all-in-one bundle that wraps several other bundles using the following ```

```-includeresource: \
	@${repo;io.vertx.core;3.2.0},\
	@${repo;com.fasterxml.jackson.core.jackson-databind;2.5.4},\
```
The wrapper bundle should export all packages that are exported by the wrapped bundles, using the same versions. Because the wrapped bundles might get updated, it's error prone to manually write the exports, so I want bnd to generate them based on the packages that the bundle inherits from the included bundles.

bjhargrave:
I spoke with Peter. We are OK with your change if called `${packages;versioned}`. Then you can do:

```
-includeresource: \
 @${repo;io.vertx.core;3.2.0},\
 @${repo;com.fasterxml.jackson.core.jackson-databind;2.5.4},\
 ...
-exportcontents: ${packages;versioned}
```

So please make this change and squash the commits.

paulbakker:
Renamed to VERSIONED, squased and updated the docs: https://github.com/bndtools/bnd.manual/blob/gh-pages/_macros/packages.md

bjhargrave:
LGTM

PR #1298, opened 2016-02-02
Close jar file streams that were set as classpath when closing

TwoStone:
This is primary a problem under windows in combination with the gradle-bundle-plugin and the gradle daemon. I was not able to write some test for verification, but it is obvious, that files which are added by the setClasspath method are not registered for close.


bjhargrave:
Please amend your commit to add a Signed-off-by to your commit message. See https://github.com/bndtools/bnd/blob/master/CONTRIBUTING.md#sign-your-work

TwoStone:
Sorry, missed that. The commit is signed now.

PR #1314, opened 2016-02-10
Avoid UnsupportedOperationException in resolver

njbartlett:
Should fix bndtools bug 1289


bjhargrave:
Fixes https://github.com/bndtools/bndtools/issues/1289.

PR #1316, opened 2016-02-11
[bnd-maven-plugin] Improve compatibility with m2e

erwint:
According to https://wiki.eclipse.org/M2E_compatible_maven_plugins a m2e
compatible Plugin must use BuildContext to skip execution when there is
no relevant workspace change.

Indeed, without this change we noticed build cycles to occur on very
complex projects.

### DIFF
--- a/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
+++ b/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
@@ -42,6 +42,7 @@ import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.plugins.annotations.ResolutionScope;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.settings.Settings;
+import org.codehaus.plexus.util.Scanner;
 import org.sonatype.plexus.build.incremental.BuildContext;
 
 import aQute.bnd.build.Project;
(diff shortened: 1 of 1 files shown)


bjhargrave:
I am not sure this does the right thing. It basically looks only at change in the project base dir (minus project target dir). Since there is bnd file inheritance from parent projects and bnd supports including bnd files from anywhere, any changes to those files must also result in a rebuild.

There is also the issue of bundle content included by the bnd instructions. Bnd can pull in packages and resources from any project dependency into the resulting bundle. If they are changed for any reason, a rebuild is needed.

erwint:
I don't think there is a way to handle changed resources outside the project directory, as they are not reported using the BuildContext for the project. The linked page clearly says that a m2e compatible plugin *MUST* use this API.

You can still trigger a (non-incremental) build in case of more complex changes. Also, we could make this either opt-in or opt-out through a property/directive. So in such scenarious you can choose if you prefer possible build lockups or possible needs of manually triggered rebuilds.

bjhargrave:
> I don't think there is a way to handle changed resources outside the project directory, as they are not reported using the BuildContext for the project. The linked page clearly says that a m2e compatible plugin MUST use this API.

I think that is a bit over the top. To the extent that the input files are in the project base dir, you can use that API to determine if they have changed. But that should not be the sole determination of whether to rebuild since there can be inputs outside of the files in the project base dir such as those possibly used by the bnd maven plugin: included files, inherited bnd.bnd files from ancestor projects, and all project dependencies.

> You can still trigger a (non-incremental) build in case of more complex changes. Also, we could make this either opt-in or opt-out through a property/directive. So in such scenarios you can choose if you prefer possible build lockups or possible needs of manually triggered rebuilds.

I don't think we should have an opt in or out. To do this properly, we need to know the full list of files (assuming it is reasonable to compute). If all the files are in the project base dir (which is basically unlikely), then was can use the API to ask if they have changed. But if any file is outside of the project base dir, then the API cannot help and we must rebuild since we cannot know if the files are unchanged.

erwint:
For a more complete processing you'd need to:

1) process the -includeresource directive, either by duplicating code from Builder or refactoring it to reuse some of the private methods
2) then still it could by a command or literal. In case of doubt a build must not be triggered IMHO, as this might then again lock up the build. It's even worse, as the constant rebuilding brings your computer to a crawl, at least on windows. 
3) a dependency change will either be a change in your POM (-> reported and handled), or from some parent POM (-> reported, not handled). You could probably roll your own change tracking through `BuildContext.get/setValue()`, but your dependency can be a directory, so you'd need to scan this for changes. 
4) if a file outside the project is changed, you won't even receive an incremental build trigger at all

With more effort you could solve 1) and maybe partially 3) above, but the others are not solvable with a maven plugin IMHO.

erwint:
Please have a look at this: https://github.com/bndtools/bnd/compare/master...GEBIT:bnd-maven-plugin-m2e-incremental-31?expand=1

Here I used a different approach, which tracks all files added to the Jar and also some more stuff (classpath changes, bnd and POM changes). 

There is more overhead with this as for an aborted incremental build still all input aQute.bnd.osgi.Jars need to be created/scanned. The full rebuild time for a ~400 project Workspace went up from about 6:00 to 7:30 minutes, but that might be tolerable.

bjhargrave:
> For a more complete processing you'd need to:
> 
> 1) process the -includeresource directive, either by duplicating code from Builder or refactoring it to reuse some of the private methods
> 2) then still it could by a command or literal. In case of doubt a build must not be triggered IMHO, 

Actually, in case of doubt, you must allow bnd to build the bundle. It is better to do too much work than to fail to do work which is necessary.

> as this might then again lock up the build. It's even worse, as the constant rebuilding brings your computer to a crawl, at least on windows.
> 3) a dependency change will either be a change in your POM (-> reported and handled), or from some parent POM (-> reported, not handled). You could probably roll your own change tracking through BuildContext.get/setValue(), but your dependency can be a directory, so you'd need to scan this for changes.
> 4) if a file outside the project is changed, you won't even receive an incremental build trigger at all

bnd maven plugin has to work, first and foremost, in the normal command line maven build. To the extent we can safely reduce work in an m2e incremental build, we should. But not at the risk of failing to properly build. Again, it is better to do too much work than to fail to build the bundle correctly when it should have been built.

> 
> With more effort you could solve 1) and maybe partially 3) above, but the others are not solvable with a maven plugin IMHO.

I would agree. So I think we are now in the state, for m2e incremental builds, where bundles are built correctly but at the expense of sometimes doing unnecessary work. We may not be able to further improve this because of how bnd includes content in a bundle and how maven itself works.

erwint:
> bnd maven plugin has to work, first and foremost, in the normal command line maven build. To the extent we can safely reduce work in an m2e incremental build, we should. But not at the risk of failing to properly build. Again, it is better to do too much work than to fail to build the bundle correctly when it should have been built.

While I generally agree (non-incremental build are not affected by this change at  all), I do disagree in this point:

>> 2) then still it could by a command or literal. In case of doubt a build must not be triggered IMHO,

> Actually, in case of doubt, you must allow bnd to build the bundle. It is better to do too much work than to fail to do work which is necessary.

I'd much prefer my build to not lock up (-> not usable) than to maybe have an outdated result (-> usable in most cases, otherwise I can clean it up). This argument could be a bit academic, as I doubt these two features are used often, anyways ;)

bjhargrave:
Please see https://github.com/bndtools/bnd/pull/1319 which I think will do what you want without having to deal with subclassing Builder or having different behavior depending upon an incremental build.

bjhargrave:
I merged #1319 into master for 3.2.

erwint:
I tried out your patch and it seems to resolve the problem. If I discover any problems with it I will leave a comment on #1319

PR #1319, opened 2016-02-12
bnd-maven-plugin: Avoid writing output if not out-of-date

bjhargrave:
The Builder collects that latest modified time of it source material
such as bnd files and included files. The Jar collects the latest
modified time of all its contents including the Builder which built it.

So we modify the expandJar method to only write files out when either the
file does not exist or the file is older than the Jar last modified time.
For incremental builds, this should avoid writing any output files if
none of the input material to the built Jar has changed. Thus we avoid
downstream dependencies from rebuilding because unchanged files were
rewritten.

Alternative to bndtools#1316

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -1332,13 +1332,12 @@ public class Analyzer extends Processor {
 	 * Set the JAR file we are going to work in. This will read the JAR in
 	 * memory.
 	 * 
-	 * @param jar
+	 * @param file
 	 * @throws IOException
 	 */
-	public Jar setJar(File jar) throws IOException {
-		Jar jarx = new Jar(jar);
-		addClose(jarx);
-		return setJar(jarx);
+	public Jar setJar(File file) throws IOException {
+		Jar jar = new Jar(file);
+		return setJar(jar);
 	}
 
 	/**
@@ -2151,8 +2150,7 @@ public class Analyzer extends Processor {
 		if (!cp.exists())
 			warning("File on classpath that does not exist: " + cp);
 		Jar jar = new Jar(cp);
-		addClose(jar);
-		classpath.add(jar);
+		addClasspath(jar);
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


erwint:
I'm afraid this does not resolve the issue (see  bndtools#1316). I'm again experiencing extremely high system loads and excessive build times in m2e. I'll try to look into this next week

bjhargrave:
I added some more debug logging to help you look into what files are being written by the plugin. <https://github.com/bndtools/bnd/commit/746fbd64fbe81de9b24b87b2ee7de10522925f1b>
So run with maven debug logging activated.

erwint:
I had a look into it yesterday, these are my findings:

* Unfortunately, after this change the plugin still does too much, as it essentially does all the processing (looking through jars, processing, ...) and then in the end maybe decides it does not need to output anything. In my test environment with about 400 bundles it's killing my machine (i7-4790K with lots of RAM and a SSD), which will build for about 15 Minutes with 100% CPU and high system load (probably due to many FS operations) - that makes my system basically unusable during this time. I guess an incremental build is triggered quite often during a full workspace build "just in case" assuming it will quickly decide if there is anything to be done (https://wiki.eclipse.org/M2E_compatible_maven_plugins),  but with bnd-maven-plugin this is just too expensive

* The check in https://github.com/bndtools/bnd/blob/master/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java#L233 essentialy causes all files to be written if only one is changed, as `outFile.lastModified()` (the time the last copy took place) is probably older then the last modification time of the newest change. So you can either drop that check or compare with the source file modification time (which might pose other problems with VCS that track the update times, so updated files are not necessarily newer than the last build time)

* I tried a combination of bndtools#1316 with this, where I looked at the source material and set a property in the `BuildContext` if all the source material is from the project. It did not work very good, as the state is only remembered across incremental builds, so when I set this flag from a non-incremental build it would not be remembered. But it gave me a new idea:

* What do you think about combining bndtools#1316 with this in such a way, that you need to set a special property to shortcut the check to only files from the project and pom.xml/bnd.bnd from parents, and then on a full build we could verify that it's ok to use this flag by looking at the actual source material. So incremental build could be very fast, but it you use source material from outside the project the next non-incremental and all command line builds would fail with an appropriate error? If you're open to the idea or have some alternative, I'd be happy to provide a PR.

bjhargrave:
Lets move discussion to #1333 since this PR is closed.

PR #1333, opened 2016-02-24
bnd-maven-plugin: Avoid building jar in memory if possible

bjhargrave:
The BuildContext.hasDelta method is used to find if any project source
has changed since the last build. The jar is built only if there is a
delta or the builder inputs are newer than the target/classes dir.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -2131,6 +2131,7 @@ public class Analyzer extends Processor {
 
 		addClose(jar);
 		classpath.add(jar);
+		updateModified(jar.lastModified(), jar.toString());
 	}
 
 	public void addClasspath(Collection< ? > jars) throws IOException {
--- a/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
+++ b/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
@@ -67,6 +67,9 @@ public class BndMavenPlugin extends AbstractMojo {
 	@Parameter(defaultValue = "${project.build.sourceDirectory}", readonly = true)
 	private File				sourceDir;
 
+	@Parameter(defaultValue = "${project.build.resources}", readonly = true)
+	private List<org.apache.maven.model.Resource>	resources;
+
 	@Parameter(defaultValue = "${project.build.outputDirectory}", readonly = true)
 	private File				classesDir;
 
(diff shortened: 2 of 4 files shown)


bjhargrave:
From https://github.com/bndtools/bnd/pull/1319#issuecomment-188164409

> I had a look into it yesterday, these are my findings:
> 
>     Unfortunately, after this change the plugin still does too much, as it essentially does all the processing (looking through jars, processing, ...) and then in the end maybe decides it does not need to output anything. In my test environment with about 400 bundles it's killing my machine (i7-4790K with lots of RAM and a SSD), which will build for about 15 Minutes with 100% CPU and high system load (probably due to many FS operations) - that makes my system basically unusable during this time. I guess an incremental build is triggered quite often during a full workspace build "just in case" assuming it will quickly decide if there is anything to be done (https://wiki.eclipse.org/M2E_compatible_maven_plugins), but with bnd-maven-plugin this is just too expensive
> 
>     The check in https://github.com/bndtools/bnd/blob/master/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java#L233 essentialy causes all files to be written if only one is changed, as outFile.lastModified() (the time the last copy took place) 

It does not matter. If one file is out of date, then you might as well write them all.

> is probably older then the last modification time of the newest change. So you can either drop that check or compare with the source file modification time (which might pose other problems with VCS that track the update times, so updated files are not necessarily newer than the last build time)
> 
>     I tried a combination of bndtools#1316 with this, where I looked at the source material and set a property in the BuildContext if all the source material is from the project. It did not work very good, as the state is only remembered across incremental builds, so when I set this flag from a non-incremental build it would not be remembered. But it gave me a new idea:
> 
>     What do you think about combining bndtools#1316 with this in such a way, that you need to set a special property to shortcut the check to only files from the project and pom.xml/bnd.bnd from parents, and then on a full build we could verify that it's ok to use this flag by looking at the actual source material. So incremental build could be very fast, but it you use source material from outside the project the next non-incremental and all command line builds would fail with an appropriate error? If you're open to the idea or have some alternative, I'd be happy to provide a PR.

I made this PR to try an avoid building the jar in memory if there are no inputs changes.

It is still critical to build correctly. We should never fail to build correctly even for performance. 

But try out this change. It includes the classpath jars in the builders last modified time and uses buildContext to detect changes in the project source since the last build.

erwint:
sorry for replying late. This change makes it better, but a considerable amount of time is spent on creating a aQute.bnd.osgi.Jar  (https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L1306)

If you defer setting the buildpath (which implicitly creates the Jar instances) and instead just call builder.updateModified with the Jar's lastModified (a Jar should never be older than any of it's content) or in case of a directory with the one from the newest file in that directory - it reduces buildtimes in my setup by about 15%. 

Also: a resource path may be the same as the source path, in this case there is no need to add it to the source directories.

erwint:
As you do not check for incremental builds, now the Manifest is not always generated. This happens because a default Manifest is generated by the archive plugin.

bjhargrave:
This PR is merged. If there are new problems, please open new issues. Thanks.

PR #1336, opened 2016-02-26
Default Bundle-SymbolicName now comes from the Maven project's artifa…

fakeh:
As discussed in https://groups.google.com/forum/#!topic/bndtools-users/u8CBmc_bojY, I have patched and tested (for my simple case) a change making the Bundle-SymbolicName instead default to the artifactId when built using the bad-maven-plugin.
I have not made a pull request before, so I've just accepted GitHub's defaults, which seem to make sense. Apologies if not correct.
Best, Dan.

### DIFF
--- a/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
+++ b/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
@@ -160,6 +160,12 @@ public class BndMavenPlugin extends AbstractMojo {
 				log.debug("builder sourcepath buildContext.hasDelta: " + delta);
 			}
 
+			// Set Bundle-SymbolicName
+			String bundleSymbolicName = builder.getProperty(Constants.BUNDLE_SYMBOLICNAME);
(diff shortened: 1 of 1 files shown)


bjhargrave:
It looks fine but you failed to sign your commit properly. Please amend your commit adding the required Signed-off-by to the commit message (see <https://github.com/bndtools/bnd/blob/master/CONTRIBUTING.md#sign-your-work>) and force push.

fakeh:
Thanks,
Sorry for not reading the regulations properly! I'd like to amend the commit to also include setting of Bundle-Name from the Maven project. Would that be OK too?
Dan.

bjhargrave:
> I'd like to amend the commit to also include setting of Bundle-Name from the Maven project.

What would the Bundle-Name default to?

fakeh:
The Maven project's name, if set.

bjhargrave:
> The Maven project's name, if set.

OK

bjhargrave:
Do you plan to update this PR? Or should I just add these changes myself?

fakeh:
Hi BJ,
I came to look at this again today and found you've already done it. Thanks.
Dan.

PR #1371, opened 2016-03-30
README - both Workspace and non-Workspace

paulvi:
it is not clear right from the beginning that both Workspace and non-Workspace Builds are supported


bjhargrave:
Thanks for the suggestion. I made slightly more extensive changes though.

paulvi:
It looks awesome now. Thank you

PR #1379, opened 2016-04-04
Maven Bnd Repository

### LINKED ISSUE #1352: support globbing during includeresource's buildpath reference resolution
BND supports a very nice ability to resolve jar files referred to in the include resource directive from the classpath (build path).

e.g.
```
-buildpath:\
    /some/dir/somelib-3.0.6.jar;version=file

-includeresource: @somelib-3.0.6.jar
```
This is a wonderful feature.

But it causes duplication of volatile information, in this case the version.

It would be really nice if this mechanism could resolve the same artifact through globbing, like so:

`-includeresource: @somelib-*.jar` OR
`-includeresource: @somelib*.jar`

bjhargrave:
This commit introduces the Maven Bnd Repository. This repository was
the first project that heavily used the HttpClient so a number of changes
are in the Http Client code and its API package. Mainly TaggedData has
grown. Further, methods were added to aQute.lib for working with
hex, strings, and urls.

It slightly modifies the Project class. Project now passes a context in
put so that the Maven bnd Repo can pickup the -maven-release function.
It also added a -installrepo instruction. Any build will be released as
to the repos listed in there.

It also accidentally includes the fix to #1352

Fixes #1352


bjhargrave:
Replaced by #1380

Issue #1385, opened 2016-04-05
bnd-indexer-gradle-plugin Docs and intro


paulvi:
https://github.com/bndtools/bnd/blob/master/maven/bnd-indexer-maven-plugin/README.md mentions OSGi repository system.

When using Gradle, how one can build one?

Please give more information about setting up OSGi repository system.

ref #1109  #1108 #1181 #997 #408

timothyjward:
The maven indexer that you reference is designed to automatically build and publish an index for a set of pom dependencies. We have looked and it is _not possible_ to do this in gradle (issue #1288). In gradle you will need to refer to existing local or fixed repositories. Work that is currently happening to improve maven/nexus integration (PR #1380)  should also help to provide scoped views in gradle.

paulvi:
@timothyjward  What is fixed repository? What is Maven Bnd Repository and how it differs from Maven (e.g. local) repository?

pkriens:
Be patient ...

timothyjward:
> @timothyjward What is fixed repository? What is Maven Bnd Repository and how it differs from Maven (e.g. local) repository?

Repositories (http://bndtools.org/repositories.html) are used as part of a workspace (http://bnd.bndtools.org/chapters/130-concepts.html#the-workspace) I strongly suggest that you try creating some basic bundles using bndtools and its templates. This will set you up with some out of the box repositories and you can try playing around with them. 

> I am confused.

Trying to work out what's going on at the sharp edge of bnd's HEAD commit before you grok the basics of OSGi repositories is unlikely to help with that. Creating and playing with a simple template workspace in bndtools will be a much better way to learn about repositories before jumping in at the deep end of pushing to maven/nexus.

pkriens:
Yup, I agree. If you're just starting, start with [OSGi enRoute][1]. There are lots of tutorials, articles, and app notes to get make you familiar. 

[1]: http://enroute.osgi.org/book/150-tutorials.html

bjhargrave:
This issue does not seem to mention a bug in bnd which can be addressed. Usage questions about bnd should be added in the bndtools-users mail list. See <https://github.com/bndtools/bnd#feedback>.


https://github.com/bndtools/bnd/issues/1288 already discussed making a gradle plugin to generate a bnd index. However research shows this is not possible in gradle since gradle does not expose the source urls for dependencies.

PR #1395, opened 2016-04-06
#1385 repositories and workspaces links

paulvi:
@bjhargrave 

#1385 give repositories and workspaces links in README `bnd-indexer-gradle-plugin` section

### DIFF
--- a/maven/bnd-indexer-maven-plugin/README.md
+++ b/maven/bnd-indexer-maven-plugin/README.md
@@ -1,7 +1,7 @@
 # bnd-indexer-maven-plugin
 
 The `bnd-indexer-maven-plugin` is a bnd based plugin that creates OSGi 
-indexes of bundles.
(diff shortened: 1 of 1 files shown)


bjhargrave:
You should use proper markdown links rather that just URLs inline. 

Also, You need to amend your commits so that each commit message includes a `Signed-off-by` as required by CONTRIBUTING. Then push force your updated commits. Thanks.

> #1385 give repositories and workspaces links in README bnd-indexer-gradle-plugin section

You have modified the README for the _maven_ plugin. There is no gradle indexer plugin.

bjhargrave:
Anyway, the maven plugins do not work with workspaces, so we should not mention then here or workspace repositories. 

The maven plugins are for maven style builds which are not bnd workspace-shaped.

paulvi:
@bjhargrave Thank for additional comments. Can this information be clearly seen on http://bnd.bndtools.org/chapters/700-tools.html page?

Issue #1350, opened 2016-03-15
Replace macro content with same delimiters


dleangen:
When bnd detects macro delimiters, but then does not recognize a macro, it replaces the macro body with the original content. However, it may use different delimiters when rewriting the content.

For example, in javascript:

var body = $('body:first');

Is being replaced by:

var body = ${'body:first'};

This is breaking the javascript.

PR #1396, opened 2016-04-06
[macro] Literal expansion

pkriens:
The ${range} macro created an error when it was used without the @ target set. Added a feature to indicate to the macro processor that the replacement should be the literal text of the macro so it can be expanded again later.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
@@ -60,20 +60,21 @@ import aQute.libg.glob.Glob;
  * pattern. ${parameter##word} Remove largest prefix pattern.
  */
 public class Macro {
-	final static String		NULLVALUE	= "c29e43048791e250dfd5723e7b8aa048df802c9262cfa8fbc4475b2e392a8ad2";
-	final static Pattern	NUMERIC_P	= Pattern.compile("[-+]?(\\d*\\.?\\d+|\\d+\\.)(e[-+]?[0-9]+)?");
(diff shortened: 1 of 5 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
```diff
@@ -933,6 +952,10 @@ public String _range(String args[]) {
 		return sb.toString();
 	}
 
+	boolean isLocalTarget(String string) {
+		return string.matches("\\$(\\{@\\}|\\[@\\]|\\(@\\)|<@>|«@»|‹@›)");
```
bjhargrave: Does this also need to handle `$@`?

biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
```diff
@@ -899,12 +912,18 @@ String version(Version version, String mask) {
...
+			if (isLocalTarget(string))
+				return LITERALVALUE;
+
+			version = new Version(string);
+		}
 		else {
 			String v = domain.getProperty("@");
-			if (v == null)
-				return null;
+			if (v == null || v.contains("@"))
```
bjhargrave: I think you need to remove the `contains("@")` test

bjhargrave:
Your commit comment should also note that this fixes https://github.com/bndtools/bnd/issues/1350.

biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
pkriens: As far as I can see $@ is illegal. You always need to encloses it.

PR #1400, opened 2016-04-07
fix null artifact file uploading

mrulli:
fixes the problem described here: https://groups.google.com/d/topic/bndtools-users/H2iRlTUDBPA/discussion


bjhargrave:

bjhargrave:
Thanks!

PR #1414, opened 2016-04-21
Correct check for fragment even if PackageAdmin is unavailable

tverbele:
Since PackageAdmin is deprecated, this is for example not included by default in the Concierge OSGi framework. This results in errors during launching since the bnd Launcher relies on PackageAdmin only to check whether a Bundle is a fragment. This can be easily resolved by checking for osgi.wiring.host requirements when PackageAdmin is unavailable.

### DIFF
--- a/biz.aQute.launcher/src/aQute/launcher/Launcher.java
+++ b/biz.aQute.launcher/src/aQute/launcher/Launcher.java
@@ -63,11 +63,14 @@ import org.osgi.framework.ServiceListener;
 import org.osgi.framework.ServiceReference;
 import org.osgi.framework.launch.Framework;
 import org.osgi.framework.launch.FrameworkFactory;
+import org.osgi.framework.wiring.BundleRevision;
+import org.osgi.service.packageadmin.PackageAdmin;
 import org.osgi.service.permissionadmin.PermissionInfo;
 
 import aQute.launcher.agent.LauncherAgent;
 import aQute.launcher.constants.LauncherConstants;
 import aQute.launcher.minifw.MiniFramework;
+
 
 /**
  * This is the primary bnd launcher. It implements a launcher that runs on Java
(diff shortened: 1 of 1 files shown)


bjhargrave:
Can you rebase your change? The PR shows merge conflicts and it did not build.

tverbele:
Oops, did a fetch upstream but forgot to merge in first. Rebased now

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -754,8 +757,12 @@ private void doSecurity() {
 
 	@SuppressWarnings("deprecation")
 	private boolean isFragment(Bundle b) {
-		return padmin != null
-				&& padmin.getBundleType(b) == org.osgi.service.packageadmin.PackageAdmin.BUNDLE_TYPE_FRAGMENT;
+		if(padmin != null){
+			return padmin.getBundleType(b) == PackageAdmin.BUNDLE_TYPE_FRAGMENT;
+		} else {
+			BundleRevision rev = b.adapt(BundleRevision.class);
+			return !rev.getDeclaredRequirements("osgi.wiring.host").isEmpty();
```
bjhargrave: This is the wrong test. According to the Core spec (section 7.4), the test should be:

```
BundleRevision rev = aBundle.adapt(BundleRevision.class);
if ( rev != null && (rev.getTypes() & BundleRevision.TYPE_FRAGMENT)!= 0 ) {
     ... // do the fragment thing
}
```

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -754,8 +757,12 @@ private void doSecurity() {
 
 	@SuppressWarnings("deprecation")
 	private boolean isFragment(Bundle b) {
-		return padmin != null
-				&& padmin.getBundleType(b) == org.osgi.service.packageadmin.PackageAdmin.BUNDLE_TYPE_FRAGMENT;
+		if(padmin != null){
+			return padmin.getBundleType(b) == PackageAdmin.BUNDLE_TYPE_FRAGMENT;
```
bjhargrave: The launcher is built against Core 4.3.1 which includes the wiring API (i.e. BundleRevision). So just remove the use of PackageAdmin entirely.

bjhargrave:
@tverbele If you want to get this in for 3.2, you should act quickly. Please just replace the use of PackageAdmin with the wiring API.

bjhargrave:
This fix does not do enough to avoid use of Package Admin in other methods. I have a new fix for which I will make a PR shortly.

PR #1420, opened 2016-04-23
Addition of target filtering in DSTestWiring

dleangen:
Thanks for introducing me to this great tool!

I noticed that there seems to be lack of support for the target filter for @References. I thought that I would use this PR as a means of getting a dialogue started around this feature. :-)

### DIFF
--- a/biz.aQute.bndlib.tests/src/aQute/bnd/testing/DSTestWiringTest.java
+++ b/biz.aQute.bndlib.tests/src/aQute/bnd/testing/DSTestWiringTest.java
@@ -21,6 +21,7 @@ public class DSTestWiringTest extends TestCase {
 	private LogService		log;
 	private String			string;
 	private A				a;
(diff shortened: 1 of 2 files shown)


bjhargrave:
Perhaps you should start with just an issue where you describe the problem you are having.

I am unclear what problem you are having here.

dleangen:
Thanks, BJ.

[Issue created](https://github.com/bndtools/bnd/issues/1421).

bjhargrave:
Perhaps we can close this PR?

dleangen:
Yes, indeed. The discussion is happening in [the related issue](https://github.com/bndtools/bnd/issues/1421) as you suggested.

By the way, I didn't realize until know how much effort you are investing in keeping all this stuff lean and mean. It is a very time-consuming and thankless job. Thanks so much for all your efforts! :-)

PR #1429, opened 2016-04-27
DIAFAMPS-4 allowing to not specify package versions

Darnok99:
The problem we encountered is - we don't usually specify ranges nor versions in the import-packages block. We didn't find a way of doing it and this change should provide us such functionality. 
What it does: if you provide an empty consumer-policy your versions won't be automatically generated.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -1713,7 +1713,9 @@ public class Analyzer extends Processor {
 					exportVersion = cleanupVersion(exportVersion);
 
 					importRange = applyVersionPolicy(exportVersion, importRange, provider);
-					importAttributes.put(VERSION_ATTRIBUTE, importRange);
+					if(!importRange.trim().isEmpty()) {
+						importAttributes.put(VERSION_ATTRIBUTE, importRange);
+					}
 				}
 
 				//


bjhargrave:

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -1713,7 +1713,9 @@ void augmentImports(Packages imports, Packages exports) throws Exception {
 					exportVersion = cleanupVersion(exportVersion);
 
 					importRange = applyVersionPolicy(exportVersion, importRange, provider);
-					importAttributes.put(VERSION_ATTRIBUTE, importRange);
+					if(importRange == null || !importRange.trim().isEmpty()) {
```
psoreide: Shouldn't that be: if(importRange != null && !importRange.trim().isEmpty()) {

bjhargrave: I don't see how `applyVersionPolicy` could ever return null. It is possible for it to return an empty string if the input importRange is null and `-provider-policy` or `-consumer-policy` is set to the empty string.

So a null check is not necessary. Just a check that importRange is empty.

bjhargrave:
Can you please squash the commits together? Also put the `Signed-off-by` on a separate line in the commit message. `git commit -s` will properly set this in the commit message. Thanks.

Darnok99:
Perfect, do you have any estimates on when this patch will be released?

2016-04-28 21:22 GMT+10:00 BJ Hargrave <notifications@github.com>:

bjhargrave:
bnd 3.2 should be released in the next 2-3 weeks.

Darnok99:
Awesome, thank you!
Have a great day!

2016-04-29 9:15 GMT+10:00 BJ Hargrave <notifications@github.com>:

Issue #1444, opened 2016-05-04
RepoIndex does not support the org.osgi.service.indexer.ResourceIndexer.STYLESHEET configuration


rotty3000:
When using the repoindex tool and passing the `-stylesheet <URL>` switch, no stylesheet is added to the index. A review of the code shows this feature is simply not implemented.

https://github.com/bndtools/bnd/blob/master/org.osgi.impl.bundle.repoindex.lib/src/org/osgi/service/indexer/impl/RepoIndex.java

PR #1448, opened 2016-05-04
repoindex: Add missing stylesheet support

bjhargrave:
Fixes https://github.com/bndtools/bnd/issues/1444


bjhargrave:
@rotty3000 Is this what you wanted?

rotty3000:
Wow that was fast!

Thx BJ!

PR #1461, opened 2016-05-13
Add skip option to maven plugins

cziegeler:
It's a common practice to enable plugins which are used in most projects in the parent pom. However, sometimes there is a need to not run a specific plugin in one of the projects. Maintaining different parent poms for this is too much work, therefore plugins usually have a skip option which can be enabled in such "special" projects.
This pull request adds the skip option to all three maven plugins


bjhargrave:

cziegeler:
Done. Sorry for not checking this earlier

maven/bnd-baseline-maven-plugin/src/main/java/aQute/bnd/maven/baseline/plugin/BaselineMojo.java
```diff
@@ -61,10 +61,17 @@
 	@Parameter(readonly = true, required = false)
 	private Base					base;
 
+    @Parameter(defaultValue = "false", readonly = true)
```
kwin: Why is this parameter readonly? That makes it impossible to use, because it will always be set to `false`.

Issue #1455, opened 2016-05-06
[3.2 RC1] cannot download bigger artifacts behind corporate-proxy


lostiniceland:
With 3.2 RC1 I get FileNotFoundException when downloading certain JARs from JPM.

This is probably caused by our corporate-proxy (McAffee) which also does virus-scanning. When the download and scanning of a artifact takes longer then the download is redirected. 

`java.io.FileNotFoundException: [http://repo1.maven.org/maven2/org/apache/cxf/cxf-rt-ws-security/3.1.5/cxf-rt-ws-security-3.1.5.jar]
	at aQute.bnd.jpm.StoredRevisionCache.download(StoredRevisionCache.java:131)
	at aQute.bnd.jpm.Repository.scheduleDownload(Repository.java:292)
	at aQute.bnd.jpm.Repository.getLocal(Repository.java:254)
	at aQute.bnd.jpm.Repository.get(Repository.java:237)
	at bndtools.model.repo.RepositoryEntry.getFile(RepositoryEntry.java:117)
	at bndtools.views.repository.RepositoriesView$5$1.run(RepositoriesView.java:313)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:55)`

First I expected that this is something which has to be configured in our proxy, but **going back to BndTools 3.1, the artifact can be downloaded**. Since there were [some changes in bnd regarding proxies](https://groups.google.com/d/msg/bndtools-users/CU0focgd4Ls/RcvsZCMpAwAJ), this must be a regression.

pkriens:
Do you explicitly setup a bnd proxy ( ~/.bnd/settings.xml)?

bjhargrave:
> Do you explicitly setup a bnd proxy ( ~/.bnd/settings.xml)?

See http://bnd.bndtools.org/instructions/connection-settings.html

lostiniceland:
No, I only set the proxy in Eclipse preferences. Will have a look at the new settings...

pkriens:
Would you be able to show a wireshark trace?

lostiniceland:
I will try to provide a trace next week, with the help from a colleague

lostiniceland:
See attached traces. I did one trace with 3.1 which works and another failing one with 3.2RC1. 

Not sure if the broken-trace has enough information, but thats what I could get out of Wireshark. I had a filter for Http and used the result of this to follow the TCP-stream.

[...]

PR #1463, opened 2016-05-13
[http] Increased timeout to 1 min instead of 5 secs

pkriens:


bjhargrave:
Fixes https://github.com/bndtools/bnd/issues/1455

PR #1473, opened 2016-05-26
add a method to check if resource is a fragment

rotty3000:
add a method on ResourceUtil to detect fragments

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
@@ -481,4 +481,12 @@ public class ResourceUtils {
 		return result;
 	}
 
+	public static boolean isFragment(Resource resource) {
+		IdentityCapability identity = getIdentityCapability(resource);
+		if (identity == null)
+			return false;
+		return IdentityNamespace.TYPE_FRAGMENT
+				.equals(identity.getAttributes().get(IdentityNamespace.CAPABILITY_TYPE_ATTRIBUTE));
+	}
+
 }


biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
```diff
@@ -474,4 +474,17 @@ public static String toProvideCapability(Capability cap) throws Exception {
 		return result;
 	}
 
+	public static boolean isFragment(Resource resource) {
+		List<Capability> capabilities = resource.getCapabilities(IdentityNamespace.IDENTITY_NAMESPACE);
+
+		if ((capabilities != null) && (capabilities.size() == 1)
+				&& (capabilities.get(0).getAttributes().get(IdentityNamespace.CAPABILITY_TYPE_ATTRIBUTE).equals(
+						IdentityNamespace.TYPE_FRAGMENT))) {
```
bjhargrave: Isn't this simply return the test in the if? That is `if (something) return true; else return false;` should just be `return something;`

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
```diff
@@ -474,4 +474,17 @@ public static String toProvideCapability(Capability cap) throws Exception {
 		return result;
 	}
 
+	public static boolean isFragment(Resource resource) {
+		List<Capability> capabilities = resource.getCapabilities(IdentityNamespace.IDENTITY_NAMESPACE);
+
+		if ((capabilities != null) && (capabilities.size() == 1)
```
bjhargrave: getCapabilities is specified to never return null. So you can skip that check.

rotty3000: ... signed-off-by

Also, is it true (caps.size() == 1) ? wasn't sure about that

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
rotty3000: it's a matter of what form you like... but changed in update

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
```diff
@@ -474,4 +474,12 @@ public static String toProvideCapability(Capability cap) throws Exception {
 		return result;
 	}
 
+	public static boolean isFragment(Resource resource) {
+		List<Capability> capabilities = resource.getCapabilities(IdentityNamespace.IDENTITY_NAMESPACE);
+
+		return (capabilities.size() == 1)
```
bjhargrave: This is checking input is valid. A resource can only provide a single identity per the spec.

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
```diff
@@ -474,4 +474,12 @@ public static String toProvideCapability(Capability cap) throws Exception {
...
+				&& (capabilities.get(0).getAttributes().get(IdentityNamespace.CAPABILITY_TYPE_ATTRIBUTE).equals(
+						IdentityNamespace.TYPE_FRAGMENT));
```
bjhargrave: Perhaps to avoid NPE, do `IdentityNamespace.TYPE_FRAGMENT.equals(capabilities.get(0).getAttributes().get(IdentityNamespace.CAPABILITY_TYPE_ATTRIBUTE))`

rotty3000: done in update!

rotty3000:
Can I get a LGTM?

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
```diff
@@ -474,4 +474,12 @@ public static String toProvideCapability(Capability cap) throws Exception {
 		return result;
 	}
 
+	public static boolean isFragment(Resource resource) {
+		List<Capability> capabilities = resource.getCapabilities(IdentityNamespace.IDENTITY_NAMESPACE);
```
bjhargrave: OK, Taking a larger look at this class, it seems to me the impl should be 

```java
IdentityCapability identity = getIdentityCapability(resource);
if (identity == null) 
  return false;
return indentity.type() == IdentityCapability.Type.fragment;
```

Then the code here uses the other parts of this class rather than being completely isolated from it.

But then I have no idea if the magic reflection code works correctly :-)

PR #1475, opened 2016-05-26
fixes #1474 - biz.aQute.resolve.ResolverValidator.check(String...) returning true when it should return false

### LINKED ISSUE #1474: biz.aQute.resolve.ResolverValidator.check(String...) returning true when it should return false
In some cases the biz.aQute.resolve.ResolverValidator.check(String...) returns true even though the validation has errors.

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
```diff
@@ -474,4 +474,17 @@ public static String toProvideCapability(Capability cap) throws Exception {
 		return result;
 	}
 
+	public static boolean isFragment(Resource resource) {
```
bjhargrave: This change was already proposed, and improved, in #1473

rotty3000: ah yes, somehow I forgot to prune the commits... fixing

PR #1476, opened 2016-05-26
A hodgepodge ...

pkriens:
A lot of new support for working with resources and repository.
Also moved to R6

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/annotationheaders/AnnotationHeadersTest.java
+++ b/biz.aQute.bndlib.tests/src/test/annotationheaders/AnnotationHeadersTest.java
@@ -30,6 +30,51 @@ import junit.framework.TestCase;
 public class AnnotationHeadersTest extends TestCase {
 
 	/**
+	 * Say I want to define a capability namespace for web applications, e.g.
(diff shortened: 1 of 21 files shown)


biz.aQute.bndlib.tests/src/test/annotationheaders/AnnotationHeadersTest.java
```diff
@@ -30,6 +30,51 @@
...
+		// b.build();
+		// assertTrue(b.check());
+		// b.getJar().getManifest().write(System.out);
+		// Parameters req = new Parameters(
+		// b.getJar().getManifest().getMainAttributes().getValue(Constants.REQUIRE_CAPABILITY));
+		// Parameters cap = new Parameters(
+		// b.getJar().getManifest().getMainAttributes().getValue(Constants.PROVIDE_CAPABILITY));
+		// assertTrue(req.get("foo") != null);
+		// assertTrue(cap.get("foo") != null);
+		// }
```
bjhargrave: What is the point of this test case?

pkriens: Is a bug but I need to fix don't want to kill the build.  #1477

PR #1481, opened 2016-05-30
[maven] Metadata parser failure on Nexus

pkriens:


biz.aQute.repository/src/aQute/maven/provider/MavenRepository.java
```diff
@@ -119,6 +123,9 @@ public void run() {
 					try {
 
 						File f = get0(archive, file);
+						if (f == null)
+							throw new FileNotFoundException("" + archive);
```
bjhargrave: Shouldn't you fail the Promise with the FNFE rather than throw it?

pkriens: exactly .... :-)

PR #1508, opened 2016-06-20
fixes #1507 XMLResourceGenerator produces bad directive entries

### LINKED ISSUE #1507: XMLResourceGenerator produces bad requirement entries
When a requirement contains a directive like resolution the generator produces entries like so:

```xml
      <requirement namespace="osgi.wiring.package">
        <directive name="filter" value="(&amp;(osgi.wiring.package=org.eclipse.jetty.jmx)(&amp;(version&gt;=9.3.9)(!(version&gt;=9.3.10))))"/>
        <directive name="resolution:" value="optional"/>
        <directive name="version" value="[9.3.9,9.3.10)"/>
      </requirement>
```
This prevents the optional from being discarded from the resolution results causing the resolution to fail.

rotty3000:


rotty3000:
@pkriens is going to take a look

pkriens:
This is the one we discussed, I don't think we should take it. I try to find the root cause

rotty3000:
If you take this repo you will see the outcome of the bug https://github.com/rotty3000/com.rotty3000.p2.http

several bundles in the Mars2 P2 repo have optional package imports which are not satisfied within the repo. the bug causes that these bundles can't resolve

PR #1512, opened 2016-06-21
fixes #1511 Domain.domain(File) not handling zip files

### LINKED ISSUE #1511: Domain.domain(File) not handling zip files
When pulling from maven central there are a lot of old artifacts which were not well formed jar files (having the manifest.mf as the first entry in the zip). However, the JVM and even OSGi frameworks seem to handle these just fine.

rotty3000:


rotty3000:
Before this change many of the older artifacts referenced in this repo via maven central could not be identified as OSGi bundles due to the fact that they are technically malformed jar files (the manifest file isn't the first entry) thus the jar API fails to find the manifest even though it exists in the file.

https://github.com/rotty3000/com.rotty3000.ejb

rotty3000:
@pkriens one thing to note!!! You will have to purge your .m2 repo of .bnd subdirs since the already existing metadata will be bad.

the command I used was:
```
find -name .bnd -exec rm -rf '{}' +
```

PR #1514, opened 2016-06-21
fixes #1513 resolving against anything wrapped by aQute.bnd.deployer.repository.wrapper.Plugin is painfully slow

### LINKED ISSUE #1513: resolving against anything wrapped by aQute.bnd.deployer.repository.wrapper.Plugin is painfully slow
The issue seems to be that on every invocation of it's `findProviders(...)` it's init method reindexes every repository that it's wrapping completely.

rotty3000:


rotty3000:
Changes can be treated on this repo https://github.com/rotty3000/com.rotty3000.ejb

Resolution went from 9 minutes to about 15 seconds. Furthermore all update operations on repos such as adding files seemed to still work as expected.

PR #1519, opened 2016-06-21
fixes #1507 XMLResourceGenerator produces bad directive entries

### LINKED ISSUE #1507: XMLResourceGenerator produces bad requirement entries

rotty3000:


rotty3000:
@pkriens I think this is it!

pkriens:
That was the right place but I also had it fixed and slightly more abstracted.

PR #1520, opened 2016-06-21
fixes #1516 Erroneous Manifest for Jar:javax.servlet-3.1.0.v201410161800 java.lang.ClassCastException: java.util.ArrayList cannot be cast to aQute.bnd.version.Version bnd.bnd

### LINKED ISSUE #1516: Erroneous Manifest for Jar:javax.servlet-3.1.0.v201410161800 java.lang.ClassCastException: java.util.ArrayList cannot be cast to aQute.bnd.version.Version	bnd.bnd
It seems that bnd does not properly handle contracts with a list of versions.

The javax.servlet bundle in this repository: http://download.eclipse.org/eclipse/updates/4.6-I-builds/I20160606-1100/
provides the contract:
```
osgi.contract=JavaServlet;uses:="javax.servlet,javax.servlet.annotation,javax.servlet.descriptor,javax.servlet.http";version:List<Version>="2.5,3.0,3.1"
```
However bnd's aQute.bnd.osgi.Contracts.collectContracts(String, Parameters) has:
https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/osgi/Contracts.java#L107
which casts the result to `Version` which fails with:
```
Erroneous Manifest for Jar:javax.servlet-3.1.0.v201410161800 java.lang.ClassCastException: java.util.ArrayList cannot be cast to aQute.bnd.version.Version
```

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/osgi/Contracts.java
```diff
@@ -166,7 +179,21 @@ void addToRequirements(Parameters requirements) {
...
+					for (Version version : c.versions) {
+						sb.append("(version=");
+						sb.append(version);
+						sb.append(")");
+					}
+					sb.append(")");
+					versionString = sb.toString();
+				}
+
+				f.format("(&(%s=%s)%s)", ContractNamespace.CONTRACT_NAMESPACE, c.name, versionString);
```
bjhargrave: This seems wrong. A osgi.contract requirement should be for one version not an or'ed list of versions. If you have more than one version, you need to pick the highest version and use that in the requirement.

rotty3000: right, I wasn't quite sure about it that's why I went this approach. So will the logic work if we simply make N clones of the Contract one per version?

bjhargrave: I don't think we want N requirements, one per version. This would be an AND over the versions.

The consumer of the contract just needs to require the contract at the highest possible version he could be using. Which is the highest version provided by the capability.

rotty3000:
@bjhargrave see updated!

bjhargrave:
LGTM

PR #1525, opened 2016-06-23
[progress] support multiple ProgressPlugins

gamerson:
Signed-off-by: Gregory Amerson <gregory.amerson@liferay.com>

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
@@ -69,7 +69,7 @@ public class HttpClient implements Closeable, URLConnector {
 	private URLCache							cache					= new URLCache(IO.getFile("~/.bnd/urlcache"));
 	private Registry							registry				= null;
 	private Reporter							reporter				= new Slf4jReporter(HttpClient.class);
-	private ProgressPlugin						progress;
+	private List<ProgressPlugin>				progressPlugins;
 
 	public HttpClient() {}
 
@@ -78,7 +78,7 @@ public class HttpClient implements Closeable, URLConnector {
 			return;
 
 		inited = true;
-		progress = registry != null ? registry.getPlugin(ProgressPlugin.class) : null;
+		progressPlugins = registry != null ? registry.getPlugins(ProgressPlugin.class) : null;
 
 		Authenticator.setDefault(new Authenticator() {
 			@Override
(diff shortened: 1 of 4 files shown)


biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -247,29 +247,35 @@ public TaggedData send0(final HttpRequest< ? > request) throws Exception {
...
+			}
 			throw t;
 		}
 	}
 
-	ProgressPlugin.Task getTask(final HttpRequest< ? > request) {
-		final ProgressPlugin.Task task;
-		if (progress != null) {
-			task = progress.startTask("Download " + request.url, 100);
+	List<Task> getTasks(final HttpRequest< ? > request) {
```
gamerson: @pkriens changing this method does that mean I need to bump the package version to 2.0?  for "default" access level methods, that is a breaking change right?

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -78,7 +78,7 @@ synchronized void init() {
 			return;
 
 		inited = true;
-		progress = registry != null ? registry.getPlugin(ProgressPlugin.class) : null;
+		progressPlugins = registry != null ? registry.getPlugins(ProgressPlugin.class) : null;
```
bjhargrave: I am not sure it is a good idea to just collect the list of ProgressPlugins in init since the list can change after that.

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -247,29 +247,35 @@ public TaggedData send0(final HttpRequest< ? > request) throws Exception {
...
 			throw t;
 		}
 	}
 
-	ProgressPlugin.Task getTask(final HttpRequest< ? > request) {
-		final ProgressPlugin.Task task;
-		if (progress != null) {
-			task = progress.startTask("Download " + request.url, 100);
+	List<Task> getTasks(final HttpRequest< ? > request) {
+		final List<Task> task = new ArrayList<>();
```
bjhargrave: This method should query the registry for the set of ProgressPlugins on each call.

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
bjhargrave: One way to minimize the broad effects of supporting multiple ProgressPlugins would be to not return a List<Task> here but instead a single Task which multiplexed the list of tasks. Then most of the other changes become unnecessary.

So if there is zero ProgressPlugins, return the noop Task. If there is only one ProgressPlugin, return the Task for it. If there are more than one ProgressPlugin, return a MultiplexerTask whose methods call all the actual Tasks methods.

bjhargrave: My suggestion would prevent the need to change the signature.

gamerson:
@bjhargrave changed based on your review https://github.com/bndtools/bnd/pull/1527

PR #1527, opened 2016-06-23
[progress] support multiple ProgressPlugins

gamerson:
Signed-off-by: Gregory Amerson <gregory.amerson@liferay.com>

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
@@ -69,7 +69,6 @@ public class HttpClient implements Closeable, URLConnector {
 	private URLCache							cache					= new URLCache(IO.getFile("~/.bnd/urlcache"));
 	private Registry							registry				= null;
 	private Reporter							reporter				= new Slf4jReporter(HttpClient.class);
-	private ProgressPlugin						progress;
 
 	public HttpClient() {}
 
@@ -78,7 +77,6 @@ public class HttpClient implements Closeable, URLConnector {
 			return;
 
 		inited = true;
-		progress = registry != null ? registry.getPlugin(ProgressPlugin.class) : null;
 
 		Authenticator.setDefault(new Authenticator() {
 			@Override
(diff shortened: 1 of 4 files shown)


biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -265,13 +263,49 @@ public TaggedData call() throws Exception {
...
+			final List<ProgressPlugin.Task> multiplexedTasks = new ArrayList<>();
+
+			for (ProgressPlugin progressPlugin : progressPlugins) {
+				multiplexedTasks.add(progressPlugin.startTask(name, size));
+			}
+
 			task = new ProgressPlugin.Task() {
+				@Override
+				public void worked(int units) {
+					for (ProgressPlugin.Task multiplexedTask : multiplexedTasks) {
```
bjhargrave: To be nitpicky, the loop variable should be called `task` so it is the same as the other method's loops.

biz.aQute.repository/src/aQute/bnd/deployer/repository/ProgressWrappingStream.java
```diff
@@ -12,8 +15,60 @@
...
+			final List<ProgressPlugin.Task> multiplexedTasks = new ArrayList<>();
+
+			for (ProgressPlugin progressPlugin : progressPlugins) {
+				multiplexedTasks.add(progressPlugin.startTask(name, size));
+			}
+
+			task = new ProgressPlugin.Task() {
+				@Override
+				public void worked(int units) {
+					for (ProgressPlugin.Task multiplexedTask : multiplexedTasks) {
```
bjhargrave: To be nitpicky, the loop variable should be called `task` so it is the same as the other method's loops.

biz.aQute.repository/src/aQute/bnd/deployer/repository/ProgressWrappingStream.java
```diff
@@ -12,8 +15,60 @@
 	private Task		task;
 
 	public ProgressWrappingStream(InputStream delegate, String name, int size, ProgressPlugin progressPlugin) {
+		this(delegate, name, size, Collections.singletonList(progressPlugin));
```
bjhargrave: What if the arg `progressPlugin` is null? This method creates a single element list with null which will blow up in the other constructor which assumes the list holds real values.

gamerson:
Thanks for review, pushed new commit.

bjhargrave:
LGTM

PR #1535, opened 2016-06-28
Better error handling for -include of URL resources.

### LINKED ISSUE #1534: Innaccurate or misleading error reporting for -include of URL resources
With `-include` of URL resources, sometimes errors are reported to the user when they shouldn't be, and the errors are not as useful as they could be, either. Pull request to follow shortly.

eliasvasylenko:
- When user supplies a valid URL, now reports specific errors relating
to failures in including the named resource.
- No longer gives spurious errors after valid URL includes.

Fixes #1534

Signed-off-by: Elias N Vasylenko <eliasvasylenko@gmail.com>

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
@@ -1020,12 +1020,14 @@ public class Processor extends Domain implements Reporter, Registry, Constants, 
 							} finally {
 								tmp.delete();
 							}
-						} catch (Exception mue) {
-							// ignore
+						} catch (MalformedURLException mue) {
+							if (fileMustExist)
+								error("Included file %s %s", file,
+										(file.isDirectory() ? "is directory" : "does not exist"));
+						} catch (Exception e) {
+							if (fileMustExist)
+								error("Error in processing included URL: %s", e, value);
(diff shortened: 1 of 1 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
```diff
@@ -1020,12 +1020,14 @@ private void doIncludes(File ubase, Properties p) {
 							} finally {
 								tmp.delete();
 							}
-						} catch (Exception mue) {
-							// ignore
+						} catch (MalformedURLException mue) {
+							if (fileMustExist)
+								error("Included file %s %s", file,
+										(file.isDirectory() ? "is directory" : "does not exist"));
+						} catch (Exception e) {
```
bjhargrave: We don't need the catch (Exception) block here since the catch (Exception) block below will cover this.

eliasvasylenko: Sure, I realise that, but I thought having a different error message for that catch encoded enough information to be worth including. I'd be happy to submit another PR without it if preferred?

bjhargrave: OK.

Issue #1547, opened 2016-07-12
aQute.bnd.osgi.Annotation.getAnnotation() doesn't really work


tomq42:
The aQute.bnd.osgi.Annotation.getAnnotation() method is implemented as:

    Class<T> c = (Class<T>) getClass().getClassLoader().loadClass(cname);

This doesn't apparently work for annotations defined outside of the bundle containing the aQute.bnd.osgi.Annotation class and things it depends on. 

I am defining my own annotations, and when I call getAnnotation I get null. I would need the loadClass() call to be made on the class loader that loaded MY class, the one that called getAnnotation, not the class loader that loaded the Annotation class itself.

djencks:
I think in your situation you are intended to call getAnnotation(Class<T> c), since your plugin does have access to the class involved?

thanks
david jencks

tomq42:
That only works though if you already know the type of the annotation you've got.

I've created a PR, #1548, with a suggested fix. This allows you to call getAnnotation(), then use "instanceof" to determine what kind of annotation you've got (like the code in bndlib does), rather than testing the annotation name against strings.

PR #1548, opened 2016-07-12
Allow specification of ClassLoader when calling Annotation.getAnnotat…

tomq42:
…ion()

To avoid having to duplicate the implementation of getAnnotation() in my own class and hence get it to use my own class loader and hence see my annotations, provide a version of getAnnotation that allows you to specify the class loader to use. Delegate the original getAnnotation to the new version, passing in getClass().getClassLoader().

See Issue #1547


bjhargrave:

PR #1549, opened 2016-07-13
Allow specification of ClassLoader when calling Annotation.getAnnotat…

tomq42:
…ion()

In order that getAnnotation() can be used for annotations other than ones built in to bndlib, it is necessary to have a mechanism such as the ability to specify the class loader through which the annotation class is loaded.

Signed-off-by: Tom Quarendon <tom@quarendon.net>


bjhargrave:
LGTM. Are you OK with this @djencks and @pkriens ?

pkriens:
LGTM

djencks:
LGTM

PR #1564, opened 2016-07-26
[httpclient] The URL caching mechanism failed

pkriens:
If a file was asked multiple times with the same URL then the locking and caching mechanisms was screwed up.

Removed the file locking for now and use an in memory cache lock.


bjhargrave:
This change is failing a numbers of tests. See https://travis-ci.org/bndtools/bnd/jobs/147443343#L447

Issue #778, opened 2015-01-15
Bnd GenericResolveContext can return invalid cache results


timothyjward:
Relates to bndtools/bndtools#971 

The GenericResolveContext returns ordered lists of Capabilities base on a supplied requirement and the configured repositories.

The important details of the requirement are:

1. Its attributes
2. Its directives
3. The resource which defines it 

Items 1. and 2. contribute both to the content of the list and its order. Item 3 contributes further to the ordering.

To improve performance (repositories can be large and slow to respond) the results of queries are cached. Unfortunately the cache key only checks items 1. and 2. from the list, and not the resource. In some cases this can result in incorrectly ordered lists being returned, which in turn can result in "strange" (although valid) resolution results. Users have complained about the duplicates in these "strange" results, and we should be consistent about applying the correct ordering

PR #1580, opened 2016-08-04
Fix: CacheKey should not include Resource in hashCode() and equals(..)

psoreide:
We want to cache the result of identical requirements, but we do not care where the requirement is coming from.
The result will be the same.


timothyjward:
Please revert this change - this is a regression which breaks the fix for bug #778

timothyjward:
Never mind - I see that you didn't merge this in the end! My bad.

PR #1618, opened 2016-08-19
[macro] Lookup macro method without generating NoSuchMethodException's

psoreide:

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
@@ -329,7 +329,18 @@ public class Macro {
 			; // System.err.println("Huh? Target should never be null " +
 		// domain);
 		else {
-			String cname = "_" + method.replaceAll("-", "_");
+			// Assume macro names do not start with '-'
+			if (method.startsWith("-")) {
+				return null;
+			}
+
+			String part = method.replaceAll("-", "_");
+			for (int i = 0; i < part.length(); i++) {
+				if (!Character.isJavaIdentifierPart(part.charAt(i)))
+					return null;
+			}
+
+			String cname = "_" + part;
 			try {
 				Method m = target.getClass().getMethod(cname, new Class[] {
 						String[].class


biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
```diff
@@ -356,6 +354,24 @@ private String doCommand(Object target, String method, String[] args) {
...
+		if (method.startsWith("-") || method.indexOf('.') > -1) {
+			return null;
+		}
+		String cname = '_' + method.replace('-', '_');
+		for (Method m : target.getClass().getMethods()) {
+			if (m.getParameterTypes().length != 1) {
+				continue;
+			}
+			if (m.getName().equals(cname)) {
+				if (m.getParameterTypes()[0].getName().equals(String[].class.getName()))
```
bjhargrave: Why not `if (m.getParameterTypes()[0] == String[].class)`. Comparing class names seems wrong.

bjhargrave:
What is the point of this change? It it to avoid the cost of creating and throwing the NSME? Doesn't this only occur in the unusual case of trying the use a macro that does not exist?

psoreide:
No, it will occur for any parameter, e.g. -pedantic, -upto, -profile etc. It creates a lot of NoSuchMethodException's

biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
```diff
@@ -356,6 +354,24 @@ private String doCommand(Object target, String method, String[] args) {
 		return null;
 	}
 
+	private Method lookupMacroMethod(String method, Object target) {
+		// Assume macro names do not start with '-' and does not contain '.'
+		if (method.startsWith("-") || method.indexOf('.') > -1) {
+			return null;
+		}
+		String cname = '_' + method.replace('-', '_');
+		for (Method m : target.getClass().getMethods()) {
```
bjhargrave: I just realized that this has a problem with inheritance. getMethod(String,Class[]) returns the overridden method. When you use getMethods, all the overrloads are in the result in no specific order. And you loop just uses the first method found which can be from a superclass. This wrong.

If you want to use getMethods, you will need to find all the matching method and the pick the method from the type closest to the target's class.

bjhargrave:
OK, we now have nested loops replacing a single method call using exception control flow. Is there data which proves this change is faster?

psoreide:
The single method call does something similar. I created this after a yourkit profiling session some time ago, where the number of NoSuchMethodException was almost at the top of the list.

bjhargrave:
> The single method call does something similar. I created this after a yourkit profiling session some time ago, where the number of NoSuchMethodException was almost at the top of the list.

My point is that with the updated code which is now nested loops, does this current code perform better than the prior code? If it is not better, then why make the change?

psoreide:
You are right. Digging deeper getMethod will be more effective. Let me rewrite the solution.

biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
```diff
@@ -329,7 +329,18 @@ private String doCommand(Object target, String method, String[] args) {
...
 		// domain);
 		else {
-			String cname = "_" + method.replaceAll("-", "_");
+			// Assume macro names do not start with '-'
+			if (method.startsWith("-")) {
+				return null;
+			}
+
+			String part = method.replaceAll("-", "_");
+			for (int i = 0; i < part.length(); i++) {
```
bjhargrave: Perhaps a regex would be more straighforward:

```java
static final Pattern JAVAIDENTIFIER = Pattern.compile("\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*");

String cname = "_" + method.replaceAll("-", "_");
if (!JAVAIDENTIFIER.matcher(cname).matches()) {
    return null;
}
```

psoreide:
For readability; true. 

Performance wise: 
100000 calls using regexp=0.298, for-loop=0.063

bjhargrave:
> Performance wise:

:-)

bjhargrave:
OK, this looks pretty safe to me now. Go ahead and merge.

psoreide:
Thanks!

PR #1619, opened 2016-08-19
[repo] more mcs tests

rotty3000:


rotty3000:
re-organizing based on feedback from Firday's call.

PR #1630, opened 2016-08-22
fixes #1557 Requirements resolution fails using aQute.bnd.repository.p2.provider.P2Repository with bundle dependencies

### LINKED ISSUE #1557: Requirements resolution fails using aQute.bnd.repository.p2.provider.P2Repository with bundle dependencies
If I define the following plugin:

    -plugin.8.Eclipse.4.5.Updates:\
    		aQute.bnd.repository.p2.provider.P2Repository;\
    			name				=		"Test";\
    			url					=		http://download.eclipse.org/eclipse/updates/4.5/

And then try to resolve the following run requirement:

    -runrequires:\
    	osgi.identity;filter:='(osgi.identity=org.eclipse.equinox.registry)'

The resolution fails, unable to satisfy the direct bundle requirement on `org.eclipse.equinox.common`.

However I can manually verify that the dependency does exist by searching for the bundle in the repositories view. If I then copy jar for `org.eclipse.equinox.common` from the cache at `cnf/cache/p2-Test` into a local indexed repository and retry the resolution it succeeds (with the requirement on `org.apache.felix.log` satisfied from Distro).

rotty3000:


rotty3000:
@pkriens @bjhargrave ? Fixes require-bundle and cleans up by reordering a little.

pkriens:
Any short description what the problem was? Things seemed to have been moved around but not sure I see what the actual priblem was?

rotty3000:
The main issue was that the osgi.wiring.bundle namespace was being added with "version" instead of "bundle-version" also, there was no "osgi.wiring.bundle" attribute, etc.

PR #1631, opened 2016-08-22
[repo] let bnd's implemetation of the OSGi RequirementBuilder API work without resources

rotty3000:


rotty3000:
@pkriens I think this is what you were looking for!

pkriens:
Yup much cleaner

PR #1633, opened 2016-08-22
[maven] if pom has no dependencies return immediately

rotty3000:


rotty3000:
@bjhargrave can you check if this messes with your changes? It's working for me to fix the no-deps issue. But sanity check required.

rotty3000:
k, I think it's fine so merged.

PR #1645, opened 2016-08-24
[repo] Make some fields protected in OSGiRepository

psoreide:


bjhargrave:
What is the point of this change? Subclasses can figure this out on their own.

psoreide:
A lot of subclasses extend FixedIndexedRepo / AbstractIndexedRepo where these fields are protected. So to simplify transition to OSGiRepository I thought this was a good idea.

bjhargrave:
But they would be better of overriding setRegistry/setReporter rather than making the fields API.

bjhargrave:
Whenever one makes fields API, it makes thing much harder to reason about correctness and concurrency.

psoreide:
OK, I'm closing this.

PR #1650, opened 2016-08-25
[xml] apply where it seemed reasonable

rotty3000:


bjhargrave:
This is too much. There has to be a more simple solution.

rotty3000:
I think this whole issue is really just a case of bad pom so going to drop this.

PR #1651, opened 2016-08-25
Add non-static Workspace getProjectFromFile method

bjhargrave:


biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -214,26 +214,35 @@ public static Workspace getWorkspace(File parent, String bndDir) throws Exceptio
...
 		workspaceDir = workspaceDir.getAbsoluteFile();
 		if (!workspaceDir.exists() && !workspaceDir.mkdirs()) {
 			throw new IOException("Could not create directory " + workspaceDir);
 		}
 		assert workspaceDir.isDirectory();
 
-		File buildDir = new File(workspaceDir, nameOfCnf).getAbsoluteFile();
+		synchronized (cache) {
+			WeakReference<Workspace> wsr = cache.get(getBase());
+			if ((wsr != null) && (wsr.get() == this)) {
```
seanbright: I don't think we care. We can skip the `.get()` and just `.remove()`/`.put()` here.

bjhargrave: We need to do the get since there can be multiple Workspace objects (via new) for a given dir. But only one can be in the cache, so we need to confirm we are that one before removing from the cache.

seanbright: Ah, right. Forgot about the public constructor.

PR #1653, opened 2016-08-25
[build] make sure the container disambiguates on deferred downloads

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/build/DownloadBlocker.java
```diff
@@ -110,4 +110,8 @@ public File getFile() {
 	public String toString() {
 		return "DownloadBlocker [stage=" + stage + ", failure=" + failure + ", file=" + file + "]";
 	}
+
+	public String getPath() {
+		return file.getAbsolutePath();
+	}
```
bjhargrave: This cannot work. The whole point is that the blocker does not have the file until sometime in the future. You cannot access the file field until getReason has been called.

rotty3000: Well we need a fix.. somehow we need the path of the file-to-be! Otherwise
we have an "equals" problem.

bjhargrave:
Closing. We agreed to make Container identity for equals/hashCode and to write a dedup method to map a Collection<Container> to Collection<File>. I'll submit a new PR.

PR #1656, opened 2016-08-26
False container duplicates

bjhargrave:
Use a more unique file name for deferred downloads in BndPomRepository. This allows Containers to be unique.


seanbright:
LGTM :thumbsup:

PR #1663, opened 2016-08-30
[resolve] when running the ProjectResolver we need to index the workspace, otherwise projects can never be part of the resolution

rotty3000:


biz.aQute.resolve/src/biz/aQute/resolve/ProjectResolver.java
```diff
@@ -81,10 +82,14 @@ private String toString(ServiceReference< ? > sr) {
...
 
-	public ProjectResolver(Project project) {
+	public ProjectResolver(Project project) throws Exception {
 		super(project);
 		getSettings(project);
 		this.project = project;
+
+		Workspace workspace = project.getWorkspace();
+		if (workspace != null)
+			project.addBasicPlugin(new WorkspaceResourcesRepository(workspace));
```
bjhargrave: This seems wrong. Everytime you create a ProjectResolver, you add a new WorkspaceResourcesRepository plugin to the workspace.

rotty3000: That's exactly how the bnd resolve operation works, and I don't disagree
because the projects are a moving target which are not indexed in real-time.

bjhargrave: But the bnd resolve operation is a run-and-done operation. This code could be run many many times in the lifetime of an Eclipse run. I am working on a PR to try and fix this.

bjhargrave: Also, you did not modify the bnd resolve operation to remove its creation of a workspace repository since it is now also done by the ProjectResolver constructor.

rotty3000: Right...

rotty3000: Uhh about the eclipse run you mention... I don't see the relevance!

bjhargrave: Eclipse can run for a long time, so a Project object can around for a long time across many resolve operations, each adding a new workspace resources repo.

Anyway, I made #1666 to fix.

PR #1665, opened 2016-08-31
[gradle] set offline mode

rotty3000:


biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BndPlugin.groovy
```diff
@@ -48,6 +48,7 @@ public class BndPlugin implements Plugin<Project> {
         if (bndWorkspace == null) {
           throw new GradleException("Unable to load bnd workspace ${rootDir}")
         }
+        bndWorkspace.setOffline(project.gradle.startParameter.offline)
```
bjhargrave: What about non-workspace model?

build.gradle
```diff
@@ -23,6 +23,7 @@ ext.bndWorkspace = new Workspace(rootDir, bnd_cnf)
 if (bndWorkspace == null) {
   throw new GradleException("Unable to load workspace ${rootDir}/${bnd_cnf}")
 }
+bndWorkspace.setOffline(project.gradle.startParameter.offline)
```
bjhargrave: Why are you doing it here? I would think this change needs to be isolated to the bnd plugin rather than requiring everyone edit their root project build.gradle file.

rotty3000: The non-workspace model doesn't have 1) a workspace 2) repos since deps are
managed in gradle, offline already works.

rotty3000:
This is only the tip of the iceburg... needs more love than I have at the moment.

PR #1667, opened 2016-09-01
Changes needed for making bndrun work standalone with maven

pkriens:


biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java
```diff
@@ -371,6 +380,61 @@ private void loadRepositories() throws Exception {
...
+			return;
+
+		Object pathObject = capability.getAttributes().get("path");
+		if (pathObject == null)
+			pathObject = "augments.bnd";
+
+		if (pathObject instanceof String) {
+			String path = (String) pathObject;
+
+			HttpClient http = project.getWorkspace().getPlugin(HttpClient.class);
```
bjhargrave: This should be registry.getPlugin, no? All other plugins come from the registry field.

biz.aQute.repository/src/aQute/bnd/deployer/repository/LocalIndexedRepo.java
```diff
@@ -541,4 +541,10 @@ public String title(Object... target) throws Exception {
 
 		return null;
 	}
+
+	public void close() {
+		// TODO Auto-generated method stub
```
bjhargrave: Should there be an implementation here? If not, lets change the comment.

biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java
pkriens: guess that would be a bit cleaner.

PR #1669, opened 2016-09-01
[gradle] enable offline mode

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -72,6 +72,7 @@
 	private URLCache							cache					= new URLCache(IO.getFile("~/.bnd/urlcache"));
 	private Registry							registry				= null;
 	private Reporter							reporter				= new Slf4jReporter(HttpClient.class);
+	private boolean								offline;
```
bjhargrave: Should this be volatile since it can be set on threads other than the thread reading it?

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -673,4 +678,8 @@ public URI makeDir(URI uri) throws URISyntaxException {
 		else
 			return uri;
 	}
+
+	public void setOffline(boolean offline) {
```
bjhargrave: Shouldn't there be a matching `isOffline()` method for completeness?

build.gradle
```diff
@@ -23,6 +23,7 @@ ext.bndWorkspace = new Workspace(rootDir, bnd_cnf)
 if (bndWorkspace == null) {
   throw new GradleException("Unable to load workspace ${rootDir}/${bnd_cnf}")
 }
+bndWorkspace.setOffline(project.gradle.startParameter.offline)
```
bjhargrave: Do we need project here? I would think gradle. is fine (like in settings.gradle.)

settings.gradle
```diff
@@ -46,6 +46,8 @@ if (workspace == null) {
   throw new GradleException("Unable to load workspace ${rootDir}/${bnd_cnf}")
 }
 
+workspace.setOffline(gradle.startParameter.offline)
```
bjhargrave: I don't think we need `gradle.` here. See the reference to `startParameter` below.

biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BndPlugin.groovy
```diff
@@ -49,6 +49,7 @@ public class BndPlugin implements Plugin<Project> {
           throw new GradleException("Unable to load bnd workspace ${rootDir}")
         }
       }
+      bndWorkspace.setOffline(project.gradle.startParameter.offline)
```
bjhargrave: This should be up inside the previous if block since we only need/want to set it IF we created the Workspace.

bjhargrave:
See https://github.com/bndtools/bnd/pull/1670. I have put the gradle changes in there including updating the test cases. So the gradle changes can be removed from this PR.

rotty3000:
@bjhargrave I've limited this change to just the HttpClient and Workspace. Once your other offline changes are in this should be all that's required to complete it.

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -725,6 +725,9 @@ public AtomicBoolean getOffline() {
 
 	public Workspace setOffline(boolean on) {
 		offline.set(on);
+		for (HttpClient client : getPlugins(HttpClient.class)) {
+			client.setOffline(on);
```
bjhargrave: I changes the code to use an AtomicBoolen to hold the offline state. You need to send that object into the HttpClient (not its current value) and then have the HttpClient call get on the object. Setting the AtomicBoolean into the HttpClient can be done when constructed and not at every all to  Workspace.setOffline.

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -72,6 +72,7 @@
 	private URLCache							cache					= new URLCache(IO.getFile("~/.bnd/urlcache"));
 	private Registry							registry				= null;
 	private Reporter							reporter				= new Slf4jReporter(HttpClient.class);
+	private volatile boolean					offline;
```
bjhargrave: Hold the AtomicBoolean from the Workspace.

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -673,4 +678,12 @@ public URI makeDir(URI uri) throws URISyntaxException {
 		else
 			return uri;
 	}
+
+	public boolean isOffline() {
+		return offline;
+	}
+
+	public void setOffline(boolean offline) {
```
bjhargrave: This method should take an AtomicBoolean.

rotty3000:
ok I got it!

rotty3000:
legitimate fail here.. fixing.

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -72,6 +73,7 @@
 	private URLCache							cache					= new URLCache(IO.getFile("~/.bnd/urlcache"));
 	private Registry							registry				= null;
 	private Reporter							reporter				= new Slf4jReporter(HttpClient.class);
+	private AtomicBoolean						offline;
```
bjhargrave: YOu need to make this volatile and initialize it in case no one else does!

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -610,6 +610,7 @@ protected void setTypeSpecificPlugins(Set<Object> list) {
 
 			try {
 				HttpClient client = new HttpClient();
+				client.setOffline(offline);
```
bjhargrave: Can you use `getOffline()` here? I prefer using methods instead of direct access to fields.

```java
client.setOffline(getOffline());
```

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -673,4 +679,15 @@ public URI makeDir(URI uri) throws URISyntaxException {
 		else
 			return uri;
 	}
+
+	public boolean isOffline() {
+		if (offline == null) {
+			return false;
+		}
+		return offline.get();
```
bjhargrave: There is a risk of NPE here if another thread calls setOffline(null) after the null check says not null. Better to copy into local var and then use that.

```java
AtomicBoolean o = offline;
if (o == null) {
	return false;
}
return o.get();
```

rotty3000:
This has been working very well for the past couple of days while I'm traveling.

gradle files in certain projects need a small change but I think it's worth it.

Can I get a LGTM?

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
bjhargrave: Adding this requested change would get a LGTM from me.

Issue #1620, opened 2016-08-19
Bndtools unable to download snapshot version from Artifactory using MavenBndRepository


Mr-Frantic:
When I am defining a dependency in artifatory.mvn with snapshot version, bndtools is not able download that dependency. I get the following exception 

An internal error occurred during: "Downloading repository entry com.ecm:catalog-interface".
java.lang.NullPointerException

I am able to download that dependency from the same artifactory with Maven when i define that in pom.xml and when it works fine when I use sharp version of the dependency

Also when I download the dependency i.e. when that dependency is in my local cache, bndtools is able to resolve that dependency.

lukewink:
I am seeing the same issue.  Additionally, when building via gradle the following error is seen:

> 14:23:55.014 [DEBUG] [aQute.bnd.build.Workspace] Reported exception
> java.io.FileNotFoundException: For Maven artifact com.test.group:test-artifact:1.0.0-SNAPSHOT
>     at aQute.maven.provider.MavenRepository$1.run(MavenRepository.java:140)
>     at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
>     at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
>     at java.lang.Thread.run(Thread.java:745)

lukewink:
I added a pull request to address this issue.  In the mean time, if you explicitly set a snapshotUrl that is the same as your releaseUrl, it should work around the problem.

bjhargrave:
I think explicitly setting a snapshot url is the solution. I don't think your PR is a good change.

pkriens:
@Mr-Frantic : I am very interested in the NPE stack trace

lukewink:
I added a PR which addresses the NPE

Mr-Frantic:
Please find the stacktrace attached
[SnapshotDownloadStactTrace.txt](https://github.com/bndtools/bnd/files/452609/SnapshotDownloadStactTrace.txt)

I tried with snapshotUrl, so my main bnd file has the repo defined as

```
-plugin.1.Artifactory = \
    aQute.bnd.repository.maven.provider.MavenBndRepository; \
    releaseUrl  = https://hostname/artifactory/proj-cpi-dev/; \
    snapshotUrl  = https://hostname/artifactory/proj-cpi-dev/; \
    name        = Artifactory
```

And I have a file called `artificatory.mvn`

[...]

PR #1671, opened 2016-09-01
fall back to releaseUrl if snapshotUrl is empty

lukewink:
Fixes: https://github.com/bndtools/bnd/issues/1620
Signed-off-by: Luke Winkenbach <lukewink@gmail.com>


bjhargrave:
If you need a snapshot repo to be always configured, then configure one.

pkriens:
There are a number if situations why I need to be able to have this separate.

lukewink:
I was making my changes based on the intended behavior written in Configuration.java.
    /**
     * The url to the remote snapshot repository. If this is not specified,
     * it falls back to the release repository or just local if this is also
     * not specified.
     */
    String snapshotUrl();

It doesn't appear that the current code falls back to the release repository which is why I made the change.

bjhargrave:
I think the comments are out of date. They reflected an earlier intent during initial development of the maven repo support.

PR #1679, opened 2016-09-05
Deduplication of plugins

pkriens:
The plugins were inherited but also recreated. This was inefficient but also could cause duplication of equals was not properly implemented.

No inherit the object references from the parent and only create plugins for the local level.


rotty3000:
Thx, Peter this made a big difference!

PR #1682, opened 2016-09-05
fixes #1681 [resolve] bndtools doesn't show the list of resources because the exception is double wrapped

### LINKED ISSUE #1681: bndtools resolve error widget doesn't show proper error and set of failed Requirements

rotty3000:


rotty3000:
@bjhargrave when this clears can we merge it to 3.3? it's tiny and a sad regression in bndtools.

PR #1688, opened 2016-09-09
[gradle] bnd build offline changes

rotty3000:


rotty3000:
OK seems we don't need most of this. And bj agreed not to delete the cache when offline, he will make the change.

PR #1692, opened 2016-09-10
[pom] A few fixes when the pom fails to download

pkriens:


bjhargrave:
@pkriens Is this to merge for 3.3?

PR #1694, opened 2016-09-12
A Requirement without a filter always matches.

### LINKED ISSUE #1687: IllegalStateException during Resolution without any cause
Resolving a workspace which worked previously fails with the lates DEV-build (3.3.0.DEV-20160908-141802-g8c80800) with the following message.

> java.lang.IllegalStateException: What?!? The capability was not found associated with any key! That is not possible at this stage.

The only bundle I know of which uses the require-capability model in this workspace is Pax-CDI  (1.0.0.RC1). Pax-CDI provides a extender-capability. 

test-workspace will follow. Unfortunately, no further information is available (error-log)

[bnd-1687.zip](https://github.com/bndtools/bnd/files/463873/bnd-1687.zip)

seanbright:
Fixes #1687

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/resource/CapReqFilterTest.java
+++ b/biz.aQute.bndlib.tests/src/test/resource/CapReqFilterTest.java
@@ -1,0 +1,52 @@
+package test.resource;
+
+import org.osgi.resource.Capability;
+import org.osgi.resource.Requirement;
+
+import aQute.bnd.osgi.resource.CapabilityBuilder;
+import aQute.bnd.osgi.resource.RequirementBuilder;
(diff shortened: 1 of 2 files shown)


seanbright:
Unit test to follow

bjhargrave:
LGTM

bjhargrave:
Add a test to verify that a requirement with no filter in a different namespace than the cap does not match.

seanbright:
You are correct. Looks like all of the current call sites for the method already know that the namespaces match. I've added a check and a unit test.

bjhargrave:
LGTM
-- 
-- 
BJ

PR #1699, opened 2016-09-15
Unnecessary null checks, unnecessary synchronization, and some thread safety

seanbright:


biz.aQute.bndlib/src/aQute/bnd/osgi/FileResource.java
```diff
@@ -51,7 +52,7 @@ public String toString() {
 	}
 
 	public void write(OutputStream out) throws Exception {
-		copy(this, out);
+		Files.copy(file.toPath(), out);
```
bjhargrave: Why this change? It was using the method below which can be overridden. I don't see the point of this change. I don't think this change should be made.

biz.aQute.repository/test/aQute/bnd/deployer/repository/FindProvidersTest.java
```diff
@@ -66,7 +66,6 @@ public static void testReadGZippedStream() throws Exception {
 		Capability identityCap = caps.iterator().next();
 
 		List<Capability> contentCaps = identityCap.getResource().getCapabilities("osgi.content");
-		assertNotNull(contentCaps);
```
bjhargrave: There is nothing wrong with asserting non null. Why remove? Same for the other nonnull checks you removed.

bjhargrave:
[Used the new GitHub PR review feature just released!]

biz.aQute.bndlib/src/aQute/bnd/osgi/FileResource.java
seanbright: Two reasons:

1. Files.copy() may use native code to perform the copy operation
2. FileResource.copy() is a static method and cannot be overriden

bjhargrave: Is there a problem with the current implementation? What does this change fix?

So why leave the old method behind then? It is not API.

It is used by ZipResource.write. But since is not API, that method can be moved to ZipResource.

biz.aQute.resolve/src/biz/aQute/resolve/AbstractResolveContext.java
```diff
@@ -135,7 +135,7 @@ protected synchronized void init() {
 	public List<Capability> findProviders(Requirement requirement) {
 		init();
 		List<Capability> result = findProviders0(requirement);
-		if (result == null || result.isEmpty()) {
+		if (result.isEmpty()) {
```
bjhargrave: This is good! If findProviders0 ever returns null, we will fail fast.

biz.aQute.bndlib/src/aQute/bnd/osgi/FileResource.java
seanbright: It fixes nothing. Reverting.

seanbright: Given the choice, if both the JCL and an external library provided a method that performed the same function, I would always choose the JCL provided method. It may have native bindings under the hood. It may be able to make better decisions about block sizes, etc.

While profiling a build, I noticed we spent a measurable amount of time in FileResource.copy(), so I assumed this was (at worst) a zero sum change.

bjhargrave: I am not opposed the change per se. My main concern was leaving dead code behind (I hate it when people do that). So make the change, but move the copy method logic into ZipResource.write.

PR #1701, opened 2016-09-16
resource: Tighten up impls

bjhargrave:
Require certain arguments to be non-null; create unmodifiable objects
earlier.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReq.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReq.java
@@ -1,6 +1,8 @@
 package aQute.bnd.osgi.resource;
 
-import java.util.Collections;
+import static aQute.bnd.osgi.resource.ResourceUtils.requireNonNull;
+import static java.util.Collections.unmodifiableMap;
+
 import java.util.HashMap;
 import java.util.Map;
 
@@ -20,14 +22,15 @@ class CapReq {
 	private final Resource				resource;
 	private final Map<String,String>	directives;
 	private final Map<String,Object>	attributes;
+	private transient int				hashCode	= 0;
 
 	CapReq(MODE mode, String namespace, Resource resource, Map<String,String> directives,
 			Map<String,Object> attributes) {
-		this.mode = mode;
-		this.namespace = namespace;
+		this.mode = requireNonNull(mode);
+		this.namespace = requireNonNull(namespace);
 		this.resource = resource;
-		this.directives = new HashMap<String,String>(directives);
-		this.attributes = new HashMap<String,Object>(attributes);
+		this.directives = unmodifiableMap(new HashMap<>(directives));
+		this.attributes = unmodifiableMap(new HashMap<>(attributes));
 	}
 
 	public String getNamespace() {
@@ -35,11 +38,11 @@ class CapReq {
 	}
 
 	public Map<String,String> getDirectives() {
-		return Collections.unmodifiableMap(directives);
+		return directives;
 	}
 
 	public Map<String,Object> getAttributes() {
-		return Collections.unmodifiableMap(attributes);
+		return attributes;
 	}
 
 	public Resource getResource() {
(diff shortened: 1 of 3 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
```diff
@@ -552,4 +552,10 @@ public static VersionedClause toVersionClause(Resource resource, String mask) {
 		return new VersionedClause(identity, attribs);
 	}
 
+	static <T> T requireNonNull(T obj) {
```
seanbright: This doesn't have any direct relation to Resources, so maybe there is a better location for it?

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReq.java
```diff
@@ -20,26 +22,27 @@
 	private final Resource				resource;
 	private final Map<String,String>	directives;
 	private final Map<String,Object>	attributes;
+	private transient int				hashCode	= 0;
```
seanbright: This is the only mutable field in the class. If we build it in the constructor instead of lazily, this class can be marked `@Immutable`

bjhargrave: The class can be marked immutable since the hashCode field is transient and always recomputes the same value. There is no need for this feild to be final or even visible across threads :-)

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
bjhargrave: I thought about putting it in aQute.lib somewhere, but it didn't seem worth it for a simple method to drag in another class/package. When we move to Java 8, we can remove this and replace with using Objects.

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReq.java
seanbright: What impact does `transient` have on memory visibility?

bjhargrave: Nothing. It is (a) notice to the programmer that the field hold a transient value (that can be recomputed) and (b) the value is not to be serialized if the class is Serializable. The latter does not apply here. I only marked it transient for the former reason.

seanbright: I've never heard of `transient` being used that (a) way before. My assumption when looking at a field is that if it's not `final`, it can be reassigned (recomputed).

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReq.java
```diff
@@ -48,14 +51,17 @@ public Resource getResource() {
 
 	@Override
 	public int hashCode() {
+		if (hashCode != 0) {
```
seanbright: Albeit unlikely, the `hashCode` could be calculated as `0`, which negates any performance optimization lazily calculating provides. Just calculate in the constructor and be done with it.

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReq.java
bjhargrave: Well it can be, but the value transient and can be recomputed from the other data of the object. It is essentially a cached value.

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReq.java
bjhargrave: Yes, in the pathological case that the actual computed has value is zero, then it will recompute every time. So what? The odds are probably 1 in 2^32 of that happening :-) So, in any practical sense, it will not happen even close to enough to be any issue.

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReq.java
bjhargrave: But I am happy to remove the keyword transient if you like.

PR #1709, opened 2016-09-21
[maven] Don't poll in batch mode

seanbright:


seanbright:
@bjhargrave - not sure if this is the most elegant approach. I also feel like we should centralize this logic instead of duplicating. If you have a better/cleaner idea, feel free to close without merging.

bjhargrave:
I think this PR is ok, but I don't think it is sufficient to really fix the issue. It only elides the problem under gradle. I made #1710 to try a more direct solution that will work even with polling in Eclipse. I want to test #1710 out first before this PR.

PR #1721, opened 2016-09-30
Set up to generate Maven plugin documentation.

benson-basis:
At least, someone can locally run mvn site and use the results. See #1722.

To see the results of this, run 'mvn site', and the look at target/site/index.html.

### DIFF
--- a/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
+++ b/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
@@ -58,6 +58,25 @@ import aQute.lib.strings.Strings;
 import aQute.lib.utf8properties.UTF8Properties;
 import aQute.service.reporter.Report.Location;
 
+/**
+ * Generate a <pre>MANIFEST.MF</pre> for an OSGi bundle using <strong>bnd</strong>.
+ * This goal runs bnd with the contents of one or more bnd files to generate
+ * <pre>${project.build.outputDirectory}/META-INF/MANIFEST.MF</pre>. The pom must then
+ * configure the maven-jar-plugin to use this generated manifest rather than generate
+ * another:
+ * <pre>{@code
+ *     <plugin>
+ *         <groupId>org.apache.maven.plugins</groupId>
+ *          <artifactId>maven-jar-plugin</artifactId>
+ *            <configuration>
+ *              <archive>
+ *                <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+ *              </archive>
+ *            </configuration>
+ *      </plugin>
+ *      }
+ * </pre>
+ */
 @Mojo(name = "bnd-process", defaultPhase = LifecyclePhase.PROCESS_CLASSES, requiresDependencyResolution = ResolutionScope.COMPILE)
 public class BndMavenPlugin extends AbstractMojo {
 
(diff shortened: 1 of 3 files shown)


bjhargrave:

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -89,6 +108,19 @@
...
+	 * The location of the bnd file that this plugin will supply to <pre>bnd</pre>.
+	 * The plugin combines the contents of bnd files from the project and its parents.
+	 * The plugin walks up the tree of parents. At each level, if the project as a <pre>basedir</pre>,
+	 * the plugin looks for a bnd file. If there is a local configuration of this parameter,
+	 * it determines the file name relative to that project's basedir; otherwise, it looks for the default,
+	 * <pre>${basedir}/bnd.bnd</pre>. For each project's bnd file, <pre>-include</pre> directives are resolved
+	 * relative to that project's basedir.
+	 */
+	@Parameter(defaultValue = "bnd.bnd")
+	String bndfile;
```
bjhargrave: This is wrong to put a field in that is unused to hang a doc comment on for a doc site we don't build or publish.

maven/bnd-maven-plugin/src/site/site.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bjhargrave: Since we don't build or publish a maven site, I don't think this should be part of the source.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -58,6 +58,25 @@
...
+ *          <artifactId>maven-jar-plugin</artifactId>
+ *            <configuration>
+ *              <archive>
+ *                <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+ *              </archive>
+ *            </configuration>
+ *      </plugin>
+ *      }
+ * </pre>
+ */
```
bjhargrave: This comment is fine.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -89,6 +108,19 @@
...
+    /* This exists only to allow documentation */
+	/**
+	 * The location of the bnd file that this plugin will supply to <pre>bnd</pre>.
+	 * The plugin combines the contents of bnd files from the project and its parents.
+	 * The plugin walks up the tree of parents. At each level, if the project as a <pre>basedir</pre>,
+	 * the plugin looks for a bnd file. If there is a local configuration of this parameter,
+	 * it determines the file name relative to that project's basedir; otherwise, it looks for the default,
+	 * <pre>${basedir}/bnd.bnd</pre>. For each project's bnd file, <pre>-include</pre> directives are resolved
+	 * relative to that project's basedir.
+	 */
```
bjhargrave: I would move this comment somewhere else such as the above comment. There is also the bnd configuration for "bnd-in-pom" bnd config.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
bjhargrave: This unused field needs to be removed.

bjhargrave:
The bnd maintainers discussed this and it was agreed that since we do not intend to publish the maven site build, we won't be building the maven site. So including the site in the source and not building it will mean it will "rot." So we agreed to close this PR. Thanks for bringing this up.

Improvements to source code comments and the README files are welcome as are improvements to the GitHub pages-based documentation in the docs/ folder.

PR #1738, opened 2016-10-10
Implement version option for ds and metatype. Closes #1728

### LINKED ISSUE #1728: No extender requirement for very simple components
bnd does not add a requirement for the component extender when the component is so simple it falls under the lowest version that specified the extender name. 

		if (options.contains(Options.extender) || maxVersion.compareTo(AnnotationReader.V1_3) >= 0) {
			maxVersion = ComponentDef.max(maxVersion, AnnotationReader.V1_3);
			addExtenderRequirement(requires, maxVersion);
		}

This does make some sense but I think we actually should always add this from now on.

djencks:
Signed-off-by: David Jencks (david_jencks@yahoo.com)

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
+++ b/biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
@@ -474,6 +474,7 @@ public class DSAnnotationTest extends BndTestCase {
 	public void testBasic() throws Exception {
 		Builder b = new Builder();
 		b.setProperty(Constants.DSANNOTATIONS, "test.component.*_basic");
+		b.setProperty(Constants.DSANNOTATIONS_OPTIONS, "version;minimum=1.0.0");
 		b.setProperty("Private-Package", "test.component");
 		b.addClasspath(new File("bin"));
 
@@ -1077,6 +1078,7 @@ public class DSAnnotationTest extends BndTestCase {
 	public void testEnums() throws Exception {
 		Builder b = new Builder();
 		b.setProperty(Constants.DSANNOTATIONS, "test.component.DSAnnotationTest*Enums");
+		b.setProperty(Constants.DSANNOTATIONS_OPTIONS, "version;minimum=1.0.0");
 		b.setProperty("Private-Package", "test.component");
 		b.addClasspath(new File("bin"));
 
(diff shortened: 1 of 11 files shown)


biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
```diff
@@ -1394,7 +1398,9 @@ public void testBinds(boolean extender) throws Exception {
 		Builder b = new Builder();
 		b.setProperty(Constants.DSANNOTATIONS, "test.component.DSAnnotationTest*CheckBinds");
 		if (extender)
-			b.setProperty(Constants.DSANNOTATIONS_OPTIONS, "extender");
```
bjhargrave: This would be a good place to test the merge properties support. So instead of an else with just version, leave the if (extender) alone and add:

```java
b.setProperty(Constants.DSANNOTATIONS_OPTIONS+".min", "version;minimum=1.2.0");
```

biz.aQute.bndlib/src/aQute/bnd/metatype/MetatypeVersion.java
```diff
@@ -20,4 +20,13 @@ public String getNamespace() {
 		return NAMESPACE_STEM + value;
 	}
 
+	static MetatypeVersion valueFor(String v) {
```
bjhargrave: I think this method should be called `valueOf`.

djencks: AFAICT calling it valueOf won't compile as it conflicts with the built in valueOf method.  In any case I think it would be too confusing to use the same name.

bjhargrave: OK

biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
djencks: I didn't know mergeProperties worked like this...

djencks:
I don't see what went wrong with the travis jdk 7 build, it doesn't look to me like it's related to my changes?

bjhargrave:
I restarted that build. Seemed like a error in gradle.

djencks:
Looks like the same error recurred.  I don't have time to figure out what went wrong right now so will leave possible merging to others.

bjhargrave:
All better now. I had to delete the Travis cache for this PR. Somehow it was bad and failing the tests of the bnd gradle plugin.

PR #1743, opened 2016-10-11
Improve method signature parsing

seanbright:


bjhargrave:
Should you add/modify a test case to verify this fix?

seanbright:
Yes

PR #1774, opened 2016-11-18
support execution level configuration

### LINKED ISSUE #1771: bnd-maven-plugin: configuration parameters bnd and bndfile are not evaluated in execution level configuration
Due to the way how bnd-maven-plugin evaluates both configuration parameters `bnd` and `bndfile` only plugin level configuration is supported but not execution level configuration (https://maven.apache.org/guides/mini/guide-default-execution-ids.html). For example the following configuration is not considered:

```
<plugin>
  <version>3.3.0</version>
  <executions>
    <execution>
      <goals>
        <goal>bnd-process</goal>
      </goals>
       <configuration>
	   <bnd><![CDATA[
			testheader:testvalue]]></bnd>
    </execution>
  </executions>
<plugin>
```

Also code completion (like provided in m2e since 1.6, http://www.eclipse.org/m2e/documentation/release-notes-16.html#improved-auto-45-completion-in-pom-46-xml-editor) is not supported, because the parameters do not appear in the `META-INF\maven\plugin.xml`.

Therefore please rather use the default way of injecting those configuration parameters into the plugin, rather than relying on evaluating the configuration from the pom.xml on your own.

kwin:
this fixes #1771

### DIFF
--- a/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
+++ b/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
@@ -36,6 +36,7 @@ import java.util.zip.ZipFile;
 
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.plugin.AbstractMojo;
+import org.apache.maven.plugin.MojoExecution;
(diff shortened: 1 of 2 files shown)


bjhargrave:
I merged this by mistake. I meant to merge another PR. So I will make a new PR for this with some changes.

PR #1787, opened 2016-11-29
[index] if the current project packaging is jar try adding it to the index

rotty3000:


maven/bnd-indexer-maven-plugin/pom.xml
```diff
@@ -37,7 +37,12 @@
 		</dependency>
 		<dependency>
 			<groupId>biz.aQute.bnd</groupId>
-			<artifactId>org.osgi.impl.bundle.repoindex.cli</artifactId>
+			<artifactId>biz.aQute.bndlib</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.osgi</groupId>
+			<artifactId>org.osgi.core</artifactId>
+			<version>5.0.0</version>
```
bjhargrave: Dependency versions should be set in the parent pom.

maven/bnd-indexer-maven-plugin/pom.xml
```diff
@@ -64,6 +69,7 @@
 					<cloneClean>true</cloneClean>
 					<projectsDirectory>src/test/resources/integration-test</projectsDirectory>
 					<settingsFile>src/test/resources/integration-test/settings.xml</settingsFile>
+					<streamLogs>true</streamLogs>
```
bjhargrave: Perhaps we should set this for all the maven plugin's integration tests?

rotty3000:
Closing for now since I've broken this down into two parts.

Issue #1792, opened 2016-12-05
Baseline suggests a "micro" bump when there is both a "minor" and a "removed" change


Ithildir:
Hi,
I think I found an issue in the baseline code. Here are the steps to reproduce it:

1. create a bundle with 2 exported packages, each one of them with 1 class in it
2. un-export one of the packages
3. add a public method to the class in the only exported package
3. try to run a semantic versioning check between this new bundle and the one created at step 1

AFAIK the change at step 2 is a `REMOVED` change, while the change at step 3 is a `MINOR` one. But baseline suggests a `MICRO` version bump instead of a `MINOR` one.

Thank you!

bjhargrave:
For future readers, a `REMOVED` change is a `MAJOR` bump. PR #1797 includes this fix.

PR #1793, opened 2016-12-05
Baseline suggests a "micro" bump when there is both a "minor" and a "removed" change

Ithildir:
Hi,
The `testMinorAndRemovedChange` test in this pull should fail. I added it to test https://github.com/bndtools/bnd/issues/1792.


bjhargrave:

Ithildir:
Done!

rotty3000:
It seems there's two issues:
1) the REMOVED delta is the highest however it never results in a MAJOR version change
https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java#L207-L209
2) the calculated diff is ignored
https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java#L212-L213

rotty3000:
The changes have been incorporated into #1797 So I'll close this one.

PR #1794, opened 2016-12-05
fixes #1779 convert bnd-indexer-maven-plugin to use bnd's newest ResourceBuilder API

### LINKED ISSUE #1779: convert bnd-indexer-maven-plugin to use bnd's newest ResourceBuilder API
the old RepoIndex doesn't produce the high quality indexed output that the ResourceBuilder API does in bndlib. So we want to migrate to use that code.

rotty3000:


rotty3000:
@timothyjward and/or @bjhargrave when you have a moment to consider this I'd like your input.

timothyjward:
@rotty3000 For the most part it looks good, however I think you've uncovered a bug in the tests as I'm pretty sure that you've broken the "add mvn urls" behaviour. The expected behaviour when mvn urls are added is that you get *two* content capabilities, one with the mvn URL and one with the HTTP url. The code as is I think only adds the mvn URL.

The "broken" test is [in the verify script](https://github.com/bndtools/bnd/blob/ba99beb6d5163cb39797f18437c11833e1cc8d38/maven/bnd-indexer-maven-plugin/src/test/resources/integration-test/test/verify.groovy#L100) which only checks for mvn URLS, and not also for other URL types. This could be easily amended with a second `(!(url=mvn*))` check in the same test, which should also match three things.

If you would like to add a mode which only adds maven urls then that's fine, but it would need to be a new config value, e.g. `<skipArtifactURLs>true</skipArtifactURLs>` which would be an error unless Maven URLs were added too. I'm not trying to be prescriptive though, so if you can come up with a better config name than my terrible suggestion that would be great 😉

rotty3000:
Got it Tim! I wasn't sure about how the mvn url behaviour should have worked. But you are correct that currently it seems to be mutually exclusive and not additive. This may actually be a bug in the ResourceBuilder to be honest. Can you explain the expected mechanics of adding the new url? Is the url field supposed to be a list? or is there a completely new attribute, or capability added?

rotty3000:
Got it Tim! I wasn't sure about how the mvn url behaviour should have worked. But you are correct that currently it seems to be mutually exclusive and not additive. This may actually be a bug in the ResourceBuilder to be honest. Can you explain the expected mechanics of adding the new url? Is the url field supposed to be a list? or is there a completely new attribute, or capability added?

timothyjward:
> Is the url field supposed to be a list? or is there a completely new attribute, or capability added?

Section 132.4 of the compendium says that a content capability has exactly one URL, but that a resource may have many different content capabilities. These are typically expected to have different mime types, but there is no restriction on having multiple mirrors with the same content. We have been treating the `mvn` URLs as an alternative mirror in this way.

rotty3000:
Please see updates @timothyjward !

maven/bnd-indexer-maven-plugin/pom.xml
```diff
@@ -37,7 +37,12 @@
 		</dependency>
 		<dependency>
 			<groupId>biz.aQute.bnd</groupId>
-			<artifactId>org.osgi.impl.bundle.repoindex.cli</artifactId>
+			<artifactId>biz.aQute.bndlib</artifactId>
+		</dependency>
+		<dependency>
+			<groupId>org.osgi</groupId>
+			<artifactId>org.osgi.core</artifactId>
```
bjhargrave: bndlib uses osgi.core;version=6, so I think we should use the same here, no?

Also use `<artifactId>osgi.core</artifactId>` which is the proper OSGi artifactId.

Finally, the version number should be set in the parent pom's `<dependencyManagement>` section.

rotty3000: fixed!

rotty3000:
FYI @bjhargrave I made the requested changes.

bjhargrave:
> FYI @bjhargrave I made the requested changes.

Did you push them? I don't see them.

rotty3000:
seems I pushed the wrong branch.

timothyjward:
👍

PR #1797, opened 2016-12-05
fixes #1792 Baseline suggests a "micro" bump when there is both a "minor" and a "removed" change

### LINKED ISSUE #1792: Baseline suggests a "micro" bump when there is both a "minor" and a "removed" change

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java
```diff
@@ -204,13 +204,19 @@ public boolean contains(Diff diff) {
 					content = pdiff.getDelta();
 					break;
 
-				case REMOVED :
+				case REMOVED : // this is a package removal not a member removal
```
bjhargrave: I think we should discuss this with @pkriens to make sure it is a good decision.

Bundle versioning is so awkward.

rotty3000: I agree it's not trivial. However, we know for a fact that we have lost many MINOR changes because of this issue.

Case in point:
http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.liferay%22%20AND%20a%3A%22com.liferay.wiki.service%22

Between version 1.2.0 to current version (1.2.18) an exception was removed from a method call, which should have caused the version to go from 1.2.x to 1.3.0 because of the API change. However it clearly has not. The change was effectively lost.

rotty3000: Furthermore, the only reason the change was lost is because it was paired with another unrelated change in the bundle which removed an exported package (not the same package as with the exception change...)

bjhargrave: This change is altering REMOVED from MAJOR to MICRO which is a big reduction. It is possible the bug in the compareTo then clause is the reason changes were lost.

rotty3000: sigh.. the case I showed above isn't exactly the right case.

rotty3000:
but we can't reorder REMOVE in Delta because for members, REMOVE is
absolutely a MAJOR change.

biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java
rotty3000: we can't reorder REMOVE in Delta because for members and types REMOVE is absolutely a MAJOR change.

pkriens: How can an exported package removal be a micro change? If it has been moved to another bundle then this should be solved in another place?

This is bundle versioning which I always recommended not to take too seriously. What problem are we solving here?

rotty3000: There's a real bug here. Let me try to explain again from the start:

You have a bundle with two packages: A and B
1) release version 1.0.0 of bundle
2) make a MINOR change in package A
3) remove package B
4) baseline the bundle against version 1.0.0

What would you expect the suggested version to be?

The actual suggested version today is 1.0.1
NOT 1.1.0, or even  2.0.0, it's 1.0.1.

Since we use baseline in CI, this failure to capture the real level of change is problematic.

pkriens: I guess I disagreed with my own comments in the code. :-(

It looks like making it MAJOR for package removal is the most logical and my comment in the code does not seem to make sense. Lets try this patch, I am ok.

bjhargrave: So this means that for `case REMOVED` we want `content = Delta.MAJOR`? Which is what it was before this patch.  I am confused.

rotty3000:
@pkriens or @bjhargrave can I get a LGTM? :D

PR #1798, opened 2016-12-06
fixes #1795 make it possible for the bnd-indexer-maven-plugin to include a project's own output

### LINKED ISSUE #1795: make it possible for the bnd-indexer-maven-plugin to include a project's own output
Currently the indexer plugin can't index it's own project output even packaging is set to `jar`.

A change should make it possible to set this through configuration.

rotty3000:


rotty3000:
@timothyjward I'd like your input on this change which is the followup adding support for the integration test scenario.

maven/bnd-indexer-maven-plugin/src/main/java/aQute/bnd/maven/indexer/plugin/IndexerMojo.java
```diff
@@ -83,6 +83,9 @@
 	@Parameter(property = "bnd.indexer.includeTransitive", defaultValue = "true")
 	private boolean						includeTransitive;
 
+	@Parameter(property = "bnd.indexer.includeCurrentIfJar", defaultValue = "false")
+	private boolean						includeCurrentIfJar;
```
bjhargrave: Can we just call it `includeJar`? This means, if the project makes a jar, that jar is included in the index.

rotty3000: I'm open to any naming changes...

bjhargrave:
You also need to add a test to test this new function.

rotty3000:
true...

rotty3000:
Test added!

timothyjward:
I'm happy for this to be merged, but I will go away and think about a better name than `includeJar`. We should also probably not restrict it to jar projects, but look for jar classifier objects attached to the current project. In the fullness of time this could be exended to allow other classifiers (e.g. configuration definitions from the Configurator).

I would also suggest a slightly more thorough investigation of the repo in the verify.groovy. Currently it just checks for a size of two, but we can be much more specific. The following would do all the current validation, and check that the `include-jar` was actually in the index

    repo = check("${basedir}/include-jar/target/index.xml", "${basedir}/include-jar/target/index.xml.gz", 2, true, true);
    check(repo, "osgi.identity", "(osgi.identity=include-jar)", "include-jar", true);

rotty3000:
@timothyjward you're right, the test was weak. I've added your suggested bit now while the time is right.

PR #1799, opened 2016-12-07
[pom repo] a test case for #1731 bnd's maven support doesn't properly handle classifier dependencies

rotty3000:


rotty3000:
haha... fixing!

PR #1803, opened 2016-12-09
this is the embedded dependency in maven plugins

rotty3000:


maven/pom.xml
```diff
@@ -124,11 +124,6 @@
 				<version>${project.version}</version>
 			</dependency>
 			<dependency>
-				<groupId>biz.aQute.bnd</groupId>
```
bjhargrave: How is this delete related to adding the embedded repo to the plugns?

maven/pom.xml
```diff
@@ -252,6 +247,33 @@
                     </lifecycleMappingMetadata>
                   </configuration>
                 </plugin>
+				<plugin>
+					<groupId>org.apache.maven.plugins</groupId>
+					<artifactId>maven-dependency-plugin</artifactId>
+					<version>2.10</version>
+					<configuration>
```
bjhargrave: It is not clear to me that this configuration should be in the parent pom. It is shared by 2 of the plugins but if we need to use the dependency plugin in another plugin, that plugin may not need the embedded repo. This config should be in each project's pom.

rotty3000: ok..

maven/pom.xml
rotty3000: it's just cleanup from a previous change! Is that not ok?

bjhargrave: OK, for some reason I thought #1794 removed the reference, but it was just in the project pom. So this is fine.

maven/bnd-export-maven-plugin/pom.xml
```diff
@@ -52,6 +51,23 @@
...
+                <artifactId>maven-dependency-plugin</artifactId>
+				<configuration>
+				    <artifactItems>
+				        <artifactItem>
+				            <groupId>biz.aQute.bnd</groupId>
+				            <artifactId>biz.aQute.bnd.embedded-repo</artifactId>
+				            <version>${project.version}</version>
+				            <type>jar</type>
+				            <overWrite>false</overWrite>
+				            <outputDirectory>${project.build.directory}/classes</outputDirectory>
```
bjhargrave: ${project.build.outputDirectory}

maven/bnd-testing-maven-plugin/pom.xml
```diff
@@ -52,6 +52,23 @@
...
+                <artifactId>maven-dependency-plugin</artifactId>
+				<configuration>
+				    <artifactItems>
+				        <artifactItem>
+				            <groupId>biz.aQute.bnd</groupId>
+				            <artifactId>biz.aQute.bnd.embedded-repo</artifactId>
+				            <version>${project.version}</version>
+				            <type>jar</type>
+				            <overWrite>false</overWrite>
+				            <outputDirectory>${project.build.directory}/classes</outputDirectory>
```
bjhargrave: ${project.build.outputDirectory}

maven/bnd-export-maven-plugin/pom.xml
bjhargrave: Yes. Using classes means failure if user uses a diff output folder.

PR #1814, opened 2016-12-13
[maven testing] better console header output

rotty3000:


maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java
```diff
@@ -131,8 +126,10 @@ private void testing(Run run) throws Exception {
 			if (run.getErrors().isEmpty())
 				return;
 
-			System.out.println(Strings.join("\n", run.getErrors()));
-			throw new MojoExecutionException("Test had errors");
+			throw new MojoExecutionException(
+					Strings.join("\n", new Object[] {
```
bjhargrave: You should not need to declare an array here since I changed join to be varargs.

maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java
```diff
@@ -131,8 +126,10 @@ private void testing(Run run) throws Exception {
...
+							"Test had errors:", Strings.join("\n", run.getErrors())
+					}));
 		} catch (Exception e) {
```
bjhargrave: Doesn't this catch the MojoExecutionException you just threw above? And rethrow it? That seems terrible.

maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java
```diff
@@ -105,7 +105,11 @@ private void testing(File runFile) throws MojoExecutionException, Exception, IOE
 
 	private void testing(Run run) throws Exception {
 		try {
-			System.out.println("Test " + run);
+			System.out.println(Strings.join("\n", new Object[] {
```
bjhargrave: This looks like a case for printf with a format string and not Strings.join.

maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java
```diff
@@ -64,7 +64,9 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 			} else {
 
 				Glob g = new Glob(testing == null ? "*" : testing);
-				logger.info("Matching glob {}", g);
+				if (!g.toString().equals("*")) {
+					logger.info("Matching glob {}", g);
```
bjhargrave: Perhaps this should just be logger.debug instead of an if test to not log.

biz.aQute.junit/src/aQute/junit/ConsoleReporter.java
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	@Override
+	public void addFailure(Test test, AssertionFailedError a) {
+		testMap.putIfAbsent(test, new State()).a = a;
+	}
+
+	@Override
+	public void endTest(Test test) {
+		StringBuilder sb = new StringBuilder();
```
bjhargrave: What is the value of using a StringBuilder in this method when you just print it out at the end of the method? Why not just System.out.print as you go instead of sb.append?

In fact, you could do it all with a single print if you add a toString method to State which handles its string value.

maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java
rotty3000: the format string would be super long and I wasn't sure how to deal with that without also having `+`s all over the place.
the varargs join doesn't work when the delimiter is the same type as the args you mean to delimit.

maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java
rotty3000: I can do that!

maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java
rotty3000: yup, as I said earlier, the varargs join doesn't work when the delimiter is the same type as the args you mean to delimit.

maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java
rotty3000: I didn't write that... this is exactly how it was, I just fixed the message to remove the duplicate output.

biz.aQute.junit/src/aQute/junit/ConsoleReporter.java
rotty3000: I can try this.

rotty3000:
a bunch of fixes as requested.

biz.aQute.junit/src/aQute/junit/ConsoleReporter.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+import aQute.lib.exceptions.Exceptions;
+import junit.framework.AssertionFailedError;
+import junit.framework.Test;
+import junit.framework.TestListener;
+
+public class ConsoleReporter implements TestListener {
```
bjhargrave: Why isn't this just done as part of BasicTestReport class? I don't see much value in making it a separate TestListener that is always added.

rotty3000: because the implementation of BasicTestReport class is doing a lot of other things that are not in any way related to console output.

bjhargrave: That does not seem a meaningful answer. Both are TestListeners, both are called during test execution. Why have 2 listeners, when one will do? Btw, the output of your code is captured by the Tee streams. Not sure if you intended that.

rotty3000: Maybe I should get rid of the Tee streams because they are conflating different concerns. I feel multiple listeners is the right model for handling different use cases of the test output.

bjhargrave: No, you cannot get rid of the Tee streams. They are necessary to capture any output of the test cases. (In fact I have a small PR to make to fix a bug I found there while reviewing this PR.)

bjhargrave: > In fact I have a small PR to make

I have merged this PR

rotty3000:
@bjhargrave this time I took a minimalist approach to get the output I really needed.

rotty3000:
updated!

biz.aQute.junit/src/aQute/junit/BasicTestReport.java
```diff
@@ -38,12 +38,20 @@ public void addError(Test test, Throwable t) {
 		activator.trace("  add error to %s : %s", test, t);
 		check();
 		errors++;
+		systemOut.capture(false);
+		System.out.printf("TEST %s <<< FAILURE:", test);
```
bjhargrave: Should be ERROR.

biz.aQute.junit/src/aQute/junit/BasicTestReport.java
```diff
@@ -38,12 +38,20 @@ public void addError(Test test, Throwable t) {
...
+		t.printStackTrace(System.out);
+		systemOut.capture(true);
 	}
 
 	public void addFailure(Test test, AssertionFailedError t) {
 		activator.trace("  add failure to %s : %s", test, t);
 		check();
 		fails++;
+		systemOut.capture(false);
+		System.out.printf("TEST %s <<< ERROR:", test);
```
bjhargrave: Should be FAILURE

PR #1817, opened 2016-12-14
processor: Fix concurrency issue with plugins Set

bjhargrave:
The plugins Set can now be safely iterated over.

Fixes https://github.com/bndtools/bndtools/issues/1610

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
@@ -26,7 +26,6 @@ import java.util.Enumeration;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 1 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
```diff
@@ -2216,14 +2218,16 @@ public static ScheduledExecutorService getScheduledExecutor() {
 	 */
 	public synchronized void addBasicPlugin(Object plugin) {
 		basicPlugins.add(plugin);
-		if (plugins != null)
-			plugins.add(plugin);
+		Set<Object> p = plugins;
```
rherrmann: Pardon my curiosity, but what is the purpose of the local variable `p`?

bjhargrave: Just a hygiene habit. I copy the mutable instance variable into a local variable and then use it knowing the local variable cannot be altered during this method's execution.

rherrmann: OK, I see. Thanks for explaining.

PR #1822, opened 2016-12-15
prevent copy if destDir is child of srcDir

gamerson:


aQute.libg/test/aQute/lib/io/IOTest.java
```diff
@@ -13,4 +15,44 @@ public void testSafeFileName() {
...
+		assertTrue(parentDir.mkdirs());
+
+		File childDir = new File("generated/test/parentDir/childDir");
+
+		try {
+			IO.copy(parentDir, childDir);
+
+			fail("Expected IllegalArgumentException");
+		} catch (IllegalArgumentException e) {
+			assertEquals(e.getMessage(), "target directory can not be child of source directory.");
```
bjhargrave: I think it is too much to test the exception message.

gamerson:
Fixed.

PR #1829, opened 2017-01-04
try to fallback to copying files on systems where symbolic links may fail

gamerson:


biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -93,14 +97,37 @@ File get(String bsn, Version version, Map<String,String> properties, DownloadLis
...
+			Files.createLink(linkPath, sourcePath);
+			createdSymlink.set(true);
+		} catch (FileSystemException e) {
+			createdSymlink.set(false);
+		}
 
 		Promise<File> go = client.build().useCache(MAX_STALE).async(url.toURL()).map(new Function<File,File>() {
 			@Override
 			public File apply(File t) {
+				// if we can't create symlinks we need to copy this into link
```
bjhargrave: I guess I don't see why we cannot also defer the link into apply so that apply is create link or copy file. Then we don't need an AtomicBoolean.

gamerson: Done and pushed!

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -93,14 +96,34 @@ File get(String bsn, Version version, Map<String,String> properties, DownloadLis
...
-		Files.createLink(link.toPath(), source.toPath());
+		final Path linkPath = link.toPath();
 
 		Promise<File> go = client.build().useCache(MAX_STALE).async(url.toURL()).map(new Function<File,File>() {
 			@Override
 			public File apply(File t) {
+				// links can fail if they span disparate Windows drives
+				// so we need to handle them slightly differently
+				try {
+					if (link.isFile() && Files.isSymbolicLink(linkPath))
```
bjhargrave: This seems like a failure point if the file exists and is not a symlink, we then try to create the symlink. And I guess expect some exception to dump us into the copy catch block. I think we should not use exception for control flow in that case.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -112,6 +135,12 @@ public File apply(File t) {
 		return link;
 	}
 
+	private boolean needsCopy(Path sourcePath, Path destPath) {
+		final File dest = destPath.toFile();
+
+		return !dest.exists() || dest.length() != sourcePath.toFile().length();
```
bjhargrave: So if we could not create a symlink, we try to shortcut the copy if the files are the same. But I think a length test is pretty weak. We really need a sha match, no?

gamerson: I agree we need more than content length, we need sha or md5 to truly determine if we need to recopy.  Because there are bad behaving p2 repos out there, that keep the version and bsn the same, but might change the content and the content length accidentally stays the same (even though the chances here are pretty remote).

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
gamerson: Yeah i don't know what happens, if you try to create a symlink to a file that already exists.  It might not through a FileSystemException but another type of IOException.

bjhargrave:
See https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/util/repository/DownloadListenerPromise.java#L52 where there is another use of `createLink` that could fail on Windows. 

And also https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L1819-L1824 where `IO.createSymbolicLink` is used.

Perhaps we need an `IO` method to make a link or copy if we cannot make the link? To encapsulate the desired behavior.

bjhargrave:
Agreed. I am not sure why we want to use hardlinks vs. symlinks. Hardlinks requires the original file and the link are on the same file system. I think we can see if both paths would be on the same file system with `Files.getFileStore(Path)`. If both paths are on the same FileStore, then we could hardlink, otherwise we could symlink?

fhuberts:
IMHO linking should not be done at all here, or anywhere else for that matter.
linking only provides value if both the source and the destination are under your control.
which they are definitely not in this case

bjhargrave:
> if both the source and the destination are under your control.
> which they are definitely not in this case

How are they not under bnd's control? It manages the source in the bnd cache and the destination.

fhuberts:
this is in the indexer, which is generic, and can therefore index anything, also things that are not under your control

bjhargrave:
> this is in the indexer

This has nothing to do with the indexer. P2Indexer is an internal class in the P2Repository implementation which manages the artifacts which are in the p2 repo.

gamerson:
superceded by https://github.com/bndtools/bnd/pull/1831

eliasvasylenko:
I think it's worth noting that in Windows, since #1831, the following `java.nio.file.FileSystemException` is thrown when attempting to create the symlinks: `A required privilege is not held by the client`.

To resolve this UAC must be turned off. Perhaps this is the reason for the original decision to use hard links instead of soft? It seems the required permissions are different.

Can any other users confirm this behaviour? It may be sensible to re-examine this PR.

bjhargrave:
> Can any other users confirm this behaviour? It may be sensible to re-examine this PR.

Yes, it has been seen by one of the bndtools devs who is a windows user. Please open a new issue for this.

PR #1835, opened 2017-01-10
Don't create symlinks on Windows

seanbright:


aQute.libg/src/aQute/lib/io/IO.java
```diff
@@ -636,6 +636,31 @@ public static boolean isSymbolicLink(File link) {
...
+	 * @param target the target of the symbolic link, or source of the copy.
+	 * @return {@code true} if the operation succeeds, {@code false} otherwise.
+	 */
+	public static boolean createSymbolicLinkOrCopy(File link, File target) {
+		try {
+			if (isWindows()) {
+				IO.copy(target, link);
+				return true;
+			}
+			return createSymbolicLink(link, target);
```
bjhargrave: I think that is the symbolic link fails, you need to fallback to copy.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -97,7 +98,7 @@ File get(String bsn, Version version, Map<String,String> properties, DownloadLis
 		if (link.isFile())
 			Files.delete(link.toPath());
 
-		Files.createSymbolicLink(link.toPath(), source.toPath());
+		IO.createSymbolicLinkOrCopy(link, source);
```
bjhargrave: Hmmm, this can get busy on a windows system since we always delete the file above and then recopy it. :-(

Deleting the symlink and recreating is (hopefully) cheap. But always deleting and recopying a potentially large file for no reason is wasteful.

@gamerson's original attempt at this tried to detect if the target file was already the proper file to shortcut the delete/copy work. So we may need something that computes the sha of the source and target and deletes/copies only when the shas differ. I guess the question is then whether the the sha computation is more costly than always delete/copy?

aQute.libg/src/aQute/lib/io/IO.java
seanbright: Addressed.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
seanbright: I'll take a look.

gamerson:
This work has been continued here: https://github.com/bndtools/bnd/pull/1878

gamerson:
Work continues here: https://github.com/bndtools/bnd/pull/1878

FYI, I just realized for committers we can push to other commiters branches (I think) so in this case I could have just pushed more commits here instead of starting a new PR.

PR #1838, opened 2017-01-10
[bnd-export-maven-plugin] fixes #1837 bnd-export-maven-plugin producing unstable outout

### LINKED ISSUE #1837: bnd-export-maven-plugin producing unstable outout
this plugin has been doing things like multiple prints of the same or different output, which is confusing. It also has two levels of change calculation which is causing it to behave erratically.

rotty3000:


maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
```diff
@@ -10,7 +10,6 @@
 import java.util.Comparator;
 import java.util.List;
 
-import org.apache.commons.collections.CollectionUtils;
```
bjhargrave: Since you are not using commons collections anymore, remove it as a dependency in the pom.

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
```diff
@@ -99,7 +103,6 @@ private void export(Run run) throws Exception {
 	private void resolve(StandaloneRun run) throws Exception, IOException {
 		try (ProjectResolver projectResolver = new ProjectResolver(run)) {
 			List<Container> runBundles = projectResolver.getRunBundles();
```
bjhargrave: This is done differently that the gradle plugin's resolve task: https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BndPlugin.groovy#L480-L491. We should use identical logic to avoid discrepancies.

I don't know which one is better, but they should be the same logic.

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
```diff
@@ -108,65 +111,46 @@ public int compare(Container o1, Container o2) {
...
+			File runFile = run.getPropertiesFile();
+			BndEditModel bem = new BndEditModel(run.getWorkspace());
+			Document doc = new Document(IO.collect(runFile));
+			bem.loadFrom(doc);
+
+			List<VersionedClause> originalBundles = bem.getRunBundles() != null ? bem.getRunBundles()
+					: new ArrayList<VersionedClause>();
+			List<VersionedClause> calculatedRunBundles = new ArrayList<>();
+			for (Container c : runBundles) {
+				ResourceBuilder rb = new ResourceBuilder();
```
bjhargrave: The way the gradle plugin works avoids the need for Container and ResourceBuilder, so perhaps it is simpler. https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BndPlugin.groovy#L480-L491

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
rotty3000: good one...

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
rotty3000: it's done differently because Groovy can use a lambda and get the resolver instance. Can we do that without java 8 in this plugin?

rotty3000: .. I think I can hack it together...

bjhargrave: It is not about groovy and closures. It is about using the keyset of the resolve result vs. calling getRunBundles which converts the Resources to Container so that you have to convert them back to Resources. Just do like the gradle plugin does and use the keyset of the resolve result. Also when you sort, you need to sort by bsn and then by version (as the gradle plugin does.)

rotty3000:
hey @bjhargrave I made several updates. Can review? thx!

PR #1841, opened 2017-01-13
Fixed format string

vintermann:
String.format uses C-style substitution, not {} as the logger does.


bjhargrave:

rotty3000:
@vintermann btw, thanks for catching this! Sorry about letting it get through!

PR #1845, opened 2017-01-17
[remote agent] index a remote framework

rotty3000:


rotty3000:
Hey @gamerson @pkriens @bjhargrave,

For your consideration, this change makes it possible to get an index of the remote framework from a remote agent and client.

I've been considering this possibility for a while. Though the distro command is somewhat similar it's missing the possibility of listing the actual bundles installed because it merges all capabilities under one single resource.

A real index of the framework gives many other possibilities. It's true that the index doesn't currently have a way to point at real resources and provide shas for the bundles, but that is just a detail that we can work out in the future, and I feel this is still pretty valuable.

pkriens:
I am not sure I understand how this works. The Agent should have no visibility to the resource indexer? The purpose of the agent is to be tiny. However, I did not see a change in the bnd file though I would've expected that?

rotty3000:
It's true that this does adds heft to the agent, which is why I wanted to
get your sanity review here. But it's working as it is, and the agent.bnd
does contain the additions required.
... on the other hand, is it really a huge deal to increase the size of the
agent really?

pkriens:
YES! The agent should be minimal. bndlib is a build time library we do not want to encumber the runtime with those requirements. E.g. bndlib will likely go 1.8 while we want to keep the agent to a lower version. 

That said, I think there are DTOs that provide the same information already. Then you can just do the storing on the client side.

rotty3000:
I'll take a look at the DTOs. Perhaps we can fudge them into resources somehow.

PR #1846, opened 2017-01-17
[remote agent] index a remote framework

rotty3000:


rotty3000:
@pkriens how's this instead?

pkriens:
Well, same problem? Why not do it as the distro command does? It gets the bundle revisions from the remote and then indexes it into a single manifest. Seems trivial to adapt this to create an index?

And why not use the distro jar? Seems easier to manipulate in the bndrun file then another indexed plugin?

rotty3000:
the distro jar loses a lot of information! It removes all the resource capabilities!

rotty3000:
It loses all of osgi.wiring.bundle, and osgi.wiring.host so that you can't resolve fragments against it.

PR #1851, opened 2017-01-22
Extra checks for Bundle-Activator in the Verifier

### LINKED ISSUE #1848: Misleading error "The default package '.' is not permitted by the Import-Package syntax"
I am facing a 'default package is not permitted' error in my Bndtools workspace. The `cnf/build.bnd` file has this contents:
```
javac.source: 1.8
javac.target: 1.8
-runee: JavaSE-1.8
Bundle-Activator: ${classes;IMPLEMENTS;org.osgi.framework.BundleActivator}
```
The `cnf/ext/repositories.bnd` declares two repositories:
```
-plugin: \
    aQute.bnd.repository.osgi.OSGiRepository; \
        name=Bndtools Hub; \
        locations=https://raw.githubusercontent.com/bndtools/bundle-hub/master/index.xml.gz; \
    , \
    aQute.bnd.repository.maven.pom.provider.BndPomRepository; \
        releaseUrls=https://repo.maven.apache.org/maven2/; \
        pom=${workspace}/cnf/maven-central.xml; \
        name=Maven Central;
```
and the `bnd.bnd` file contents of the only project in this workspace is:
```
-testpath: \
    junit:junit;version=latest
```

The `Bundle-Activator` header in `cnf/build.bnd` seems to cause the confusion. After removing the line, the workspace builds just fine.

Event though the project does not provide an activator class, from reading the [Automatic Bundle-Activator FAQ entry](http://bnd.bndtools.org/chapters/920-faq.html?#automaticActivator) I assume that it should be legal to do so.

But even if an activator must be provided, given the `Bundle-Activator` header is present, a more revealing error message would help.

The full error message is:
> The default package '.' is not permitted by the Import-Package syntax.
 This can be caused by compile errors in Eclipse because Eclipse creates
valid class files regardless of compile errors.
The following package(s) import from the default package null

Though I am using bnd through Bndtools, I got the impression that this is an issue with bnd itself. Please let me know if this issue doesn't belong here.

timothyjward:
Fixes #1848

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/AnalyzerTest.java
+++ b/biz.aQute.bndlib.tests/src/test/AnalyzerTest.java
@@ -742,6 +742,108 @@ public class AnalyzerTest extends BndTestCase {
 	}
 
 	/**
+	 * Scan for a BundleActivator, but there are no matches!
+	 * 
+	 * @throws Exception
+	 */
+	public static void testBundleActivatorNoType() throws Exception {
+		Builder a = new Builder();
+		try {
+			Properties p = new Properties();
+			p.put("Private-Package", "test.api");
+			p.put("Bundle-Activator", "");
+			a.addClasspath(new File("bin"));
+			a.setProperties(p);
+			a.build();
+			Manifest manifest = a.getJar().getManifest();
+	
+			assertEquals(a.getErrors().toString(), 0, a.getErrors().size());
+			assertEquals(a.getWarnings().toString(), 1, a.getWarnings().size());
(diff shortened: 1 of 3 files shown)


timothyjward:
Note that this fix makes the following changes:

* It is an error to have a Bundle-Activator which isn't a valid type name
* It is an error to have a list of type names in the Bundle-Activator header
* It is an error to have an empty Bundle-Activator header
* It is a warning to have macro which results in an empty Bundle-Activator

PR #1858, opened 2017-01-24
Handle protected changes in Provider type classes

timothyjward:
In a provider type class protected fields/methods should never be accessible to the consumer. Therefore if these fields are deleted it should not be a breaking change, therefore a MINOR bump should be sufficient...


rotty3000:
Isn't a publicly visible deletion always breaking whether the thing being deleted was protected or not?

timothyjward:
@pkriens - what do you think of this change?

Binary compatibility for classes has always been more confusing than interfaces, but in this case I think the proposed change is correct. Specifically, if I have a `@ProviderType` class (e.g. an Exception type thrown by a library) then I don't think the removal of a `protected` field/method should count as a consumer breaking change. 

The logic is as follows:

1. `protected` methods and fields are not accessible unless you subclass the type
2. If you subclass the type then you are acting as a provider
3. If you do not subclass the type then you have *no visibility* of any additions or deletions

As a result consumers, who can never see the field/method, cannot be broken by its removal. The change is therefore provider breaking, and should result in a MINOR bump.

timothyjward:
> Isn't a publicly visible deletion always breaking whether the thing being deleted was protected or not?

@rotty3000 yes, but the point is that these changes are not publicly visible. The class is `@ProviderType` so protected members are not visible to consumers.

rotty3000:
Ok, I'm sold! But I'd like to get both @pkriens' and @bjhargrave's feedback on this for sanity.

timothyjward:
> Ok, I'm sold! But I'd like to get both @pkriens' and @bjhargrave's feedback on this for sanity.

I already spoke to @bjhargrave in person, and we all agree that we want @pkriens to weigh in ;)

pkriens:
I agree your reasoning makes sense. It is possible that we break people but sometimes that is necessary for progress. I suggest you add an explanation to the ProviderType annotation that elucidates this new rule. I.e. a subclass of a ProviderType is considered a provider.

Issue #1834, opened 2017-01-10
FileSystemException creating cache symlinks for P2Repository due to insufficient priviliges in Windows


eliasvasylenko:
Since the changes in #1831 the following `java.nio.file.FileSystemException` is thrown when attempting to create cache symlinks in Windows: `A required privilege is not held by the client`.

Currently to resolve this users must turn off UAC.

This could be addressed with a fall-back behaviour which copies instead of links, as suggested in #1829

To give the best behaviour for the greatest number of users the ideal solution may be to implement a series of fallbacks: `Soft link -> Hard link -> Copy`.

bjhargrave:
I think for windows, we abandon all hope of links and always make a copy. Windows FS is too brain damaged. Making symlinks requires a permission windows users don't normally have or, if they are admin, requires run as admin to make symlinks <https://technet.microsoft.com/en-us/library/cc766301(WS.10).aspx>.

Hard links don't work across file systems anywhere, so that can fail on a regular basis.

So I think we need to use symlinks if not windows. If windows or symlinking fails, then copy.

PR #1878, opened 2017-02-04
Don't create symlinks on Windows

gamerson:
Closes https://github.com/bndtools/bnd/pull/1835
Fixes https://github.com/bndtools/bnd/issues/1834

### DIFF
--- a/aQute.libg/src/aQute/lib/io/IO.java
+++ b/aQute.libg/src/aQute/lib/io/IO.java
@@ -636,6 +636,31 @@ public class IO {
 		return Files.isSymbolicLink(link.toPath());
 	}
 
+	/**
+	 * Creates a symbolic link from {@code link} to the {@code target}, or
+	 * copies {@code target} to {@code link} if running on Windows.
+	 * <p>
+	 * Creating symbolic links on Windows requires administrator permissions, so
+	 * copying is a safer fallback.
+	 *
+	 * @param link the location of the symbolic link, or destination of the
+	 *            copy.
+	 * @param target the target of the symbolic link, or source of the copy.
+	 * @return {@code true} if the operation succeeds, {@code false} otherwise.
+	 */
+	public static boolean createSymbolicLinkOrCopy(File link, File target) {
+		try {
+			if (isWindows()) {
+				IO.copy(target, link);
+				return true;
+			}
+			return createSymbolicLink(link, target);
+		} catch (Exception ignore) {
+			// ignore
+		}
+		return false;
+	}
+
 	static public OutputStream	nullStream	= new OutputStream() {
 
 												@Override
--- a/aQute.libg/src/aQute/lib/io/IO.java
+++ b/aQute.libg/src/aQute/lib/io/IO.java
@@ -650,11 +650,10 @@ public class IO {
 	 */
 	public static boolean createSymbolicLinkOrCopy(File link, File target) {
 		try {
-			if (isWindows()) {
+			if (isWindows() || !createSymbolicLink(link, target)) {
 				IO.copy(target, link);
-				return true;
 			}
-			return createSymbolicLink(link, target);
+			return true;
 		} catch (Exception ignore) {
 			// ignore
 		}
--- a/aQute.libg/src/aQute/lib/io/IO.java
+++ b/aQute.libg/src/aQute/lib/io/IO.java
@@ -28,6 +28,7 @@ import java.net.URL;
 import java.net.URLConnection;
 import java.nio.ByteBuffer;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.security.MessageDigest;
 import java.util.Collection;
 import java.util.LinkedHashSet;
@@ -623,6 +624,16 @@ public class IO {
 	}
 
 	public static boolean createSymbolicLink(File link, File source) throws Exception {
+		if (isSymbolicLink(link)) {
+			Path linkSource = Files.readSymbolicLink(link.toPath());
+
+			if (source.toPath().equals(linkSource)) {
+				return true;
+			} else {
+				link.delete();
+			}
+		}
+
 		try {
 			Files.createSymbolicLink(link.toPath(), source.toPath());
 			return true;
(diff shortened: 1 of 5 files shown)


gamerson:
hey @seanbright  i pulled your commits, rebased them to master and add a couple of more.  This is based on bndtools dev call discussion where we decided to add a check for lastModified and length before we copy.  @pkriens  mentioned this should be "good enough" for our usage.

aQute.libg/src/aQute/lib/io/IO.java
```diff
@@ -639,6 +636,35 @@ public static boolean isSymbolicLink(File link) {
...
+	 * @param target the target of the symbolic link, or source of the copy.
+	 * @return {@code true} if the operation succeeds, {@code false} otherwise.
+	 */
+	public static boolean createSymbolicLinkOrCopy(File link, File target) {
+		try {
+			if (isWindows() || !createSymbolicLink(link, target)) {
+				// only copy if target length and timestamp differ
+				if (target.lastModified() != link.lastModified() || target.length() != link.length()) {
+					IO.copy(target, link);
+					target.setLastModified(link.lastModified());
```
bjhargrave: This is backwards. `link` is what is being created by the `IO.copy`. So the last modified of `link` needs to be set to the last modified value of the `target`. (target is a terrible variable name here.)

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -97,7 +98,7 @@ File get(String bsn, Version version, Map<String,String> properties, DownloadLis
 		if (link.isFile())
 			Files.delete(link.toPath());
```
bjhargrave: I think we also need to remove this delete since we want the createSymbolicLinkOrCopy method to be able to avoid the copy if the file was previously copied and there is no change to the source file.

aQute.libg/src/aQute/lib/io/IO.java
```diff
@@ -622,9 +622,6 @@ public static PrintWriter writer(OutputStream out) throws IOException {
 		return writer(out, "UTF-8");
 	}
 
-	/**
-	 * Reflective way to create a link. This assumes Java 7+
-	 */
 	public static boolean createSymbolicLink(File link, File target) throws Exception {
 		try {
 			Files.createSymbolicLink(link.toPath(), target.toPath());
```
bjhargrave: I think this code needs to handle the case where the link already exists and is pointing to the requested target. In that case, return true. (Alternately, if could test if link is a symbolic link already, and it so, delete it before creating a new link that points to the target file.)

I think the current code will catch an exception and return false which means createSymbolicLinkOrCopy will then copy.

This change is needed since we cannot always delete the link before calling createSymbolicLinkOrCopy since it may be a copy which is unchanged from the target.

aQute.libg/src/aQute/lib/io/IO.java
gamerson: Yes I did that backwards, partly because of confusing variable name.  I'll switch this up.

gamerson: Unix's `ln` command uses the terms 'source' and 'target'.  Which to me makes more sense then the JDK's api.

aQute.libg/test/aQute/lib/io/IOTest.java
```diff
@@ -54,4 +55,70 @@ public void testDestDirIsChildOfSource() throws Exception {
...
+
+		link.delete();
+
+		assertFalse(Files.isSymbolicLink(link.toPath()));
+
+		Files.createSymbolicLink(link.toPath(), originalSource.toPath());
+
+		File newSource = new File("testresources/zipped.dat");
+
+		IO.createSymbolicLinkOrCopy(link, newSource);
```
bjhargrave: On Windows this will always copy instead of link.

gamerson: Ah yes so i guess that is why this is failing on windows?

gamerson: BTW, I could not scroll to the bottom of that Appveyor log, it was crashing my chrome browser on macOS

bjhargrave:
```
:aQute.libg:test
aQute.lib.io.IOTest > testCreateSymlinkOrCopyWillDeleteOriginalLink FAILED
    junit.framework.AssertionFailedError
        at junit.framework.Assert.fail(Assert.java:55)
        at junit.framework.Assert.assertTrue(Assert.java:22)
        at junit.framework.Assert.assertTrue(Assert.java:31)
        at junit.framework.TestCase.assertTrue(TestCase.java:201)
        at aQute.lib.io.IOTest.testCreateSymlinkOrCopyWillDeleteOriginalLink(IOTest.java:122)
182 tests completed, 1 failed
```

gamerson: OK, I force pushed a new commit that makes this test skip windows.

bjhargrave: I am not sure that is the right fix. I think the test is still somewhat interesting, but I think the expected results are in different. On either Windows or not windows, the link should be replaced.

aQute.libg/test/aQute/lib/io/IOTest.java
```diff
@@ -54,4 +55,74 @@ public void testDestDirIsChildOfSource() throws Exception {
...
+
+	public void testCreateSymlinkOrCopyWillDeleteOriginalLink() throws Exception {
+		File originalSource = new File("testresources/unzipped.dat");
+		File link = new File("generated/test/originalLink");
+
+		link.delete();
+
+		assertFalse(Files.isSymbolicLink(link.toPath()));
+
+		Files.createSymbolicLink(link.toPath(), originalSource.toPath());
```
bjhargrave: Is there a risk this `Files.createSymbolicLink` call will fail on some users' systems? It seems to work on the appveyor machines. But some other person building on Windows may not run as admin. I am thinking this call should be

```java
assertTrue(IO.createSymbolicLinkOrCopy(link, originalSource));
```

You should probably also `assertTrue` on the next call to `IO.createSymbolicLinkOrCopy` too.

gamerson: agreed.

PR #1884, opened 2017-02-08
fix connection-settings

gamerson:


gamerson:
Lets give it a whirl!

PR #1887, opened 2017-02-09
fixes #1886 fragment host versions are not properly parsed into requirement filter

### LINKED ISSUE #1886: fragment host versions are not properly parsed into requirement filter
the ResourceBuilder is wrongly attempting to use the `version` attribute instead of the `bundle-version` attribute which applies to `osgi.wiring.host` capability.

rotty3000:


biz.aQute.bndlib.tests/src/test/ResourceTest.java
```diff
@@ -204,7 +206,7 @@ public void testResourceToVersionedClause() throws Exception {
 		VersionedClause versionClause = ResourceUtils.toVersionClause(resource, "[===,==+)");
 		StringBuilder sb = new StringBuilder();
 		versionClause.formatTo(sb);
-		assertEquals("demo-fragment;version='[3.3.0,3.3.1)'", sb.toString());
+		assertEquals("demo-fragment;version='[3.4.0,3.4.1)'", sb.toString());
```
bjhargrave: I am thinking the demo-fragment project should set it's version to a fixed 1.0 version so that each time we need to rebuild demo-fragment, we don't have to change this assert.

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java
```diff
@@ -462,11 +462,11 @@ public void addFragmentHost(String bsn, Attrs attrs) throws Exception {
...
 		filter.append("(").append(HostNamespace.HOST_NAMESPACE).append("=").append(bsn).append(")");
 
-		String v = attrs.getVersion();
+		String v = attrs.get(HostNamespace.CAPABILITY_BUNDLE_VERSION_ATTRIBUTE);
 		if (v != null && VersionRange.isOSGiVersionRange(v)) {
 			VersionRange range = VersionRange.parseOSGiVersionRange(v);
 			filter.insert(0, "(&");
-			filter.append(range.toFilter());
+			filter.append(range.toFilter().replaceAll("version", "bundle-version"));
 			filter.append(")");
```
bjhargrave: Lets not do a replaceAll. Instead, fix `VersionRange.toFilter()` (or overload it) to take the attribute name as a argument. See OSGi's `VerionRange.toFilterString(String)`. The then call here becomes `range.toFilter(HostNamespace.CAPABILITY_BUNDLE_VERSION_ATTRIBUTE)`.

demo-fragment/bnd.bnd
```diff
@@ -1,3 +1,3 @@
-Fragment-Host: demo
+Fragment-Host: demo;bundle-version='[1.1,1.2)'
 Provide-Capability: some.namespace;somenamespace=hello;version:Version=1.0.0
 -resourceonly: true
```
bjhargrave: As I mentioned above, add `Bundle-Version: 1.0` here to fix the bundle version.

rotty3000:
Will do!

rotty3000:
Updated!

PR #1889, opened 2017-02-11
bnd distro command creates jar with lastModified value of 0

gamerson:


pkriens:
Why you write the JAR before you set the modified date? There could be a reason if you use the Jar later on but it looks kind of weird and could use some comment?

gamerson:
Ah I see the write api now, Jar has an 'updateModified' that looks like the right method to you.  I've pushed an updated commit.

pkriens:
LGTM!

PR #1891, opened 2017-02-13
[resolve] always use the EE defined in the configuration.

rotty3000:


bjhargrave:
Not sure I understand `-distro` enough to render an opinion here.

pkriens:
I am not sure I undestand why the EE should be used in Distro? This should come from the target/distro shouldn't it?

rotty3000:
The problem with the distro is that it falls short of the work done in the non-distro use case. The BndrunResolveContext does extra work to fill out the value `-runee` to include every compatible lower order execution environment.

However, when the distro adds the ee it only adds the exact current ee under which it's running. And since EEs are not inclusive by default, this causes a situation where the distro contains let's say, a 1.7 ee and so a bundle requiring a 1.6 ee fails to resolve.

So, we'd then have to do the same work in distro generation to fill out the ee compatible values which could be doable.

However, what about the scenario where a developer wants to test an existing distro against a higher versioned EE? This is impossible if the ee is encoded into the distro. The distro would have to be completely regenerated. Which at that moment might not be entirely easy.

pkriens:
Not sure I understand. The remote system must have all the backward compatible capabilities for the EE? So why don't we pick them up automatically? The remote framework must have calculated those?

rotty3000:
I'll retest just to make sure I didn't miss something!

rotty3000:
Thanks @pkriens for the sanity check. You were right of course. I had broken the ee handling without realizing it. It's fixed and I added a test to assert it. I've also pruned out the none spec osgi.native attributes because it was leaking system properties and environment data.

I've updated. Please review.

pkriens:
:-) LGTM

Issue #1625, opened 2016-08-21
Gradle export task to create distribution of runbundles does not create distribution with all jars


Mr-Frantic:
When the bndrun file has runbundles have some jars which have the same name i.e.

-runbundle: \
    org.app.common.api;version='[8.0.0,8.0.1)',\
    org.app.dataenquiry.api;version='[8.0.0,8.0.1)'

The actual coordinates of jars in maven is:
org.app.common:api:8.0.0
org.app.dataenquiry:api:8.0.0
--- i.e. both jars are named as api.jar

So when we run the export task to create the distribution, the distribution contains only one api.jar

I think the export task should export the bundles with full bundle symbolic name instead of the jar name.

Also when we use the bundles generated by bndtools, the distribution generated doesn't contain the version of the bundles, it only contains the name of the bundle,

rotty3000:
Could you please retest this scenario after https://github.com/bndtools/bnd/commit/98a4af11a55b7a5703abc01e33fee388d235d21c ?

this may solve your issue!

rotty3000:
Sorry, the issue is different!

zhsc:
Not sure if this is your issue, but I sadly banged my head on this.  Make sure you run the "**jar**" task before running the export or runbundles tasks...

i.e. gradle clean **jar** export.dev

It quietly just doesn't pickup the bundles if they're not built already.

Seems like before it would implicitly do the jar task on all the required projects or something.

zhsc:
This may have been because I was trying to use the -p directive to specify a specific project.  This doesn't appear to be compatible with the export task.  So, this probably isn't your issue.

Mr-Frantic:
Sorry I am a bit confused with the discussion. 

I think the problem is because when we exportRunbundles, we copy the jar by the name of the jar.

In **aQute.bnd.build.Project.java**, method **exportRunbundles** there is the following code:

		for (Container container : runbundles) {
			File bundle = container.getFile();
			IO.copy(bundle, new File(outputDir, bundle.getName()));
		}

Instead we that can be changed use **bundleSymbolicName with version.**

Do you think this is the correct way to fix the export issue?

bjhargrave:
Project.export (via ProjectLauncherImpl.executable) has the same issue. Both export and exportBundles assume the file name of each file is unique when creating the output. We will need someway to make unique file names. BSN+Version may not work if the file is not a bundle (in the case of -runpath for Project.export).

rotty3000:
[...]

PR #1908, opened 2017-02-21
[maven] export bundles only

rotty3000:


biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java
```diff
@@ -161,6 +165,23 @@ public int compare(VersionedClause a, VersionedClause b) {
 	}
 
 	/**
+	 * Copy the calculated <code>-runbundles</code> into a directory.
+	 *
+	 * @param targetDir the directory in which <code>-runbundles</code> are
+	 *            placed
+	 * @throws Exception
+	 */
+	public void runbundles(File targetDir) throws Exception {
```
bjhargrave: Why don't you just call `Project.exportRunbundles(null, targetDir)` from  the mojo? This a duplicate of that method. You have made some improvements here using `Files`. Better to fix the original method. 

Note https://github.com/bndtools/bnd/issues/1625 which these methods suffer from. So better to have one method we can fix once.

rotty3000: Fixed to use the project method.

Also fixed the issue for potential file name collisions. 
However, fixing the issue for the launcher required slightly different code which I couldn't centralize in one place because the API for talking to file systems and jars wasn't close enough to warrant it. But, I did solve the issue, but in the simplest way possible. I added a _Windows_ style renaming logic.

biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
```diff
@@ -260,6 +261,11 @@ public Jar executable() throws Exception {
 					runpathShas.add(sha + ";name=\"" + file.getName() + "\"");
 				} else {
 					String newPath = "jar/" + file.getName();
+					String[] parts = Strings.extension(file.getName());
+					int i = 1;
+					while (jar.exists(newPath)) {
+						newPath = String.format("%s[%d].%s", parts[0], i++, parts[1]);
```
bjhargrave: Doesn't this lose the `jar/` prefix? Shouldn't the format string be `"jar/%s[%d].%s"`?

biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
```diff
@@ -260,6 +261,11 @@ public Jar executable() throws Exception {
 					runpathShas.add(sha + ";name=\"" + file.getName() + "\"");
 				} else {
 					String newPath = "jar/" + file.getName();
+					String[] parts = Strings.extension(file.getName());
```
bjhargrave: `Strings.extension` can return `null` if the file name has no `.` in it. Not likely but possible. The code should handle this eventuality to avoid future NPE bug reports.

biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
```diff
@@ -283,6 +289,11 @@ public Jar executable() throws Exception {
 					actualPaths.add("${JPMREPO}/" + sha);
 				} else {
 					String newPath = "jar/" + file.getName();
+					String[] parts = Strings.extension(file.getName());
```
bjhargrave: `Strings.extension` can return `null`.

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -2002,13 +2005,26 @@ public void exportRunbundles(String runFilePath, File outputDir) throws Exceptio
...
+			Path source = container.getFile().toPath();
+			Path target = conflictFreeFilePath(outputDir.toPath(), source);
+			Files.copy(source, target, StandardCopyOption.COPY_ATTRIBUTES);
 		}
 	}
 
+	private Path conflictFreeFilePath(Path outputDir, Path source) {
+		String fileName = source.getFileName().toString();
+		Path target = outputDir.resolve(fileName);
+		String[] parts = Strings.extension(fileName);
```
bjhargrave: `Strings.extension` can return `null`

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -2002,13 +2005,26 @@ public void exportRunbundles(String runFilePath, File outputDir) throws Exceptio
 		}
 		packageProject.clear();
 
+		outputDir.mkdirs();
 		Collection<Container> runbundles = packageProject.getRunbundles();
 		for (Container container : runbundles) {
-			File bundle = container.getFile();
-			IO.copy(bundle, new File(outputDir, bundle.getName()));
+			Path source = container.getFile().toPath();
+			Path target = conflictFreeFilePath(outputDir.toPath(), source);
```
bjhargrave: Hoist `outputDir.toPath()` out of the loop.

biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
```diff
@@ -283,6 +289,11 @@ public Jar executable() throws Exception {
...
+					int i = 1;
+					while (jar.exists(newPath)) {
+						newPath = String.format("%s[%d].%s", parts[0], i++, parts[1]);
```
bjhargrave: Same as above re losing `jar/` prefix.

rotty3000:
Updated!

bjhargrave:
Jumped the gun on the merge before the CI builds finished... :-)

rotty3000:
crap...

Issue #477, opened 2014-02-26
Baseline doesn't recognize changes in generic types


jepet:
When changing a method signature in generic types, say from:

```
public Collection<A> getAll()
```

to:

```
public Collection<B> getAll()
```

will not be detected as a change at all when running the Baseline.baseline(..) method, rather it passes through as unchanged. This is bad, as the change is completely incompatible and should be marked as a Major change!

PR #1915, opened 2017-02-24
baseline: Compare on erased types and handle void/Object return types

bjhargrave:
Baselining should be about binary compatibility and the only the erased
signature matters. Prior to this fix, adding generics to code which did
not alter the erased signature would result in a MAJOR baseline change
as the differ saw a method added and a method removed.

Also, the diff failed to properly rate changing the return type from
void or Object to some other type. Changing from void to non-void or
from Object to some other type is equivalent to replacing the method.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/differ/JavaElement.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/differ/JavaElement.java
@@ -30,8 +30,6 @@ import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.jar.Manifest;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
 import aQute.bnd.header.Attrs;
 import aQute.bnd.header.OSGiHeader;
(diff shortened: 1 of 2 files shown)


bjhargrave:
@pkriens This fixes baselining for the OSGi build. Wanted you to review.

biz.aQute.bndlib/src/aQute/bnd/differ/JavaElement.java
```diff
@@ -685,26 +685,28 @@ private String toString(TypeRef[] prototype) {
...
+	static final Element	SHORT_R		= new Element(RETURN, "short");
+	static final Element	CHAR_R		= new Element(RETURN, "char");
+	static final Element	INT_R		= new Element(RETURN, "int");
+	static final Element	LONG_R		= new Element(RETURN, "long");
+	static final Element	FLOAT_R		= new Element(RETURN, "float");
+	static final Element	DOUBLE_R	= new Element(RETURN, "double");
+	static final Element	OBJECT_R	= new Element(RETURN, "java.lang.Object");
 
 	private void getCovariantReturns(Collection<Element> elements, TypeRef type, String generics) throws Exception {
-		if (type == null || type.isObject())
```
bjhargrave: Discusses with Peter. We do need to add Object and void as return types to detect changes to the return types from void to non-void and from Object to another type. Without added Object and void as return types, a change between void and Object is unnoticed and a change from void or Object to some other return type is only seen as an add only. The remove of void or Object is unnoticed.

stillalex:
hi, @bjhargrave are you sure this change is a good idea, from my (very limited) understanding of the code it seems you are effectively removing checks for generic types (reverting #477), which is pretty unfortunate.

> Baselining should be about binary compatibility and the only the erased
signature matters.

has this been discussed anywhere? I'd love to understand the reasoning behind this decision.

bjhargrave:
> are you sure this change is a good idea

I am quite sure. As you quoted, baselining is about binary compatibility and generics do not play a part in binary compatibility. The reasoning is very straight forward in that generics are "erased" as the source code is compiled to byte code. While the generics signatures are retained in the class files and are useful in compilation and reflection, they are not used by the JVM when selecting and invoking methods at runtime.

You are interested in source compatibility which is beyond the scope of baselining and OSGi's semantic versioning of packages.

Issue #1919, opened 2017-02-27
launcher: does not run under compact1 profile


fhuberts:
The bnd launcher uses the compact3 profile for its agent (see below).

I'd really like the bnd launcher to run under the compact1 profile.

Would it be possible that the agent part is split off, be an extension/plugin, or that there be different launchers - one with and one without the agent?

```
$ jdeps -P jar/biz.aQute.launcher-3.3.0.jar |grep -vE '(compact1|biz.aQute.launcher-3.3.0.jar|not found)$'
biz.aQute.launcher-3.3.0.jar -> /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.121-1.b14.fc25.x86_64/jre/lib/rt.jar (compact3)
   aQute.launcher (biz.aQute.launcher-3.3.0.jar)
      -> java.lang.instrument                               compact3
   aQute.launcher.agent (biz.aQute.launcher-3.3.0.jar)
      -> java.lang.instrument                               compact3
   aQute.launcher.constants (biz.aQute.launcher-3.3.0.jar)
   aQute.launcher.minifw (biz.aQute.launcher-3.3.0.jar)
   aQute.launcher.plugin (biz.aQute.launcher-3.3.0.jar)
   aQute.launcher.pre (biz.aQute.launcher-3.3.0.jar)
   aQute.lib.collections (biz.aQute.launcher-3.3.0.jar)
   aQute.lib.hex (biz.aQute.launcher-3.3.0.jar)
   aQute.lib.io (biz.aQute.launcher-3.3.0.jar)
   aQute.lib.strings (biz.aQute.launcher-3.3.0.jar)
   aQute.lib.utf8properties (biz.aQute.launcher-3.3.0.jar)
   aQute.libg.cryptography (biz.aQute.launcher-3.3.0.jar)
   aQute.libg.glob (biz.aQute.launcher-3.3.0.jar)
   aQute.libg.tuple (biz.aQute.launcher-3.3.0.jar)
```

bjhargrave:
I think perhaps we just wrap the references to the instrumentation code in `aQute.launcher.Launcher` in a `try/catch(NoClassDefFoundError)` block so that of the Instrumentation class is not present, we just ignore the `LauncherAgent`.

fhuberts:
If so, then make it very clear somewhere that this is done this way, because jdeps will then still hapily indicate compact3

bjhargrave:
[...]

PR #1921, opened 2017-02-27
launcher: Run on compact1 without java.lang.instrument

bjhargrave:
Fixes https://github.com/bndtools/bnd/issues/1919


bjhargrave:
Waiting on confirmation of fix from @fhuberts before merging.

Will also need to document this somewhere too.

fhuberts:
Seems to work for us

PR #1927, opened 2017-03-02
[baseline] make sure to include manifest in jar Analyzer

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/differ/DiffPluginImpl.java
```diff
@@ -100,9 +100,8 @@ public Tree tree(File newer) throws Exception {
 	 * @see aQute.bnd.service.diff.Differ#tree(aQute.bnd.osgi.Jar)
 	 */
 	public Tree tree(Jar newer) throws Exception {
-		Analyzer anewer = new Analyzer();
+		Analyzer anewer = new Analyzer(newer);
 		try {
-			anewer.setJar(newer);
 			return tree(anewer);
 		} finally {
 			anewer.setJar((Jar) null);
```
bjhargrave: You can remove the `setJar(null)`.

Might also want to tidy this up into a try-with-resources block to avoid the finally with a close call.

```java
try (Analyzer anewer = new Analyzer(newer)) {
  return tree(anewer);
}
```

bjhargrave:
Also, I'd like to test this PR on the OSGi build before we merge. The OSGi build is using baselining now.

biz.aQute.bndlib/src/aQute/bnd/differ/DiffPluginImpl.java
rotty3000: I think this was there because the jar is still used later on in processing. If we leave the jar in the analyzer, might it not be closed also?

bjhargrave: When the analyzer is closed, the Jar will be closed because setJar calls addClose.

bjhargrave: But I suppose the set to null is to avoid closing the Jar when closing the analyzer. You are right. This method should not close the jar.

bjhargrave:
I tried this PR on the OSGi build and it seems to work fine. No unexpected baseline issues.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -124,6 +125,17 @@
...
 	}
 
+	public Analyzer(Jar jar) {
+		setJar(jar);
+		try {
+			Manifest manifest = jar.getManifest();
+			if (manifest != null)
+				copyFrom(Domain.domain(manifest));
+		} catch (Exception e) {
+			throw Exceptions.duck(e);
```
pkriens: The common style is to throw Exception ...

Issue #1929, opened 2017-03-03
Stream issues marked as Security Vulnerability not properly closed


bsulich2:
Found some issues in the code, where stream is open, but not properly closed.

1. [Workspace.java](https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java#L688)
2. [Jar.java](https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java#L93)
3. [Repository.java](https://github.com/bndtools/bnd/blob/master/biz.aQute.repository/src/aQute/bnd/jpm/Repository.java#L807)
4. [HttpBasicAuthURLConnector.java](https://github.com/bndtools/bnd/blob/master/biz.aQute.repository/src/aQute/bnd/deployer/http/HttpBasicAuthURLConnector.java#L183)
5. [SpringComponent.java](https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/lib/spring/SpringComponent.java#L47)
6. [Builder.java](https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java#L117)
7. [Project.java](https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L2869)
8. [OSGI_CORE.java](https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/build/model/OSGI_CORE.java#L15)

List above point to lines in classes where I found problems.
Most of them I fixed in branch (is inPullRequests). 
According to classes listed above:
 - 1, 2, 3, 5, 6, 7, 8 are fixed within the same class

 - 4 is fixed in [CachingUriResourceHandle.java](https://github.com/bndtools/bnd/blob/master/biz.aQute.repository/src/aQute/bnd/deployer/repository/CachingUriResourceHandle.java#L198)

bjhargrave:
> 1- Workspace.java:688 - didn't fix. Could you tell me where is this stream closed? Is this class used in runtime, or only by the bnd plugin for maven/gradle?

This appears to just need a similar fix as https://github.com/bndtools/bnd/pull/1930/files#diff-d42bdfe7f7601ee1423a8e5e7bb6c022

PR #1930, opened 2017-03-03
Streams are not properly closed

bsulich2:
Refactoring classes to handle autoclosable inside try with resources block.

Fixes https://github.com/bndtools/bnd/issues/1929

Signed-off-by: Bartlomiej Sulich <bartlomiej.sulich@intel.com>

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
@@ -113,10 +113,8 @@ public class Builder extends Analyzer {
 			File mff = getFile(mf);
 			if (mff.isFile()) {
 				updateModified(mff.lastModified(), "Manifest " + mff);
-				try {
-					InputStream in = new FileInputStream(mff);
+				try (InputStream in = new FileInputStream(mff)){
 					manifest = new Manifest(in);
-					in.close();
 				} catch (Exception e) {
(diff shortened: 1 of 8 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
```diff
@@ -113,8 +113,7 @@ public Jar build() throws Exception {
 			File mff = getFile(mf);
 			if (mff.isFile()) {
 				updateModified(mff.lastModified(), "Manifest " + mff);
-				try {
-					InputStream in = new FileInputStream(mff);
+				try (InputStream in = new FileInputStream(mff)){
 					manifest = new Manifest(in);
 					in.close();
```
bjhargrave: There is already a close here. If you want to use try-with-resource, then remove the explicit close call.

bjhargrave:
You should also fix the commit comment to say 'Streams' and not 'Stream' :-)

bsulich2:
- fixed comment. 
- added try with resources to [Workspace.java](https://github.com/bsulich2/bnd/blob/1929_refactor_streams_close/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java#L688)
- removed close() from try with resources block in [Builder.java](https://github.com/bsulich2/bnd/blob/1929_refactor_streams_close/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java#L117)

PR #1942, opened 2017-03-12
git ignore  intellij module and  project files (*.iml, .idea , and *.ipr)

sesuncedu:
Add intellij project files to .gitignore file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ buildfiles
 /.gradle/
 /reports/
 /generated/
+*.iml
+.idea
+*.ipr


.gitignore
```diff
@@ -8,3 +8,6 @@ buildfiles
 /.gradle/
 /reports/
 /generated/
+*.iml
+.idea
```
bjhargrave: This should be `/.idea/`.

.gitignore
```diff
@@ -8,3 +8,6 @@ buildfiles
 /.gradle/
 /reports/
 /generated/
+*.iml
```
bjhargrave: Do we really need .iml and .ipr? They seem to be legacy metadata formats. https://intellij-support.jetbrains.com/hc/en-us/articles/206544839

bjhargrave:

.gitignore
lostiniceland: Intellij still creates *.iml files within modules in some cases. Just had this issue last week with IDEA 2016.3.5 but dont know where it came from.

sesuncedu:
I'd backburnered this issue because I noticed that iml files were under /.idea and figured it must be my imagination, but then found new ones appeared in other projects and had no, um, idea. (happens with 2017.1 as well). 
I had an unsupported hypothesis that there may be a connection with having the idea plugin enabled in gradle (which sometimes is needed with plugins like jmh), but that's not the only case, since it's happened with maven imported projects.

sesuncedu:
Also, will add an SOB either by ammend/force, or a fresh PR referencing this one.
 (I have never been convinced that SOB has any additional legal force in the US, but do think that a policy requiring them could create liability if not applied consistently. 
  I also do not believe that I would have any enforceable IP rights in this particular submission due to the mechanical nature of the change,  and consequent lack of creativity.) 
(I am not a lawyer)

bjhargrave:
Any progress on this?

bjhargrave:
Closing due to lack of progress. Please submit a new PR if you can address the requested changes. Thanks.

Issue #1952, opened 2017-03-20
Executable JAR working directory


pkriens:
The executable JAR working directory is ~/bnd/<name>. This cannot be overridden.

The runtime should have an option to set the framework directory

pkriens:
We probably should allow org.osgi.framework.storage tobe set with -D from the command line. This should override any bnd options

bjhargrave:
Fixed by faa3825

PR #1955, opened 2017-03-20
[launcher] #1952  Executable JAR working directory

pkriens:
Allow override of any of the launcher constants as 
well as allowing to override calculated org.osgi.framework.storage
property when set as system property. That is, the following
forces the framework dir regardless of other settings.

   java -Dorg.osgi.framework.storage=fw -jar foo.jar

### DIFF
--- a/biz.aQute.launcher/src/aQute/launcher/Launcher.java
+++ b/biz.aQute.launcher/src/aQute/launcher/Launcher.java
@@ -129,6 +129,9 @@ public class Launcher implements ServiceListener {
 
 				Properties properties = new Properties();
 				load(in, properties);
+
+				augmentWithSystemProperties(properties);
+
 				Launcher target = new Launcher(properties, propertiesFile);
 				exitcode = target.run(args);
 			} catch (Throwable t) {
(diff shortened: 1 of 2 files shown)


biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -824,7 +837,10 @@ public void run() {
 		if (!workingdir.isDirectory())
 			throw new IllegalArgumentException("Cannot create a working dir: " + workingdir);
 
-		p.setProperty(Constants.FRAMEWORK_STORAGE, workingdir.getAbsolutePath());
+		if (System.getProperty(Constants.FRAMEWORK_STORAGE) == null)
+			p.setProperty(Constants.FRAMEWORK_STORAGE, workingdir.getAbsolutePath());
+		else
+			p.setProperty(Constants.FRAMEWORK_STORAGE, System.getProperty(Constants.FRAMEWORK_STORAGE));
```
bjhargrave: Should this be done in LauncherConstants.getProperties so that method always returns a storageDir?

pkriens: No, I want it to only make a difference when people use -D in the launching for backward compatibility. And the storagedir gets processed. I think this is fine.

bjhargrave: So you want to process the storageDir set and then ignore that when -Dorg.osgi.framework.storage? What about the other  org.osgi.framework.* setting on the command line? Like clean?

pkriens: storage is differently and was treated wrongly in the generation of the executable JAR. The code sets the storagedir to null because the different file systems make it impossible to create an always correct path. The other values are not platform dependent. However, feel free to override the others as well.

Issue #1976, opened 2017-03-31
use aQute.bnd.repository.fileset.FileSetRepository in maven export, resolver and testing plugins


rotty3000:

rotty3000:
fixed by #1976

PR #1978, opened 2017-04-01
implements #1976 use aQute.bnd.repository.fileset.FileSetRepository in maven export, resolver and testing plugins

rotty3000:


rotty3000:
@bjhargrave bump!

maven/bnd-export-maven-plugin/README.md
```diff
@@ -7,24 +7,38 @@ The `bnd-export-maven-plugin` is a bnd based plugin to export bndrun files.
...
+```
+
+Here's an example setting the `bundles` used for resolution.
+
+```
+    ...
+    <configuration>
+        ...
+        <bundles>
+            <bundle>bundles/org.apache.felix.eventadmin-1.4.8.jar</bundle>
```
bjhargrave: What directory are the bundles relative to?

maven/bnd-export-maven-plugin/pom.xml
```diff
@@ -41,6 +41,10 @@
             <groupId>biz.aQute.bnd</groupId>
             <artifactId>biz.aQute.repository</artifactId>
         </dependency>
+        <dependency>
+            <groupId>biz.aQute.bnd</groupId>
+            <artifactId>bnd-maven-plugin-lib</artifactId>
```
bjhargrave: Perhaps we can call this `bnd-shared-maven-lib`? This seems to fit the existing naming better. After `bnd-` is the _purpose_ and at the end is the type. This is a _lib_ rather than a _plugin_.

Another idea (which may be stupid given my general maven lack-of-knowledge), would be to put the shared code in the parent project? Crazy?

maven/bnd-export-maven-plugin/src/test/resources/integration-test/test/export-from-dependencies/test.bndrun
```diff
@@ -1,4 +1,4 @@
--standalone: ${.}/index/index.xml.gz
+-standalone:
```
bjhargrave: Remove standalone all together. You don't need it unless you specify an index.

maven/bnd-export-maven-plugin/src/test/resources/integration-test/test/export-from-inputbundles/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>biz.aQute.bnd</groupId>
+                <artifactId>bnd-export-maven-plugin</artifactId>
+                <configuration>
+                    <bundlesOnly>true</bundlesOnly>
+                    <bundles>
+                        <bundle>bundles/org.apache.felix.eventadmin-1.4.8.jar</bundle>
```
bjhargrave: I think we should encourage full paths, no? 

`<bundle>${project.basedir}/bundles....`

maven/bnd-export-maven-plugin/src/test/resources/integration-test/test/export-from-inputbundles/test.bndrun
```diff
@@ -0,0 +1,4 @@
+-standalone:
```
bjhargrave: Don't need -standalone

maven/bnd-export-maven-plugin/src/test/resources/integration-test/test/export-bundles-only/pom.xml
```diff
@@ -20,8 +20,6 @@
                 <artifactId>bnd-export-maven-plugin</artifactId>
                 <configuration>
                     <bundlesOnly>true</bundlesOnly>
-                    <failOnChanges>false</failOnChanges>
-                    <resolve>true</resolve>
```
bjhargrave: Why remove these from the test?

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
```diff
@@ -9,27 +9,46 @@
...
+	private RepositorySystemSession		repositorySession;
+
 	@Parameter(readonly = true, required = true)
 	private List<File>	bndruns;
 
 	@Parameter(defaultValue = "${project.build.directory}", readonly = true)
 	private File		targetDir;
 
+	@Parameter(readonly = true, required = false)
+	private List<File>					bundles;
```
bjhargrave: I guess with this configuration setup, all bndruns must share the same FileSetRepo?

maven/bnd-maven-plugin-lib/.project
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>bnd-maven-plugin-lib</name>
```
bjhargrave: Lets change the name to `bnd-shared-maven-lib` unless we put all this code in parent!

maven/bnd-maven-plugin-lib/src/main/java/aQute/bnd/maven/lib/resolve/DependencyResolver.java
```diff
@@ -0,0 +1,170 @@
...
+			String name, Collection<File> bundlesInputParameter, boolean useDefaults)
+		throws Exception {
+
+		Collection<File> bundles = new ArrayList<>();
+		if (useDefaults) {
+			Map<File,ArtifactResult> dependencies = resolve();
+			bundles.addAll(dependencies.keySet());
+
+			// TODO Find a better way to get all the artifacts produced by this project!!!
+			File current = new File(project.getBuild().getDirectory(), project.getBuild().getFinalName() + ".jar");
```
bjhargrave: This probably good enough.

maven/bnd-resolver-maven-plugin/pom.xml
```diff
@@ -84,7 +88,7 @@
                       <jsse.enableSNIExtension>false</jsse.enableSNIExtension>
                     </properties>
                     <goals>
-                        <goal>bnd-resolver:resolve</goal>
+                        <goal>package</goal>
```
bjhargrave: This is a big change, no? And the README still mentions the resolve goal.

maven/bnd-resolver-maven-plugin/src/test/resources/integration-test/test/pom.xml
```diff
@@ -2,30 +2,39 @@
     xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <groupId>biz.aQute.bnd-test</groupId>
-    <artifactId>export-test</artifactId>
+    <artifactId>resolver-test</artifactId>
```
bjhargrave: Nice catch!

maven/bnd-resolver-maven-plugin/src/test/resources/integration-test/test/resolve-from-dependencies/test.bndrun
```diff
@@ -0,0 +1,4 @@
+-standalone:
```
bjhargrave: Dont need -standalone

maven/bnd-resolver-maven-plugin/src/test/resources/integration-test/test/resolve-from-inputbundles/test.bndrun
```diff
@@ -0,0 +1,4 @@
+-standalone:
```
bjhargrave: Dont' need -standalone

maven/bnd-testing-maven-plugin/src/test/resources/integration-test/test/test-with-resolve-from-dependencies/test.bndrun
```diff
@@ -0,0 +1,4 @@
+-standalone:
```
bjhargrave: dont need -standalone

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
bjhargrave: Actually, I think we need to change how we configure these plugins. Instead of allowing multiple bndruns, we allow only a single bndrun and you use once execution per bndrun if you need more. Then you can specify different bundles for each bndrun.

Old way:
```
            <plugin>
                <artifactId>bnd-export-maven-plugin</artifactId>
                <version>${bnd.version}</version>
                <configuration>
                    <failOnChanges>false</failOnChanges>
                    <bndruns>
                        <bndrun>mylaunch.bndrun</bndrun>
                        <bndrun>otherlaunch.bndrun</bndrun>
                    </bndruns>
                    <bundles>
                       <bundle>xxxxx</bundle>
                       <bundle>yyyy</bundle>
                       <bundle>zzz</bundle>
                    </bundles>
                </configuration>
                <executions>
                    <execution>
                        <goals>
                            <goal>export</goal>
                        </goals>
                    </execution>
                </executions>
            </plugin>
```

New way:

```
            <plugin>
                <artifactId>bnd-export-maven-plugin</artifactId>
                <version>${bnd.version}</version>
                <configuration>
                    <failOnChanges>false</failOnChanges>
                </configuration>
                <executions>
                    <execution>
                       <id>mylaunch</id>
                       <goals>
                            <goal>export</goal>
                       </goals>
                       <configuration>
                         <bndrun>mylaunch.bndrun</bndrun>
                         <bundles>
                            <bundle>xxxxx</bundle>
                            <bundle>yyyy</bundle>
                         </bundles>
                      </configuration>
                    </execution>
                    <execution>
                       <id>otherlaunch</id>
                       <goals>
                            <goal>export</goal>
                       </goals>
                       <configuration>
                         <bndrun>otherlaunch.bndrun</bndrun>
                         <bundles>
                            <bundle>zzz</bundle>
                            <bundle>yyyy</bundle>
                         </bundles>
                      </configuration>
                    </execution>
                </executions>
            </plugin>
```

I think this is cleaner for normal one bndrun configuration (Don't need <bndruns>) and allow better control for <bundles> for each bndrun when you have multiple bndruns.

maven/bnd-export-maven-plugin/README.md
rotty3000: should I just add this to the documentation? The paths are relative to the project dir as it is. This is handled by maven internals as it is.

maven/bnd-export-maven-plugin/pom.xml
rotty3000: The parent is just a *reactor*. I think it would make it painful to put it in the parent.

maven/bnd-export-maven-plugin/src/test/resources/integration-test/test/export-from-dependencies/test.bndrun
rotty3000: I guess we're explicitly creating standalone bndruns so we don't rely on the `-standalone` directive to actually exist to enforce that so ... ok.

maven/bnd-export-maven-plugin/src/test/resources/integration-test/test/export-from-inputbundles/pom.xml
rotty3000: The relativity to project basedir is already implied for file collections. Would documenting that be enough?

maven/bnd-export-maven-plugin/src/test/resources/integration-test/test/export-bundles-only/pom.xml
rotty3000: because those are already set in the shared config.

maven/bnd-resolver-maven-plugin/pom.xml
rotty3000: This is only so that the test can trigger the target which is not bound by default.

maven/bnd-export-maven-plugin/README.md
bjhargrave: Yeah, the docs should be clear on relative paths.

maven/bnd-export-maven-plugin/pom.xml
bjhargrave: OK, then just rename.

maven/bnd-export-maven-plugin/src/test/resources/integration-test/test/export-from-dependencies/test.bndrun
bjhargrave: Yes. For the maven plugins, all the bndrun files are standalone by definition. There is no workspace they can be a member of. The same is true for the non-workspace gradle task types.

maven/bnd-export-maven-plugin/src/test/resources/integration-test/test/export-from-inputbundles/pom.xml
bjhargrave: Yes. But for testing purposes, we should use some absolute paths too.

maven/bnd-export-maven-plugin/src/test/resources/integration-test/test/export-bundles-only/pom.xml
bjhargrave: Ah!

maven/bnd-resolver-maven-plugin/pom.xml
bjhargrave: OK, but shouldn't the invoker invoke the `bnd-resolver:resolve` target? That is, should the test case look normal and the test runner run the test in the expected way?

rotty3000: No because the invoker is invoking a reactor... not indivdual projects of which there are more than one.

rotty3000: i.e. the top level reactor pom doesn't have the plugin configuration for the resolver. Only the children do.

rotty3000:
Updated!

njbartlett:
What does the new <bundle> configuration setting do?

bjhargrave:
> What does the new `<bundle>` configuration setting do?

It lets you specify specific bundles to include in or replace the bundles uses for the FileSetRepository which is added to the standalone workspace. The bndrun files can then get bundles from that repo for the `-run*` instructions.

rotty3000:
Also, the properties are described in the README.md of each plugin.

njbartlett:
Thanks BJ and Ray. At first glance this looked terrifying because I don't want to list all the available bundles for resolution... that's what the index is for! But if this is just supplementary to the index contents, that's fine I guess, though I don't really see much need for it.

bjhargrave:
Well you don't even need an index now. It will automatically make all your compile+runtime depenedencies available to the bndrun file via the FileSetRepository. So normally you just add what ever you need to your pom as a compile or runtime dependency, and your bndrun file can use it. The `<bundles>` is just for some special cases if you have them.

njbartlett:
So what does a bndrun look like now? Just a `-standalone:` line? In which case, how would you resolve that in Bndtools?

bjhargrave:
`-standalone` is only required (1) when there is a workspace to disavow it or (2) when you need to specify an index.

The standalone workspace has a FileSetRepository added to it which includes the indexed compile+runtime depenencies of the pom (if `useDefault` is true (which is the default) plus any specified `<bundles>`.

You can also specify `-standalone` to supply further repos using the specified indexes.

Issue #1985, opened 2017-04-06
Bnd OSGi Test Launcher (JUnit) dont support OSGi version=4.2


jwausle:
As developer I want to use bnd/bndtools (>=3.4) to build and test my bundles for an OSGi container version=4.2.

Currrently is version=42. not supported. 
- biz.aQute.tester bundle is always INSTALLED

----
Java version 1.7, 1.8
Bnd(tools) version 3.3.0
Os mac and windows

PR #1986, opened 2017-04-06
#1985 Support OSGi version = 4.2 for Bnd OSGi Test Launcher.

jwausle:


jwausle:
The purpose is testing of bundles in OSGi container version 4.2 with bndtools. Our target hardware is currently restricted to Java 1.7. 

I know the generic case. But it works for me. I use only the testing bundles to find an schedule the cases.

bjhargrave:
I have made an alternate PR to properly address #1985. See PR #1989.

PR #1987, opened 2017-04-06
add new api to agent supervisor to allow subclasses to set a connect timeout

gamerson:


biz.aQute.remote/src/aQute/remote/util/AgentSupervisor.java
```diff
@@ -41,15 +42,23 @@
 	}
 
 	protected void connect(Class<Agent> agent, Supervisor supervisor, String host, int port) throws Exception {
-		while (true)
+		connect(agent, supervisor, host, port, null);
+	}
+
+	protected void connect(Class<Agent> agent, Supervisor supervisor, String host, int port, Integer timeout)
```
bjhargrave: Using Integer, to get null as a possible input, means a lot of boxing/unboxing for comparisons and `-=` below. Can't we use -1 to mean no timeout?

gamerson: Yes i thought about -1 for no timeout, but then it would require more calculations for decrementing the integer for when it goes below -1.  But that shouldn't require too much more code.  I can send that change in.

gamerson:
@bjhargrave thoughts on latest push?

biz.aQute.remote/src/aQute/remote/util/AgentSupervisor.java
```diff
@@ -41,15 +42,26 @@
...
 			try {
 				Socket socket = new Socket(host, port);
 				link = new Link<Supervisor,Agent>(agent, supervisor, socket);
 				this.setAgent(link);
 				link.open();
 				return;
 			} catch (ConnectException e) {
-				Thread.sleep(200);
+				if (timeout > 0) {
+					timeout -= connectWait;
```
bjhargrave: I think you changes this to `timeout = Math.max(timeout - connectWait, 0);` That way `0` is the floor. The you can remove the `timeout == -1` check below.

biz.aQute.remote/src/aQute/remote/util/AgentSupervisor.java
```diff
@@ -41,15 +42,26 @@
 	}
 
 	protected void connect(Class<Agent> agent, Supervisor supervisor, String host, int port) throws Exception {
-		while (true)
+		connect(agent, supervisor, host, port, -1);
+	}
+
+	protected void connect(Class<Agent> agent, Supervisor supervisor, String host, int port, int timeout)
+			throws Exception {
+		while (timeout == -1 || timeout > 0)
```
bjhargrave: I think you can leave this as a `while (true)` loop. If successful, it always returns from the try block. The same should be done from the catch block: If the time out is expired, then return. Otherwise loop back up. This avoids the awkward test in the while statement.

biz.aQute.remote/src/aQute/remote/util/AgentSupervisor.java
gamerson: @bjhargrave  you made an A in geometry class didn't you!

biz.aQute.remote/src/aQute/remote/util/AgentSupervisor.java
```diff
@@ -41,16 +42,26 @@
...
 				link = new Link<Supervisor,Agent>(agent, supervisor, socket);
 				this.setAgent(link);
 				link.open();
 				return;
 			} catch (ConnectException e) {
-				Thread.sleep(200);
+				if (timeout == 0) {
+					return;
+				}
+				timeout = Math.max(timeout - connectWait, 0);
```
bjhargrave: Need to handle `timeout == -1`. We don't want to convert it to 0 and then return on the next iteration.
```java
if (timeout > 0) {
   timeout = Math.max(timeout - connectWait, 0);
}
```

gamerson: Done.

biz.aQute.remote/src/aQute/remote/util/AgentSupervisor.java
```diff
@@ -41,16 +42,28 @@
 	}
 
 	protected void connect(Class<Agent> agent, Supervisor supervisor, String host, int port) throws Exception {
-		while (true)
+		connect(agent, supervisor, host, port, -1);
+	}
+
+	protected void connect(Class<Agent> agent, Supervisor supervisor, String host, int port, int timeout)
```
gamerson: @bjhargrave if a client calls `connect(agent, supervisor, host, port, -100)` it will hang indefinitely.  Is this OK?

bjhargrave: Any negative number is the same as -1. It is an indefinite wait. The alternative is to check timeout value on entrance to the method and throw an exception if timeout < -1.

Looking at `Object.wait(long)` it uses 0 for indefinite wait. Negative value is an error.

pkriens:
Ahum ... I do not want a timeout in bndtools since I can cancel it with a button. I am ok with a timeout but the bndtools must continue to try to connect. I often find out the Raspberry Pi is not running and then it is convenient that bndtools keeps trying until I've set it up.

biz.aQute.remote/src/aQute/remote/util/AgentSupervisor.java
gamerson: Ok we can leave it as is.

gamerson:
hey @pkriens no plans to change bndtools.  This is for a Liferay application that would like a timeout.

PR #1990, opened 2017-04-08
gradle: Add bnd property to Builder task type

### LINKED ISSUE #1966: Allow to pass merged instructions to the Bnd Builder Gradle Plugin
It's currently impossible to pass a merged instruction like this:
```gradle
jar {
	manifest {
		attributes('-includeresource.foo': 'foo.txt')
	}
}
```

Gradle complains that the attribute name is invalid:
```
The Key=-includeresource.foo violates the Manifest spec!
```

bjhargrave:
This allows for inline specification of the bnd file in the build.gradle file.

Fixes #1966


bjhargrave:
@rotty3000, I think this will solve the need to specify the bnd file "inline" in the build.gradle file instead of in a separate file.

biz.aQute.bnd.gradle/README.md
```diff
@@ -252,15 +253,23 @@ task bundle(type: Bundle) {
 }
 ```
 
-In either usage mode, there are three properties, in addition to the
+In either usage mode, there are four properties, in addition to the
 `Jar` task properties, which can be configured.
 
 ### bndfile
 
 This is the File object to use for the bnd file to use to create the
```
rotty3000: > to use ... to use ...

 sounds redundant

biz.aQute.bnd.gradle/README.md
```diff
@@ -252,15 +253,23 @@ task bundle(type: Bundle) {
...
-bundle. The default value is `project.file('bnd.bnd')`. If the bnd file
-does not exist, this is OK. But without some instructions to Bnd, your
-bundle will not be very interesting.
+bundle. If this property is not set or the bnd file does not exist,
+then the `bnd` property will be used.
+
+### bnd
+
+This is a string to use for the bnd file to use to create the
```
rotty3000: > to use ... to use ...

PR #2026, opened 2017-05-02
implement shell function in Agent

AndyWu2015:


AndyWu2015:
Hey @gamerson  , we can use this for executing "lb -s some-bsn" instead of getting all bundles.

pkriens:
LGTM

biz.aQute.remote/src/aQute/remote/agent/RedirectOutput.java
```diff
@@ -132,4 +135,17 @@ public PrintStream getOut() {
...
 
+	public String getLastOutput() {
+		return lastOutput;
+	}
+
+	public void setLastOutput(String out) {
+		if (!"".equals(out) && out != null) {
+			out = out.replaceAll("^>.*$", "");
+
+			if (!"".equals(out) && !out.startsWith("true")) {
```
gamerson: if the shell command returns 'true' , do we not want to just let "true" be returned?

also the "^>.*$" assumes that the gogo prompt will be standard, but what if its modified?  with

```
g! prompt= '$ '
$   
```

AndyWu2015: yes , but I think we can only assume that user will not change the prompt string when they calling Agent.shell("string") , also it makes no sense that user change the prompt string cause they just want to get gogo shell result and they are not using command line and don't need to see the prompt in UI.

gamerson: I think that we can go ahead and merge this and if we need improved implementation in the future, we can always improve it.

gamerson:
hey @AndyWu2015  can you test this new commit I pushed to your branch?  if it works ok I will merge

AndyWu2015:
Sure , I tested it on my side , it works well.

Issue #2027, opened 2017-05-03
P2Indexer reloads and reindexes artefacts unnecessarily


eliasvasylenko:
There are two components to this problem.

Currently the indexer skips over artefacts URIs which are listed multiple times. This fails to account for instances where the same artefact can be found at multiple URIs, such as with composite repositories.

The solution is to skip based on encountered *id/version pairs* instead of matching URIs. This should be safe, p2 assumes each id/version to be unique within a repository by design and composites have [build checks](https://wiki.eclipse.org/Equinox/p2/Composite_Repositories_(new)#Attributes) to make sure it holds.

The second problem is that even if we already have capability/requirement information in our index, when we refresh the indexer throws this information away and starts from scratch. It would be a lot quicker to only load artefacts which are new to us.

bjhargrave:
> Currently the indexer skips over artefacts URIs which are listed multiple times. This fails to account for instances where the same artefact can be found at multiple URIs, such as with composite repositories.

This seems to be confused. The same URI always points to the same artifact. While the same artifact can be found at multiple URI, each of those URIs is processed. So I do not see a problem here. Processing the same URI multiple times is a waste. But processing each unique URI is necessary since we cannot know in advance they point to the same artifact.

> 
> The solution is to skip based on encountered id/version pairs instead of matching URIs. This should be safe, p2 assumes each id/version to be unique within a repository by design and composites have build checks to make sure it holds.

Based upon my comment above, I don't think this is correct. 

> 
> The second problem is that even if we already have capability/requirement information in our index, when we refresh the indexer throws this information away and starts from scratch. It would be a lot quicker to only load artefacts which are new to us.

Yes, it would be faster as the expense of correctness. There is no guarantee that the artifact for the bsn+version is unchanged and thus you can avoid processing it. For example, the Bndtools build makes a P2 update site each build. So that site has artifacts with the same bsn+version across builds but the artifacts potentially change each build. You would need something stronger like a hash to know the artifact is unchanged and thus can avoid processing again.

eliasvasylenko:
> But processing each unique URI is necessary since we cannot know in advance they point to the same artifact.

[...]

PR #2028, opened 2017-05-03
#2027 p2 indexing fix

eliasvasylenko:
Fix for #2027, this knocks huge amounts of time off certain builds. Tens of minutes easily on an old HDD.

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/ArtifactID.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/ArtifactID.java
@@ -1,0 +1,51 @@
+package aQute.bnd.repository.p2.provider;
+
+class ArtifactID {
+	private final String	id;
+	private final String	version;
+	private final String	md5;
+
+	public ArtifactID(String id, String version, String md5) {
+		super();
+		this.id = id;
+		this.version = version;
+		this.md5 = md5;
+	}
+
+	@Override
+	public int hashCode() {
+		final int prime = 31;
+		int result = 1;
+		result = prime * result + ((id == null) ? 0 : id.hashCode());
+		result = prime * result + ((md5 == null) ? 0 : md5.hashCode());
+		result = prime * result + ((version == null) ? 0 : version.hashCode());
+		return result;
+	}
+
+	@Override
+	public boolean equals(Object obj) {
+		if (this == obj)
+			return true;
+		if (obj == null)
+			return false;
+		if (getClass() != obj.getClass())
+			return false;
+		ArtifactID other = (ArtifactID) obj;
+		if (id == null) {
+			if (other.id != null)
+				return false;
+		} else if (!id.equals(other.id))
+			return false;
+		if (md5 == null) {
+			if (other.md5 != null)
+				return false;
+		} else if (!md5.equals(other.md5))
+			return false;
+		if (version == null) {
+			if (other.version != null)
+				return false;
+		} else if (!version.equals(other.version))
+			return false;
+		return true;
+	}
+}
--- a/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/ArtifactID.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/ArtifactID.java
@@ -1,11 +1,13 @@
 package aQute.bnd.repository.p2.provider;
 
+import aQute.bnd.version.Version;
+
 class ArtifactID {
 	private final String	id;
-	private final String	version;
+	private final Version	version;
 	private final String	md5;
 
-	public ArtifactID(String id, String version, String md5) {
+	public ArtifactID(String id, Version version, String md5) {
 		super();
 		this.id = id;
 		this.version = version;
--- a/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
@@ -1,11 +1,15 @@
 package aQute.bnd.repository.p2.provider;
 
+import static aQute.bnd.osgi.resource.ResourceUtils.toVersion;
+import static java.util.Collections.singleton;
+
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -29,9 +33,13 @@ import aQute.bnd.osgi.repository.BridgeRepository;
 import aQute.bnd.osgi.repository.ResourcesRepository;
 import aQute.bnd.osgi.repository.XMLResourceGenerator;
 import aQute.bnd.osgi.repository.XMLResourceParser;
+import aQute.bnd.osgi.resource.CapabilityBuilder;
+import aQute.bnd.osgi.resource.FilterBuilder;
+import aQute.bnd.osgi.resource.RequirementBuilder;
 import aQute.bnd.osgi.resource.ResourceBuilder;
 import aQute.bnd.osgi.resource.ResourceUtils;
 import aQute.bnd.osgi.resource.ResourceUtils.ContentCapability;
+import aQute.bnd.osgi.resource.ResourceUtils.IdentityCapability;
 import aQute.bnd.service.RepositoryPlugin.DownloadListener;
 import aQute.bnd.util.repository.DownloadListenerPromise;
 import aQute.bnd.version.Version;
@@ -51,6 +59,11 @@ class P2Indexer implements Closeable {
 	private final File						indexFile;
 	private volatile BridgeRepository		bridge;
 	private static final Resource		RECOVERY	= new ResourceBuilder().build();
+	private static final String			P2_CAPABILITY_NAMESPACE	= "bnd.p2";
+	private static final String			MD5_ATTRIBUTE			= "md5";
+	private static final Requirement	MD5_REQUIREMENT			= new RequirementBuilder(P2_CAPABILITY_NAMESPACE)
+			.addFilter(new FilterBuilder().isPresent(MD5_ATTRIBUTE))
+			.buildSyntheticRequirement();
 
 	P2Indexer(Reporter reporter, File location, HttpClient client, URI url, String name) throws Exception {
 		this.reporter = reporter;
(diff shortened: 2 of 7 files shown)


bjhargrave:

eliasvasylenko:
Oops yeah forgot about that

eliasvasylenko:
Sorry I forgot about that, done.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -130,12 +132,29 @@ private Repository readRepository() throws Exception {
 		P2Impl p2 = new P2Impl(client, this.url, Processor.getExecutor());
 		List<Artifact> artifacts = p2.getArtifacts();
 		List<Promise<Resource>> fetched = new ArrayList<>(artifacts.size());
-		Set<URI> visited = new HashSet<>(artifacts.size());
+		Set<BundleId> visited = new HashSet<>(artifacts.size());
+
+		final Map<BundleId,Resource> existing = new HashMap<>();
+		if (getBridge() != null) {
```
bjhargrave: You have a circularity problem here. This method is called to create the Repository which the BridgeRepository wraps. So you cannot use the wrapper to create the thing it wraps. So when called from the constructor, there is no BridgeRepository yet (obviously). But when called from `refresh` you are using the outgoing BridgeRepository's to create the data for the incoming BridgeRepository. That hardly seems like a refresh operation. 

So this whole thing seems horribly broken. It is a no-op (`getBridge()` returns `null`) when constructing a P2Indexer object. And when refreshing the object, it uses stale date from the outgoing BridgeRepository. So, yes, the refresh is a whole lot faster at the expense of being potentially wrong since you don't actually refresh. It is not sufficiently safe to assume that the artifact is unchanged even though it has the same bsn+version.

eliasvasylenko: It's a refresh in that it removes artefacts which are nolonger in the repo and adds artefacts which have been added to the repo since it was last indexed. That's not a no-op.

If a refresh has to build the index from scratch then there's no point in even keeping the index file in the cache for Eclipse users. There is no point in a cache that is discarded as soon as you start the IDE.

There's also no point in "refreshing" a resource if we can demonstrate it's not changed. In this case the refresh *should be* a no-op and just validate the existing contents. If we suspect the index has become corrupted or that the previous indexing failed for some reason then yes we need a hard reset. But I do not accept that this needs to be done every time I open the IDE just in case. Just delete the cache if this happens.

As for whether that validation is currently sufficient ... I happily accept that it's not, but I think "horribly broken" is a bit exaggerated. With hashes added as discussed in the issue comments I see no further problems with this approach.

eliasvasylenko: It's a refresh in that it removes artefacts which are nolonger in the repo and adds artefacts which have been added to the repo since it was last indexed. It's only a no-op for artefacts which it validates as unchanged.

As for whether that validation is currently sufficient ... I happily accept that it's not, but I think "horribly broken" is a bit exaggerated. With hashes considered as discussed in the issue comments I think this approach is fine.

If a refresh has to build the index from scratch then there's no point in even keeping the index file in the cache for Eclipse users. There is no point in a cache that is discarded as soon as you start the IDE.

There's also no point in "refreshing" a resource if we can validate that it's unchanged. In this case the refresh *should be* a no-op. If we suspect the index has become corrupted or that the previous indexing failed for some reason then yes we need a hard reset. But I do not accept that this needs to be done every time I open the IDE just in case. Just delete the cache if the cache is broken.

bjhargrave: To validate it is unchanged, you need to compare the new hash to the old hash. If they are the same, then you can reuse the Resource object. Otherwise you need to process the artifact again.

eliasvasylenko: Yes I agree, I have been looking at adding md5 support and it seems feasible with (I hope) fairly minimal additional complexity. It would be so much easier if p2 already supported sha-256! But I think this is probably a fair ways off.

eliasvasylenko:
I think it should be okay now, please let me know if there are further problems.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -131,10 +138,29 @@ private Repository readRepository() throws Exception {
...
+		Map<String,Resource> existing = new HashMap<>();
+
+		if (getBridge() != null) {
+			Requirement requirement = new RequirementBuilder(P2_CAPABILITY_NAMESPACE).addAttribute(MD5_ATTRIBUTE, "*")
+					.buildSyntheticRequirement();
+			for (Capability capability : getBridge().getRepository().findProviders(asList(requirement)).get(
+					requirement)) {
+				Resource resource = capability.getResource();
+				String md5 = (String) capability.getAttributes().get(MD5_ATTRIBUTE);
+				existing.put(md5, resource);
```
bjhargrave: I am concerned that MD5 alone is not strong enough be the sole key. The `existing` map here is meant to be a map from some unique bundle key to the unique bundle resource. Since MD5 is known to be weak, perhaps the key here should be composed of the bsn+version+md5. The key is the bundle as uniquely identified by its bsn and version and also a hash to detect changes.

This comment also applies to the `visitedHash` set. I think the current `visited` set should be renamed `visitedURI` and the new set `visitedArtifact`. The `existing` name is also not descriptive enough. I think a better name is `knownResources`?

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -131,10 +138,29 @@ private Repository readRepository() throws Exception {
 		List<Artifact> artifacts = p2.getArtifacts();
 		List<Promise<Resource>> fetched = new ArrayList<>(artifacts.size());
 		Set<URI> visited = new HashSet<>(artifacts.size());
+		Set<String> visitedHash = new HashSet<>(artifacts.size());
+		Map<String,Resource> existing = new HashMap<>();
+
+		if (getBridge() != null) {
+			Requirement requirement = new RequirementBuilder(P2_CAPABILITY_NAMESPACE).addAttribute(MD5_ATTRIBUTE, "*")
```
bjhargrave: This requirement is static. It does not change upon each invocation. So why not move it to a static final field? In fact the static final field should be: 
```
private static final List<Requirement> hasMD5 = Collections.singletonList(new RequirementBuilder(P2_CAPABILITY_NAMESPACE).addAttribute(MD5_ATTRIBUTE, "*").buildSyntheticRequirement());
```

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -131,10 +138,29 @@ private Repository readRepository() throws Exception {
...
+				Resource resource = capability.getResource();
+				String md5 = (String) capability.getAttributes().get(MD5_ATTRIBUTE);
+				existing.put(md5, resource);
+			}
+		}
 
 		for (final Artifact a : artifacts) {
 			if (!visited.add(a.uri))
 				continue;
+			if (a.md5 != null && !visitedHash.add(a.md5))
```
bjhargrave: I would wrap all this inside the `if (a.md5 != null) { ... }` block.

biz.aQute.repository/src/aQute/p2/provider/ArtifactRepository.java
```diff
@@ -107,7 +107,8 @@ public Domain getParent() {
 
 		NodeList artifactNodes = getNodes("repository/artifacts/artifact");
 		for (int i = 0; i < artifactNodes.getLength(); i++) {
-			final XMLArtifact xmlArtifact = getFromType(artifactNodes.item(i), XMLArtifact.class);
+			final Node artifactNode = artifactNodes.item(i).cloneNode(true);
```
bjhargrave: Why do you have to clone the node?

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
eliasvasylenko: MD5 is known to be weak against deliberate collisions, but for accidental ones it is more than sufficient for this sort of application afaiu. I did consider a composite key like you suggested, I just wanted to minimise the change footprint ... but sure I suppose it makes sense to make it as safe as possible with the information we have since it doesn't really cost anything. Okay.

biz.aQute.repository/src/aQute/p2/provider/ArtifactRepository.java
eliasvasylenko: It's a variant on a common XPath optimisation. Resolving relative paths is extremely slow normally as it resolves against the entire document. Cloning divorces the resulting node from its parent, so the resolve operates only on the relevant section. I think that's the gist of it anyway, and it certainly worked.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
eliasvasylenko: Sure that makes sense, thanks.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
eliasvasylenko: Yes absolutely, good point.

eliasvasylenko:
Okay thanks for taking the time to go through this. I'll take another shot at it when I get home.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
eliasvasylenko: This is a little awkward since the addAttribute throws checked exceptions. Either everything has to be bogged down with unnecessary safety logic or we just rethrow as an assertionerror in the static initialiser block, since nothing should ever go wrong. I'll do the latter for now, please advise if this isn't okay.

eliasvasylenko:
Okay I've taken another crack at it. Not sure about that AssertionError or if it'd be better to initialise it as a non-static final at the use site or something. Should really be fine though. Cheers for bearing with me.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/ArtifactID.java
```diff
@@ -0,0 +1,51 @@
+package aQute.bnd.repository.p2.provider;
+
+class ArtifactID {
+	private final String	id;
+	private final String	version;
```
bjhargrave: Why not make `version` of type `Version`? Then we avoid `toString()` calls below.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -51,6 +57,17 @@
 	private final File						indexFile;
 	private volatile BridgeRepository		bridge;
 	private static final Resource		RECOVERY	= new ResourceBuilder().build();
+	private static final String			P2_CAPABILITY_NAMESPACE	= "bnd.p2";
+	private static final String			MD5_ATTRIBUTE			= "md5";
+	private static final Requirement	MD5_REQUIREMENT;
+	static {
+		try {
+			MD5_REQUIREMENT = new RequirementBuilder(P2_CAPABILITY_NAMESPACE).addAttribute(MD5_ATTRIBUTE, "*")
```
bjhargrave: Also, I'm curious why this is not:

```java
new RequirementBuilder(P2_CAPABILITY_NAMESPACE).addFilter("(" + MD5_ATTRIBUTE + "=*)").buildSyntheticRequirement();
```

I would assume you need a filter in the requirement to match since attributes on Requirements are generally ignored. Using `addFilter` also avoids the thrown exception so you don't need a try/catch any more.

See https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BaseRepository.java#L37-L40

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/ArtifactID.java
eliasvasylenko: Two different kinds of `Version`. Figured it was just a tad simpler than converting between `aQute.bnd.version.Version` and `org.osgi.framework.Version`.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
eliasvasylenko: Oh okay thanks, I didn't realise it would be ignored the way I did it. I just thought it felt a bit neater than formatting as a filter string, but yes that all makes sense I'll sort it out.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/ArtifactID.java
bjhargrave: @pkriens This seems a bit of a mess. Artifact uses OSGi Version while IdentityCapability uses bnd Version. Can we somehow avoid this and converge on one Version class in Bnd code?

eliasvasylenko:
Yet another attempt, with bonus single character commit/bug-fix :)

Issue #1864, opened 2017-01-27
Bnd gets confused by ill-formed bundle references


rherrmann:
I am using bnd through Bndtools 3.4.0.DEV and specified an ill-formed macro/variable as follows:
```
test-dependencies: \
    org.junit;version='[4.12.0,5.0.0)' \
    org.mockito;version='[1.9.5,2.0.0)'
```
Note the missing comma to separate the two bundle references.

When referencing the variable, I see these error markers (in Bndtools):

> Cannot find /error/org.junit;version=[1.9.5,2.0.0) Not found in [bnd-cache, Bndtools Hub, aQute.bnd.repository.p2.provider.P2Repository@66927bb2, Maven Central]

> org.junit;version=[1.9.5,2.0.0) Not found in [bnd-cache, Bndtools Hub, aQute.bnd.repository.p2.provider.P2Repository@d48dbd2, Maven Central]

Wouldn't it be clearer if there was an error message saying that `org.junit;version='[4.12.0,5.0.0)' org.mockito;version='[1.9.5,2.0.0)'` is not a valid list of bundle references?

As far as I understand, variables in bnd are resolved on a pure textual base. However, at that point in time, it should be possible to detect which actual format the variable value must have and ensure the formal correctness of the variable.

bjhargrave:
It depends upon where the variable is used. Not what the variable contains. It is just text substitution. 

Of course error reporting can always be improved and such PRs are always welcome.

bjhargrave:
Project.getBundles needs to pass itself to Parameters as a Reporter. OSGiHeaders.parseHeaders can then report the end of a quoted string that is not followed by , ; or end of string.

bjhargrave:
> Project.getBundles needs to pass itself to Parameters as a Reporter.

I have made this change in #1877.

PR #2033, opened 2017-05-11
Bnd gets confused by ill-formed bundle references #1864

pkriens:
The properties parser now detects missing delimiters after 
a quoted string. We could still do checks when we know
what the header means but I think this catches more
of the problems.


pkriens:
Should fix #1864

Issue #1971, opened 2017-03-29
-connection-settings does not support maven password encryption


jwausle:
Authentication to nexus fail silent if I use
`-connection-settings: ~/.m2/settings.xml, ~/.m2/settings-security.xml`
with encrypted maven password (https://maven.apache.org/guides/mini/guide-encryption.html)
---

bnd-3.3.0 with gradlew

**Workaround**

> -connection-settings: my-settings.xml
```
<settings>
  <servers>
    <server>
      <id>https://my-secure-host</id>
      <username>me</username>
      <password>decrypted-password</password>
      <verify>true</verify>
      <trust></trust>
    </server>
  </servers>
</settings>
```
- Thats solve the problem. 
- But every developer of the team download with my credentials.
- Special system-user is not possible (security constraints)

jwausle:
Authentication to nexus fail silent if I use 
> -connection-settings: ~/.m2/settings.xml, ~/.m2/settings-security.xml

with encrypted maven password.
- https://maven.apache.org/guides/mini/guide-encryption.html

## Workaround 
-connection-settings: my-settings.xml

```
<settings>
  <servers>
    <server>
      <id>https://my-secure-host</id>
      <username>me</username>
      <password>decrypted-password</password>
      <verify>true</verify>
      <trust></trust>
    </server>
  </servers>
</settings>
```

Thats solve the problem. But every developer of the team download with my credentials. 
- special system-user is not possible (security constraints)

pkriens:
I've read up on the maven 'secure' passwords and I am a bit puzzled. Why do you share your settings.xml in the first place? This file should be local to every user?

http://stackoverflow.com/questions/30769636/how-does-maven-3-password-encryption-work/43118084#43118084

jwausle:
It was not my first approach. 

1) First I try to use (without success :( )
```
-connection-settings: ~/.m2/settings.xml, ~/.m2/settings-security.xml 
```

2) The simplest fix (for me) is to provide my-settings.xml in connection-settings format like here explained http://bnd.bndtools.org/instructions/connection-settings.html 
- the trick was to use the real decrypt password from nexus and not the encrypted password from maven in '~/.m2/settings.xml'

pkriens:
I will make the fix but after analyzing the Maven code I am utterly puzzled. Therefore, my question is about 'why' you need this since I cannot see how you get any extra protection with this scheme.

[...]

PR #2034, opened 2017-05-11
Encrypted Maven passwords #1971

pkriens:


pkriens:
Should fix #1971

aQute.libg/src/aQute/lib/concurrentinit/ConcurrentInitialize.java
```diff
@@ -0,0 +1,92 @@
...
+						throw new IllegalStateException(
+								"Cycle:  ConcurrentInitialize's create returns to same instance");
+					do {
+						try {
+							lock.wait();
+						} catch (InterruptedException e) {
+							return null;
+						}
+					} while (state == State.CREATING);
+					return value;
```
bjhargrave: What if it the state is `ERROR` here? The exception should be thrown.

aQute.libg/src/aQute/lib/concurrentinit/ConcurrentInitialize.java
```diff
@@ -0,0 +1,92 @@
...
+
+				case ERROR :
+					throw exception;
+
+				case DONE :
+					return value;
+			}
+		}
+		try {
+			value = create();
```
bjhargrave: This is not safe, you set `value` here. You should set a local var and then set value from the local var in the synchronized block. Just like how exception is done.

```java
T v = create();
synchronized (lock) {
  value = v;
```

aQute.libg/src/aQute/lib/concurrentinit/ConcurrentInitialize.java
```diff
@@ -0,0 +1,92 @@
...
+			}
+		} catch (Exception e) {
+			synchronized (lock) {
+				exception = e;
+				creatingThread = null;
+				state = State.ERROR;
+				lock.notifyAll();
+			}
+		}
+		if (exception != null)
```
bjhargrave: You should only read and write mutable state under guard. Here you read exception and value without guard. I know this is only run on INIT, so it is OK. It would be must better if the value was returned and the exception thrown from within the synchronized regions above:

```java
try {
  T v = create();
  synchronized (lock) {
    creatingThread = null;
    state = State.DONE;
    lock.notifyAll();
    return value = v;
  }
} catch (Exception e) {
  synchronized (lock) {
    creatingThread = null;
    state = State.ERROR;
    lock.notifyAll();
    throw exception = e;
  }
}```

File: aQute.libg/src/aQute/lib/xpath/XPathParser.java [hunk 4]
```diff
@@ -0,0 +1,72 @@
...
+	static DocumentBuilder				db;
+	static XPath						xp;
+	final Document						doc;
+
+	static {
+		try {
+			db = dbf.newDocumentBuilder();
+			xp = xpf.newXPath();
+		} catch (Throwable e) {
+			// ignore
```
bjhargrave: Seem serious. May convert to thrown Error?

biz.aQute.bndlib/src/aQute/bnd/connection/settings/ConnectionSettings.java
```diff
@@ -323,6 +351,35 @@ private void parse(File file) throws Exception {
...
+		} else {
+
+			if (!MavenPasswordObfuscator.isObfuscatedPassword(master)) {
+				warning("Master password in %s was not obfuscated, using actual value", path);
+				return master;
+			}
+			try {
+				return MavenPasswordObfuscator.decrypt(master, M2_SETTINGS_SECURITY_PROPERTY);
+			} catch (Exception e) {
+				error("Could not decrypt the master password from %s with key %s", e, path,
```
bjhargrave: Use `exception` instead of `error` when you have an exception object to report.

aQute.libg/src/aQute/lib/concurrentinit/ConcurrentInitialize.java
pkriens: I fixed it but in this case it is not a problem. Any other thread is blocked until the state progresses from CREATING. Since it is a private variable as far as I can see nobody can use the object before the subsequent synchronised block has exited?

Anyway, I moved the synchronized blocks into one method to make it simpler

aQute.libg/src/aQute/lib/xpath/XPathParser.java
bjhargrave: I meant `throw new Error(e);`

aQute.libg/src/aQute/lib/concurrentinit/ConcurrentInitialize.java
```diff
@@ -69,24 +68,29 @@ public T get() throws Exception {
...
-			throw exception;
-		return value;
 	}
 
+	private void set(T value, Exception e, State state) {
+		synchronized (lock) {
+			this.value = value;
+			this.state = state;
+			this.creatingThread = null;
+			this.lock.notifyAll();
```
bjhargrave: You fail to set`exception = e;`

Issue #1970, opened 2017-03-29
Bnd OSGi test launcher ignore Tests inside test.fragments


jwausle:
My annotated RunWith-Tests ignored if I declare them inside a test.fragment.
- Same Tests works well inside real test.bundle.

bnd-3.3.0 with gradlew

## Buggy-setup (tests ignored)
- host.bundle/bnd.bnd
>  Bundle-SymbolicName: host.bundle

- fragment.bundle/bnd.bnd
> Fragment-Host: host.bundle
> Test-Cases: ${classes;CONCRETE;ANNOTATED;org.junit.runner.RunWith}

## Workaround-setup (tests executed)

- host.bundle/bnd.bnd
> -sub: *.bnd
- host.bundle/host.bnd
>  Bundle-SymbolicName: host.bundle
- host.bundle/fragment.bnd
> Fragment-Host: host.bundle
> Export-Package: host.bundle.internal.*

- test.bundle/bnd.bnd
> Test-Cases: ${classes;CONCRETE;ANNOTATED;org.junit.runner.RunWith}

## Background
- I want to get access to my internal classes while bundle-testing like in RCP-development.

bjhargrave:
Fixed by https://github.com/bndtools/bnd/pull/2035

PR #2035, opened 2017-05-11
[testing] Fragment testing #1970

pkriens:
If the test bundle is a fragment then its host is
used for class loading.


pkriens:
Should fix #1970

biz.aQute.junit/src/aQute/junit/Activator.java
```diff
@@ -359,6 +371,46 @@ public void check() {
...
+				}
+			}
+		}
+
+		return found == null ? bundle : found;
+	}
+
+	private boolean isFragment(Bundle bundle) {
+		return bundle != null && bundle.getState() != Bundle.ACTIVE
+				&& bundle.getHeaders().get(Constants.FRAGMENT_HOST) != null;
```
bjhargrave: Why don't you use `PackageAdmin.getBundleType`? You already track the package admin service now.

Issue #2038, opened 2017-05-11
Support DynamicImport-Package expansion


magnet:
It can be useful to have DynamicImport-Packages fully expanded -- classpath expansion + versioning info to have strong compatibility guarantees while allowing more a more dynamic handling than optional packages (like an optional, greedy reference vs. an optional, reluctant reference in DS).

In [1], @bjhargrave proposed using `resolution:=dynamic` on packages in the `Import-Package` instruction to tell Bnd to move those packages in DynamicImport-Package.

https://groups.google.com/forum/#!topic/bndtools-users/94Wb2uQghf8

bjhargrave:
Fixed by #2039

PR #2039, opened 2017-05-11
Support DynamicImport-Package expansion (issue #2038)

magnet:
Packages in Import-Package marked with resolution:=dynamic will be moved to the DynamicImport-Package entry after expansion takes place.

The test demonstrates packages get their version attribute added and are properly expanded, and moved to DynamicImport-Package (added after the existing instruction, and removed from Import-Package), and that other imported packages are not affected.

I see no regression risk unless someone used resolution:=dynamic in an existing build... but then again, as far as I know the directive "resolution" is reserved by the OSGi spec.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -87,6 +87,7 @@ import aQute.libg.cryptography.SHA1;
 import aQute.libg.generics.Create;
 import aQute.libg.glob.Glob;
 import aQute.libg.reporter.ReporterMessages;
+import aQute.libg.tuple.Pair;
 
 public class Analyzer extends Processor {
 	private final static Logger						logger					= LoggerFactory.getLogger(Analyzer.class);
@@ -713,11 +714,22 @@ public class Analyzer extends Processor {
 			else
 				main.remove(EXPORT_PACKAGE);
 
-			// Remove all the Java packages from the imports
-			if (!imports.isEmpty()) {
-				main.putValue(IMPORT_PACKAGE, printClauses(imports));
+			// Divide imports with resolution:=dynamic to DynamicImport-Package
+			// and add them to the existing DynamicImport-Package instruction
+			Pair<Packages,Parameters> regularAndDynamicImports = divideRegularAndDynamicImports();
+			Packages regularImports = regularAndDynamicImports.getFirst();
+
+			if (!regularImports.isEmpty()) {
+				main.putValue(IMPORT_PACKAGE, printClauses(regularImports));
 			} else {
 				main.remove(IMPORT_PACKAGE);
+			}
+
+			Parameters dynamicImports = regularAndDynamicImports.getSecond();
+			if (!dynamicImports.isEmpty()) {
+				main.putValue(DYNAMICIMPORT_PACKAGE, printClauses(dynamicImports));
+			} else {
+				main.remove(DYNAMICIMPORT_PACKAGE);
 			}
 
 			Packages temp = new Packages(contained);
@@ -817,7 +829,8 @@ public class Analyzer extends Processor {
 				}
 
 				if (header.equals(BUNDLE_CLASSPATH) || header.equals(EXPORT_PACKAGE) || header.equals(IMPORT_PACKAGE)
-						|| header.equals(REQUIRE_CAPABILITY) || header.equals(PROVIDE_CAPABILITY))
+						|| header.equals(DYNAMICIMPORT_PACKAGE) || header.equals(REQUIRE_CAPABILITY)
+						|| header.equals(PROVIDE_CAPABILITY))
 					continue;
 
 				if (header.equalsIgnoreCase("Name")) {
@@ -1749,6 +1762,25 @@ public class Analyzer extends Processor {
 		if (isPedantic() && noimports.size() != 0) {
 			warning("Imports that lack version ranges: %s", noimports);
 		}
+	}
+
+	Pair<Packages,Parameters> divideRegularAndDynamicImports() {
+		Packages regularImports = new Packages(imports);
+		Parameters dynamicImports = getDynamicImportPackage();
+
+		Iterator<Entry<PackageRef,Attrs>> regularImportsIterator = regularImports.entrySet().iterator();
+		while (regularImportsIterator.hasNext()) {
+			Entry<PackageRef,Attrs> packageEntry = regularImportsIterator.next();
+			PackageRef packageRef = packageEntry.getKey();
+			Attrs attrs = packageEntry.getValue();
+			String resolution = attrs.get(Constants.RESOLUTION_DIRECTIVE);
+			if (Constants.RESOLUTION_DYNAMIC.equals(resolution)) {
+				attrs.remove(Constants.RESOLUTION_DIRECTIVE);
+				dynamicImports.put(packageRef.fqn, attrs);
+				regularImportsIterator.remove();
+			}
+		}
+		return new Pair<>(regularImports, dynamicImports);
 	}
 
 	String applyVersionPolicy(String exportVersion, String importRange, boolean provider) {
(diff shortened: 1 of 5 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
```diff
@@ -307,6 +307,7 @@
 	String							LINK_ATTRIBUTE								= "link";
 	String							LITERAL_ATTRIBUTE							= "literal";
 	String							NAME_ATTRIBUTE								= "name";
+	String							DYNAMIC_ATTRIBUTE							= "dynamic";
```
bjhargrave: It is not an attribute name. It is a directive value. See https://osgi.org/javadoc/r6/core/org/osgi/framework/namespace/PackageNamespace.html#RESOLUTION_DYNAMIC. So I would prefer to name it `RESOLUTION_DYNAMIC`.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -1751,6 +1755,23 @@ void augmentImports(Packages imports, Packages exports) throws Exception {
...
+		Iterator<Entry<PackageRef,Attrs>> importsIterator = imports.entrySet().iterator();
+		while (importsIterator.hasNext()) {
+			Entry<PackageRef,Attrs> packageEntry = importsIterator.next();
+			PackageRef packageRef = packageEntry.getKey();
+			Attrs attrs = packageEntry.getValue();
+			String resolution = attrs.get(Constants.RESOLUTION_DIRECTIVE);
+			if (Constants.DYNAMIC_ATTRIBUTE.equals(resolution)) {
+				attrs.remove(Constants.RESOLUTION_DIRECTIVE);
+				dynamicImportPackages.add(packageRef.getFQN(), attrs);
+				importsIterator.remove();
```
bjhargrave: I am not sure this is exactly the right way to go to remove these from `imports`. There are other places where `imports` is checked and it should still contain all the imports. You should do the division into `Import-Package` and `DynamicImport-Package` in `calcManifest()` but `imports` should always retain all the imported packages so plugins which run after the analyzer can see all the imports.

bjhargrave: The code here: https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L717-L721 is where you should make changes to divide the imports into IP and DIP headers.

magnet: I'll update the PR, but as far I could see plugins can not run at this point of the execution. They are invoked earlier.

magnet:
I applied the requested fixes.

About the doc, I tried not to go too much in details and just show how to use it.

Regarding the logic to move the imports, I just moved the call to moveDynamicImports() to calcManifest() to avoid changing calcManifest() too much. So I did not locally change exports to keep this logic intact:

```java
			// Remove all the Java packages from the imports
			if (!imports.isEmpty()) {
				main.putValue(IMPORT_PACKAGE, printClauses(imports));
```

At that stage no plugin can interfere. (also the comment about Java packages is outdated, I did not remove it but the removal happens much earlier in:

```java
				// Remove any Java references ... where are the closures???
				for (Iterator<PackageRef> i = referredAndExported.keySet().iterator(); i.hasNext();) {
					if (i.next().isJava())
						i.remove();
				}
```
( https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L280 ) 

I think it should do it. Regular imports are printed using printClauses() but only to strip exported packages that are imported from bnd export directives. I don't think it makes sense to do it for DIPs which truly have no reason to come from exports, but just in case I have the code ready to use printClauses() for DIPs too. It's slightly more complex:

```java
	void moveDynamicImports() throws IOException {
		Packages movedDynamicImportPackages = new Packages();
		Iterator<Entry<PackageRef,Attrs>> importsIterator = imports.entrySet().iterator();
		while (importsIterator.hasNext()) {
			Entry<PackageRef,Attrs> packageEntry = importsIterator.next();
			PackageRef packageRef = packageEntry.getKey();
			Attrs attrs = packageEntry.getValue();
			String resolution = attrs.get(Constants.RESOLUTION_DIRECTIVE);
			if (Constants.RESOLUTION_DYNAMIC.equals(resolution)) {
				attrs.remove(Constants.RESOLUTION_DIRECTIVE);
				movedDynamicImportPackages.put(packageRef, attrs);
				importsIterator.remove();
			}
		}
		if (!movedDynamicImportPackages.isEmpty()) {
			String explicitDynamicImportPackages = getProperty(DYNAMICIMPORT_PACKAGE);
			String movedDIPClauses = printClauses(movedDynamicImportPackages);

			if (explicitDynamicImportPackages != null) {
				setDynamicImportPackage(explicitDynamicImportPackages + ',' + movedDIPClauses);
			} else {
				setDynamicImportPackage(movedDIPClauses);
			}
		}
	}
```

(`printClauses()` throws IOException for no reason)

magnet:
I applied the requested fixes.

About the doc, I tried not to go too much in details and just show how to use it.

Regarding the logic to move the imports, I just moved the call to moveDynamicImports() to calcManifest() to avoid changing calcManifest() too much. So I did not locally change exports to keep this logic intact:

```java
			// Remove all the Java packages from the imports
			if (!imports.isEmpty()) {
				main.putValue(IMPORT_PACKAGE, printClauses(imports));
```

At that stage no plugin can interfere. (also the comment about Java packages is outdated, I did not remove it but the removal happens much earlier in:

```java
				// Remove any Java references ... where are the closures???
				for (Iterator<PackageRef> i = referredAndExported.keySet().iterator(); i.hasNext();) {
					if (i.next().isJava())
						i.remove();
				}
```
( https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L280 ) 

I think it should do it. Regular imports are printed using printClauses() but only to strip exported packages that are imported from bnd export directives. I don't think it makes sense to do it for DIPs which truly have no reason to come from exports, but just in case I have the code ready to use printClauses() for DIPs too. It's slightly more complex:

```java
	void moveDynamicImports() throws IOException {
		Packages movedDynamicImportPackages = new Packages();
		Iterator<Entry<PackageRef,Attrs>> importsIterator = imports.entrySet().iterator();
		while (importsIterator.hasNext()) {
			Entry<PackageRef,Attrs> packageEntry = importsIterator.next();
			PackageRef packageRef = packageEntry.getKey();
			Attrs attrs = packageEntry.getValue();
			String resolution = attrs.get(Constants.RESOLUTION_DIRECTIVE);
			if (Constants.RESOLUTION_DYNAMIC.equals(resolution)) {
				attrs.remove(Constants.RESOLUTION_DIRECTIVE);
				movedDynamicImportPackages.put(packageRef, attrs);
				importsIterator.remove();
			}
		}
		if (!movedDynamicImportPackages.isEmpty()) {
			String explicitDynamicImportPackages = getProperty(DYNAMICIMPORT_PACKAGE);
			String movedDIPClauses = printClauses(movedDynamicImportPackages);

			if (explicitDynamicImportPackages != null) {
				setDynamicImportPackage(explicitDynamicImportPackages + ',' + movedDIPClauses);
			} else {
				setDynamicImportPackage(movedDIPClauses);
			}
		}
	}
```

(`printClauses()` throws IOException for no reason)

magnet:
OK, it will be a bit more verbose but I'll fix the PR that way.
How do you feel about using printClauses() to print the DIPs? It strips export directives, I don't think it's useful.

bjhargrave:
> How do you feel about using printClauses() to print the DIPs? It strips export directives, I don't think it's useful.

Yes, I do think you need to use `printClauses`. It does a number of other things and will provide consistent output to the manifest.

magnet:
@bjhargrave I applied changes you suggested.
I divide regular Import-Packages from DynamicImport-Packages using a method. It returns a Pair<Packages, Parameter> where first is the (filtered) Import-Packages, and Parameter is the header for the Dynamic-ImportPackage entry. I can't make it a Packages instance because it contains the original DIP header which may contain invalid packages (e.g containing wildcard) and they are not PackageRefs.

I use the same (existing) code pattern to add Import-Packages and Dynamic-ImportPackages.

I added an extra testcase when there's no DIP header present and rebased the PR on HEAD.

bjhargrave:
Excellent!

> If it's OK with you I'll squash the commits into one.

Yes, please squash them together.

magnet:
@bjhargrave Thanks, I squashed the commits and cleaned up the commit a bit (removed changed not needed anymore)

Issue #2040, opened 2017-05-12
bnd-baseline-maven-plugin fails on projects with no <distributionManegement>


magnet:
```[DEBUG] Found versions [1.0.0-SNAPSHOT, 1.0.0]
[INFO] The baseline version was found to be 1.0.0
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 6.961 s
[INFO] Finished at: 2017-05-12T10:04:09+02:00
[INFO] Final Memory: 21M/301M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal biz.aQute.bnd:bnd-baseline-maven-plugin:3.4.0-SNAPSHOT:baseline (baseline) on project primeval-commons: An error occurred while calculating the baseline: NullPointerException -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal biz.aQute.bnd:bnd-baseline-maven-plugin:3.4.0-SNAPSHOT:baseline (baseline) on project primeval-commons: An error occurred while calculating the baseline
```

The problem happens because the plugin inserts the distributionManagement in the repository list by default. When it is null, Aether fails with a NPE.

magnet:
There is a workaround with:
```xml
<properties>		
   <bnd.baseline.include.distribution.management>false</bnd.baseline.include.distribution.management>
</properties>
```

timothyjward:
@magnet I'll let you verify this before we mark it as closed

magnet:
Looks good to me, thanks.

PR #2041, opened 2017-05-12
Issue #2040: Fix NPE when a project has no distributionManagement

magnet:
On projects without distributionManagement, releaseDistroRepo is null, inserted at the head of the list and it causes Aether to throw a NPE.

It is difficult to test because the current maven baseline tests rely on distributionManagement with a local repo to find a baseline, and this happens only after a baseline has been found. It might be possible by setting settings to offline, and provisioning a test m2 local repository, but I hope a NPE fix does not warrant so much maven pain. :grin: 

There is a workaround with:
```xml
<properties>		
   <bnd.baseline.include.distribution.management>false</bnd.baseline.include.distribution.management>
</properties>
```

### DIFF
--- a/maven/bnd-baseline-maven-plugin/src/main/java/aQute/bnd/maven/baseline/plugin/BaselineMojo.java
+++ b/maven/bnd-baseline-maven-plugin/src/main/java/aQute/bnd/maven/baseline/plugin/BaselineMojo.java
@@ -143,7 +143,11 @@ public class BaselineMojo extends AbstractMojo {
 				releaseDistroRepo = RepositoryUtils.toRepo(project.getDistributionManagementArtifactRepository());
 			}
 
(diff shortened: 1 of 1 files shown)


timothyjward:
I see your predicament when it comes to testing. Adding distribution management (when it exists) is important because some (slightly odd) projects release to a repository which isn't a configured repository for dependencies. A null check is definitely the best we can do here without a major rewrite of the tests, and I have no idea how to reproduce some of the test cases without a distribution management section!

magnet:
Thanks for merging.

Yes I understand why you are doing this. Even though it should be up to the projects themselves to deal with this, Maven does not make it easy. Running Nexus just for that is a bit overkill.

(I avoid distributionManagement because I find attaching a build to an FTP server is heresy. Fortunately there is decent support for doing without it)

PR #2042, opened 2017-05-14
Add --write option to 'bnd resolve' command

njbartlett:
This option writes runbundles back to the specified file, e.g.:
   $ bnd resolve resolve --write foo.bndrun

### DIFF
--- a/biz.aQute.bnd/src/aQute/bnd/main/ResolveCommand.java
+++ b/biz.aQute.bnd/src/aQute/bnd/main/ResolveCommand.java
@@ -36,6 +36,7 @@ import aQute.lib.getopt.Description;
 import aQute.lib.getopt.Options;
 import aQute.lib.io.IO;
 import aQute.lib.strings.Strings;
+import biz.aQute.resolve.Bndrun;
 import biz.aQute.resolve.ProjectResolver;
 import biz.aQute.resolve.ResolverValidator;
 import biz.aQute.resolve.ResolverValidator.Resolution;
@@ -238,6 +239,9 @@ public class ResolveCommand extends Processor {
 
 		@Description("Print out the bundles")
 		boolean bundles();
+
+		@Description("Print -runbundles instruction to the file")
+		boolean write();
 	}
 
 	@Description("Resolve a bndrun file")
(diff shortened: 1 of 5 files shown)


biz.aQute.bnd/src/aQute/bnd/main/ResolveCommand.java
```diff
@@ -268,28 +272,33 @@ public void _resolve(ResolveOptions options) throws Exception {
...
-							System.out.printf("# %-50s ok\n", f.getName());
-							if (options.bundles()) {
-								for (Resource r : resolution.keySet()) {
-									IdentityCapability id = ResourceUtils.getIdentityCapability(r);
-									List<ContentCapability> content = ResourceUtils.getContentCapabilities(r);
-									System.out.printf("  %-50s %40s %s\n", id.osgi_identity(),
-											content.get(0).osgi_content(), content.get(0).url());
+				if (options.write()) {
+					Bndrun bndrun = new Bndrun(run.getWorkspace(), f);
+					String resolved = bndrun.resolve(false, true);
```
bjhargrave: It seems to me this block should be used for both --write and not --write. The only difference is that you pass `option.write()` as the second arg to the `resolve` method. Then if `option.write()` is false, then print out the result `resolved`.

njbartlett: I agree, however that would change the output format drastically, which would be a backward-incompatible change. Currently it produces output like this:

    # bundle-name       content_SHA     content_URL

bjhargrave: How is the output format API? Also, it does not seem useful. You have to massage it into a useful format to put back in the bndrun file. I guess @pkriens can comment here.

njbartlett: It's arguably API because somebody could have built a toolchain that post-processes that output, e.g. with awk or sed. I agree it's not a very useful format aside from that -- and also inconsistent with the failure output of the Maven plugins such as resolve-, export- and testing-maven-plugin.

pkriens: It does seem cleaner to not duplicate code. Why can't you keep the output and the new option from the same block?

njbartlett: I don't understand the question. If we call Bndrun.resolve() we get different output. If I change the output of Bndrun.resolve() to the original/existing output of this command, then it will break other users of that method.

pkriens: ? Bndrun does exactly the same as the current body. Since Bndrun itself does not do any output, you can keep the current body and, as BJ seems to suggest, just use write as a flag to resolve?Just refactoring the old body and use Bndrun seems the best solution or am I missing something?

njbartlett: To do this I will have to change the return type from Bndrun.resolve(). It currently returns a flat String, not anything I can transform into the desired output.

njbartlett:
Implemented suggested changes. To preserve compatibility of the API, I have added a new method `resolveResources` to Bndrun and deprecated the old `resolve` method.

njbartlett:
Updated again after discussing with BJ.

PR #2049, opened 2017-05-18
[macro] Properly report position of the error and warning macros.

pkriens:
If the macro is used multiple times it finds the first occurrence. Attempting to find the reported string is very complex since this string can contain macros so the source is different from what is reported. 

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
@@ -38,6 +38,7 @@ import javax.script.ScriptContext;
 import javax.script.ScriptEngine;
 import javax.script.ScriptEngineManager;
 
+import aQute.bnd.osgi.Processor.FileLine;
 import aQute.bnd.version.MavenVersion;
 import aQute.bnd.version.Version;
 import aQute.bnd.version.VersionRange;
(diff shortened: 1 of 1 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
```diff
@@ -607,16 +609,22 @@ public String _replace(String args[]) {
 		return sb.toString();
 	}
 
-	public String _warning(String args[]) {
+	public String _warning(String args[]) throws Exception {
 		for (int i = 1; i < args.length; i++) {
-			domain.warning("%s", process(args[i]));
+			SetLocation warning = domain.warning("%s", process(args[i]));
+			FileLine header = domain.getHeader(Pattern.compile(".*"), Pattern.compile("\\$\\{warning;"));
```
bjhargrave: Would it make sense to put the patterns into `private static final` variables to avoid repeated pattern compilation?

pkriens: Didn't think it was worth the effort, this is a very low frequency event

bjhargrave: ok, I restarted the appveyor build since it failed on a network issue

PR #2051, opened 2017-05-19
Use RepositoryPlugin::getName to get repository names where applicable

psoreide:
Some RepositoryPlugin::toString returns stateful information of the repository so they won't
be picked up if you have ordered/selected repositories in .bnd(run) files

### DIFF
--- a/biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java
+++ b/biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java
@@ -47,6 +47,7 @@ import aQute.bnd.osgi.resource.RequirementBuilder;
 import aQute.bnd.osgi.resource.ResourceBuilder;
 import aQute.bnd.osgi.resource.ResourceUtils;
 import aQute.bnd.service.Registry;
+import aQute.bnd.service.RepositoryPlugin;
 import aQute.bnd.service.Strategy;
 import aQute.bnd.service.resolve.hook.ResolverHook;
 import aQute.lib.converter.Converter;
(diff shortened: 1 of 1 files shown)


biz.aQute.repository/src/aQute/bnd/repository/osgi/OSGiRepository.java
```diff
@@ -331,4 +332,12 @@ public void close() throws IOException {
 			throw Exceptions.duck(e);
 		}
 	}
+
+	public static String getRepositoryName(Repository repo) {
```
bjhargrave: This seems an invalid place to put this method. It has nothing to do with the OSGiRepository type. This is bad coupling.

And it is only called from one place. I would put the logic there.

biz.aQute.repository/src/aQute/bnd/deployer/repository/wrapper/Plugin.java
```diff
@@ -153,7 +153,11 @@ public void close() throws java.io.IOException {
 	}
 
 	public String toString() {
-		return wrapper != null ? wrapper.toString() : "<wrapper not set>";
+		if (wrapper != null) {
+			wrapper.init();
```
bjhargrave: This is wrong. Calling toString should not init a repo. toString is called in debuggers and it should not have the side effect like this.

Also, the wrapper is basically deprecated. As is jpm repo.

biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java
```diff
@@ -367,7 +368,7 @@ private Processor loadRepositories() throws Exception {
 
 			Map<String,Repository> repoNameMap = new HashMap<String,Repository>(allRepos.size());
 			for (Repository repo : allRepos)
-				repoNameMap.put(repo.toString(), repo);
```
bjhargrave: The logic to use getName() or getString() as the repo name should be here.

biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java
```diff
@@ -366,8 +367,18 @@ private Processor loadRepositories() throws Exception {
 			// Map the repository names...
 
 			Map<String,Repository> repoNameMap = new HashMap<String,Repository>(allRepos.size());
-			for (Repository repo : allRepos)
-				repoNameMap.put(repo.toString(), repo);
+			for (Repository repo : allRepos) {
+				String name = repo.toString();
```
bjhargrave: The toString call can be defered to a final else block to avoid calling if not needed. And also to avoid double calling for wrapper plugin.

biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java
```diff
@@ -366,8 +367,18 @@ private Processor loadRepositories() throws Exception {
...
+				if (repo instanceof RepositoryPlugin) {
+					name = ((RepositoryPlugin) repo).getName();
+				} else if (repo instanceof aQute.bnd.deployer.repository.wrapper.Plugin) {
```
bjhargrave: You may want to order the tests such that the wrapper plugin is tested for first since it is the most specific type. Right now it is not a RepositoryPlugin, but some super type could be changed to implement RepositoryPlugin and then it would fall into the if RepositoryPlugin test and miss the, I assume necessary, call to init before toString.

PR #2059, opened 2017-05-24
[JartoolSigner] Add support for Time Stamping Authority (TSA)

psoreide:

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/signing/JartoolSigner.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/signing/JartoolSigner.java
@@ -45,6 +45,12 @@ public class JartoolSigner implements Plugin, SignerPlugin {
 		String sigFile() default "";
 
 		String digestalg() default "";
+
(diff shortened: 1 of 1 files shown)


biz.aQute.bndlib/src/aQute/bnd/signing/JartoolSigner.java
```diff
@@ -45,6 +45,12 @@
 		String sigFile() default "";
 
 		String digestalg() default "";
+
+		String tsaurl() default "";
```
bjhargrave: The other option names generally match the names of the corresponding jarsigner tool option. So this option should be `tsa`

biz.aQute.bndlib/src/aQute/bnd/signing/JartoolSigner.java
```diff
@@ -45,6 +45,12 @@
...
+
+		String tsacertalias() default "";
```
bjhargrave: This should be `tsacert`.

PR #2062, opened 2017-05-26
Support abbreviated syntax for -runrequires

### LINKED ISSUE #1996: Support abbreviated syntax for -runrequires
We have more and more users who need to edit bndrun files outside Bndtools. Unfortunately the syntax of -runrequires is very cumbersome for direct editing, being based on the full specification of a Require-Capability header.

We should introduce an optional abbreviated syntax for the most common case of requiring a bundle by bsn and version. I'm open to suggestions but propose something like:

    -runrequires:  bundle; bsn=org.example.foo; version=1.2

This does not require changing the parser, since it is still a valid requirement. It just needs post-processing to translate the namespace and filter expression.

njbartlett:
Fixes #1996

Example:
  -runrequires: bundle; bsn=org.example.foo; version=1.0

which translates to:

  -runrequires: osgi.identity; filter:=‘(&(osgi.identity=org.example.foo)(version>=1.0.0))’

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReqBuilder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReqBuilder.java
@@ -5,9 +5,12 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Set;
 import java.util.regex.Pattern;
 
 import org.osgi.framework.Constants;
@@ -37,6 +40,10 @@ import aQute.libg.filters.LiteralFilter;
 import aQute.libg.filters.SimpleFilter;
 
 public class CapReqBuilder {
+
+	private static final String			REQ_ALIAS_BUNDLE				= "bundle";
+	private static final String			REQ_ALIAS_BUNDLE_NAME_ATTRIB	= "bsn";
+	private static final String REQ_ALIAS_BUNDLE_VERSION_ATTRIB = "version";
 
 	private final String				namespace;
 	private Resource					resource;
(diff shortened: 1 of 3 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReqBuilder.java
```diff
@@ -38,6 +41,10 @@
 
 public class CapReqBuilder {
 
+	private static final String			REQ_ALIAS_BUNDLE				= "bundle";
```
bjhargrave: `bundle` seems too short. We should at least scope with `bnd.bundle`. This also makes it clear this is a bnd-ism.

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReqBuilder.java
```diff
@@ -260,33 +267,121 @@ public CapReqBuilder filter(CharSequence f) {
...
+	 *            Requirement in the unspecified namespace "bundle".
+	 * @return The list of parsed requirements.
+	 * @throws Exception
+	 */
+	public static List<Requirement> getRequirementsFrom(Parameters rr, boolean unalias) throws Exception {
 		List<Requirement> requirements = new ArrayList<Requirement>();
 		for (Entry<String,Attrs> e : rr.entrySet()) {
-			requirements.add(getRequirementFrom(Processor.removeDuplicateMarker(e.getKey()), e.getValue()));
+			Requirement req = getRequirementFrom(Processor.removeDuplicateMarker(e.getKey()), e.getValue());
+			if (unalias) {
```
bjhargrave: Why don't you just pass `unalias` to the overload of `getRequirementFrom`?

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReqBuilder.java
bjhargrave: Also `bundle` seems the wrong alias name. To me, a `bundle` requirement is `osgi.wiring.bundle` namespace. That is, `Require-Bundle`. You are doing an alias for `osgi.identity`, so I would think the better alias name is `bnd.identity`.

PR #2063, opened 2017-05-30
Allows a transitive false|true option for a BndPomRepository

pkriens:
---


pkriens:
Should handle #1571

PR #2066, opened 2017-06-02
maven: Rename option to useMavenDependencies

bjhargrave:
This is to make the option more self-documenting. The old name,
useDefaults, was too vague.


bjhargrave:
FYI @azzazzel

PR #2072, opened 2017-06-22
[test] make it possible to use the maven/gradle system properties for specifying which integration test class to execute

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -2112,6 +2112,15 @@ public void test(File reportDir, List<String> tests) throws Exception {
 			IO.delete(reportDir);
 			tester.setReportDir(reportDir);
 		}
+		if (tests == null) {
+			// check system properties: gradle
+			String singleTest = System.getProperty("integTest.single");
```
bjhargrave: This is not the right property name. See <https://docs.gradle.org/4.0/userguide/java_plugin.html#sec:single_test_execution_via_system_properties>. The property name is based upon the gradle task name. And you cannot really know the gradle task name here in bndlib. The standard task name for the BndPlugin is `testOSGi`.

Furthermore, value of the property in gradle is a globbing pattern.

Finally, this is the wrong place to put maven and gradle related support. Such support needs to go in the appropriate plugins.

rotty3000: https://docs.gradle.org/3.3/userguide/java_plugin.html#sec:single_test_execution_via_system_properties

rotty3000: OR https://docs.gradle.org/4.0/userguide/java_plugin.html#sec:single_test_execution_via_system_properties

bjhargrave: I put that link in my comment.

rotty3000: also we can't glob like gradle or maven do unless we change many more pieces.

rotty3000: are you suggesting we just use yet another system property specific to bnd's test infra?

bjhargrave: But this is wrong in many ways.

rotty3000: i.e. "tester.names" is already defined, just need to add the logic to retrieve it from the sys props.

bjhargrave: > are you suggesting we just use yet another system property specific to bnd's test infra?

No. I am suggesting it is done properly in the proper place. And here is not the proper place and it is not done in the proper way.

rotty3000: we can't use the globbing feature of gradle or maven at the plugin itself, because there are no classes there to filer. The class names are not available until WAY later during the test launcher framework execution. So we need a single form a globbing that is neither gradle nor maven supported anyway... so do we just check the system properties for the property bnd already defines (a.k.a. `tester.names`)?

rotty3000: finally, these ARE integration tests.. that's why I applied the integration test properties defined by maven and gradle and not the "unit" test properties.

bjhargrave: `tester.names` is not for us to use. It if for the project to communicate to the launched VM. 

> finally, these ARE integration tests.. that's why I applied the integration test properties defined by maven and gradle and not the "unit" test properties.

Yes, but you did not properly read the gradle description:

```
Setting a system property of taskName.single = testNamePattern will only execute tests that match the specified testNamePattern. The taskName can be a full multi-project path like “:sub1:sub2:test” or just the task name.

gradle -Dtest.single=:proj1:test:Customer build
gradle -DintegTest.single=c/d/ :proj1:integTest
```
The property name depends upon the task name. `integTest` is not some special name. It is just the name of the task in the example.

rotty3000: Got it! What do you propose then?

bjhargrave: First, this is for after 3.4. 

Second, we can modify the BndPlugin for gradle and bnd-testing-maven-plugin for maven to support a property which can be used to set the second argument to `Project.test(File reportDir, List<String> tests)`.

PR #2075, opened 2017-06-29
Fix bugs found by Eclipse Oxygen java compiler

bjhargrave:


pkriens:
LGTM

Just curious, what did the compiler change?

bjhargrave:
> Just curious, what did the compiler change?

The compiler has some new warnings that find strange objects being passed to Collection.remove. All these warnings found bad code which I fixed in this PR.

PR #2081, opened 2017-07-10
Fix #2080: incorrect osgi.native req. from Bundle-NativeCode

### LINKED ISSUE #2080: bnd generates incorrect osgi.native requirement from Bundle-NativeCode header
Given the following Bundle-NativeCode header:

    Bundle-NativeCode: lib/libhttp.so; osname=Linux; processor=x86_64

bnd generates the following requirement filter in an index:

    ...(osgi.native.osname=Linux)(osgi.native.processor=x86_64)...

This is incorrect according to OSGi Core R6 Section 3.10, which asserts that the approximate equality operator should be used, i.e. `(osgi.native.osname~=Linux)`. This will allow for case-insensitive matching.

njbartlett:


njbartlett:
Any objections to merging this for 3.4? It's low impact.

timothyjward:
This is exactly as described in section 8.7 of the OSGi Core specification, so I think we should be doing it.

njbartlett:
Haven't heard any objections... merging.

PR #2091, opened 2017-07-18
Revert "fixes #2088 injecting an interface which extends and replaces a pre-existing `implements` decleration causes a MAJOR change in baseline"

### LINKED ISSUE #2088: injecting an interface which extends and replaces a pre-existing `implements` decleration causes a MAJOR change in baseline
Given the following classes (in the same bundle):
```
interface Foo {}
class Bar implements Foo {}
```
after changing the hierarchy, in a compatible way, like so:
```
interface Foo {}
interface Fum extends Foo {}
class Bar implements Fum {}
```
the baseline operation reports a MAJOR change claiming that Bar no longer implements Foo!

rotty3000:


bjhargrave:
More work is needed on this change. It caused Bnd's build to fail.

```
:biz.aQute.repository:jarC:\projects\bnd\biz.aQute.repository\src\aQute\maven\api\packageinfo:0: error: Baseline mismatch for package aQute.maven.api, MAJOR change. Current is 1.1.0, repo is 1.0.0, suggest 2.0.0 or -
MAJOR                PACKAGE    aQute.maven.api
  MINOR              CLASS      aQute.maven.api.IPom$Dependency
    ADDED            METHOD     bindToVersion(aQute.maven.provider.MavenRepository)
      ADDED          RETURN     void
  MAJOR              CLASS      aQute.maven.api.Revision
    ADDED            IMPLEMENTS java.lang.Comparable
    ADDED            METHOD     compareTo(java.lang.Object)
      ADDED          ACCESS     abstract
      ADDED          RETURN     int
  MINOR              INTERFACE  aQute.maven.api.Release
    ADDED            METHOD     force()
      ADDED          ACCESS     abstract
      ADDED          RETURN     void
    ADDED            ANNOTATED  org.osgi.annotation.versioning.ProviderType
  REMOVED            VERSION    1.0.0
  ADDED              VERSION    1.1.0
```

PR #2092, opened 2017-07-18
fixes #2088 injecting an interface which extends and replaces a pre-existing `implements` decleration causes a MAJOR change in baseline

### LINKED ISSUE #2088: injecting an interface which extends and replaces a pre-existing `implements` decleration causes a MAJOR change in baseline

rotty3000:


rotty3000:
.. was tested with `./rebuild-with-local-plugin`

biz.aQute.bndlib/src/aQute/bnd/differ/JavaElement.java
```diff
@@ -299,9 +299,7 @@ public void implementsInterfaces(TypeRef names[]) throws Exception {
 				Arrays.sort(names); // ignore type reordering
 				for (TypeRef name : names) {
 
-					String comment = null;
-					if (clazz.isInterface() || clazz.isAbstract())
-						comment = inherit(members, name);
+					String comment = inherit(members, name);
 					members.add(new Element(IMPLEMENTS, name.getFQN(), null, MINOR, MAJOR, comment));
```
seanbright: Drop the temporary reference and just pass the result of `inherit()` directly to the `Element` constructor

rotty3000: updated and pushed!

PR #2101, opened 2017-07-26
Analyzer: Sort private references for consistent error messages

### LINKED ISSUE #2100: Private References from exported package message is not stable between VMs
Bnd produces helpful warnings of the form:

    Export org.quartz, has 3, private references [org.quartz.spi, org.quartz.impl.matchers, org.quartz.utils]

This is produced by the Bnd analyser
https://github.com/bndtools/bnd/blob/1992478dc0e7b83dda28b1e73ba711cbd2b822a5/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L342

Unfortunately, because a `HashSet` is used the private references are not sorted by name, and their order changes when using different VMs. This makes it particularly challenging to write a `-fixupmessages:` entry to silence the warning without resorting to wildcards. Elsewhere bnd sorts the output by name to avoid this problem.

seanbright:
Fixes #2100


bjhargrave:
@seanbright  Try rebasing your commit on the new master commit and force pushing.

PR #2114, opened 2017-08-06
BndPomRepository polling for changes

### LINKED ISSUE #2113: Changes to BndPomRepository not reflecting in UI
When using a BndPomRepository in a Bnd-workspace, adding a dependency to the pom does not update the repositories-view. The user has to manually press the "refresh" action in the view to be able to use the new dependency.

The issue has already been discussed: https://groups.google.com/forum/#!topic/bndtools-users/dVRGypk7DS4

lostiniceland:
Fixes #2113

Turns out, the MvnBndRepository implements a polling mechanism where the BndPomRepository does not. 
This is a simple implementation using the lastmodified-attribute from the configured POM files (basically a slightly modified version of the polling used in MavenBndRepository).

Signed-off-by: Marc Schlegel <marc.schlegel@gmx.de>

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
@@ -7,7 +7,7 @@ import java.io.InputStream;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -18,7 +18,12 @@ import java.util.concurrent.atomic.AtomicBoolean;
 
 import org.osgi.resource.Capability;
 import org.osgi.resource.Requirement;
+import org.osgi.util.promise.Deferred;
+import org.osgi.util.promise.Failure;
 import org.osgi.util.promise.Promise;
+import org.osgi.util.promise.Success;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import aQute.bnd.annotation.plugin.BndPlugin;
 import aQute.bnd.build.Workspace;
@@ -35,6 +40,7 @@ import aQute.bnd.service.Registry;
 import aQute.bnd.service.RegistryPlugin;
 import aQute.bnd.service.RepositoryListenerPlugin;
 import aQute.bnd.service.RepositoryPlugin;
+import aQute.bnd.service.url.TaggedData;
 import aQute.bnd.util.repository.DownloadListenerPromise;
 import aQute.bnd.version.Version;
 import aQute.lib.converter.Converter;
@@ -54,27 +60,31 @@ import aQute.service.reporter.Reporter;
 @BndPlugin(name = "PomRepository")
 public class BndPomRepository extends BaseRepository
 		implements Plugin, RegistryPlugin, RepositoryPlugin, Refreshable, Actionable, Closeable {
-	static final String			MAVEN_REPO_LOCAL	= System.getProperty("maven.repo.local", "~/.m2/repository");
+	private final static Logger		logger				= LoggerFactory.getLogger(BndPomRepository.class);
+	static final String				MAVEN_REPO_LOCAL	= System.getProperty("maven.repo.local", "~/.m2/repository");
+	static int						DEFAULT_POLL_TIME	= (int) TimeUnit.MINUTES.toSeconds(5);
 
-	private boolean				inited;
-	private PomConfiguration	configuration;
-	private Registry			registry;
-	private String				name;
-	private Reporter			reporter			= new Slf4jReporter(BndPomRepository.class);
-	private InnerRepository	repoImpl;
-	private List<Revision>		revisions;
-	private BridgeRepository	bridge;
+	private HttpClient				client;
+	private boolean					inited;
+	private PomConfiguration		configuration;
+	private Registry				registry;
+	private String					name;
+	private Reporter				reporter			= new Slf4jReporter(BndPomRepository.class);
+	private InnerRepository			repoImpl;
+	private List<Revision>			revisions;
+	private BridgeRepository		bridge;
 	private Map<URI,Long>			pomFiles;
-	private String				query;
-	private String				queryUrl;
+	private String					query;
+	private String					queryUrl;
 	private ScheduledFuture< ? >	pomPoller;
+
 
 	public synchronized void init() {
 		try {
 			if (inited)
 				return;
 			Workspace workspace = registry.getPlugin(Workspace.class);
-			HttpClient client = registry.getPlugin(HttpClient.class);
+			this.client = registry.getPlugin(HttpClient.class);
 			File localRepo = IO.getFile(configuration.local(MAVEN_REPO_LOCAL));
 			File location = workspace.getFile(getLocation());
 
@@ -99,6 +109,7 @@ public class BndPomRepository extends BaseRepository
 				repository.close();
 				throw new IllegalStateException("We have neither a pom, revision, or query set!");
 			}
+
 			bridge = new BridgeRepository(repoImpl);
 			inited = true;
 		} catch (Exception e) {
@@ -114,39 +125,100 @@ public class BndPomRepository extends BaseRepository
 			return;
 		}
 		final AtomicBoolean busy = new AtomicBoolean();
-		pomPoller = Processor.getScheduledExecutor().scheduleAtFixedRate(new Runnable() {
+		int polltime = configuration.poll_time(DEFAULT_POLL_TIME);
+		if (polltime > 0) {
+			pomPoller = Processor.getScheduledExecutor().scheduleAtFixedRate(new Runnable() {
 
-	@Override
-			public void run() {
-				if (busy.getAndSet(true))
-					return;
-				try {
-					boolean needsUpdate = false;
-					for (Entry<URI,Long> pomFilesEntry : pomFiles.entrySet()) {
-						if (pomFilesEntry.getValue() == null) {
-							// not interested
-							continue;
+				@Override
+				public void run() {
+					if (busy.getAndSet(true))
+						return;
+					try {
+						boolean needsUpdate = false;
+						for (Entry<URI,Long> pomFilesEntry : pomFiles.entrySet()) {
+							final URI pomURI = pomFilesEntry.getKey();
+							if ("file".equalsIgnoreCase(pomURI.getScheme())) {
+								if (pomFilesEntry.getValue() == null) {
+									// not interested
+									continue;
+								}
+								File f = new File(pomURI);
+								if (f.isFile() && f.lastModified() > pomFilesEntry.getValue()) {
+									needsUpdate = true;
+									pomFilesEntry.setValue(f.lastModified());
+								}
+							} else {
+								// check for staleness of non-file-URI
+								if (isStale(pomURI))
+									needsUpdate = true;
+							}
 						}
-						File f = new File(pomFilesEntry.getKey());
-						if (f != null && f.isFile() && f.lastModified() > pomFilesEntry.getValue()) {
-							needsUpdate = true;
-							pomFilesEntry.setValue(f.lastModified());
+						if (needsUpdate) {
+							refresh();
+							for (RepositoryListenerPlugin listener : registry
+									.getPlugins(RepositoryListenerPlugin.class))
+								listener.repositoryRefreshed(BndPomRepository.this);
 						}
-					}
-					if (needsUpdate) {
-						refresh();
-						for (RepositoryListenerPlugin listener : registry.getPlugins(RepositoryListenerPlugin.class))
-							listener.repositoryRefreshed(BndPomRepository.this);
-					}
 
-				} catch (Exception e) {
-					reporter.error("Error when polling for %s for change", this);
-				} finally {
-					busy.set(false);
+					} catch (Exception e) {
+						reporter.exception(e, "Error when polling for %s for change", this);
+					} finally {
+						busy.set(false);
+					}
 				}
-			}
 
-		}, 5000, 5000, TimeUnit.MILLISECONDS);
+			}, 1, polltime, TimeUnit.SECONDS);
+		}
+	}
+
+	/**
+	 * Check any of the URL POMs are stale.
+	 * 
+	 * @param uri remote-URI, must not be "file"
+	 * @return true, when URI resolved to be stale
+	 * @throws Exception InterruptedException if the tread was canceled during
+	 *             waiting
+	 */
+	boolean isStale(final URI uri) throws Exception {
+		final Deferred<Void> freshness = new Deferred<>();
+		Promise<Void> promise;
+		try {
+			Promise<TaggedData> async = client.build().useCache().asTag().async(uri);
+			promise = async.then(new Success<TaggedData,Void>() {
+				@Override
+				public Promise<Void> call(Promise<TaggedData> resolved) throws Exception {
+					switch (resolved.getValue().getState()) {
+						case OTHER :
+							// in the offline case
+							// ignore might be best here
+							logger.debug("Could not verify {}", uri);
+							break;
+
+						case UNMODIFIED :
+							break;
+
+						case NOT_FOUND :
+						case UPDATED :
+						default :
+							logger.debug("Found {} to be stale", uri);
+							freshness.fail(new Exception("stale"));
+					}
+					freshness.resolve(null);
+					return null;
+				}
+			}, new Failure() {
+				@Override
+				public void fail(Promise< ? > resolved) throws Exception {
+					logger.debug("Could not verify {}: {}", uri, resolved.getFailure());
+					freshness.fail(resolved.getFailure());
+				}
+			});
+		} catch (Exception e) {
+			logger.debug("Checking stale status: {}: {}", uri, e);
+		}
+
+		// Block until freshness is resolved
+		return freshness.getPromise().getFailure() != null;
 	}
 
 	@Override
@@ -169,7 +241,7 @@ public class BndPomRepository extends BaseRepository
 
 		if (configuration.pom() != null) {
 			List<String> parts = Strings.split(configuration.pom());
-			pomFiles = new HashMap<>();
+			pomFiles = new LinkedHashMap<>();
 			for (String part : parts) {
 				File f = IO.getFile(part);
 				if (f.isFile()) {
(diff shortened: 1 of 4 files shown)


biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -88,13 +99,57 @@ public synchronized void init() {
...
+				if (busy.getAndSet(true))
+					return;
+				try {
+					boolean needsUpdate = false;
+					for (Entry<URI,Long> pomFilesEntry : pomFiles.entrySet()) {
+						if (pomFilesEntry.getValue() == null) {
+							// not interested
+							continue;
+						}
+						File f = new File(pomFilesEntry.getKey());
```
bjhargrave: The File constructor will throw an exception if the URI is not a file URI. Since non-file URIs are supported, using them should not result in this Runnable constantly reporting errors.

bjhargrave: Also, since non-file URIs are allowed, shouldn't they be checked for staleness? See https://github.com/bndtools/bnd/blob/master/biz.aQute.repository/src/aQute/bnd/repository/osgi/OSGiIndex.java#L134-L188 for how the OSGiRepository checks.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -88,13 +99,57 @@ public synchronized void init() {
...
+					}
+
+				} catch (Exception e) {
+					reporter.error("Error when polling for %s for change", this);
+				} finally {
+					busy.set(false);
+				}
+			}
+
+		}, 5000, 5000, TimeUnit.MILLISECONDS);
```
bjhargrave: Rather than having a fixed poll time and no way to disable polling, the polling time should be configurable. https://github.com/bndtools/bnd/blob/master/biz.aQute.repository/src/aQute/bnd/repository/osgi/OSGiRepository.java#L134-L152

lostiniceland:
I've updated my changes with a test and configuration-option.

@bjhargrave Regarding the URIs: what kind of non-file URIs are supported in the BndPomRepository? I only expected some local pom-files. And how can those become stale?

bjhargrave:
> @bjhargrave Regarding the URIs: what kind of non-file URIs are supported in the BndPomRepository? I only expected some local pom-files. And how can those become stale?

http: for example. One could reference a pom file directly on the web. And since the content of a URL can change, the reference can become stale. 

For example of an http URI, see the test case:
https://github.com/bndtools/bnd/blob/54963f51234b2d894de1ea768129633d67a19a6c/biz.aQute.repository/test/aQute/bnd/repository/maven/pom/provider/PomRepositoryTest.java#L221-L222

lostiniceland:
While trying to implement the stale-check for non-uri-files, I was wondering if this i really necessary for this automatic-update of the repositories. Remote files are probably not changing very often and opening a connection every few seconds to see if it changed will slow down the whole update-experience for the user (think about bad connection or company-proxies)

I tend to ignore URIs which are not "file://" during the update

bjhargrave:
It is already done for the OSGiRepository repo type.

https://github.com/bndtools/bnd/blob/f1aeacda844d28ea717d24c8291e0c023afe468f/biz.aQute.repository/src/aQute/bnd/repository/osgi/OSGiIndex.java#L140-L188

The code avoids downloading the content of the content is unmodified.

Also, you would need a way to disable polling, like a negative poll time, when the use knows the remote URI will never change such as at Maven Central.

https://github.com/bndtools/bnd/blob/f1aeacda844d28ea717d24c8291e0c023afe468f/biz.aQute.repository/src/aQute/bnd/repository/osgi/OSGiRepository.java#L135

lostiniceland:
Since I cannot access the test-results in the CI I have no idea why the build broke. Locally the build completed without errors

bjhargrave:
> Since I cannot access the test-results in the CI I have no idea why the build broke. Locally the build completed without errors

The stack trace is in the log 
https://travis-ci.org/bndtools/bnd/jobs/268532239#L528
https://ci.appveyor.com/project/bndtools/bnd/build/2218/job/kd44670t9126u11d#L250

```aQute.bnd.repository.maven.pom.provider.PomRepositoryTest > testBndPomRepoFilePolling FAILED
    junit.framework.AssertionFailedError: expected:<3> but was:<2>
        at junit.framework.Assert.fail(Assert.java:57)
        at junit.framework.Assert.failNotEquals(Assert.java:329)
        at junit.framework.Assert.assertEquals(Assert.java:78)
        at junit.framework.Assert.assertEquals(Assert.java:234)
        at junit.framework.Assert.assertEquals(Assert.java:241)
        at junit.framework.TestCase.assertEquals(TestCase.java:409)
        at aQute.bnd.repository.maven.pom.provider.PomRepositoryTest.testBndPomRepoFilePolling(PomRepositoryTest.java:291)
```

lostiniceland:
My bad. I used `./gradlew :dist:build` which somehow did not execute all tests.
Anyhow, everything up and running now.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -88,13 +109,119 @@ public synchronized void init() {
...
+						if (needsUpdate) {
+							refresh();
+							for (RepositoryListenerPlugin listener : registry
+									.getPlugins(RepositoryListenerPlugin.class))
+								listener.repositoryRefreshed(BndPomRepository.this);
+						}
+
+					} catch (Exception e) {
+						reporter.error("Error when polling for %s for change", this);
+						e.printStackTrace();
```
bjhargrave: Remove the printStackTrace.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -88,13 +109,119 @@ public synchronized void init() {
...
+						}
+						if (needsUpdate) {
+							refresh();
+							for (RepositoryListenerPlugin listener : registry
+									.getPlugins(RepositoryListenerPlugin.class))
+								listener.repositoryRefreshed(BndPomRepository.this);
+						}
+
+					} catch (Exception e) {
+						reporter.error("Error when polling for %s for change", this);
```
bjhargrave: Use reporter.exception() to include the exception caught.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -115,13 +242,13 @@ public void setProperties(Map<String,String> map) throws Exception {
 
 		if (configuration.pom() != null) {
 			List<String> parts = Strings.split(configuration.pom());
-			pomFiles = new ArrayList<>();
+			pomFiles = new HashMap<>();
```
bjhargrave: Use a LinkedHashMap to preserve iteration order. This is important.

biz.aQute.repository/test/aQute/bnd/repository/maven/pom/provider/PomRepositoryTest.java
```diff
@@ -259,6 +264,58 @@ public void testBndPomRepoRefresh() throws Exception {
...
+				"testdata/pomrepo/simple.xml"
+						+ ",https://repo1.maven.org/maven2/org/apache/felix/org.apache.felix.gogo.shell/0.12.0/org.apache.felix.gogo.shell-0.12.0.pom");
+		config.put("snapshotUrls", "https://repo1.maven.org/maven2/");
+		config.put("releaseUrls", "https://repo1.maven.org/maven2/");
+		config.put("name", "test");
+		config.put("pollTime", "100");
+		bpr.setProperties(config);
+
+		List<String> list = bpr.list(null);
+		assertNotNull(list);
```
bjhargrave: Since the test case can abnormally end here, the try/finally block which closes the BndPomRepository should wrap the whole test logic. That is, always close the BndPomRepository.

biz.aQute.repository/test/aQute/bnd/repository/maven/pom/provider/PomRepositoryTest.java
```diff
@@ -259,6 +264,58 @@ public void testBndPomRepoRefresh() throws Exception {
...
+			for (String line : lines) {
+				if (line.contains("</dependencies>")) {
+					sb.append(
+							"<dependency><groupId>org.slf4j</groupId><artifactId>slf4j-api</artifactId><version>1.7.25</version></dependency>");
+				}
+				sb.append(line);
+			}
+			Files.write(path, sb.toString().getBytes(StandardCharsets.UTF_8), StandardOpenOption.WRITE,
+					StandardOpenOption.TRUNCATE_EXISTING);
+			Thread.sleep(500); // might be necessary to increase this on slow machines (or busy CI)
```
bjhargrave: Or wrap in a retry loop. Say 3 times with a 500 ms wait each time before declaring failure.

biz.aQute.repository/test/aQute/bnd/repository/maven/pom/provider/PomRepositoryTest.java
```diff
@@ -259,6 +264,58 @@ public void testBndPomRepoRefresh() throws Exception {
...
+		try {
+			StringBuilder sb = new StringBuilder();
+			for (String line : lines) {
+				if (line.contains("</dependencies>")) {
+					sb.append(
+							"<dependency><groupId>org.slf4j</groupId><artifactId>slf4j-api</artifactId><version>1.7.25</version></dependency>");
+				}
+				sb.append(line);
+			}
+			Files.write(path, sb.toString().getBytes(StandardCharsets.UTF_8), StandardOpenOption.WRITE,
```
bjhargrave: You should not write over the original file. If the test needs to change the file, first copy to a temp area and the mutate the temp file as needed for the test.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
lostiniceland: Out of curiosity: why is the iteration order important for Bnd?

bjhargrave: For consistent across builds across different systems. You are replacing a List, which is ordered, by a HashMap whose keySet is this unordered. This is a source of entropy we need to avoid. Thus you will see that Bnd almost always uses LinkedHashMap in preference to just HashMap to provide a consistent iteration order.

lostiniceland:
One last problem: PomRepositoryTest has side-effects on the new test. When running only testBndPomRepoFilePolling everything is fine. But when the whole test-class runs, the new test fails due to an not-updated Repo.

I can only assume that the VM caches something from the previous test-methods. For example, the other tests do not close the Repo.

biz.aQute.repository/test/aQute/bnd/repository/maven/pom/provider/PomRepositoryTest.java
```diff
@@ -259,6 +266,71 @@ public void testBndPomRepoRefresh() throws Exception {
...
+
+		BndPomRepository bpr = new BndPomRepository();
+		Workspace w = Workspace.createStandaloneWorkspace(new Processor(), tmp.toURI());
+		w.setBase(tmp);
+		bpr.setRegistry(w);
+
+		Map<String,String> config = new HashMap<>();
+		config.put("pom",
+				path.toString()
+						+ ",https://repo1.maven.org/maven2/org/apache/felix/org.apache.felix.gogo.shell/0.12.0/org.apache.felix.gogo.shell-0.12.0.pom");
```
bjhargrave: Why do you include gogo.shell in here? Doesn't seem to be relevant.

biz.aQute.repository/test/aQute/bnd/repository/maven/pom/provider/PomRepositoryTest.java
```diff
@@ -259,6 +266,71 @@ public void testBndPomRepoRefresh() throws Exception {
...
+		bpr.setProperties(config);
+
+		try {
+			List<String> list = bpr.list(null);
+			assertNotNull(list);
+			assertEquals(1, list.size()); // this seems to be a bit flaky (the dependencies change from time to time)
+
+			List<String> lines = Files.readAllLines(path, StandardCharsets.UTF_8);
+
+			StringBuilder sb = new StringBuilder();
```
bjhargrave: It would be cleaner to have another file in the testdata folder and then copy it over the output file created above rather then messing with trying to edit the file.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/PomConfiguration.java
```diff
@@ -62,4 +62,10 @@
 	 */
 	boolean transitive(boolean deflt);
 
+	/**
+	 * Sets the time in milliseconds when to check for changes in the pom-files
+	 */
+	// default: 5000ms
+	int pollTime(int pollTimeInSecs);
```
bjhargrave: Can we call this 

https://github.com/bndtools/bnd/blob/5d8500fd174eb7ec4228f383baba910d4d5573f8/biz.aQute.repository/src/aQute/bnd/repository/osgi/OSGiRepository.java#L67

Then we use the same property name `poll.time` as in other repos.

biz.aQute.repository/test/aQute/bnd/repository/maven/pom/provider/PomRepositoryTest.java
```diff
@@ -259,6 +266,71 @@ public void testBndPomRepoRefresh() throws Exception {
...
+		bpr.setRegistry(w);
+
+		Map<String,String> config = new HashMap<>();
+		config.put("pom",
+				path.toString()
+						+ ",https://repo1.maven.org/maven2/org/apache/felix/org.apache.felix.gogo.shell/0.12.0/org.apache.felix.gogo.shell-0.12.0.pom");
+		config.put("snapshotUrls", "https://repo1.maven.org/maven2/");
+		config.put("releaseUrls", "https://repo1.maven.org/maven2/");
+		config.put("name", "test");
+		config.put("pollTime", "300");
```
bjhargrave: This is now `"poll.time`. The property is described as poll time in seconds. 

See the following example of poll testing, It uses a plugin to detect that refresh has been called. You will need to add the RepositoryListenerPlugin to the workspace since the workspace is set as the registry for bpr.

 https://github.com/bndtools/bnd/blob/5d8500fd174eb7ec4228f383baba910d4d5573f8/biz.aQute.repository/test/aQute/bnd/repository/osgi/OSGiRepositoryTest.java#L125-L192

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -88,13 +109,118 @@ public synchronized void init() {
...
+						}
+
+					} catch (Exception e) {
+						reporter.exception(e, "Error when polling for %s for change", this);
+					} finally {
+						busy.set(false);
+					}
+				}
+
+			}, polltime, polltime, TimeUnit.MILLISECONDS);
```
bjhargrave: This should be `TimeUnit.SECONDS` since that is what we use for other repos. It is not useful to configure for smaller than second increments in the normal case.

https://github.com/bndtools/bnd/blob/5d8500fd174eb7ec4228f383baba910d4d5573f8/biz.aQute.repository/src/aQute/bnd/repository/osgi/OSGiRepository.java#L152

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -44,27 +59,32 @@
...
-	private InnerRepository	repoImpl;
-	private List<Revision>		revisions;
-	private BridgeRepository	bridge;
-	private List<URI>			pomFiles;
-	private String				query;
-	private String				queryUrl;
+		implements Plugin, RegistryPlugin, RepositoryPlugin, Refreshable, Actionable, Closeable {
+	private final static Logger		logger				= LoggerFactory.getLogger(BndPomRepository.class);
+	static final String				MAVEN_REPO_LOCAL	= System.getProperty("maven.repo.local", "~/.m2/repository");
+	static int						DEFAULT_POLL_TIME	= (int) TimeUnit.SECONDS.toMillis(5);
```
bjhargrave: This should be in seconds.

https://github.com/bndtools/bnd/blob/5d8500fd174eb7ec4228f383baba910d4d5573f8/biz.aQute.repository/src/aQute/bnd/repository/osgi/OSGiRepository.java#L53

And a default time of 5 seconds is way too short. Use 5min like in OSGiRepository.

biz.aQute.repository/test/aQute/bnd/repository/maven/pom/provider/PomRepositoryTest.java
lostiniceland: Just c&p from another test to have some remote uri

lostiniceland:
@bjhargrave I will include your review comments.

Though I am having other problems with this test (for weeks now) which I cannot get rid of. When running the test in isolation, it passes. Running all tests, causes it to fail. 

I've identified the isStale to be the cause (removing the check makes all tests pass), but I dont see anything wrong with it. 
Could it be that the HttpClient provided by Bnd is locking a Socket from a previous test? I am out of ideas...for example I've tried adding a finally-block to all tests which closes the repository afterwards, but this didn't help either. But there must be some state in the JVM which causes this behaviour

bjhargrave:
> I've identified the isStale to be the cause (removing the check makes all tests pass), but I dont see anything wrong with it.

That does not mean it is the cause. Removing my alleviate the problem without it being the cause.

> Could it be that the HttpClient provided by Bnd is locking a Socket from a previous test? I am out of ideas...for example I've tried adding a finally-block to all tests which closes the repository afterwards, but this didn't help either. But there must be some state in the JVM which causes this behaviour

I doubt this. HttpClient is used all over my many other parts of Bnd so it seem unlikely it is broken for your new function. But it is possible.

Once you get all of the other requested changes made, I can take a look at the problem and see if I can help identify it.

bjhargrave:
Also, you will need to rebase your work onto the latest master and resolve any conflicts. I made some minor changes there recently.

lostiniceland:
The failing builds were caused by the test being to fast. A repository-refresh can be quite slow (on my current machine 2-3 seconds), hence I had to raise the wait and retry-loop in the test.

PR #2116, opened 2017-08-08
Avoid NPE when ID capability lacks version attrib

njbartlett:


biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
```diff
@@ -546,6 +546,8 @@ public static VersionedClause toVersionClause(Resource resource, String mask) {
 		if (resource.getCapabilities(WORKSPACE_NAMESPACE).isEmpty()) {
 			Macro macro = new Macro(new Processor());
 			Version version = getVersion(idCap);
+			if (version == null)
+				version = Version.emptyVersion;
```
bjhargrave: I am not sure this is the right answer. I think we should test for null and throw an exception indicating malformed osgi.identity capability. Just replacing with 0 seems wrong.

njbartlett: It's a common practice everywhere else in OSGi (missing Bundle-Version in manifest, missing version attrib on export, etc, all imply 0.0.0).

bjhargrave: True. But they are specified to default to 0 when absent. The osgi.identity capability is specified that the version attribute is mandatory. The type attribute is also mandatory. If the indexed bundle has no Bundle-Version header, then the index generator must write out the osgi.identity version attribute with a 0 value.

pkriens: I think I agree with BJ. The right solutions is to validate the resources before they are used like this. I think we need a method in ResourceUtils that validates all the capabilities and requirements in the resource, namespace specific?

bjhargrave: Also, I will note that `XMLResourceGenerator` is partly to blame for this mess. It simply writes out what it has in the memory Resource objects without regard to the validity of the document to the schema. So if the osgi.identity Resource object does not have a version attribute, `XMLResourceGenerator` does not write out `version:Version=0` as required by the schema.

njbartlett:
@pkriens @bjhargrave Could you take another look at this? I have added code on the generation side to ensure that mandatory attributes on the capabilities are always generated. This includes some that we have always missed such as `bundle-version` on `osgi.wiring.package`.

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java
```diff
@@ -509,22 +513,44 @@ public Capability addProvideCapability(String namespace, Attrs attrs) throws Exc
 		return addCapability0(capb);
 	}
 
+	/**
+	 * Adds a set of package exports assuming they are from the system bundle.
+     *
+	 * @deprecated Use {@link #addExportPackages(Parameters, String, String)} in preference.
+	 */
+	@Deprecated
```
bjhargrave: Why add a new method that is deprecated. That does not make sense.

biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java
```diff
@@ -187,7 +187,7 @@ public synchronized void init() {
 
 				Parameters systemPackages = new Parameters(properties.mergeProperties(Constants.RUNSYSTEMPACKAGES),
 						project);
-				system.addExportPackages(systemPackages);
+				system.addSystemBundleExportPackages(systemPackages);
```
bjhargrave: This is the new deprecated method. Use a non-deprecated method here.

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/CapabilityValidation.java
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	private static class Directive extends Attribute<String> {
+		Directive(String namespace, String name, Function<Object, String> check) {
+			super(namespace, name, check);
+		}
+	}
+
+	private final Map<String, List<Attribute<?>>>	validationAttributes	= new HashMap<>();
+	private final Map<String, List<Directive>>		validationDirectives	= new HashMap<>();
```
bjhargrave: I would have thought these should be static final unmodifiable. But you build these maps for each CapabilityValidation object. Couldn't they just be made once?

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java
njbartlett: The intention was to create a method that can be used in contexts where the bundle identity is not known, to avoid a more significant refactoring.

biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java
njbartlett: It's not clear what non-deprecated method I can invoke. I don't know the bundle identity/version. I can just default to `system.bundle`/empty here but then that defaulting is duplicated here and in `ResourceBuilder`.

njbartlett:
I just noticed that I already bumped the package version to 3.0.0, which gives me the ability to just remove the deprecated methods. `BndrunResolveContext` will have to either work out or guess the bundle identities. If you prefer me not to bump the MAJOR version on that package then we need to work out a better way of deprecating the relevant methods.

bjhargrave:
I don't have a problem with deprecating methods. But it seems wrong to add a new, deprecated method.

njbartlett:
I have modified the commit, now I just deleted the bad methods and bumped the MAJOR version. Please take a look.

If you prefer not to bump the MAJOR then I don't see a completely clean way to achieve this. There will have to be some duplication I guess. Let me know which route you prefer. The commit history of this branch is a bit of a mess now so I will rebase and squash the commits before merging.

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/CapabilityValidation.java
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	private static class Directive extends Attribute<String> {
+		Directive(String namespace, String name, Function<Object, String> check) {
+			super(namespace, name, check);
+		}
+	}
+
+	private static final Map<String, List<Attribute<?>>>	VALIDATION_ATTRIBUTES	= new HashMap<>();
+	private static final Map<String, List<Directive>>		VALIDATION_DIRECTIVES	= new HashMap<>();
```
bjhargrave: I guess this is in preparation for adding validation rules for directives at some point?

rotty3000:
wondering if this should get merged before it's overly stale!

bjhargrave:
Closing as an old PR which is out of date.

PR #2118, opened 2017-08-10
utf8: Avoid sharing CharsetDecoders between threads

bjhargrave:
CharsetDecoders are not thread-safe. We use thread locals to avoid
always having to create new decoders for each invocation.

Reported by @rotty3000.


seanbright:
👍

PR #2120, opened 2017-08-17
Launcher with keep

pkriens:
If -runkeep is specified then restarting an executable
JAR should not update bundles that have not been changed.
Also, bundles should be updated before the framework
is started so that the changes are more atomic. Before
this patch there was a lot of bouncing start/stop because bundles
were updated even if they were identical.

This patch changes the starting order. It will update the
bundles (update/install/remove) before the framework
is started. It will also not update a bundle when it is
already installed.

To be installed is defined as:

* The uber jar has a SHA for the bundle
* In the system bundle data file there is a
  file .bnd.launcher/<bundle-id> that contains the sha
* The sha’s match

To test this, there is also now a magic exit code 197 that 
will not exit the launcher with System.exit() but it
just returns.


rotty3000:
:+1: sounds very nice!

rotty3000:
Can the value stored value be overridden with a system property?

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -438,7 +451,7 @@ private int start(BundleContext systemContext, int result, BundleActivator activ
 	 * 
 	 * @param begin
 	 */
-	void update(long before) throws Exception {
+	List<Bundle> update(long before) throws Exception {
```
bjhargrave: Calling this method `update` is confusing. It should be called `start` since it is all about starting the specified list of bundle which may have already been updated.

pkriens:
Not sure what value you mean?

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -650,22 +671,80 @@ private String toNativePath(String s) {
...
+			return true;
+
+		File shafile = shafile(bundle);
+
+		if (shafile == null || !shafile.isFile() || !shafile.canRead())
+			return true;
+
+		try {
+			String storedSha = IO.collect(shafile);
+			if (storedSha == null || !storedSha.equals(sha))
```
bjhargrave: Shouldn't this use `equalsIgnoreCase` to be safe? Since you are using the manifest header value and its case is not defined.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -1173,6 +1251,7 @@ public boolean implies(Permission permission) {
 	@SuppressWarnings("unchecked")
 	public synchronized void serviceChanged(ServiceEvent event) {
 		if (event.getType() == ServiceEvent.REGISTERED) {
+			trace("service even " + event);
```
bjhargrave: The trace text is misspelled.

rotty3000:
@pkriens the value of runkeep! Since I envision cases where, when this value is true in the executable, I wish to run the executable with it set to false using like a system property.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
pkriens: I wanted to mimimise changes ...

biz.aQute.launcher/src/aQute/launcher/Launcher.java
pkriens: yup, you're right though they are basically opaque since they always come from the manifest. But if they switch tools it might change

pkriens:
@rotty3000 that is already support just `java -Dlaunch.keep=true|false -jar exec.jar`. You can override all run properties and any parameter in LauncherConstants.

rotty3000:
@pkriens :+1:

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -447,18 +456,11 @@ void update(long before) throws Exception {
...
-			// Will be reset by the Framework listener we added
-			// when we created the framework.
-			while (inrefresh)
-				Thread.sleep(100);
+		return tobestarted;
+	}
 
-		} else
-			trace("cannot refresh the bundles because there is no Package Admin");
+	void update(List<Bundle> tobestarted) throws Exception {
```
bjhargrave: This method should be named `start` since it does not update.

biz.aQute.remote/test/biz/aQute/remote/AgentTest.java
```diff
@@ -185,10 +185,11 @@ public void testAgentUpdateBundleFromURL() throws Exception {
 	}
 
 	public void testAgentShell() throws Exception {
-		String result = supervisor.getAgent().shell("lb");
-		assertNotNull(result);
-		assertTrue(result.contains("START"));
-		assertTrue(result.contains("LEVEL"));
+		// This test requires a gogo to be added to the framework
```
bjhargrave: If this method is not useful, comment out or remove then entire method. Having an empty test method inflates the test numbers.

bjhargrave:
@pkriens why is the test case in bnd and not bndlib.tests where there are already launcher tests?

pkriens:
Same as for the JUnitFramework. I need to have access to the bnd build environment repository plugins.

bjhargrave:
So add the proper projects to the -testpath.

pkriens:
Yeah, and then of course wait forever until Eclipse resolves the cycle?

PR #2121, opened 2017-08-18
[junit] An OSGi Framework to be used in JUnit classes

pkriens:
Added a surprisingly small class that makes it
possible to do standard JUnit tests with a full
blown framework. 

The framework runs in the JUnit process and exports
all packages in the test directory + what the test directory
imports. You can add bundles just like you can in
a bnd(run) file. 

There is some simple support to build temporary bundles.
This is very nice because you can use classes from the
test directory, no jarring required.

I really need this many projects. I started to copy
it between projects but it is always a big pain
in the ass to get all the required dependencies right
because it requires the full bnd + repositories used
in the build.


biz.aQute.bndlib/src/aQute/bnd/junit/JUnitFramework.java
```diff
@@ -0,0 +1,331 @@
...
+			Project p = getProject();
+			File bin_test = p.getTestOutput();
+			this.bin_test = new Jar(bin_test);
+
+			String extra = getExtra();
+
+			Map<String,String> props = new HashMap<>();
+			props.put(Constants.FRAMEWORK_SYSTEMPACKAGES_EXTRA, extra);
+
+			File storage = IO.getFile("generated/fw");
```
bjhargrave: This is a fixed folder name in the current dir. Seems like a bad idea. Since you have the project, you should make a folder in the target folder.

biz.aQute.bnd/test/aQute/bnd/junit/JUnitFrameworkTest.java
```diff
@@ -0,0 +1,150 @@
+package aQute.bnd.junit;
```
bjhargrave: If the new class is in bndlib, why is the test for it in bnd? That seems wrong. It should be in bndlib.tests

biz.aQute.bnd/test/aQute/bnd/junit/JUnitFrameworkTest.java
```diff
@@ -0,0 +1,150 @@
...
+import org.osgi.service.startlevel.StartLevel;
+import org.osgi.util.tracker.BundleTracker;
+
+import aQute.bnd.junit.JUnitFramework.BundleBuilder;
+import aQute.bnd.osgi.EmbeddedResource;
+
+@SuppressWarnings("deprecation")
+public class JUnitFrameworkTest extends Assert {
+	static {
+		System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", "warn");
```
bjhargrave: This is a wide side effect for all tests in this VM.

biz.aQute.bnd/test/aQute/bnd/junit/JUnitFrameworkTest.java
```diff
@@ -0,0 +1,150 @@
...
+	}
+	static JUnitFramework framework = new JUnitFramework();
```
bjhargrave: Why is this static and not per test method to avoid side effects between test methods?

biz.aQute.bndlib/src/aQute/bnd/junit/JUnitFramework.java
pkriens: makes sense, I already did that for the test bin folder.

biz.aQute.bnd/test/aQute/bnd/junit/JUnitFrameworkTest.java
pkriens: Yeah, that sucks a bit but I do not see another way. I need to have all the repositories on the classpath + the embedded-repo. This creates a cycle. Although the code does not depend on it, our build depends on it. 

I would be nice if we had a single project that created the uber bnd jar that was then shared with bnd and gradle, and maybe bndtools.

biz.aQute.bnd/test/aQute/bnd/junit/JUnitFrameworkTest.java
pkriens: Alternative?

biz.aQute.bnd/test/aQute/bnd/junit/JUnitFrameworkTest.java
pkriens: Because creating this sucker is not cheap. You need to analyse the classpath of the project. So most tests can be written to uninstall bundles and make sure they are not in each other's way. You can of course do it instance based but you just wait a lot longer.

biz.aQute.bndlib/src/aQute/bnd/junit/JUnitFramework.java
```diff
@@ -0,0 +1,331 @@
...
+		executor.shutdownNow();
+		bin_test.close();
+	}
+
+	public BundleContext getBundleContext() {
+		return context;
+	}
+
+	public <T> List<T> getServices(Class<T> class1) throws InvalidSyntaxException {
+		BundleContext context = framework.getBundleContext();
```
rotty3000: don't you have the context already?

pkriens: Yup, but I like a fresh copy :-)

PR #2134, opened 2017-08-31
Add a new "-readme" bnd instruction to keep in sync readme files 

### LINKED ISSUE #2123: Avoid duplicate information between the readme file and the bnd file
README.md files contains a lot of information that are also present in the bundle manifest such as the bundle name, the description, the developer contacts, licences...

In order to keep this information in sync, the enroute project puts some variables in the readme file : 

> \# ${Bundle-Name}
${Bundle-Description}

This works and the jar contains a proper readme file when include with brackets, but the variables remain unchanged when we access the readme file from GitHub.

I do the reverse by using html tags and bnd macros to extract the manifest headers value, for exemple: 

> \# \<bnd-name\>Great Bundle\</bnd-name\> --> Bundle-Name: Great Bundle

This works but users must know the html tags that I used and the macros are a little complicated. 

A new bnd instruction such as **-readme: true**  could be more useful. This instruction could automatically create/update the readme file at the root of the project/workspace based on the manifest headers and include the readme in the jar at the same time.

Is this feature could be useful ?

cdecode:
Closes   #2123

Add a new bnd instruction to keep in sync the manifest headers values, the component declarations and the metatype declarations with the readme file of a project.

This instruction allows a user to automatically generate /update a markdown file with the information in the manifest. If the bundle provide a declarative service, It can also generate a documentation for it (configuration, lifecyle....)

The core implementation provides 5 public classes:

* ReadmeConfiguration: Used to configure the readme generation.
* ReadmeInformation: Used to represent input data for the readme. 
* ReadmeUpdater: Take an out-of-date readme file content, a current configuration and the data input and return an updated readme file content.
* ReadmeProjectOption: Helper to build the configuration and the data input based on the instruction and the jar.
* ReadmeWorkspaceOption: Helper to build the configuration and the data input based on the instruction and the workspace configuration (not yet implemented).

For the project readme file generation, I have added a doReadme method in the Builder class (like the -pom instruction is implemented, sorry I didn't find a plugin interface that meets the requirements)

Also, I have wrote a documentation of this instruction for further description : [here](https://github.com/cdecode/bnd/blob/2123-readme/docs/_instructions/readme.md)

I am aware of this:
> We're trying very hard to keep Bnd lean and focused. We don't want it to do everything for everybody. This means that we might decide against incorporating a new feature. However, there might be a way to implement that feature on top of Bnd.

If this feature is not expected in bnd, could you give me some hints to integrate this in another way ?! Thanks

Signed-off-by: Clement Delgrange cde@cdecode.org


bjhargrave:
This is a big PR. I will review but it may be a little while til I have time for such a large PR.

Thanks for contributing!

cdecode:
For the review you can find on this repo [https://github.com/cdecode/org.cdecode.firebase](https://github.com/cdecode/org.cdecode.firebase) what the instruction does on a real project.
The instruction is set globally in the build.bnd

PR #2139, opened 2017-09-12
Metadata Generator

### LINKED ISSUE #2138: Generate a file that reports the bundle metadata
This issue derives from the issue #2123 

The goal is to generate a file that resume some metadata extracted from a jar. The metadata could be the manifest headers, the component declarations, the metatype definition, etc.

The preferred format for this file is Json, and is intended to facilitate the access to the metadata of a jar to any processes that need them. So, the generator should not preprocess too much the metadata to keep them generic but can improve the usability by computing default values where possible, preprocessing the localization, etc...

cdecode:
closes #2138 

Contains the metadata generator.

### Packages

* aQute.bnd.metadata.dto:

 Contains a set of DTO that represents the OSGI Alliance manifest headers Core 6, the Bundle-Developers header, the component declarations and the metatype definitions.

 All this DTO expect the default values to be set. All attributes and directives handled by the framework can be accessed directly. The localizations are computed and inserted in DTO that support localization (see DTOUtil). 

* aQute.bnd.metadata: public class
  * MetadataAnalyzer: Take a jar and return the metadata as DTO
  * MetadataJsonIO: Can convert a metadata Json file to metadata DTO and vise versa
  * DTOUtil: can create a deep copy of any DTO (from enroute DTOs implementation) and can translate any localizable DTO to a specific locale (following the OSGI rules). 

* aQute.bnd.metadata: private class

Contains a set of extractor that read the jar and convert to DTO and verify DTO spec. The code is divided by "things to extract" (headers, components, ...) and not by tasks (read jar, verify) so it contains a lot of class.


cdecode:
As for now this PR does not contain a command and is not integrated in the build process, I need to know how you want to do this before. You can see a json result at this location : [https://github.com/cdecode/bnd/blob/2138-metadata/biz.aQute.bndlib.tests/testresources/metadata/bundleResult.json](https://github.com/cdecode/bnd/blob/2138-metadata/biz.aQute.bndlib.tests/testresources/metadata/bundleResult.json)

cdecode:
Ok, so finally I added a command in bnd, you can test on a real jar with:

`bnd metadata -e org.mybundle.jar`

this will generate the json file org.mybundle.json in the current directory. To just print, don't use *-e*

PR #2143, opened 2017-09-19
[reporter] Build structure for reporter.

pkriens:
This implements a command. In the bnd root dir
try:

  java -jar biz.aQute.bnd/generated/biz.aQute.bnd.jar report generate -t biz.aQute.bnd.reporter/xslt/simple.xslt biz.aQute.bnd/generated/biz.aQute.bnd.j
ar

It also implements a ReportGenerator skeleton and it 
will participate in a build when the ReportPostBuildPlugin is added to the
set of plugins (this should happen in Gradle later).

There is also a simplistic testcase.


cdecode:
I think you can close this PR, this was a base for #2169

PR #2145, opened 2017-09-22
[gogo] A diagnostics bundle

pkriens:


rotty3000:
Cool! We just finished writing something like this.

PR #2147, opened 2017-09-27
Fix bndtools#1781: Cannot select JavaSE-9 as EE

njbartlett:


njbartlett:
Fixes https://github.com/bndtools/bndtools/issues/1781

bjhargrave:
There some changes missed. I am investigating...

PR #2164, opened 2017-10-12
[1769] Update cnf/cache/p2/index.xml.gz on any URL change.

jwausle:
Fixes https://github.com/bndtools/bndtools/issues/1769

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
@@ -9,6 +9,7 @@ import java.io.IOException;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


biz.aQute.repository/test/aQute/bnd/repository/p2/provider/P2IndexerTest.java
```diff
@@ -146,4 +155,37 @@ public void testRefresh() throws Exception {
...
+				.contains("http://starts-not-with-host/plugins/dummy_1.0.0.jar"));
+		byte[] md5 = md5(invalidCache.resolve("index.xml.gz"));
+
+		// new P2Indexer force a (re)creation of the index.xml.gz if URL check fails.
+		try (P2Indexer indexer = new P2Indexer(new Slf4jReporter(), invalidCache.toFile(), new HttpClient(),
+				Paths.get("http://starts-not-with-host/plugins/").toUri(), "invalid")) {}
+		byte[] md5AfterP2IndexerInit = md5(invalidCache.resolve("index.xml.gz"));
+
+		Assert.assertFalse(!Arrays.areEqual(md5, md5AfterP2IndexerInit));
+	}
```
bjhargrave: I don't see the test method ever reference `index-p2-starts-with-host.xml`?

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -77,16 +78,49 @@
...
 			r = readRepository();
 			save(r);
+		} else {
+			r = readFile();
 		}
 
 		bridge = new BridgeRepository(r);
 	}
 
+	private boolean startsAnyResourceUrlWithPrefix(URI prefix) throws Exception {
```
bjhargrave: This seems like a very convoluted way of deciding the URI changed. Since `name` is written into the saved index file

https://github.com/bndtools/bnd/blob/5d8500fd174eb7ec4228f383baba910d4d5573f8/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java#L227

We should use a name based upon the URI 

https://github.com/bndtools/bnd/blob/5d8500fd174eb7ec4228f383baba910d4d5573f8/biz.aQute.bndlib/src/aQute/bnd/http/URLCache.java#L149-L151

as the name attribute of the repository element, we just need to see if the name based upon the current URI matches the name in the xml. But to do this, we will need to modify XMLResourceParser since it does not currently expose the name (and increment) attribute of the top level resource element.

With this change, the check for changed URI  become easy.

bjhargrave:
I have made an alternate PR for this. #2165 

So if your are ok with #2165, you can either close this or change it to be a new test case.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
bjhargrave: I made the above suggested fix in PR #2165

jwausle:
@bjhargrave I´m fine with Your fix #2165. 

- I checked successful the change with my build 👍

biz.aQute.repository/test/aQute/bnd/repository/p2/provider/P2IndexerTest.java
```diff
@@ -1,14 +1,23 @@
...
+import java.nio.file.Paths;
+import java.security.DigestInputStream;
+import java.security.MessageDigest;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Set;
 import java.util.concurrent.Semaphore;
 import java.util.concurrent.atomic.AtomicReference;
 
+import org.bouncycastle.util.Arrays;
```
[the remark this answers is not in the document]
jwausle: Right that`s was mistake.

biz.aQute.repository/test/aQute/bnd/repository/p2/provider/P2IndexerTest.java
```diff
@@ -146,4 +155,37 @@ public void testRefresh() throws Exception {
 			assertEquals(1, p2.versions("name.njbartlett.eclipse.macbadge").size());
 		}
 	}
+
+	@SuppressWarnings("restriction")
+	@Test
```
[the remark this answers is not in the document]
jwausle: Thank You. For the future :)

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
jwausle: More elegant way. #PR 2165 is very helpful for that.

PR #2183, opened 2017-10-25
Support the path to local Maven repository to be a symbolic link

manandbytes:
When path to the local Maven repository is a symlink:

    file ~/.m2/repository

returns

    /home/user/.m2/repository: symbolic link to /home/user/.cache/m2/repository/

MavenRepository fails with:

    java.nio.file.FileAlreadyExistsException: /home/user/.m2/repository
	at sun.nio.fs.UnixException.translateToIOException(UnixException.java:88)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:102)
	at sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:107)
	at sun.nio.fs.UnixFileSystemProvider.createDirectory(UnixFileSystemProvider.java:384)
	at java.nio.file.Files.createDirectory(Files.java:674)
	at java.nio.file.Files.createAndCheckIsDirectory(Files.java:781)
	at java.nio.file.Files.createDirectories(Files.java:727)
	at aQute.lib.io.IO.mkdirs(IO.java:753)
	at aQute.lib.io.IO.mkdirs(IO.java:749)
	at aQute.maven.provider.MavenRepository.<init>(MavenRepository.java:61)
	at aQute.bnd.repository.maven.provider.MavenBndRepository.init(MavenBndRepository.java:494)

This happens because Java NIO Files' createDirectory method explicitly
doesn't allow the directory itself to be a symbolic link:

    try {
	createDirectory(dir, attrs);
    } catch (FileAlreadyExistsException x) {
	if (!isDirectory(dir, LinkOption.NOFOLLOW_LINKS))
	    throw x;
    }

At the same time, it is OK when the path includes symbolic link.

Add tests for cases when a) location is a symbolic link and b) one of
the location's parents is a symbolic link. Having these tests in
MavenBndRepository makes them kind of integration tests but this way
they reproduce the actual problem as close as possible.

Handle this case in MavenRepository instead of IO's mkdirs method
(which is riskier as it has has a lot more usages) by creating the
directory using target directory that link points to and a link
itself.

Signed-off-by: Mykola Nikishov <mn@mn.com.ua>

### DIFF
--- a/aQute.libg/src/aQute/lib/io/IO.java
+++ b/aQute.libg/src/aQute/lib/io/IO.java
@@ -750,7 +750,11 @@ public class IO {
 	}
 
 	public static void mkdirs(Path dir) throws IOException {
-		Files.createDirectories(dir);
+		if (Files.isSymbolicLink(dir)) {
+			mkdirs(Files.readSymbolicLink(dir));
+		} else {
+			Files.createDirectories(dir);
+		}
 	}
 
 	public static long drain(InputStream in) throws IOException {
--- a/aQute.libg/test/aQute/lib/io/IOTest.java
+++ b/aQute.libg/test/aQute/lib/io/IOTest.java
@@ -1,9 +1,12 @@
 package aQute.lib.io;
 
 import java.io.File;
+import java.io.IOException;
 import java.io.InputStream;
 import java.nio.ByteBuffer;
 import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
 
 import junit.framework.TestCase;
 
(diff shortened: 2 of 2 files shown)


biz.aQute.repository/src/aQute/maven/provider/MavenRepository.java
```diff
@@ -58,7 +59,10 @@ public MavenRepository(File base, String id, List<MavenBackingRepository> releas
 
 		this.executor = new PromiseExecutor(executor);
 		this.localOnly = this.release.isEmpty() && this.snapshot.isEmpty();
-		IO.mkdirs(base);
+		if (Files.isSymbolicLink(this.base.toPath()))
```
bjhargrave: I would suggest that instead we change `IO.mkdirs` to:

```java
	public static void mkdirs(Path dir) throws IOException {
		if (Files.isSymbolicLink(dir)) {
			mkdirs(Files.readSymbolicLink(dir));
		} else {
			Files.createDirectories(dir);
		}
	}
```
 Which correctly handles symlinks which point to symlinks which point to a real folder.

Putting the fix in IO means everyone benefits. The tests should move to the aQute.lib project too.

manandbytes: Not sure about the best way to address your comment. I'm going to push second commit on top of the original one. Is it OK? Or is it better to amend an existing commit?

bjhargrave: Just amend your commit and push force as we discussed at EclipseCon.

manandbytes:
@bjhargrave Ping

bjhargrave:
Thanks!

PR #2184, opened 2017-10-26
URLResource can hold jar files open after being closed

timothyjward:

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/URLResource.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/URLResource.java
@@ -4,6 +4,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.net.JarURLConnection;
 import java.net.URL;
 import java.net.URLConnection;
 import java.nio.ByteBuffer;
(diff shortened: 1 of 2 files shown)


timothyjward:
Based on user feedback at the OSGi Community Event we identified a place in bnd where Windows file locks can be acquired and then not cleanly released.

The commit 2ea99c3 adds a test (which failed on appveyor) and a change to URLResource which fixes the test. The change to URLResource does, however, make the build take an extra 5 minutes.

biz.aQute.bndlib/src/aQute/bnd/osgi/URLResource.java
```diff
@@ -32,9 +33,18 @@ private ByteBuffer getBuffer() throws Exception {
...
+			File file = new File(url.toURI());
 			lastModified = file.lastModified();
 			return buffer = IO.read(file.toPath());
+		} else if (url.getProtocol().equals("jar")) {
+			String subUrl = url.toString();
+			int entryStart = subUrl.indexOf("!/");
+			String entry = subUrl.substring(entryStart + 2);
+			subUrl = subUrl.substring(4, entryStart);
+
+			try (Jar jar = Jar.fromResource("tmp", new URLResource(new URL(subUrl)))) {
```
bjhargrave: This seems like a bad performance idea. This reads the entire jar into memory. Perhaps this is why the build takes 5 min longer. I did a bunch of debugging on this and have a more complete and performant solution I will make a PR for.

PR #2194, opened 2017-11-09
[bnd] cli baseline - verbose API diff

rotty3000:


timothyjward:
Does this need equivalent configuration/work to be available in the bnd-baseline-maven-plugin?

bjhargrave:
> Does this need equivalent configuration/work to be available in the bnd-baseline-maven-plugin?

Well, this PR does the work all on the command line UI rather than in bndlib where gradle, maven and the command line can all use the changes.

bjhargrave:
It is also possible the gradle and maven output is already this detailed and that the command line UI was behind. :-)

rotty3000:
Well, many moons ago I proposed exactly this type of enhancement to the baseline "reporting". However it was rejected for reasons I don't remember. Let's merge this in and @timothyjward if you try it and like what you see then perhaps I could be convinced to elevate this directly up to the baseline engine.

The one major caveat is that right now each and every baseline "client" handles presentation pretty much how they choose. So that would have to be corrected to provide a more unified reporting model.

bjhargrave:
I think each driver will need to control presentation. Maven/gradle will need to log to their logger and/or write a report to some file in the appropriate output folder. So I don't think we can push all of presentation down into bndlib.

rotty3000:
Got it! Would it be an option is to make just the logic reusable?

PR #2225, opened 2017-12-05
Only warn about missing annotations when pedantic

timothyjward:

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
@@ -19,6 +19,8 @@ import org.osgi.annotation.bundle.Header;
 import org.osgi.annotation.bundle.Headers;
 import org.osgi.annotation.bundle.Requirement;
 import org.osgi.annotation.bundle.Requirements;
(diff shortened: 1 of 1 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
```diff
@@ -283,9 +289,19 @@ public void annotation(Annotation a) throws Exception {
...
-					fqn, current.getFQN());
+			// Don't repeatedly log for the same missing annotation
+			if (loggedMissing.add(fqn)) {
+				// Only issue a warning if pedantic
+				if (analyzer.isPedantic()) {
+					analyzer.warning(
+							"Unable to determine whether the meta annotation %s applied to type %s provides bundle annotations as it is not on the project build path. If this annotation does provide bundle annotations then it must be present on the build path in order to be processed",
+							fqn, current.getFQN());
+				} else {
+					analyzer.getLogger().info(
```
bjhargrave: Don't use the Processor's Logger since you lose the class doing the logging as the source of the logging. (We should really get rid of Processor.getLogger().) I suggest that this class create its own Logger and then use that Logger to log.

PR #2226, opened 2017-12-06
[export] Attach the generated runnable jar

timothyjward:

### DIFF
--- a/maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
+++ b/maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
@@ -5,6 +5,8 @@ import static org.apache.maven.plugins.annotations.LifecyclePhase.PACKAGE;
 import java.io.File;
 import java.util.List;
 
+import org.apache.maven.artifact.DefaultArtifact;
+import org.apache.maven.artifact.handler.DefaultArtifactHandler;
 import org.apache.maven.execution.MavenSession;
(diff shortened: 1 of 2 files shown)


maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
```diff
@@ -160,4 +166,22 @@ private void report(Bndrun run) {
...
+					"The export plugin has been configured not to attach the generated application to the project.");
+			return;
+		} else if (bundlesOnly) {
+			logger.debug("The export plugin will not attach a bundles-only output to the project.");
+			return;
+		}
+
+		DefaultArtifactHandler handler = new DefaultArtifactHandler("jar");
+		handler.setExtension("jar");
+		DefaultArtifact artifact = new DefaultArtifact(project.getGroupId(), project.getArtifactId(),
```
rotty3000: Won't N executable jar files end up having the same `artifactId`? I wonder if the `artifactId` should maybe incorporate the file name to make it unique?

timothyjward: We could use the bndrun file name as the classifier to ensure uniqueness. Changing the artifactId is probably a bad plan as it breaks the link between the module GAV and its published output...

rotty3000: yes that makes more sense.

Issue #1343, opened 2016-03-01
Add support for new OSGi Bundle Annotations (RFC 220)


bjhargrave:
OSGi is working on new bundle annotations which require support in bnd. See https://github.com/osgi/design/tree/master/rfcs/rfc0220.

We need to add support for these annotations to bnd.

cschneider:
I am also interested in this feature. My main reason is that the current bnd annotations are not repeatable.
Is anyone working on this?

bjhargrave:
> Is anyone working on this?

Not at the moment. This is a placeholder for the work.

cschneider:
Would it make sense that I take I a look at the current annotations and try to add support for repeatable annotations?

bjhargrave:
By current annotations you mean the Bnd annotations? I would rather put the effort into supporting the OSGi standardized annotations.

cschneider:
I guess that the standard annotations will work very similar to the bnd ones. Can you give me some pointers where the current bnd annotations are implemented and where the new support should reside?

timothyjward:
So far the following standard bundle annotation have *some* support:

- `@Requirement`
- `@Capability`
- `@Export`

Things known not to work:

- `@Header` annotations are parsed and added to the Analyzer, but don't show up in the manifest.
- The `substitition` attribute of the `@Export` annotation is not processed
- The `@Attribute` and `@Directive` annotations are not processed

bjhargrave:
> The substitition attribute of the @Export annotation is not processed

I have the change for this in progress. Just need to write more tests before I make a PR.

bjhargrave:
The bundle annotations support appears complete.

PR #2237, opened 2017-12-19
Add support for the standard @Export annotation

timothyjward:
This does not include anything for the substitution attribute of the annotation, as I have no idea how to implement that, but it does cover the attributes and uses behaviour. The commit also deprecates the bnd annotation.

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/export/annotation/package-info.java
+++ b/biz.aQute.bndlib.tests/src/test/export/annotation/package-info.java
@@ -1,0 +1,7 @@
+@org.osgi.annotation.bundle.Export(uses = {
+		"foo", "bar"
+}, attribute = {
+		"fizz=buzz", "foobar:=fizzbuzz"
+})
+@org.osgi.annotation.versioning.Version("1.0.0")
+package test.export.annotation;
--- a/biz.aQute.bndlib/src/aQute/bnd/header/Parameters.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/header/Parameters.java
@@ -5,6 +5,7 @@ import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collector;
 
 import aQute.lib.collections.SortedList;
 import aQute.service.reporter.Reporter;
@@ -249,5 +250,11 @@ public class Parameters implements Map<String,Attrs> {
 
 	public boolean allowDuplicateAttributes() {
 		return allowDuplicateAttributes;
+	}
+
+	public static Collector<String,StringBuilder,Parameters> toParameters() {
+		return Collector.of(StringBuilder::new, (b, s) -> b.append(b.length() == 0 ? "" : ",").append(s),
+				(a, b) -> new StringBuilder(a).append(a.length() == 0 ? "" : ",").append(b),
+				b -> new Parameters(b.toString()));
 	}
 }
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -389,9 +389,7 @@ public class Analyzer extends Processor {
 				.map(Clazz::getClassName)
 				.map(TypeRef::getPackageRef)
 				.map(PackageRef::getFQN)
-				.reduce((a, b) -> a + "," + b)
-				.map(Parameters::new)
-				.orElse(new Parameters());
+				.collect(Parameters.toParameters());
 	}
 
 	private Clazz getPackageInfoClazz(PackageRef pr) {
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -254,12 +254,14 @@ public class Analyzer extends Processor {
 
 				Instructions filter = new Instructions(getExportPackage());
 				filter.append(getExportContents());
+				filter.append(getExportedByAnnotation());
 
 				exports = filter(filter, contained, unused);
 
 				if (!unused.isEmpty()) {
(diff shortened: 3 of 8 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -563,6 +607,38 @@ public void annotation(Annotation a) {
...
+							for (Object use : usesClause) {
+								sb.append(del);
+								sb.append(use);
+								del = ",";
+							}
+							info.put(USES_DIRECTIVE, sb.toString());
+						}
+
 						break;
 					case "aQute.bnd.annotation.Export" :
```
bjhargrave: I would put the warning for using the Bnd Export annotation here rather than a separate stream check like above.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -357,6 +360,47 @@ public void analyze() throws Exception {
 						uses.transpose().get(Descriptors.DEFAULT_PACKAGE));
 			}
 
+			// Check for use of the deprecated bnd @Export annotation
+
+			TypeRef bndAnnotation = descriptors.getTypeRefFromFQN(aQute.bnd.annotation.Export.class.getName());
```
bjhargrave: Rather than this stream forEach, put the warning for using the Bnd Export annotation in the case statement below.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -357,6 +360,47 @@ public void analyze() throws Exception {
...
+		return contained.keySet()
+				.stream()
+				.map(this::getPackageInfoClazz)
+				.filter(clz -> clz != null)
+				.filter(clz -> clz.annotations != null)
+				.filter(clz -> clz.annotations.contains(exportAnnotation))
+				.map(Clazz::getClassName)
+				.map(TypeRef::getPackageRef)
+				.map(PackageRef::getFQN)
+				.reduce((a, b) -> a + "," + b)
```
bjhargrave: `.collect(Collectors.joining(","))`

Or better yet, modify Parameters to add a method which returns a Collector you can use here. See <https://github.com/bndtools/bnd/blob/5eee8ed041f0eb9e089590c1d257d43c97a7317b/biz.aQute.bndlib/src/aQute/bnd/osgi/repository/ResourcesRepository.java#L66-L68> for example.

biz.aQute.bndlib.tests/src/test/export/annotation/package-info.java
```diff
@@ -0,0 +1,7 @@
+@org.osgi.annotation.bundle.Export(uses = {
+		"foo", "bar"
+}, attribute = {
+		"fizz=buzz", "foobar:=fizzbuzz"
+})
```
bjhargrave: I think we also need test for the `substitution` annotation element.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -563,6 +607,38 @@ public void annotation(Annotation a) {
 								// Ignore
 							}
 						}
+						break;
+					case "org.osgi.annotation.bundle.Export" :
```
bjhargrave: I think we also need support for the `substitution` annotation element.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
timothyjward: This code is called to scan every entry on the class path - I originally tried putting the warning here but it caused bnd to complain about classpath libraries that were nothing to do with the jar being built. 

Due to the order in which things are done by bnd I was unable to find a way to determine whether the package is contained at the time of doing this scan (we only update the contained Packages after calling this method). It seems preferable to only warn for classes included in the bundle.

bjhargrave: Fair enough!

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
timothyjward: That does need to be done, but I have no idea how best to implement it. There isn't an obvious equivalent from the bnd annotations, and I'm not sufficiently familiar with how bnd determines version range policies to feel comfortable making up a solution

bjhargrave: Then we need to open an issue to make sure we do not forget to implement this support at some point.

timothyjward: I have commented in #1343 to say which items are outstanding (there are a few).

biz.aQute.bndlib/src/aQute/bnd/header/Parameters.java
```diff
@@ -250,4 +251,10 @@ public void mergeWith(Parameters other, boolean override) {
 	public boolean allowDuplicateAttributes() {
 		return allowDuplicateAttributes;
 	}
+
+	public static Collector<String,StringBuilder,Parameters> toParameters() {
```
bjhargrave: Here is a slightly simplified version using StringJoiner.

```java
public static Collector<String, StringJoiner, Parameters> toParameters() {
	return Collector.of(() -> new StringJoiner(","), StringJoiner::add, StringJoiner::merge,
		j -> new Parameters(j.toString()));
}
```

An alternate version avoids string concatenation:

```java
public static Collector<String, Parameters, Parameters> toParameters() {
	return Collector.of(Parameters::new, Parameters::accumulator, Parameters::combiner);
}
private static void accumulator(Parameters p, String s) {
	p.mergeWith(OSGiHeader.parseHeader(s), true);
}
private static Parameters combiner(Parameters t, Parameters u) {
	t.mergeWith(u, true);
	return t;
}
```

PR #2240, opened 2017-12-20
Don't duplicate properties in standalone Workspace

njbartlett:
There does not seem to be a good reason to copy the content of the Run file into
the standalone "Workspace". When we do this, the properties are duplicated at
both levels.

As a result it becomes impossible to remove a key from a BndEditModel... in
`BndEditModel#loadFrom(InputStream)` we do a "clear" which involves repopulating
the map from the content of the workspace, but in the standalone case this just
reloads from whatever the content of the file was when first opened. So you
cannot delete a key from the map.

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/BndEditModelTest.java
+++ b/biz.aQute.bndlib.tests/src/test/BndEditModelTest.java
@@ -1,5 +1,6 @@
 package test;
 
+import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.util.LinkedList;
 import java.util.List;
@@ -8,6 +9,7 @@ import org.osgi.framework.namespace.IdentityNamespace;
 import org.osgi.resource.Namespace;
 import org.osgi.resource.Requirement;
 
+import aQute.bnd.build.Run;
 import aQute.bnd.build.Workspace;
 import aQute.bnd.build.model.BndEditModel;
 import aQute.bnd.build.model.clauses.ExportedPackage;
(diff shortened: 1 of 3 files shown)


pkriens:
LGTM

pkriens:
how do you delete a key?

pkriens:
I propose the 'true' as the tie breaker. I do not see how you can have true and not copy the properties

And I see I am an idiot since I did LGTM :-(

njbartlett:
The bug manifests when you do the following in Bndtools:

1. Open a bndrun file in the editor.
2. Switch to the source tab and delete a key, e.g. `-runbundles`.
3. Switch back to the UI tab and note that runbundles is still populated.

This is reproduced by the test added in this commit, `testRemovePropertyFromStandalone`.

I don't understand your comment about "true as the tiebreaker", can you clarify?

I also don't understand the motivation for copying the properties into two levels at all. If the bndrun is standalone there should be nothing in the Workspace.

timothyjward:
> I also don't understand the motivation for copying the properties into two levels at all. If the bndrun is standalone there should be nothing in the Workspace.

This is certainly how the maven integration expects things to be.

pkriens:
@njbartlett Ok, I see it is an edit issue. Could we not detect this in the edit model?

As  I said, the code checks for -standalone: true. I added this because I needed to define more repositories that are not OSGi repos. Repositories only work on Workspace level. So if you remove the properties there are no longer any repositories defined. 

For example, this bndrun allows me to do a lot of work with these repos without being forced to create a special workspace. 

	-standalone           : true
	-groupid              : com.xyz.thirdparty
	-connection-settings  : ~/.bnd/xyz.xml

	-plugin.2.Central-Runtime: \
	         releaseUrl="http://repo1.maven.org/maven2/" ; \
	         index =${.}/../buildconfig/central-runtime.mvn; \
	         name="Central-Runtime"
	
	-plugin.3.Central-Build: \
	         releaseUrl="http://repo1.maven.org/maven2/" ; \
	         index =${.}/../buildconfig/central-build.mvn; \
	         name="Central-Build"
	
	-plugin.4.Partners: \
	         releaseUrl=", \
	            https://${nexus}/content/repositories/s_2ndparty/, \
	            https://${nexus}/content/repositories/public/, \
	            https://${nexus}/content/repositories/2ndparty/,\
	            https://${nexus}/content/repositories/snapshots/, \
	            https://${nexus}/content/repositories/s_snapshots/" ; \
	         index =${.}/../buildconfig/partners.mvn; \
	         name="Partners"

There are not many, but there are a few, things that work on workspace mode only. 

By using the 'true' option, I made it standalone but did not have to define any repos in the -standalone instruction.  So I propose to copy the properties only when -standalone=true. 

For maven, I assume your bndrun file points to XML files so properties are not be copied.

njbartlett:
It may be possible to fix in the `BndEditModel`. The problem is that the dummy "workspace" is initialised once only, and calling `loadFrom` cannot remove any properties that were loaded into that workspace upon the first load (though it can replace them with non-null values).

I think it would be necessary to make the `loadFrom` also reload the workspace, or disconnect from the initial workspace and create a new one.

If you can think of a way to support your use case while still passing `testRemovePropertyFromStandalone` then I would be happy.

I don't know what `-standalone: true` is intended to communicate over and above just `-standalone`. It's not documented and it doesn't obviously mean anything.

pkriens:
Yeah, the test passes with -standalone true, see 

Could you merge https://github.com/bndtools/bnd/pull/2463 if you're ok? The result that it passes is here: https://travis-ci.org/bndtools/bnd/builds/388663692?utm_source=github_status&utm_medium=notification

PR #2247, opened 2017-12-26
docs: fix typo runrequire -> runrequires

tomioka:
This pull-request fixes the first example of http://bnd.bndtools.org/chapters/300-launching.html where runrequires was missing "s".


bjhargrave:
Thanks for the PR!

PR #2255, opened 2018-01-03
Avoid using upper case for generated requirement directives

timothyjward:
Using the cardinality or resolution attributes of the Requirement annotation resulted in an upper cased value in the manifest which caused bnd to complain.


bjhargrave:
I think the correct answer here would be to fix the OSGi annotations so their toString is a lower case value. This is done for DS annotations but I failed to do it for the Bundle annotations. I will fix this for OSGi.

PR #2260, opened 2018-01-07
Handle relative redirects.

### LINKED ISSUE #2259: HttpClient does not handle redirects properly
I'm using (actually trying to use) a maven repository that answers some requests with relative redirects ("`Location: <relative URL>`"). This complies to the [latest standard](https://tools.ietf.org/html/rfc7231#section-7.1.2). The HttpClient in bnd currently only handles absolute URLs in this header.
 
The error is:

`java.net.MalformedURLException: no protocol: /com/...`

which is caused by calling `new URL(location)` with `location` being (only) a path (i.e. a relative URL).

mnlipp:
Closes #2259. I'd love to see this in a 3.5.1, if it will ever come out, as this really causes some problems in my environment.

### DIFF
--- a/biz.aQute.bndlib.comm.tests/test/aQute/bnd/comm/tests/HttpClientTest.java
+++ b/biz.aQute.bndlib.comm.tests/test/aQute/bnd/comm/tests/HttpClientTest.java
@@ -183,6 +183,14 @@ public class HttpClientTest extends TestCase {
 		}
 	}
 
+	public void testRedirectRelative() throws Exception {
+		try (HttpClient hc = new HttpClient();) {
+			TaggedData tag = hc.build().get(TaggedData.class).go(httpServer.getBaseURI("redirect/3/200?relative=true"));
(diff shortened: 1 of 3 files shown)


bjhargrave:

Also, please add some test cases to verify the change works as desired. The test cases would go in the `biz.aQute.bndlib.comm.tests` project. You will probably need to also modify `biz.aQute.http.testservers` to do redirects with a relative URL.

mnlipp:
I tried to, but GitHub doesn't allow me to do this, because the only branches are "master" and "next".

grafik

I can search for a tag 

grafik

and see the "form", but I cannot enter anything and submit it.

grafik

bjhargrave:
Use master. That is what I asked for.

mnlipp:
Sorry, I thought I had chosen master.

bjhargrave:
You will probably need to checkout the master branch and make you fix to it before redoing this PR since your PR must just have your changes. Thanks.

PR #2262, opened 2018-01-08
Handle relative redirects.

### LINKED ISSUE #2259: HttpClient does not handle redirects properly

mnlipp:
Handle relative redirects. Closes #2259.

Signed-off-by: Michael Lipp <mnl@mnl.de>

### DIFF
--- a/biz.aQute.http.testservers/src/aQute/http/testservers/Httpbin.java
+++ b/biz.aQute.http.testservers/src/aQute/http/testservers/Httpbin.java
@@ -166,7 +166,9 @@ public class Httpbin extends HttpTestServer {
 	public void _redirect(Request rq, Response rsp, int count, int response) throws URISyntaxException {
 		if (count > 0) {
 			System.out.println("redirect " + count);
-			rsp.headers.put("Location", getBaseURI() + "/redirect/" + (--count) + "/" + response);
+			String location = (!rq.args.containsKey("relative") ? getBaseURI() : "") + "/redirect/" + (--count) + "/"
+					+ response;
+			rsp.headers.put("Location", location);
 			rsp.code = 301;
 		} else {
 			rsp.code = response;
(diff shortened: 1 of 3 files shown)


biz.aQute.http.testservers/src/aQute/http/testservers/Httpbin.java
```diff
@@ -166,7 +166,9 @@ public void _index(Request rq, Response rsp) throws IOException {
 	public void _redirect(Request rq, Response rsp, int count, int response) throws URISyntaxException {
 		if (count > 0) {
 			System.out.println("redirect " + count);
-			rsp.headers.put("Location", getBaseURI() + "/redirect/" + (--count) + "/" + response);
+			String location = (!rq.args.containsKey("relative") ? getBaseURI() : "") + "/redirect/" + (--count) + "/"
```
bjhargrave: Shouldn't this be more than just `containsKey` and instead check the value for truthiness? In the test case you use `?relative=true`. So there seems a small disconnect there.

mnlipp: There are no other possible values for key "relative". But AFAIK, the syntax for the query part must be "key=value", even if the value doesn't matter.

bjhargrave:
And now the AppVeyor build works!

biz.aQute.http.testservers/src/aQute/http/testservers/Httpbin.java
bjhargrave: An alternate would be a different path `/redirectrelative/`

bjhargrave:
Thanks for the PR!

biz.aQute.http.testservers/src/aQute/http/testservers/Httpbin.java
mnlipp: Maybe, I'm not as deep in your test server design as you. But I don't like duplicating code and this seemed the easiest way to modify the behavior of _redirect in that particular aspect.

PR #2272, opened 2018-01-24
clarify bnd instruction inheritance/merging

kwin:
also document all mojo configuration parameters

### DIFF
--- a/maven/bnd-maven-plugin/README.md
+++ b/maven/bnd-maven-plugin/README.md
@@ -77,16 +77,31 @@ although it will be valid, the bundle may not be *useful*.
 For further usage information, see the integration test projects under the included
 `src/test/resources/integration-test/test` directory.
 
-## Configuration Properties
+## Configuration Parameters
 
-|Configuration Property | Description |
+|Configuration Parameter | Description |
 | ---                   | ---         |
+|`bndfile`            | File path to a bnd file containing bnd instructions for this project. The file path can be either absolute or relative to the project directory. _Defaults to `bnd.bnd`_.|
+|`bnd`            | Bnd instructions for this project specified directly in the pom file. This is generally be done using a {@code <![CDATA[]]>} section. If the projects has a `bndfile` configuration property or a file in the default location `bnd.bnd`, then this configuration element is ignored. |
 |`targetDir`            | The director into which to export the result. _Defaults to `${project.build.directory}`._|
 |`sourceDir`            | Specify an alternative source directory. _Defaults to `${project.build.sourceDirectory}`._|
 |`resources`            | Specify an alternative resources directory. _Defaults to `${project.build.resources}`._|
 |`classesDir`           | Specify an alternative classes directory. _Defaults to `${project.build.outputDirectory}`._|
 |`manifestPath`         | Specify the path to a manifest file to use. _Defaults to `${project.build.outputDirectory}/META-INF/MANIFEST.MF`._|
 |`skip`                 | Skip the index process altogether. _Defaults to `false`._ Override with property `bnd.skip`.|
+
+## Inheritance of bnd instructions
+
+Bnd instructions in general are being merged with the ones being set on the Maven project's parent. This works slightly differently than [default plugin configuration merging in Maven](http://blog.sonatype.com/2011/01/maven-how-to-merging-plugin-configuration-in-complex-projects/). From the following places the bnd instructions are collected from all pom.xmls in the hierarchy (starting with the top level parent pom and ending with the leaf project's pom):
+
+1. From the `<pluginManagement>` section for this plugin (both configuration within the current execution and the global one)
+   1. the bnd file (given through configuration parameter `bndfile` or the implicit default `bnd.bnd` in pluginManagement. _The implicit default can only be used where the parent is not resolved from the Maven repository but rather from a parent directory in the filesystem_
+   2. the bnd instructions given explicitly via `bnd`
+1. From the `<plugins>` section for this plugin(both configuration within the current execution and the global one)
+   1. the bnd file (given through configuration parameter `bndfile` or the implicit default `bnd.bnd` in pluginManagement. _The implicit default can only be used where the parent is not resolved from the Maven repository but rather from a parent directory in the filesystem_
+   2. the bnd instructions given explicitly via `bnd`
+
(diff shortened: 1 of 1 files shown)


maven/bnd-maven-plugin/README.md
```diff
@@ -77,17 +77,32 @@ although it will be valid, the bundle may not be *useful*.
...
 |`resources`            | Specify an alternative resources directory. _Defaults to `${project.build.resources}`._|
 |`classesDir`           | Specify an alternative classes directory. _Defaults to `${project.build.outputDirectory}`._|
 |`manifestPath`         | Specify the path to a manifest file to use. _Defaults to `${project.build.outputDirectory}/META-INF/MANIFEST.MF`._|
 |`skip`                 | Skip the index process altogether. _Defaults to `false`._ Override with property `bnd.skip`.|
 
+## Inheritance of bnd instructions
+
+Bnd instructions in general are being merged with the ones being set on the Maven project's parent. This works slightly differently than [default plugin configuration merging in Maven](http://blog.sonatype.com/2011/01/maven-how-to-merging-plugin-configuration-in-complex-projects/). From the following places the bnd instructions are collected from all pom.xmls in the hierarchy (starting with the top level parent pom and ending with the leaf project's pom):
+
+1. From the `<pluginManagement>` section for this plugin (both configuration within the current execution and the global one)
```
bjhargrave: I don't think this is quite right. We only look for pluginManagement if the project does not have a plugins for the bnd-maven-plugin. So, starting at the top level pom, if the project has the bnd-maven-plugin configured we merge the project's configuration in. If the project does not have the bnd-maven-plugin configured, if the project has bnd-maven-plugin in pluginManagement, we merge the pluginManagement configuration in.

At each project, if the bnd file is present, it is used for bnd configuration. If the bnd file is not present, then the bnd element is used if present.

maven/bnd-maven-plugin/README.md
```diff
@@ -77,17 +77,32 @@ although it will be valid, the bundle may not be *useful*.
 For further usage information, see the integration test projects under the included
 `src/test/resources/integration-test/test` directory.
 
-## Configuration Properties
+## Configuration Parameters
```
bjhargrave: You need to amend your commits so that each commit message includes a `Signed-off-by` as required by CONTRIBUTING.

maven/bnd-maven-plugin/README.md
kwin: > At each project, if the bnd file is present, it is used for bnd configuration. If the bnd file is not present, then the bnd element is used if present.

Ok I will modify
> _The implicit default can only be used where the parent is not resolved from the Maven repository but rather from a parent directory in the filesystem_

to

> _bnd files can only be used where the Maven project is not resolved from the Maven repository but rather from a parent directory in the filesystem_.

 WDYT?

kwin: > I don't think this is quite right. We only look for pluginManagement if the project does not have a plugins for the bnd-maven-plugin. So, starting at the top level pom, if the project has the bnd-maven-plugin configured we merge the project's configuration in. If the project does not have the bnd-maven-plugin configured, if the project has bnd-maven-plugin in pluginManagement, we merge the pluginManagement configuration in.

From a code perspective you are right but you haven't considered that Maven already does some merging for you. So I think from a functional perspective it is correctly described. It first considers the configuration from `pluginMgmt` and then merges/overwrites with the configuration from `plugins`. How would you like to rephrase this. What do you consider exactly wrong from the description above?

bjhargrave: But the section is called `Inheritance of bnd instructions` not how maven merges maven configuration. I think as I described in my comment above is better. You descend to the top level pom and then look for execution configurations. Only if the project does not have an execution configuration do you look for pluginManagement configuration. pluginManagement is not checked if the project has an execution configuration. Perhaps another way to update this section is to simply say nothing about pluginManagement. Users will just expect it to work, and it (now) does. :-)

rotty3000:
@kwin did you see the comment about adding the signed-off-by to your commit?

maven/bnd-maven-plugin/README.md
kwin: > Only if the project does not have an execution configuration do you look for pluginManagement configuration. pluginManagement is not checked if the project has an execution configuration. 

This is not correct from a functional perspective. PluginMgmt is always being merged by Maven and both plugin as well as pluginMgmt support execution and global configuration!

bjhargrave: If the project has an execution configuration, maven has already merged the pluginManagement configuration into it. So we do not need to look directly at the pluginManagement configuration.

kwin: Maven allows a execution configuration both below the pluginMgmt as well as below the plugins section. Consider the following example

```
<project ....>
 <build>
  <pluginManagement>
   <plugins>
    <plugin>
     <configuration>some global config in pluginMgmt</configuration>
     <executions>
      <execution>
        <id>someid</id>
        <goals>
         <goal>bnd-process</goal>
        </goals>
        <configuration>some exec config in pluginMgmt</configuration>
       </execution>
      </executions>
    <plugin>
   </plugins> 
  </pluginManagement>
  <plugins>
   <plugin>
    <configuration>some global config in plugins</configuration>
    <executions>
     <execution>
       <id>someid</id>
        <configuration>some exec config in plugin</configuration>
      </execution>
     </executions>
   <plugin>
  </plugins>
 </build>
</project>
```
I would expect the documentation regarding the inheritance cover all those cases.

maven/bnd-maven-plugin/README.md
kwin: I just did that

maven/bnd-maven-plugin/README.md
bjhargrave: Yes, the documentation should. But the documentation should not describe the precise algorithm that does this. The documentation should describe the result, not how the result is achieved. So in the above example, the configuration is the merge of the two since maven already does the merge, so bnd-maven-plugin does not have to do it.

kwin: Maven does not only do by default the merge between plugin and pluginMgmt (both global config and execution config) but usually also between parent pom and current pom. That is what users expect when they use a Maven Plugin!

If your Maven plugin would rely on injection of the parameters only (as 99% of the Maven plugins do) that would work as expected. In your case you explicitly overwrite the inheritance logic for parent poms. Therefore it is reasonable to explain both mechanisms here, to make clearer what exactly is different here.

Anyhow since I fear we don't get a common ground here, feel free to adjust the text after merging. IMHO having a paragraph about inheritance is much better than not saying about that at all as it is not working as expected! Even if the documentation is complicated it is better than not documenting that at all (as long as the description is correct).

PR #2278, opened 2018-01-26
InfoRepositoryWrapper: Ensure we always return (at least) an empty list

seanbright:

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/deployer/repository/wrapper/InfoRepositoryWrapper.java
+++ b/biz.aQute.repository/src/aQute/bnd/deployer/repository/wrapper/InfoRepositoryWrapper.java
@@ -168,10 +168,12 @@ public class InfoRepositoryWrapper extends BaseRepository {
 			throws Exception {
 		init();
 
-		nextReq: for (Requirement req : requirements) {
+		for (Requirement req : requirements) {
+			result.putIfAbsent(req, new ArrayList<Capability>(1));
+
 			String f = req.getDirectives().get("filter");
 			if (f == null)
-				continue nextReq;
+				continue;
 
 			Filter filter = new Filter(f);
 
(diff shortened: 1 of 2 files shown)


biz.aQute.repository/src/aQute/bnd/deployer/repository/wrapper/InfoRepositoryWrapper.java
```diff
@@ -168,21 +168,21 @@ public void findProviders(Map<Requirement,List<Capability>> result, Collection<
 			throws Exception {
 		init();
 
-		nextReq: for (Requirement req : requirements) {
+		for (Requirement req : requirements) {
+			result.putIfAbsent(req, new ArrayList<Capability>(1));
```
seanbright: FYI - This is `putIfAbsent()` instead of `put()` because I don't know that we are guaranteed to have unique `Requirement`s being passed in.

seanbright:
Looks like the test is wrong. I am addressing.

seanbright:
@bjhargrave changes to the test look correct?

seanbright:
This is not specific to InfoRepositoryWrapper

bjhargrave:
> This is not specific to InfoRepositoryWrapper

Well, I think the error is here. It may exists in other Repository implementations too.

seanbright:
@bjhargrave How would you deal with aQute.bnd.repository.maven.provider.IndexFile.findProviders? My stream-fu is lacking.

bjhargrave:
You need to post-process the map to add missing requirements:

```java
	public Map<Requirement,Collection<Capability>> findProviders(Collection< ? extends Requirement> requirements) {
		Map<Requirement, Collection<Capability>> result = descriptors.values()
				.stream()
				.map(BundleDescriptor::getResource)
				.filter(Objects::nonNull)
				.flatMap(res -> requirements.stream()
						.flatMap(req -> res.getCapabilities(req.getNamespace())
								.stream()
								.filter(cap -> ResourceUtils.matches(req, cap))
								.map(cap -> new ReqCap(req, cap))))
				.collect(groupingBy(ReqCap::req, mapping(ReqCap::cap, toCollection(ArrayList::new))));
		requirements.forEach(req -> result.computeIfAbsent(req, r -> new ArrayList<>()));
		return result;
	}
```

I'll make another PR for this. You can merge this PR.

seanbright:
I wonder if it would be beneficial to create a `RepositoryUtil.ensureValidity(Map<Req,Collection<Cap>> meh)` that can be leveraged by the various `Repository` implementations.

bjhargrave:
I looked at all the others and they all use ResourceRepository which does not suffer from this.

PR #2279, opened 2018-01-26
 maven: findProviders must always include the requirements in result 

bjhargrave:
We change the IndexFile.findProviders method to now use ResourcesRepository since it always includes the input requirements in the result.


seanbright:
LGTM & WFM 👍

PR #2280, opened 2018-01-26
URLCache: Reduce synchronization, improve concurrency.

seanbright:


seanbright:
I'm not entirely sure how I merged this. Should I revert?

bjhargrave:
You could use computeIfAbsent if you duck the exception.

```java
info = infos.computeIfAbsent(file, f -> {
  try {
    return new Info(file, uri);
  } catch (Exception e) {
    throw Exceptions.duck(e);
  }
});
```

PR #2281, opened 2018-01-28
Various cleanups, mostly around Java language features

seanbright:
There should be no user-visible changes here - just lots of cleanups and boilerplate code removal.


aQute.libg/src/aQute/lib/converter/Converter.java
```diff
@@ -339,10 +339,10 @@ private Number number(Object o) {
 			return (Number) o;
 
 		if (o instanceof Boolean)
-			return ((Boolean) o).booleanValue() ? 1 : 0;
+			return (Boolean) o ? 1 : 0;
```
bjhargrave: I don't care much for this commit. I prefer explicit boxing/unboxing.

aQute.libg/src/aQute/configurable/Configurable.java
```diff
@@ -316,7 +317,7 @@ Object convertArray(Type componentType, Object o) throws Exception {
 				else
 					return unescape(s);
 			}
-			return Arrays.asList(o);
+			return Collections.singletonList(o);
 		}
```
bjhargrave: This commit is wrong. A singletonList is immutable. asList is mutable.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -196,7 +196,7 @@ boolean refresh() throws Exception {
 		}
 		return descriptors.values()
 				.stream()
-				.filter(descriptor -> {
+				.anyMatch(descriptor -> {
```
bjhargrave: This is wrong since anyMatch short circuits the stream. ALL elements in the stream must be processed. This is why count is used!

bjhargrave: The side effects are important.

seanbright: If side effects are important, we shouldn't be using a stream operation that is intended to be stateless. This should probably just be a foreach loop. I'll update the PR today.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -194,23 +194,25 @@ boolean refresh() throws Exception {
...
-							File f = descriptor.promise.getValue();
-							if ((f != null) && f.isFile() && (f.lastModified() != descriptor.lastModified)) {
-								updateDescriptor(descriptor, f);
-								return true;
-							}
-						}
-					} catch (InvocationTargetException | InterruptedException e) {
-						throw Exceptions.duck(e);
+
+		boolean updated = false;
```
bjhargrave: This is a pointless change and adds no value. It was a for loop before I changed it to streams like the rest of the descriptors operations in this source file.

Using a for loop imposes an iteration order that is not required. Using streams would allow the streams implementation to potentially optimize the iteration order in the future.

seanbright: > This is a pointless change and adds no value.

I disagree.

> It was a for loop before I changed it to streams like the rest of the descriptors operations in this source file.

So? Every other use of streams in the file is concise and obvious and has no bearing on this usage.

> Using a for loop imposes an iteration order that is not required. Using streams would allow the streams implementation to potentially optimize the iteration order in the future.

I would err on the side of code clarity over _potential_ optimization. If the performance of this particular method is that much of a consideration then we should explicitly parallelize it and not just cross our fingers that the JVM will do it for us at the expense of readable code.

bjhargrave: I guess we will have to disagree here. I don't disagree with the other commits but this one I disagree with.

I am not sure how a for loop (external iteration) has more clarity than stream processing (internal iteration).

seanbright: > I am not sure how a for loop (external iteration) has more clarity than stream processing (internal iteration).

I think were it using `forEach` (with an external flag) instead of `filter`/`count` it wouldn't bother me as much. As it is, the specific count of items that satisfy the filter predicate is not relevant beyond "is it 0 or is it greater than 0" but we can't use a short circuiting terminal operation because we need to ensure that we process the filter predicate for all items. Simple!

On the other hand, what the for loop is doing is obvious.

seanbright: Also, I've rebased that commit away.

Issue #2287, opened 2018-01-31
IPom.bindToVersion does not work with "mixed content" maven repositories


mnlipp:
If `IPom.bindToVersion` is called with a maven repository that contains both a release version and a snapshot of that version, it binds to the wrong one. It simply calls `Collections.sort(revisions)`, but this does not sort "1.0.1-SNAPSHORT" as less than "1.0.1" (as it should be with maven versioning).

Issue #2286, opened 2018-01-31
MavenVersion.isInteger is buggy


mnlipp:
Integer.MAX_VALUE is not considered an integer by `MavenVersion.isInteger`. The problematic line (bnd 3.5.0) is MavenVersion.java:262. Instead of `minor.compareTo("2147483647") < 0` it must be `minor.compareTo("2147483647") <= 0`.

The "big" impact of this is that `MavenVersion.HIGHEST.toString()` cannot be parsed successfully.

PR #2291, opened 2018-02-03
Use canonical maven version comparison and range check.

### LINKED ISSUE #2285: MavenVersionRange.includes is buggy
According to the [maven specs](http://maven.apache.org/enforcer/enforcer-rules/versionRanges.html) the maven version range "[1.0.1,)" means v >= 1.0.1. However, feeding this range and "1.0.1" to `MavenVersionRange.includes` returns `false`. 

The "missing" upper limit is converted to 0.0.0. This is wrong, because `MavenVersion.compare` (used in `MavenVersionRange.includes`) quite correctly does not consider "1.0.1" to be smaller than "0.0.0". So "0.0.0" may not be used as replacement for the missing upper bound (or special handling must be implemented).

mnlipp:
Fixes #2285, #2286, #2287.

This fix embeds the maven-artifact library, because it would be ridiculous to re-implement the [maven version ordering rules](https://maven.apache.org/pom.html#Version_Order_Specification) and I didn't want to copy the code.

Test case `test.TestSelfBuild` fails, but IMHO this is due to the fact that I had to push the package version. I'd perfectly understand if this is not acceptable for a "fix". I'd also understand if you don't want to depend on the maven-artifact library and therefore reject the pull request. Nevertheless, I think that using the "canonical" library provides the best compatibility with maven tooling. 

The bugs referenced showed up in one of my projects, but I have implemented a work-around. So, as long as nobody else is affected, they are definitely low priority.

Signed-off-by: Michael N. Lipp <mnl@mnl.de>


bjhargrave:
I don't think this is best way to solve the issues for Bnd. I have instead made PRs #2292, #2293, and #2295 to address the issues. The pointer to the ComparableVersion class in maven-artifact was very helpful. Thanks.

PR #2299, opened 2018-02-06
Exporter improvements

bjhargrave:


bjhargrave:
It is not removed. It was moved to the biz.aQute.bnd.exporters project a long time ago and the file I deleted had been disabled anyway (getTypes returned an empty array). I was just removing some dead wood. So work can continue on the file in the biz.aQute.bnd.exporters project if anyone is interested.

eliasvasylenko:
Ah that's good to hear, thanks for the explanation.

PR #2337, opened 2018-03-05
fixes #2335 - Use ResourcesRepository.toResourcesRepository collector instead of forEach

### LINKED ISSUE #2335: Use ResourcesRepository.toResourcesRepository collector instead of forEach
The forEach on the stream should be replaced with a collect using the ResourcesRepository.toResourcesRepository collector to return a ResourcesRepository object. The work in the forEach will need to be moved to map/filter operations.

https://github.com/bndtools/bnd/blob/463c7cdf6f0eb223c0da29bd001cfc4c34368121/biz.aQute.bndlib/src/aQute/bnd/osgi/repository/SimpleIndexer.java#L110-L129

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/osgi/repository/SimpleIndexer.java
```diff
@@ -107,26 +122,24 @@ public static void index(Collection<File> files, OutputStream outputStream, URI
...
-						resourcesRepository.add(resourceBuilder.build());
+						return resourceBuilder.build();
 					}
 				} catch (Exception e) {
 					logger.error("Could not index file {}", f, e);
 				}
-			});
+				return null;
+			})
+			.filter(r -> Objects.nonNull(r))
```
bjhargrave: You do not need to filter for null. ResourcesRepository ignores adding null. :-)

https://github.com/bndtools/bnd/blob/eda1aa4fa34b785baea0cb5683d4b2d163b6c2b0/biz.aQute.bndlib/src/aQute/bnd/osgi/repository/ResourcesRepository.java#L48-L52

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/SimpleIndexer.java
```diff
@@ -141,4 +154,170 @@ public static void index(Collection<File> files, OutputStream outputStream, URI
 
 		repository.save(outputStream);
 	}
+
+	private static class SafeResourceBuilder extends ResourceBuilder {
```
bjhargrave: Since ResourceBuilder is a concrete class, we will add methods to it from time to time. So this class may be ignore and leak new methods. Perhaps it would be better to move this class to ResourceBuilder and add a `ResourceBuilder safeResourceBuilder()` method to ResourceBuilder which returns an object of this implementation. Then anyone changing ResourceBuilder will see this class and thus also change it?

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java
```diff
@@ -710,4 +710,170 @@ public boolean addFile(File file, URI uri) throws Exception {
 		addContentCapability(uri, sha256, file.length(), mime);
 		return hasIdentity;
 	}
+
+	public ResourceBuilder safeResourceBuilder() {
+		return new SafeResourceBuilder();
+	}
+
+	private static class SafeResourceBuilder extends ResourceBuilder {
```
bjhargrave: This cannot be static and have no reference to the real ResourceBuilder. :-)

I think you want to make this non-static, so it contain a reference to the real ResourceBuilder and replace the `super` calls with calls to `ResourceBuilder.this.`...

PR #2341, opened 2018-03-06
fixes #2135 - snapshots and versions installed by bnd workspace into a maven local repository are ignored by maven

### LINKED ISSUE #2135: snapshots and versions installed by bnd workspace into a maven local repository are ignored by maven
When a maven build searches for dependencies, if there is a remote repository which provides the version/snapshot, the local one is always ignored. The only workaround is to use offline mode.

rotty3000:


bjhargrave:
Looks ok to me, but @pkriens needs to review.

pkriens:
Don't have time to check it out but it looks harmless to me. So if it works ... I noticed there was no test case?

rotty3000:
sure I can add a test!

All this change does is allow the metadata files to be written when a snapshot is built locally (i.e. not fetched from a repo).

pkriens:
:-)

rotty3000:
what this change boils down to is that even when building local snapshots, we need to create the metadata files. Caveat is that locally build snapshots have slightly different metadata.

biz.aQute.repository/src/aQute/maven/provider/Releaser.java
```diff
@@ -71,24 +71,30 @@ protected void updateMetadata() throws Exception, InterruptedException {
 		int n = 0;
```
bjhargrave: Seems like there is still an NPE risk here. You now alway call `updateMetadata` and it always calls `isUpdateProgramMetadata` whose implementation on `Releaser` assumes `repo` is non-null. I think you also need to fix `Releaser.isUpdateProgramMetadata` to guard against repo being null.

rotty3000: no because the implementation overrides that method:
```
	protected boolean isUpdateProgramMetadata() {
		return true;
	}
```

rotty3000: now i did consider that maybe the Snapshot releaser should just provide a completely different path for the non-repo case as opposed to having all these null checks when driving through the same logical path.

rotty3000: ok... now I see what you mean...

rotty3000: maybe there should be a new path for SnapshotReleaser with no repo?

bjhargrave: That is possible. I thought of something like that. But I am not sure that is just too risky.

bjhargrave: Also, I think you can release a non-snapshot version to a local only repository and so Releaser will be used with repo==null. So the fix cannot be only in SnapshotReleaser.

bjhargrave: I think you only need up update the `Releaser.isUpdateProgramMetadata` method to handle repo==null. I dont think it is a big change.

rotty3000:
the required change was much less intrusive in the end.

biz.aQute.repository/src/aQute/maven/provider/MetadataParser.java
```diff
@@ -108,9 +108,12 @@ public Tag toTag() {
 			Tag versioning = new Tag(top, "versioning");
 
 			Tag snapshot = new Tag(versioning, "snapshot");
```
bjhargrave: I think you need a test to not emit the snapshot tag if all of the fields are null. Now you can end up with an empty snapshot element.

rotty3000: got it, the two sets are supposed to be kind of mutually exclusive.

PR #2363, opened 2018-03-15
[build] quiet compiler warning

rotty3000:


rotty3000:
Closing since we're not changing the output sadly...

PR #2368, opened 2018-03-17
[osgi] parse bundle header output for macros

rotty3000:


rotty3000:
@timothyjward this was something you had wanted also right?

PR #2370, opened 2018-03-17
[header annotations] place analyzer entries at beginning of list rather than in arbitrary sorted order

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
```diff
@@ -807,10 +809,11 @@ public String getHeader(String name) {
 			//
 			// Remove duplicates and sort
 			//
-			Set<String> set = new TreeSet<>(headers.get(name));
-			if (value != null)
-				set.add(value);
-			return Strings.join(set);
+			List<String> list = new ArrayList<>(new TreeSet<>(headers.get(name)));
```
bjhargrave: I guess I am unclear on why the header value from annotations are any more or less important than header values in the bnd file. That itself seems arbitrary. 

In some ways the developer of the bundle should have more control over the manifest (via her bnd file) than the code being included in the bundle which may come from 3rd parties.

rotty3000: > In some ways the developer of the bundle should have more control over the manifest (via her bnd file) than the code being included in the bundle which may come from 3rd parties.

That's exactly why I made this change. Otherwise there is no rhyme or reason to how the requirements are aggregated. In fact the only logic at all is natural ordering of strings via `TreeSet`, which makes it REALLY hard for the developer to figure out how to give their bnd file precedence over values from annotations. The above does exactly that.

bjhargrave: Is there precedence in the order of clauses in the Require-Capability header?

Also your change puts the annotation value first before the bnd file values, so does that mean you think later values have precedence over earlier values?

rotty3000: that's incorrect, it puts the bnd file value first.

bjhargrave: OK, I see that now. 

```java
		String value = analyzer.getProperty(name);
		if (headers.containsKey(name)) {
			//
			// Remove duplicates and sort
			//
			Set<String> set = new TreeSet<>(headers.get(name));
			List<String> result = new ArrayList<>(set.size()+1);
			if (value != null)
				result.add(value); // analyzer value at start of list
			result.addAll(set);
			return Strings.join(result);
		}
		return value;
```

This avoids the awkward insert at 0 which array copies the entire list.

rotty3000:
@bjhargrave updated

PR #2374, opened 2018-03-19
Processor: capture error/warning locations as well as the messages.

kriegfrj:
Make sure that when error() and warning() call location(), they do so on the current owner process instead of on themselves.

Includes a test case.

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/ProcessorTest.java
+++ b/biz.aQute.bndlib.tests/src/test/ProcessorTest.java
@@ -1,5 +1,7 @@
 package test;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
@@ -15,6 +17,7 @@ import aQute.bnd.osgi.resource.RequirementBuilder;
 import aQute.bnd.osgi.resource.ResourceBuilder;
 import aQute.bnd.osgi.resource.ResourceUtils;
 import aQute.lib.strings.Strings;
+import aQute.service.reporter.Reporter.SetLocation;
 import junit.framework.TestCase;
 
 public class ProcessorTest extends TestCase {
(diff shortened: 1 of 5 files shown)


biz.aQute.bndlib.tests/src/test/ProcessorTest.java
```diff
@@ -188,6 +191,44 @@ private void assertNative(String in, String filter, String... fixup) throws Exce
 		}
 	}
 
+	static class SeethroughProcessor extends Processor {
```
bjhargrave: Not sure what the purpose of this subclass is. It does not change the behavior of Processor at all. I think you can remove this subclass.

biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
```diff
@@ -290,7 +290,7 @@ public SetLocation warning(String string, Object... args) {
 		if (!p.warnings.contains(s))
 			p.warnings.add(s);
 		p.signal();
-		return location(s);
+		return p.location(s);
```
bjhargrave: Great job finding this subtle bug!

biz.aQute.bndlib.tests/src/test/ProcessorTest.java
kriegfrj: The accessibility on `Processor.beginHandleErrors()`/`endHandleErrors()` is protected. We can't invoke these methods from within the test method on a bare `Processor` object. Subclassing it was the easiest way to gain direct access to the methods for testing, as subclasses are allowed to widen the accessibility of overridden methods (hence calling it "Seethrough" - you can see the methods that ordinarily would be).

To be fair, this is not immediately obvious. If it helps, I can add a comment to this effect to explain to the next developer why this was done.

Another alternative would be to use reflection to bypass the accessibility check, but unless you have a good reflection helper API on the test classpath this code wouldn't be much easier to read.

bjhargrave: Of course. I did not see that. I guess a comment would help to explain. Thanks for educating me!

kriegfrj: Actually, perhaps a better alternative would be to move the unit test into the `aQute.bnd.osgi` package, and then the protected methods would be accessible from the test method without me needing to override it. I'm happy to implement whatever you suggest.

biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
kriegfrj: It's small fries compared with what you guys have done on the rest of the project, so I don't really have a right to accolades.

biz.aQute.bndlib.tests/src/test/ProcessorTest.java
bjhargrave: Use the subclass with a comment to explain why.

kriegfrj:
I've added the comment to the test class. Let me know if you have any other suggestions.

bjhargrave:
> I've added the comment to the test class.

Are you sure you pushed it? I don't see the comment in the commit in the PR.

kriegfrj:
Well, it looks like I pushed _something_ - but apparently it didn't include any changes! I must have rebased and then forgotten to actually commit the change or something. Apologies.

Should be there now.

Issue #2377, opened 2018-03-19
Issues with Ignored Integration Tests


karusher:
I have been executing Junit 4 integration tests inside Eclipse with bndtools v3.5.0-REL and Java 8. While doing so, I have noted the following issues with the handling of ignored testcases:

- Ignored testcases are not included in execution statistics
- Testsuites are displayed as being nested when they are not

Here is a sample of the Eclipse report:

![nested](https://user-images.githubusercontent.com/4839040/37612564-eb6343c8-2b7b-11e8-82cb-5c381746fd44.png)

# Root Cause Analysis

## Ignored testcases excluded from execution statistics
From what I can see, the issue stem from interactions between the plugin biz.aQute.junit and junit.framework.JUnit4TestAdapter.java

Within biz.aQute.junit, aQute.junit.Activator is responsible for building the list of testcases and testsuites. In aQute.junit.Activator.flatten, JUnit4TestAdapter provides a list of contained testcases:

From Activator.java:612 - commit c4b7830:

```
if (test instanceof JUnit4TestAdapter) {
  list.add(test);

  for (Test t : ((JUnit4TestAdapter) test).getTests()) {
    if (t instanceof TestSuite) {
      realCount += flatten(list, (TestSuite) t);
...
```

JUnit4TestAdaper.getTests() does not return a list of all testcases. It only returns a list of testcases which are **not** ignored. 

From JUnit4TestAdapter.java:52 - commit 25495b3

```
// reflective interface for Eclipse
public List<Test> getTests() {
    return fCache.asTestList(getDescription());
}

...

public Description getDescription() {
    Description description = fRunner.getDescription();
    return removeIgnored(description);
}
```
As a result of this behavior, ignored testcases are excluded from the flattened list. I believe this is why ignored testcases are not included in the execution statistics.

## Testsuites are displayed as falsely being nested

The flattened list is used by aQute.junit.JUnitEclipseReport to provide Eclipse a list of testcases via a message.

The entries of the message have the following format:

[ counter, test.stringOf(), (Is TestSuite or JUnit4TestAdapter), test.countTestCases() ]

For JUnit4TestAdapter, the implementation of JUnit4TestAdapter.countTestCases() gets the number of testcases including ignored. The implementation is below:

From JUnit4TestAdapter.java:43 - commit 25495b3
```
public int countTestCases() {
    return fRunner.testCount();
}
```
[...]

PR #2378, opened 2018-03-19
junit: Use getTests().size() for JUnit4TestAdapter

bjhargrave:
countTestCases() includes ignored tests.

Fixes https://github.com/bndtools/bnd/issues/2377


bjhargrave:
We are shooting for late April for the next release. You can test the fix now using the latest builds from cloudbees.

PR #2381, opened 2018-03-20
Fix formatting of iclause definition

Azquelt:
This section is rather important and collapses into an unreadable mess on the website.

Hopefully being in a code block will retain the formatting.

See http://bnd.bndtools.org/heads/include_resource.html for the current page.


bjhargrave:
Thanks!

Issue #2380, opened 2018-03-20
Bnd doesn't handle src.includes that aren't directories


kriegfrj:
This was the root cause of bndtools/bndtools#1862.

I'm porting an existing PDE project to Bnd. The PDE project had a build.properties file with a src.includes directive that includes several directories and files.

It appears that `Project` is assuming that all the entries in this property should be directories, and trying to create them if they don't exist:

https://github.com/bndtools/bnd/blob/920be8252daa3850755de049644102276fff6bf0/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L323-L325

Of course, the files in my PDE project did actually exist, hence the attempt to create the directory threw an exception, which caused the attempt to build the Project model to abort. Because just about everything in Bnd depends on the model, this basically means that nothing works.

There are two distinct issues here, I believe:

1. Bnd should not die so abruptly, but should handle the exception and move on as gracefully as possible.
2. More specifically in this case, how to properly handle the case when build.properties contains source paths that point to files.

It seems to be legal syntax for `src.includes` to contain files as well as directories. It is used by PDE to copy files into the final bundle. However, I'm not sure how bnd should handle it.

kriegfrj:
Of course there is a workaround, as I am porting to Bnd anyway, that I can use Bnd directives to copy all of the relevant files into the output jar instead of using the Eclipse build.properties file. But if this is supposed to work, then it should do so cleanly.

bjhargrave:
There is no `src.includes` property in Bnd. That is from PDE and has no special meaning to Bnd. Bnd has a `src` property which holds the source folders for a project. `src` is a merged property.

https://github.com/bndtools/bnd/blob/920be8252daa3850755de049644102276fff6bf0/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L307

This means that the `src` property value and the values from all the properties whose names start with `src.` (which `src.includes` does) are merged into the value of the `src` property. So Bnd expects that `src.includes` specifies additional source _folders_ for the project.

This is not a Bnd bug but a configuration error on your part. You cannot include a PDE build.properties into a bnd file when there are disjoint meanings for property values like `src`.

kriegfrj:
[...]

PR #2383, opened 2018-03-21
#2380 handle bnd files with src directories misconfigured to point at files

kriegfrj:
Gracefully handle source dirs that point at files, or that can't be successfully created.

Prior to this fix, the `error()` call that reported if it wasn't a directory was effectively unreachable - the mkdir() call would throw an exception and abort the entire Project.


biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -320,8 +320,21 @@ public void prepare() throws Exception {
 						continue;
 					}
 
-					if (!dir.isDirectory()) {
-						IO.mkdirs(dir);
+					if (!dir.exists()) {
+						try {
+							IO.mkdirs(dir);
+						} catch (Exception ex) {
+							error("could not create src directory (in src property) %s: %s", dir, ex)
```
bjhargrave: When you have an exception object, use the `exception` method instead of `error`.

biz.aQute.bndlib.tests/src/test/ProjectTest.java
```diff
@@ -762,6 +765,26 @@ public void testMultidirsrc() throws Exception {
...
+	/*
+	 * Verify that this also works when you have multiple directories
+	 */
+	public void testSrcPointsToFile() throws Exception {
+		Workspace ws = getWorkspace("testresources/ws");
+		Project p = ws.getProject("pfilesrc");
+		Collection<File> sourcePath = p.getSourcePath();
+		File a = p.getFile("a");
+		File b = p.getFile("b");
+		String expectedError = "the src path (src property) contains an entry that is not a directory "
```
bjhargrave: Generally our tests use the `check` method to test for the state of the processor and ensure certain messages are present. You will find examples of the use of the check method in many other tests.

kriegfrj: Happy to use `check()` if that's what you decide, but I offer the following observations of pros & cons vs the AssertJ approach that I used:

1. `check()` and `containsExactly()` will both give essentially the same diagnostic information about why the check failed (ie, which patterns were missing, and which elements were found that shouldn't have been). However, `containsExactly()` will report the diagnostic in the actual `AssertionError` instead of to stderr. Thus it won't pollute stderr with output that is subsequently hard to find (especially when running tests in bulk). When using it integrated with Eclipse JUnit runner, it also means that if the assertion fails, you'll be able to click on the failed test and immediately see the corresponding diagnostic message instead of having to wade through stderr to find it.
2. `check()` doesn't check to see if the error(s)' location(s) have been set properly, like this implementation does.
3. `check()` doesn't facilitate checking the severity level of the individual problems - it matches patterns against both errors and warnings. As the test currently stands, the test is a guarantee that the problem was reported as an error; if I use `check()` instead then the test can't guarantee that the problem wasn't reported as a warning instead of an error.

Anyway, I am new on this project and you're the boss - if you feel strongly about using `check()`, or if there are other reasons for using `check()` that I am not aware of, I'm more than happy to do it that way instead.

bjhargrave: I know your check is extremely precise. But precision is non-free. Do we really need to the test here to be that precise? Since the test is very focused, we only really need to see that there is an error. The precises wording of the message may get improved later and then this precise test case will break. Some lubrication is nice to reduce friction between the code and its tests. check with its pattern matching allows for some lubrication. Sometimes you need to be very precise but not always.

kriegfrj: I have relaxed it & used check. Would be nice if we could find an error using an RE and then have that method return the Location object associated with the matched error, so that we could check the right location is set (best of both worlds).

Issue #2404, opened 2018-04-06
bnd-baseline-maven-plugin: don't mark the base parameter as read-only


kwin:
The base should not be marked as readonly as otherwise features like code completion in Eclipse's pom.xml editor do not work correctly (https://github.com/bndtools/bnd/blob/d7511968b0ec96f1bec5bd3de59f8010cb3cb9f1/maven/bnd-baseline-maven-plugin/src/main/java/aQute/bnd/maven/baseline/plugin/BaselineMojo.java#L69). In addition that should be documented as explicit parameter also in https://github.com/bndtools/bnd/tree/master/maven/bnd-baseline-maven-plugin#configuration-properties.

Also it would be nice if all the parameter would have a short javadoc as that is evaluated/exposed as well in IDEs when using code completion.

bjhargrave:
Can you please make a PR with these fixes?

bjhargrave:
> Can you please make a PR with these fixes?

Nevermind, I made the PR.

PR #2405, opened 2018-04-06
maven: Remove readonly=true for better Eclipse pom editing experience

bjhargrave:
Also add some javadoc to configuration methods.

Fixes https://github.com/bndtools/bnd/issues/2404


bjhargrave:
> Eclipse\Maven takes the javadoc from the annotated field not the setter method. Could you please add a minimal documentation to each field annotated with @Parameter?

Sigh. Let do that in another PR.

PR #2434, opened 2018-05-01
Add tests for using the Directive meta-annotation

timothyjward:
Fix the identified problem in Attrs

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/annotationheaders/StdAnnotationHeadersTest.java
+++ b/biz.aQute.bndlib.tests/src/test/annotationheaders/StdAnnotationHeadersTest.java
@@ -148,6 +148,14 @@ public class StdAnnotationHeadersTest extends TestCase {
 			assertTrue(p.containsKey("number:Long"));
 			assertEquals("42", p.get("number:Long"));
 			assertFalse(p.containsKey("ignoredName"));
+			assertTrue(p.containsKey("x-open:"));
+			assertEquals("seed", p.get("x-open:"));
+			assertTrue(p.containsKey("x-anotherUsedName:"));
+			assertEquals("foo,bar", p.get("x-anotherUsedName:"));
+			assertTrue(p.containsKey("x-anotherNumber:"));
+			assertEquals("17", p.get("x-anotherNumber:"));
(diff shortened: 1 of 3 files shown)


biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
```diff
@@ -214,7 +214,13 @@ public void putTyped(String key, Object value) {
 			} else {
 				type = getObjectType(value);
 			}
-			key += ":" + type.toString();
+			if (key.endsWith(":")) {
+				// This is a directive
+				key += type.toString() + ":";
```
bjhargrave: This is wrong. Directives are never typed. They are always String type. Only attributes can be typed. See https://osgi.org/specification/osgi.core/7.0.0/framework.module.html#d0e2821.

timothyjward: That's good to know - in which case the Attrs is still doing the wrong thing. Should it be an exception to putTyped with a Directive (ending in a `:`), or should it simply not add a type?

timothyjward: For reference, the existing behaviour is to output

    x-directiveName::List<String>="foo, bar"

when it encounters something like:

    @Directive("x-directiveName")
    String[] value;

bjhargrave: I am working on a change to Attrs now...

bjhargrave: My change is in https://github.com/bndtools/bnd/pull/2435

bjhargrave: @timothyjward I merged #2435 so you can rebase on the tip of master removing your changes to Attrs. My change just avoids writing the type for directives since they must be String. So 

```
@Directive("x-directiveName")
String[] value;
```

should be emitted as:

```
x-directiveName:="foo,bar"
```

But frankly using a type other than String for an @Directive annotation element is strange.

PR #2460, opened 2018-05-29
Modify the Tag class to accept DTO object

cdelg:
See issue #2138 

### DIFF
--- a/aQute.libg/src/aQute/lib/tag/Tag.java
+++ b/aQute.libg/src/aQute/lib/tag/Tag.java
@@ -2,6 +2,9 @@ package aQute.lib.tag;
 
 import java.io.PrintWriter;
 import java.io.StringWriter;
+import java.lang.reflect.Array;
+import java.lang.reflect.Field;
+import java.lang.reflect.Modifier;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -10,8 +13,12 @@ import java.util.Date;
 import java.util.Formatter;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.regex.Pattern;
+import java.util.stream.Stream;
 
 /**
  * The Tag class represents a minimal XML tree. It consist of a named element
@@ -21,16 +28,19 @@ import java.util.regex.Pattern;
  */
 public class Tag {
 
-	final static String				NameStartChar	= ":A-Z_a-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF]\uFDF0-\uFFFD";
-	final static String				NameChar		= "[" + NameStartChar + "0-9.\u00B7\u0300-\u036F\u203F-\u2040\\-]";
-	final static String				Name			= "[" + NameStartChar + "]" + NameChar + "*";
-	final public static Pattern		NAME_P			= Pattern.compile(Name);
+	final private static String		ARRAY_ELEMENT_NAME	= "element";
 
-	Tag								parent;																																											// Parent
-	String							name;																																											// Name
-	final Map<String, String>		attributes		= new LinkedHashMap<>();
-	final List<Object>				content			= new ArrayList<>();																																			// Content
-	final static SimpleDateFormat	format			= new SimpleDateFormat("yyyyMMddHHmmss.SSS");
+	final static String				NameStartChar		= ":A-Z_a-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02FF\u0370-\u037D\u037F-\u1FFF\u200C-\u200D\u2070-\u218F\u2C00-\u2FEF\u3001-\uD7FF\uF900-\uFDCF]\uFDF0-\uFFFD";
+	final static String				NameChar			= "[" + NameStartChar
+		+ "0-9.\u00B7\u0300-\u036F\u203F-\u2040\\-]";
+	final static String				Name				= "[" + NameStartChar + "]" + NameChar + "*";
+	final public static Pattern		NAME_P				= Pattern.compile(Name);
+
+	Tag								parent;																																												// Parent
+	String							name;																																												// Name
+	final Map<String, String>		attributes			= new LinkedHashMap<>();
+	final List<Object>				content				= new ArrayList<>();																																			// Content
+	final static SimpleDateFormat	format				= new SimpleDateFormat("yyyyMMddHHmmss.SSS");
 	boolean							cdata;
 
 	/**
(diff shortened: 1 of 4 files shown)


aQute.libg/test/aQute/lib/tag/TagTest.java
```diff
@@ -0,0 +1,289 @@
...
+public class TagTest extends TestCase {
+
+	public void testDtoToTag() {
+		Tag t;
+		ListIterator<Object> l;
+		Tag c;
+		Map<String,Object> map;
+		Map<String,Object> map2;
+
+		// Test Scalar
```
bjhargrave: Can you break this test method in a  number of test methods? Each // comment seems like a good split point.

aQute.libg/src/aQute/lib/tag/Tag.java
```diff
@@ -131,6 +137,84 @@ public Tag addContent(Tag tag) {
 		return this;
 	}
 
+	/**
+	 * Add new content tags derived from the DTO value.
+	 */
+	public Tag addContent(String name, Object dto) {
```
bjhargrave: This method is quite different than the other add methods. They leave all the "work" to the print time logic. I would think we should do the same for new content types.

aQute.libg/src/aQute/lib/tag/Tag.java
```diff
@@ -131,6 +137,84 @@ public Tag addContent(Tag tag) {
...
+		return false;
+	}
+
+	private boolean isComplex(Object a) {
+		return a instanceof Map || a instanceof Collection || a.getClass().isArray()
+				|| getFields(a.getClass()).length > 0;
+	}
+
+	private Field[] getFields(Class< ? > c) {
+		List<Field> publicFields = new ArrayList<>();
```
bjhargrave: This method is better implemented using Streams.

aQute.libg/src/aQute/lib/tag/Tag.java
bjhargrave: Or, in the alternate, perhaps this method be a Tag constructor like `Tag(String name, Object... contents)`. The current approach is a hybrid.

aQute.libg/src/aQute/lib/tag/Tag.java
```diff
@@ -131,6 +137,84 @@ public Tag addContent(Tag tag) {
...
+
+		for (Field field : c.getFields()) {
+			if (field.isEnumConstant() || field.isSynthetic() || Modifier.isStatic(field.getModifiers()))
+				continue;
+			publicFields.add(field);
+		}
+
+		return publicFields.toArray(new Field[publicFields.size()]);
```
bjhargrave: Using size in the array allocation as input to `toArray` is bad practice for performance reasons. Use `0` as the size.

aQute.libg/test/aQute/lib/tag/TagTest.java
```diff
@@ -0,0 +1,289 @@
...
+		t = getTag(null);
+		assertTrue(t.content.size() == 0);
+
+		t = getTag("");
+		assertTrue(t.content.size() == 1);
+		assertTrue(((Tag) t.content.listIterator().next()).content.size() == 1);
```
bjhargrave: You may want to consider using [`assertj`](http://joel-costigliola.github.io/assertj/assertj-core.html) for these tests since it is easier to test things about lists like size, content, etc.

aQute.libg/src/aQute/lib/tag/Tag.java
cdelg: So what do you prefer: first, second or both ?

bjhargrave: I think I would prefer the constructor approach since there are already a number of constructors for making Tags from a variety if inputs. A DTO-like object can be an another input. In fact, I think the current `Tag(String name, Object... contents)` constructor can be adapted to handle contents objects which are DTO-like.

cdelg: After looking at, I don't see the specific "work" that is done at print time and not in the add* methods, there are simple (no work to be done). The constructor  `Tag(String name, Object... contents)` already exists, do you want to change its behavior?

The current implementation only convert complex objects to Tag, what other solution is possible while keeping the `getContent` working?

bjhargrave: Well the constructor could convert a DTO-like object to a set of Tags which would be visible then via getContents. Or the work can be done in `print` to write the appropriate XML output for a DTO-like content object.

cdelg: The actual constructor allows any objects as soon as they have a `toString` method that produce a valid Xml element. I will not be able to determine if the object must be treated as a standard object or as a DTO (eg; a subclass of List that redefines the `toString`). The other issue with that is when the user provide a List of **n** objects. The current implementation adds **n** tags with the provided tag name as element name. With the constructor I cannot do that.

This problem occurs also if we do the conversion work in the `print` method. 

An other solution could be to rename the current method with a more explicite name? Any idea?

bjhargrave: Perhaps the answer would be a static method instead of a constructor. The static method would return a Tag object from the DTO-like argument. The returned Tag can then be added to another Tag with `addContent(Tag)`.

cdelg: Yes, except that the method have to return a list of tags for the case that I describe previously (a list of object). The signature can be `List<Tag> toTags(String name, Object dto)` and consumers will have an extra step to add all these Tags to another Tag. If you see a scenario where users want to convert a DTO to Tags but not add them to another Tag...

bjhargrave: Perhaps I don't fully understand what you are trying to do here. You suggest the signature `List<Tag> toTags(String name, Object dto)` which supplies one name but returns multiple Tags. Do all these Tags share that same name? I was thinking that one DTO-like object would be mapped to one Tag object. Can you show some examples of the different mappings that would need to return multiple Tag objects for one DTO-like object?

cdelg: A List is a DTO-Like, but don't have a key to be used as Tag name.

#### `List<Tag> toTags("person", [{name:Bob, age:55}, {name:Rick, age:42}])`:
**Result:**
```xml
<person>
    <name> Bob </name>
    <age> 55 </age>
</person>
<person>
    <name> Rick </name>
    <age> 42 </age>
</person>
```


#### `Tag toTag("person", [{name:Bob, age:55}, {name:Rick, age:42}])`:
**Result:**
```xml
<person>
    < ?? >
        <name> Bob </name>
        <age> 55 </age>
    </ ?? >
    < ?? >
        <name> Rick </name>
        <age> 42 </age>
    </ ?? >
</person>
```

or

```xml
<person>
    <name> Bob </name>
    <age> 55 </age>
    <name> Rick </name>
    <age> 42 </age>
</person>
```

bjhargrave: I would expect

#### `Tag toTag("person", [{name:Bob, age:55}, {name:Rick, age:42}])`:
**Result:**
```xml
<persons>
    <person>
        <name>Bob</name>
        <age>55</age>
    </person>
    <person>
        <name>Rick</name>
        <age>42</age>
    </person>
</persons>
```

But even so, having a single name argument falls apart with a DTO-like input:

`{friends: [{name:Bob, age:55}, {name:Rick, age:42}]}`

For that DTO, I would like:

```xml
<friends>
    <friend>
        <name>Bob</name>
        <age>55</age>
    </friend>
    <friend>
        <name>Rick</name>
        <age>42</age>
    </friend>
</friends>
```
as the XML.

Here, we are really getting into deserializing a DTO-like object into XML and there are many ways...

We may need two methods. One for List-like things (arrays and collections) which need some name for the top level tag and the element tags(e.g. `fromList(String elementName, Object listLike)`. And one for Map-like things (maps and public fields) where each item has a name (e.g. `fromMap(Object mapLike)`.

The `fromList` method can pluralize/depluralize the elementName to make the group tag (plural; e.g. `friends`) and the element tags (non-plural; e.g. `friend`).

bjhargrave: Building upon the ideas above, I like the method signature:

`fromDTO(String rootName, String arrayElementName, Object dto)`

(`fromDTO` is a better method name since it describes the source object and if we add more such static methods for other source object types, we will need to use different method names. `toTag` is too non-specific. All of these static methods are about making Tag objects.)

However, for arrays which have a name, we should use the depluralized name for the array elements rather than just always using the arrayElementName value.

Example: 
`fromDTO("MyTagNames", "element", [ "Bob", {"sistersNames": ["Amy", "Emily"] }, ["one", "two"]])`

This would give:

```xml
<MyTagNames>
  <MyTagName>Bob</MyTagName>
  <MyTagName>
    <sistersNames>
      <sistersName>Amy</sistersName>               
      <sistersName>Emily</sistersName>
    </sistersNames>
  </MyTagName>
  <MyTagName> 
    <element>one</element>
    <element>two</element>
  </MyTagName>
</MyTagNames>
```

where the top level array has the name "MyTagNames" and thus each element is "MyTagName". Similarly for the array named "sistersNames", each element can have the name "sistersName". Only for an array which has no name itself, do we fall back to the arrayElementName.

cdelg: Ok for the method name.

For the depluralized part, I find it a bit brittle:
* The user of the resulting XML may not be in control of the data source, eg; Json (from someone else) -> DTO -> XML. If in the Json file  keys are not pluralized, elements will have a truncated name.
* Some words are invariable some others are exceptions (child -> children)
* A key can be a single character.

Another solution will be to capitalize the first letter of `arrayElementName` and append it for Arrays with names:

```xml
<MyTagNames>
  <MyTagNamesElement>Bob</MyTagNamesElement>
  <MyTagNamesElement>
    <sistersNames>
      <sistersNamesElement>Amy</sistersNamesElement>               
      <sistersNamesElement>Emily</sistersNamesElement>
    </sistersNames>
  </MyTagNamesElement>
  <MyTagNamesElement> 
    <element>one</element>
    <element>two</element>
  </MyTagNamesElement>
</MyTagNames>
```

I don't know if it is more readable but for processing, the elements of a list can be found  more easily. That still involves  to modify a tag name which can be problematic if the convention is not camel case but for example `MY_TAG_NAME`. (but less difficult to manage than depluralized)

cdelg: I agree that for Lists of Lists (which is rather unusual) it is useful to enclose the elements of the Lists to not lose the list structure , but for the other cases?
I mean, is it for a Human readability, XML validation or for computer processing? Because:
* XML validation: I think this is not the purpose of the Tag class? For a general DTO to Xml convertion the process should rely on dtd rules or Xml schema to make good choices.
*  Computer processing: Is there any advantage to enclose lists?
* Human readability: There is no real definite choice on the subject.

bjhargrave: > Another solution will be to capitalize the first letter of arrayElementName and append it for Arrays with names:

This could be done IF depluralization cannot be done. But attempt to depluralize first.

cdelg: One more thing. For null values and null keys in maps, is the following good ?

* If the DTO in argument is null, return an empty element.
* If a DTO contains null values (List, Map, Object) do not create an element (skip)
* If a Map constains a null key create the tag with a null name.

bjhargrave: > If a Map constains a null key create the tag with a null name.

Does that mean the name "null"? I guess I would chose to omit the map entry from the xml.

bjhargrave:
Also, see https://github.com/bndtools/bnd/pull/2476/commits/f212eb6969d9069180f5fd97f214f99255922953#diff-77abe44342e718f977ef3c84e1897693 where Peter has implemented what could be considered the reverse of this. That class is reading xml into a DTO object.

cdelg:
done

cdelg:
Do you know why the build failed? It seems to be the maven plugin.

PR #2475, opened 2018-06-15
[libg] Addition of a number of utilities in lib

pkriens:
- Helper to set annotation values
- Mangle method name -> property key updated to OSGi
- Files check for directory before descent
- UTF8Properties from file constructor
- A DTO parser on a XML DOM
- SHA512 class for crypto
- Multiple glob checking
- Configuration interface proxy

### DIFF
--- a/aQute.libg/src/aQute/libg/glob/Glob.java
+++ b/aQute.libg/src/aQute/libg/glob/Glob.java
@@ -2,6 +2,7 @@ package aQute.libg.glob;
 
 import java.io.File;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.regex.Matcher;
@@ -140,5 +141,21 @@ public class Glob {
 			} else if (recursive && sub.isDirectory())
 				getFiles(sub, result, recursive, usePath);
 		}
+	}
+
+	public static boolean in(Glob[] globs, String key) {
+		for (Glob g : globs) {
+			if (g.matcher(key).matches())
+				return true;
+		}
+		return false;
+	}
+
+	public static boolean in(Collection< ? extends Glob> globs, String key) {
+		for (Glob g : globs) {
+			if (g.matcher(key).matches())
+				return true;
+		}
+		return false;
 	}
 }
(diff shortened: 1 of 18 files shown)


pkriens:
I did them in a row because I do not want to it one by one over the next weeks. At least I've got them in sequence now. (You are sure the second PR is not automatically ok?)

aQute.libg/src/aQute/libg/glob/Glob.java
```diff
@@ -141,4 +142,20 @@ public void getFiles(File root, List<File> result, boolean recursive, boolean us
 				getFiles(sub, result, recursive, usePath);
 		}
 	}
+
+	public static boolean in(Glob[] globs, String key) {
+		for (Glob g : globs) {
```
bjhargrave: Isn't this more simply?

```java
return Stream.of(globs).anyMatch(g -> g.matcher(key).matches());
```

aQute.libg/src/aQute/libg/glob/Glob.java
```diff
@@ -141,4 +142,20 @@ public void getFiles(File root, List<File> result, boolean recursive, boolean us
...
+			if (g.matcher(key).matches())
+				return true;
+		}
+		return false;
+	}
+
+	public static boolean in(Collection< ? extends Glob> globs, String key) {
+		for (Glob g : globs) {
```
bjhargrave: Isn't this more simply?

```java
return globs.stream().anyMatch(g -> g.matcher(key).matches());
```

pkriens: Maybe ... but I hate streaming when you debug. I do not think my code is less efficient (on the contrary I think) or harder to read. I loved streams but somehow I often revert back to old style to be able to single step it.

And I am not sure we should review these PRs for these coding styles because that can become a lot of work. I've no problem if you want to change these things since all works now but I really do not want to make these changes in this commit.

bjhargrave: You are right. This is a very complicated function... (sigh)

Never mind about my review.

PR #2476, opened 2018-06-15
[repo] repositoryupdates

pkriens:
repository] Updates of P2 and MavenBndRepository

- Option to allow redeploy on maven repo
- Option to ignore maven metadata in bundle when put in RepositoryPlugin
- If META-INF/maven/<group>/artifact is used in a put, the POM in there is ignored and the properties are used. The POM could have invalid macro references.
- Nexus ls + OSGi index facility of nexus repos with crawl and REST API
- P2 support for target platform repo
- NPE fixes


eliasvasylenko:
Thank you for the target platform support, that's a great Idea. I assume this just means the repository contents are specified via a target platform file? This will really help keep resolver times from exploding I'm sure...

pkriens:
Yup, the XML of a target platform is (or should be) directly be reusable in a P2 Repository

biz.aQute.repository/src/aQute/maven/provider/POM.java
```diff
@@ -58,6 +58,9 @@
 	private Map<Program, Dependency>	dependencies			= new LinkedHashMap<>();
 	private Map<Program, Dependency>	dependencyManagement	= new LinkedHashMap<>();
 	private XPath						xp;
+	private String						JAR_PACKAGING[]			= {
```
bjhargrave: We normally use `String[] var;` not `String var[];`.

biz.aQute.repository/src/aQute/maven/provider/POM.java
```diff
@@ -347,7 +350,7 @@ private String replaceMacros(String value) {
 					property = replaceMacros(property);
 
 				if (property == null) {
-					System.out.println("?? prop: " + key);
+					l.info("Undefined prop erty in {} : key {}", this, key);
```
bjhargrave: Space in `prop erty`.

biz.aQute.repository/src/aQute/p2/api/package-info.java
```diff
@@ -1,4 +1,4 @@
-@Version("1.1.0")
+@Version("1.3.0")
```
bjhargrave: Why bump minor by 2?

biz.aQute.repository/src/aQute/p2/provider/P2Impl.java
```diff
@@ -50,21 +51,40 @@ public P2Impl(HttpClient c, URI base, PromiseFactory promiseFactory) throws Exce
...
 	}
 
 	private Promise<List<Artifact>> getArtifacts(Set<URI> cycles, URI uri) {
 		if (!cycles.add(uri)) {
-			return promiseFactory
-				.failed(new IllegalStateException("There is a cycle in the p2 setup : " + cycles + " -> " + uri));
+			return promiseFactory.resolved(Collections.emptyList());
+
+			//
+			// TODO discuss with BJ
```
bjhargrave: When will this discussion occur?

biz.aQute.repository/test/SomeTest.java
```diff
@@ -0,0 +1,11 @@
+import org.junit.Test;
+
+public class SomeTest {
```
bjhargrave: What is the point of this file? It does nothing except add one more passing test.

biz.aQute.repository/test/aQute/bnd/repository/p2/provider/P2RepositoryTest.java
```diff
@@ -36,4 +42,21 @@ public void testSimple() throws Exception {
 		}
 	}
 
+	// @Test does not work for some reason
+	public void testXtext() throws Exception {
```
bjhargrave: So this is a test which is not executed?

biz.aQute.repository/src/aQute/maven/provider/POM.java
pkriens: I've changed it because I do not want to fight over it but if you want to enforce these things you should start a style guide where we can agree on these things because this feels very arbitrary.

biz.aQute.repository/src/aQute/p2/provider/P2Impl.java
pkriens: We can do it here. I think the cycle detection should not throw an exception. Aborting that branch of the tree suffices as far as I can see. Do I miss something?

biz.aQute.repository/test/aQute/bnd/repository/p2/provider/P2RepositoryTest.java
pkriens: Thanks for keeping me honest. There was actually quite a fundamental flaw in the P2Impl ... It resolved against the base repo, not against the repo that contained the compositeArtifacts.xml we were parsing. A lot of complex P2 repos were not working, only simple P2 repos worked well.

biz.aQute.repository/test/aQute/bnd/repository/p2/provider/P2RepositoryTest.java
```diff
@@ -13,13 +13,15 @@
...
 import aQute.bnd.osgi.Processor;
+import aQute.bnd.service.progress.ProgressPlugin;
 import aQute.lib.io.IO;
 
 public class P2RepositoryTest {
 	File tmp = IO.getFile("generated/tmp");
 
 	@Before
 	public void setUp() {
+		System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", "warn");
```
bjhargrave: This is kind of a big side effect. If you need it in setUp, you better restore the original property in `tearDown`. But I am guessing we don't really need it, it was just for test debugging.

biz.aQute.repository/test/aQute/bnd/repository/p2/provider/P2RepositoryTest.java
```diff
@@ -42,19 +44,47 @@ public void testSimple() throws Exception {
 		}
 	}
 
-	// @Test does not work for some reason
+	@Test
 	public void testXtext() throws Exception {
+
 		try (P2Repository p2r = new P2Repository()) {
 			Workspace w = Workspace.createStandaloneWorkspace(new Processor(), tmp.toURI());
+			w.addBasicPlugin(new ProgressPlugin() {
```
bjhargrave: Is this plugin all just for debugging since it does not test anything. If it is just for test debugging, we should probably not leave it in.

biz.aQute.repository/src/aQute/p2/provider/P2Impl.java
```diff
@@ -75,20 +75,11 @@ private URI normalize(URI base) throws Exception {
 	private Promise<List<Artifact>> getArtifacts(Set<URI> cycles, URI uri) {
 		if (!cycles.add(uri)) {
 			return promiseFactory.resolved(Collections.emptyList());
```
bjhargrave: This should go back to how it was (returning a failed promise). The failed promise is properly handled at https://github.com/bndtools/bnd/blob/ce27a28a42a4cc3e9ca4740417ed99c4d3c8ffd2/biz.aQute.repository/src/aQute/p2/provider/P2Impl.java#L128-L133 where a message may be logged and promise result converted to an empty list.

pkriens: But I do not want a message?

pkriens: Looking at P2 cycles seem to be common in complex repos

bjhargrave: So is a cycle a normal possibility and not an error condition?

pkriens: Yup, they nowadays composite repositories that point to a very large number of other repositories. So it is likely not uncommon. It would be extremely hard to control there are no cycles since they often just point to a versionless URI.

So it should not generate errors imho.

bjhargrave: OK, then your change makes sense.

PR #2488, opened 2018-06-20
[repo] Repo update

pkriens:
Same as #2476 but fixes empty POM problems  that caused it to be reverted in #2486


pkriens:
Thanks, sorry for the mistake.

PR #2169, opened 2017-10-14
Add report generation feature

cdelg:
Closes #2138

Add the report generation feature. This feature allows to generate various kind of reports from the bnd command line.

## Use case

* Generate readme.md files for the workspace and its projects.
* Generate a data model file of the entire workspace that can be used by a site generator such as Jekyll.
* Generate an html file that compiles all the metatypes defined in the workspace.
* Generate any reports you like from a Jar.  
* ...  

## Extraction and Generation

The report generation is a two steps process. First metadata of a bnd workspace, a bnd project or a bundle are extracted, then, reports are generated either by directly serialize this metadata (the model) to the desired format or by applying a transformation with a template.

### Metadata extraction

You can configure the extraction process of a workspace, a project and a bundle with plugins and respectively with the following instructions: `-report.model.workspace`,`-report.model.project `,`-report.model.bundle`.

#### Common properties

* `imports`: a list of file paths to include into the report. By default, the file name will be used as entry name into the report and its extension to call the proper deserializer plugin but this can be overridden with the following construct: `<path>:entryName:format`
* `@ anyprop`: any properties that starts with `@` are added to the report.
* `includes`: a list of plugin names which will contribute to the report. Default is all the declared plugins.
* `excludes`: a list of plugin names which will be ignored when building the report. Default is empty.

#### `-report.model.bundle` properties

* `imports`: if a path starts with `@`, it will be resolved from the root of the extracted bundle.
* `locales`: a list of locales, eg "en_US", used to extract the bundle metadata in different languages.  

#### Plugin APIs

* `aQute.bnd.service.reporter.BundleEntryPlugin`: Extract and contribute to a bundle report. 
* `aQute.bnd.service.reporter.ProjectEntryPlugin`: Extract and contribute to a project report. 
* `aQute.bnd.service.reporter.WorkspaceEntryPlugin`: Extract and contribute to a workspace report. 
* `aQute.bnd.service.reporter.ReportImportDeserializerPlugin`: Convert a file into a DTO representation.

#### Plugin Implementations

* `biz.aQute.bnd.reporter.plugins.ManifestEntryPlugin`: Extract the bundle manifest. *Set by default*.
* `biz.aQute.bnd.reporter.plugins.ComponentsEntryPlugin`: Extract the bundle declarative services.
* `biz.aQute.bnd.reporter.plugins.MetatypesEntryPlugin`: Extract the bundle metatypes.
[...]

PR #2491, opened 2018-06-20
[bndlib] Add reporter plugins APIs

cdelg:
- There is 3 plugins  to extract metadata, one at the bundle level, one at the project level and one for the workspace level.
- 1 to serialize the metadata given a file extension. (xml, json, ..)
- 1 to dezerialize imported files given their extensions. (xml, json, property file, ...)
- 1 to transform metadata (xslt, twig, ...)

See #2138 


biz.aQute.bndlib/src/aQute/bnd/service/reporter/BundleEntryPlugin.java
```diff
@@ -0,0 +1,41 @@
...
+	 * locale must be as specified in {@link Locale}, an empty {@code String}
+	 * means {@code unlocalized}.
+	 * 
+	 * @param jar the Jar to inspect, must not be {@code null}
+	 * @param locale the {@code String} representation of a {@code Locale}, must
+	 *            not be {@code null}
+	 * @param reporter used to report error, must not be {@code null}
+	 * @return a DTO representation or {@code null} if no data is available
+	 */
+	public Object extract(final Jar jar, final String locale, final Processor reporter) throws Exception;
```
bjhargrave: The use of Object as the return type in these plugins is a bit disconcerting. It means that only reflective access to the DTOs is possible. Why would we not want to define the DTO types and then return them here?

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportImportDeserializerPlugin.java
```diff
@@ -0,0 +1,31 @@
...
+
+	/**
+	 * Deserialize the input stream into a DTO representation.
+	 * 
+	 * @param input the stream to deserialize, must not be {@code null}
+	 * @return a DTO representation of the input stream content or {@code null}
+	 *         if the stream is empty
+	 * @throws Exception if any errors occur during the deserialization process
+	 */
+	public Object deserialyze(InputStream input) throws Exception;
```
bjhargrave: I would have thought the method would take the DTO type to create. Something like.

```java
public <D> D deserialize(Class<D> type, InputStream input);
```

Otherwise, what is the type of the returned object? Or are you planning on encoding that into the serialized form like Java serialization?

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportSerializerPlugin.java
```diff
@@ -0,0 +1,31 @@
...
+
+	/**
+	 * Serialize the metadata DTO into the output stream.
+	 * 
+	 * @param metadata the metadata to serialize, must not be {@code null}
+	 * @param output the output stream to write the serialization result, must
+	 *            not be {@code null}
+	 * @throws Exception if any errors occur during the serialization process
+	 */
+	public void serialize(Map<String,Object> metadata, OutputStream output) throws Exception;
```
bjhargrave: Why isnt this

```java
public void serialize(Object dto, OutputStream output);
```
 Since this is about serializing a DTO-like object and not just a map?

bjhargrave: Since the plugin can handle multiple formats, what format is the serialize method to generate? Perhaps the requested format should be a parameter to the serialize method?

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportTransformerPlugin.java
```diff
@@ -0,0 +1,47 @@
...
+	 * Transform the model by applying the template on it and write the result
+	 * to the output stream.
+	 * 
+	 * @param model an input stream that contains the model, must not be
+	 *            {@code null}
+	 * @param template an input stream that contains the template, must not be
+	 *            {@code null}
+	 * @param output the output stream to write the transformation result, must
+	 *            not be {@code null}
+	 * @param parameters a map of parameters and their value that must be
```
bjhargrave: This is not a very type safe way to pass parameter. You might be better of accepting an Object argument which is a DTO type  defined by the template engines to pass parameters in a type safe way.

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportImportDeserializerPlugin.java
```diff
@@ -0,0 +1,31 @@
...
+import java.io.InputStream;
+
+import org.osgi.annotation.versioning.ProviderType;
+
+/**
+ * This plugin deserialize a stream of a specific format into a DTO
+ * representation.
+ */
+@ProviderType
+public interface ReportImportDeserializerPlugin {
```
bjhargrave: Isn't this more generally `ReportDeserializerPlugin`?

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ProjectEntryPlugin.java
```diff
@@ -0,0 +1,32 @@
...
+public interface ProjectEntryPlugin {
+
+	/**
+	 * Extracts a piece of information from the BND project in arguments.
+	 * 
+	 * @param project the project to inspect, must not be {@code null}
+	 * @param reporter used to report error, must not be {@code null}
+	 * @return a DTO representation or {@code null} if no data is available
+	 */
+	public Object extract(final Project project, final Processor reporter) throws Exception;
```
bjhargrave: The use of Object as the return type in these plugins is a bit disconcerting. It means that only reflective access to the DTOs is possible. Why would we not want to define the DTO types and then return them here?

biz.aQute.bndlib/src/aQute/bnd/service/reporter/WorkspaceEntryPlugin.java
```diff
@@ -0,0 +1,32 @@
...
+public interface WorkspaceEntryPlugin {
+
+	/**
+	 * Extracts a piece of information from the BND workspace in arguments.
+	 * 
+	 * @param workspace the workspace to inspect, must not be {@code null}
+	 * @param reporter used to report error, must not be {@code null}
+	 * @return a DTO representation or {@code null} if no data is available
+	 */
+	public Object extract(final Workspace workspace, final Processor reporter) throws Exception;
```
bjhargrave: The use of Object as the return type in these plugins is a bit disconcerting. It means that only reflective access to the DTOs is possible. Why would we not want to define the DTO types and then return them here?

biz.aQute.bndlib/src/aQute/bnd/service/reporter/BundleEntryPlugin.java
```diff
@@ -0,0 +1,41 @@
...
+	/**
+	 * Extracts a piece of information from the Jar in arguments.
+	 * <p>
+	 * If the Jar contains localized data, it will be extracted for the
+	 * specified locale or a less specific if not found. The syntax of the
+	 * locale must be as specified in {@link Locale}, an empty {@code String}
+	 * means {@code unlocalized}.
+	 * 
+	 * @param jar the Jar to inspect, must not be {@code null}
+	 * @param locale the {@code String} representation of a {@code Locale}, must
```
bjhargrave: Shouldn't this param be of type `Locale`?

bjhargrave: Why don't the extract methods of the project and workspace plugins also need locale params?

biz.aQute.bndlib/src/aQute/bnd/service/reporter/BundleEntryPlugin.java
cdelg: The consumer just uses these plugins to collect DTO and serialization.

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportImportDeserializerPlugin.java
cdelg: Same here, this plugin is made to import a file and to return a DTO which is collected for the report, then the report will be serialized (with a serializer plugin).

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportSerializerPlugin.java
cdelg: > Since this is about serializing a DTO-like object and not just a map?

The metadata root is of type Map<String,Object>, this plugin is here to serialize the report in this context, it is not a general plugin that serializes DTO.

>Since the plugin can handle multiple formats, what format is the serialize method to generate? Perhaps the requested format should be a parameter to the serialize method?

The provider implements a specific serialization format and provide one or multiple standard extension names for that format (with the `getHandledExtensions`, perhaps this would be component properties in DS). Multiple implementations of this plugin are collected by the report generator processor and based on user configuration, it selects the good implementation (which matches the desired format).

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportTransformerPlugin.java
cdelg: The parameters are not for the template engine itself, but to be inserted as parameter in the template file for this specific transformation. So the type of the parameters are handled by the template engine that read the template file. For xslt, parameters correspond to `<xsl:param>` for the jtwig engine there is no concept of parameters, so this map is added to the data model for this specific transformation.

Parameters are defined by the user in bnd.bnd files.

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportImportDeserializerPlugin.java
cdelg: We do not deserialize a report, but user can import file content to their report (eg, pom.xml, or pom.properties). So this plugin is made to deserialize imported file.

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ProjectEntryPlugin.java
cdelg: Same as for BundleEntryPlugin.

biz.aQute.bndlib/src/aQute/bnd/service/reporter/WorkspaceEntryPlugin.java
cdelg: Same as for BundleEntryPlugin.

biz.aQute.bndlib/src/aQute/bnd/service/reporter/BundleEntryPlugin.java
cdelg: > Shouldn't this param be of type Locale?

This is because, as far as I remember, a Local object cannot represent an unlocalized form. So in the case where user is not interested in localized data, the argument as to be null. But I think we can switch if you prefer.

> Why don't the extract methods of the project and workspace plugins also need locale params?
I know that bundle defines standard way to provide localized data (manifest, metatype) but not for projects or workspaces!?

cdelg:
Do you need more clarification? Let me know if I can do something.

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportSerializerPlugin.java
bjhargrave: If the serializer takes a map as input, why doesn't the deserializer produce a map as output? Seems a symmetry problem.

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportImportDeserializerPlugin.java
bjhargrave: I guess I don't see the larger design into which all these plugins fit. Perhaps some high level documentation in package-info is in order to explain how all these plugins interact and are expected to be used. The lack of this may be the cause of many of my misunderstandings.

cdelg: I think the global use is described in the PR #2169, but for the deserializer plugin, here is a specific example:

- Suppose you want to document all your projects that produce an executable application jar, your applications use the OSGI Configuator  service (with a `OSGI-INF/configurator/configuration.json` ressource) and you want to document the provided configuration data into a readme file.
- A Json ReportImportDeserializerPlugin implementation will be used to import the `OSGI-INF/configurator/configuration.json` ressource into the report DTO.
- The user is then free to use XSLT or twig (or any other) template files to generate its readme file. If he uses XSLT, the report DTO will be serialized to XML with the ReportSerializerPlugin and the result will be provided to a XSLT implementation of the ReportTransformerPlugin. The user can also just serialize the report into a file and use this file with an external tool.

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportSerializerPlugin.java
cdelg: The deserializer does not produce a report but a DTO representation of a file that is added to the report (see comments ReportImportDeserializer).

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportImportDeserializerPlugin.java
bjhargrave: I may be described in that issue, but the issue is not part of the code base (and this PR). I think we need that sort of overall doc in the package's package-info.java file.

bjhargrave: So the name of the plugin confused me. It is not deserializing a report. It is converting some file into a DTO.

In reading the text in #2169, is says

    aQute.bnd.service.reporter.ReportImportDeserializerPlugin: Convert a file into a DTO representation. Just like ProjectEntryPlugin converts a project into a DTO. There is no report here, no import and no deserialization here. 

Perhaps a better name for this type is ResourceEntryPlugin whose method is

`public Object extract(InputStream input) throws Exception;`

cdelg: Yes it is a better name.

cdelg:
Some times ago you said that one point of discussion will be the number of plugins, so except for the ReportImport bad name and package documentation are you agree with the APIs ?

Do you think that the BundleEntryPlugin, the ProjectEntryPlugin and the WorkspaceEntryPlugin shoud be replaced by one generic EntryPlugin interface. Now that I am using maven I think it will be useful to be able to build a report on other type of resources.

cdelg:
I don't know what you mean by `HandledExtensions`, I was thinking to make the extract method generic too. In the implementation I have 3 ReportGenerator one for each type workspace, project, bundle that the user needs to instantiate to generate the report. Another solution could be make those generators generic also and add them to the plugin registry and to have a single global generator class that could generate a report from an Object and a specified clazz. This global generator could look up for a specialized generator for that clazz which in turn call the proper EntryPlugin to build the report.

bjhargrave:
> I don't know what you mean by HandledExtensions, 

You have a getHandledExtensions method on the plugin so Bnd can find a plugin which supports the extension (I assume things like xml, json, etc.)

> I was thinking to make the extract method generic too. In the implementation I have 3 ReportGenerator one for each type workspace, project, bundle that the user needs to instantiate to generate the report. Another solution could be make those generators generic also and add them to the plugin registry and to have a single global generator class that could generate a report from an Object and a specified clazz. This global generator could look up for a specialized generator for that clazz which in turn call the proper EntryPlugin to build the report.

I guess the question is how to match an object to an EntryPlugin which can understand that object.

cdelg:
I thing the only way is to add a method like `public Class<T> getHandledType()` to the generic EntryPluing<T> interface. Do you see another solution?

cdelg:
Plugins in bnd doesn't have property like DS, no?

cdelg:
Is there an interface to define plugins with properties (essentially to filter)? I only saw the Plugin interface with the ability to set a map of properties.

bjhargrave:
> Is there an interface to define plugins with properties (essentially to filter)? I only saw the Plugin interface with the ability to set a map of properties.

See https://github.com/bndtools/bnd/blob/e2229347cdda31a0f451e24ad00ed6751ef9a381/biz.aQute.repository/src/aQute/bnd/repository/osgi/OSGiRepository.java#L281-L283 for an example of using the Converter to convert the map into an interface.

cdelg:
ha sorry I think that's not what I meant. I was looking for something like this:

An interface
```java
public interface PropPlugin{
   public Map<String,String> getProperties();
}
```

that other plugins API can subtype

```java
public interface MyPluginDef extends PropPlugin{
   ...
}
```

to be able to to do something like that (a kind of target filter for Reference in DS)

```java
public <T extends PropPlugin> List<T> getPlugins(Class<T> clazz, Filter filter) {
		List<T> plugins = getPlugins(clazz).stream()
			.filter(p->filter.matchMap(p.getProperties))
			.collect(toList());
		return plugins;
	}
```
Is there another way to do to this in bnd?

cdelg:
In addition:

1. The current API addresses the use case, the user can configure a report such as `-exportreport: <destionation path>;locale=<my-LOCALE>;template=<template-path>` and control what will be extracted by defining plugins such as `-plugin.ReportConfig: biz.aQute.reporter.plugins.entries.bundle.ManifestPlugin, ...`. They can also fine tune the content by providing filter in this way: `-exportreport:readme.md;filter='(customeParam=bundle)'` and configuring plugins like this:  `-plugin.ReportConfig: biz.aQute.reporter.plugins.entries.bundle.ComponentsPlugin(will be ignored for the current exportreport), biz.aQute.reporter.plugins.entries.bundle.ManifestPlugin;customeParam=bundle, biz.aQute.reporter.plugins.entries.bundle.ImportResourcePlugin;url=<my url>;customeParam=bundle`. Doing in this way works but it is not user friendly, the user will have to understand the filter and know the class name of many entry plugins. Later, we could add an instruction for the `ReportExporterService` such as `-reportconfig: <config-name>;<user friendly name>;..`  (eg; `-reportconfig: api-bundle;manifest;metatype:resourceimport='url=http://xxx',provider-bundle;manifest;components`) and add a property to the `-exportreport` like `configName=bundle-api`. In this way the configuration is easier to write and can be inherited from the workspaces. The implementation of this will be just to generate the good scoped configuration.

2. I thing @pkriens generally found the name of the instructions /commands related to this feature too generic, so you can tell me which names you thing is more appropriate.

cdelg:
For 1. and 2., I moved instructions to the implementation, so if you agree with these APIs, no more changes from me.

cdelg:
I created a branch with the impl and a bnd command if you need to see how these APIs are used:[](https://github.com/cdelg/bnd/tree/reporter-api-impl-demo).

From there you can uncomment line at the end of the build.bnd and use the command line.

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportEntryPlugin.java
```diff
@@ -0,0 +1,57 @@
...
+	 * <p>
+	 * If the source contains localized data, it will be extracted for the
+	 * specified locale or a less specific if not found.
+	 * 
+	 * @param source the source to inspect, must not be {@code null}
+	 * @param locale the {@code String} representation of a {@code Locale}, must
+	 *            not be {@code null}
+	 * @return a DTO representation or {@code null} if no data is available
+	 */
+	public Object extract(final T source, final Locale locale) throws Exception;
```
bjhargrave: As a nit, I would not put `final` in the API definition in an interface. This applies to all the methods here.

biz.aQute.bndlib/src/aQute/bnd/service/reporter/ReportGeneratorService.java
```diff
@@ -0,0 +1,65 @@
...
+	 * 
+	 * @param source the source object from which the report is generated, if
+	 *            {@code null} an empty report is returned.
+	 * @param filter a filter used to filter which {@link ReportEntryPlugin}
+	 *            plugins will contribute to the report, if not specified all
+	 *            configured {@link ReportEntryPlugin} plugins that handle the
+	 *            source object type will contribute.
+	 * @return the report, never {@code null}
+	 */
+	public Map<String, Object> generateReportOf(final Object source, final String filter);
```
bjhargrave: Is this an OSGi LDAP filter expression?

bjhargrave: I see from looking at your impl branch that it is. The javadoc here should be more clear.

bjhargrave:
> I have created a branch with the impl and a bnd command if you need to see how these APIs are used:https://github.com/cdelg/bnd/tree/reporter-api-impl-demo.

This looks really good. I like how the main bndlib code is unchanged.

One comment on the impl, bnd tries very hard to minimize dependencies. Your impl uses google collections which would add a new dependency. When we move to the next PR to add the impl, please avoid adding more dependencies unless we can convince ourselves that it is necessary. I think google collections can be avoided easily enough.

cdelg:
I have removed the final modifiers, updated the doc for LDAP and rebase/squash the commits. Thanks for your comments on the impl.

PR #2504, opened 2018-06-27
[multirepo] Add support for nested zip/par files

pkriens:
I needed the bnd repo to handle ZIP and PAR files who embed JARs. Initially I hacked this in but it looked like a mess, the IndexFile class was not up to it. This class was written half way through the tools we've developed in bnd over time to handle resources.

So I rewrote the IndexFile class and adjusted the repo. I added a number of testcases and took great effort to not change the existing tests. This class is now much simpler and should be more reliable.


biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
```diff
@@ -29,19 +35,30 @@
  * This class ignores duplicate bsn/version entries
  */
 public class BridgeRepository {
+	static Logger						logger		= LoggerFactory.getLogger(BridgeRepository.class);
```
bjhargrave: this should be final

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
```diff
@@ -140,25 +166,70 @@ public boolean isError() {
...
+		find(resources, allBndInfo);
+
+		resources.forEach(this::index);
+	}
+
+	private void find(Set<Resource> resources, Requirement req) {
+		repository.findProviders(Collections.singleton(req))
+			.get(req)
+			.stream()
+			.map(c -> c.getResource())
```
bjhargrave: Generally we use method references. So this would be `Capability::getResource`.

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
```diff
@@ -221,13 +292,25 @@ public Repository getRepository() {
...
+	public static void addInformationCapability(ResourceBuilder rb, String name, Version version, String from,
+		String error) {
+
 		try {
 			CapabilityBuilder c = new CapabilityBuilder("bnd.info");
 			c.addAttribute("name", name);
 			if (from != null)
 				c.addAttribute("from", from);
 			if (error != null)
 				c.addAttribute("error", error.toString());
```
bjhargrave: `error` is a String argument, so no need for `.toString()`.

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
```diff
@@ -122,13 +154,7 @@ private synchronized void init() {
...
 
 		public InfoCapability getInfo() {
-			List<Capability> capabilities = resource.getCapabilities("bnd.info");
-			InfoCapability info;
-			if (capabilities.size() >= 1) {
-				info = ResourceUtils.as(capabilities.get(0), InfoCapability.class);
-			} else
-				info = null;
-			return info;
+			return BridgeRepository.getInfo(resource);
```
bjhargrave: Since this is the BridgeRepository class, do you need to name the class before the static method name?

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
```diff
@@ -198,9 +198,14 @@ public static ContentCapability getContentCapability(Resource resource) {
 
 	public static List<ContentCapability> getContentCapabilities(Resource resource) {
 		List<ContentCapability> result = new ArrayList<>();
-
-		for (Capability c : resource.getCapabilities(ContentNamespace.CONTENT_NAMESPACE)) {
-			result.add(as(c, ContentCapability.class));
+		if (resource == null) {
+			System.err.println("null resource");
```
bjhargrave: A print here is wrong. Log if you need to say something. And the next line dereferences resource so it would NPE anyway. Perhaps this print is part of your debugging?

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceUtils.java
```diff
@@ -198,9 +198,14 @@ public static ContentCapability getContentCapability(Resource resource) {
...
+		}
+		List<Capability> capabilities = resource.getCapabilities(ContentNamespace.CONTENT_NAMESPACE);
+		if (capabilities != null) {
```
bjhargrave: This can never be null. The contract for the getCapabilities method says it cannot return null.

biz.aQute.repository/src/aQute/maven/api/Archive.java
```diff
@@ -11,7 +11,7 @@
 																																																								// whitespace
 			+ "(?<program>[^:]+:[^:]+)         # program\n"																																										//
 			+ "(:(?<extension>[^:]+)         # optional extension\n"																																							//
-			+ "    (:(?<classifier>[^:]+))?  # optional classifer (must be preceded by extension)\n"																															//
+			+ "    (:(?<classifier>[^:]*))?  # optional classifer (must be preceded by extension)\n"																																			//
```
bjhargrave: So the classifier can be empty?

pkriens: yup. The default archive actually has an empty ("") classifier.  This was supported in the API but you could not set it from a string.

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
pkriens: Get an error otherwise ...

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -13,403 +11,527 @@
...
+					Archive nextArchive = new Archive(archive.revision, null, "jar",
+						String.format("%s%04d", archive.classifier, n));
+					n++;
+					File dest = repo.toLocalFile(nextArchive);
+					if (!dest.isFile() || dest.lastModified() < multi.lastModified()) {
+						IO.copy(entry.getValue()
+							.openInputStream(), dest);
+						dest.setLastModified(multi.lastModified());
+					}
+					parseSingleOrMultiFile(nextArchive, dest);
```
bjhargrave: Since you recurse here, does it matter that the recursion uses n=0 again for classifiers?

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -13,403 +11,527 @@
...
+					.recover(p -> {
+						// TODO log
+						return failed(a, getMessage(p.getFailure()));
+					});
+
+				results.add(promise);
+			}
 		}
+
+		Collection<Resource> snapshot = archives.values();
```
bjhargrave: Since archives is a concurrent map, there is no point in the snapshot variable. Better to just pass `archives.values()` in the map lambda below.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -13,403 +11,527 @@
...
+						dest.setLastModified(multi.lastModified());
+					}
+					parseSingleOrMultiFile(nextArchive, dest);
 				}
 			}
 		}
+	}
+
+	boolean refresh() throws Exception {
+		if (indexFile.lastModified() != lastModified && last + 10000 < System.currentTimeMillis()) {
```
bjhargrave: Why the 10sec delay? You are limiting refreshes. Seems restrictive if I want to hit the refresh button in Bndtools.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -13,403 +11,527 @@
...
+			if (parts.length >= 2) {
+				String classifier = parts.length > 2 ? parts[parts.length - 1] : "";
+
+				Program p = Program.valueOf(parts[0], parts[1]);
+				if (p != null) {
+					Set<Version> collect = archives.keySet()
+						.stream()
+						.filter(archive -> archive.revision.program.equals(p) && archive.classifier.equals(classifier))
+						.map(archive -> archive.revision.version.getOSGiVersion())
+						.collect(Collectors.toSet());
```
bjhargrave: Collect straight into a TreeSet. `.collect(Collectors.toCollection(TreeSet::new));`

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -13,403 +11,527 @@
...
+		if (resource == null)
+			return null;
+
+		Archive archive = archives.entrySet()
 			.stream()
-			.map(BundleDescriptor::getResource)
-			.collect(toResourcesRepository())
-			.findProviders(requirements);
+			.filter((Map.Entry<Archive, Resource> e) -> e.getValue() == resource)
+			.map(e -> e.getKey())
```
bjhargrave: `.map(Map.Entry::getKey)`

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -13,403 +11,527 @@
...
+	//
+	// Deprecated to not have to increase the version with major
+	//
+	@Deprecated
+	public class BundleDescriptor extends ResourceDescriptor {
+		public long		lastModified;
+		public Archive	archive;
+		public boolean	merged;
+
+		public synchronized Resource getResource() {
```
bjhargrave: synchronized?

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -13,403 +11,527 @@
...
+	private Archive getArchive(Resource resource) {
+		if (resource == null)
+			return null;
+
+		Archive archive = archives.entrySet()
 			.stream()
-			.map(BundleDescriptor::getResource)
-			.collect(toResourcesRepository())
-			.findProviders(requirements);
+			.filter((Map.Entry<Archive, Resource> e) -> e.getValue() == resource)
```
bjhargrave: Not sure why you need the cast, I could delete it safely in Eclipse.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -13,403 +11,527 @@
...
-				.onFailure(failure -> reporter.exception(failure, "Failed to sync %s", entry.getKey())));
-		}
-		promiseFactory.all(sync)
-			.getFailure(); // block until all promises resolved
-	}
+	/*
+	 * Persistently add an archive to the index file and refresh. If the archive
+	 * is not kept then this is a noop
+	 */
+	synchronized void add(Archive archive) throws Exception {
```
bjhargrave: I am not sure why some methods are synchronized and other which seem to manipulate the same archives map are not. It also seems like there is an awful lot done inside the synchronized block.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -13,403 +11,527 @@
...
+				.entrySet()) {
+				String path = entry.getKey()
+					.toLowerCase();
+				if (path.endsWith(".jar")) {
+					Archive nextArchive = new Archive(archive.revision, null, "jar",
+						String.format("%s%04d", archive.classifier, n));
+					n++;
+					File dest = repo.toLocalFile(nextArchive);
+					if (!dest.isFile() || dest.lastModified() < multi.lastModified()) {
+						IO.copy(entry.getValue()
```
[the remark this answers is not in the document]
pkriens: That is a bit subtle ... requires you to understand quite well how the Resource in this implementation actually works.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
pkriens: documentation ..

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
pkriens: add and the removes methods must be synchronized. Hard work is done in update() that needs to know the delta. If the add/remove methods could be called overlapping the delta could be off. Quite theoretical though. The update does the parsing synchronous if it is local as an optimization. Could do it in the background but that seems more overhead than gain. Any download code is done in another thread.

PR #2529, opened 2018-07-23
Fix some problems while iterating over all WorkspaceProjects

stbischof:
- remove wrong return; from perProject mehod,
- clean perProject,
- prevent nullpointer in release method


bjhargrave:
Thanks for the PR!

Issue #2503, opened 2018-06-27
Duplicate Export


cdelg:
I have an issue with packages being exported twice using bnd Bnd-4.0.0.201805111645 and the bnd-maven-plugin. I have an API bundle A which exports all its package and a provider bundle B that import and exports the API packages it provides.
Packages in A are exported with the Export OSGI annotation. When I configure the bundle B to export these packages with:

Export-Package: \
    com.test.api.auth

the resulting manifest looks like this:

Export-Package:
  com.test.api.auth {version=1.0.0, imported-as=[1.0,1.1)},
  com.test.api.auth~ {version=1.1.0}       <- version of the bundle B

If I make com.test.api.auth a private package, and remove the export clause the bundle correctly exports the package (which is weird!).

## Reproduce

1. Clone the enRoute repository [https://github.com/osgi/osgi.enroute.git](link)
2. Modify the dao-impl project to export the `org.osgi.enroute.examples.microservice.dao` package
3. Observe that the manifest contains a duplicate entry in the export clause.
4. Modify the dao-impl project to make the `org.osgi.enroute.examples.microservice.dao` package private.
5. Observe that the manifest is good.

cdelg:
If I do not use the Export annotation but bnd instructions in the API project the problem disappears .

cdelg:
If I do not use the Export annotation but bnd instructions in the API project the problem disappears .

PR #2533, opened 2018-07-24
Fix double export of package

pkriens:
Addresses #2503

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/ExportAnnotationTest.java
+++ b/biz.aQute.bndlib.tests/src/test/ExportAnnotationTest.java
@@ -1,5 +1,7 @@
 package test;
 
+import static org.assertj.core.api.Assertions.assertThat;
+
 import java.io.File;
 import java.util.Properties;
 import java.util.jar.Manifest;
@@ -12,10 +14,42 @@ import aQute.bnd.header.OSGiHeader;
 import aQute.bnd.header.Parameters;
 import aQute.bnd.osgi.Builder;
 import aQute.bnd.osgi.Jar;
+import aQute.bnd.osgi.Packages;
 import junit.framework.TestCase;
 
 @SuppressWarnings("resource")
 public class ExportAnnotationTest extends TestCase {
+
+	public void testExportAnnotationOverlapsExportPackageHeader() throws Exception {
+		try (Builder builder = new Builder()) {
+			builder.addClasspath(new File("bin"));
+
+			builder.setExportPackage("test.export.annotation.testCalculated;version=42");
+			builder.build();
+			Packages exports = builder.getExports();
+			assertThat(exports.size()).isEqualTo(1);
+			Attrs attrs = exports.getByFQN("test.export.annotation.testCalculated");
+
+			// Check override from manifest
+			assertThat(attrs.getVersion()).isEqualTo("42");
+
+			// Check if we still have the stuff on the
+			// annotation
+			assertThat(attrs.get("fizz")).isEqualTo("buzz");
+
+		}
+	}
+
+	public void testRemoveExportAnnotation() throws Exception {
+		try (Builder builder = new Builder()) {
+			builder.addClasspath(new File("bin"));
+
+			builder.setPrivatePackage("test.export.annotation.testCalculated");
+			builder.build();
+			Packages exports = builder.getExports();
+			assertThat(exports.size()).isEqualTo(0);
+		}
+	}
 
 	public void testCalculated() throws Exception {
 		try (Builder builder = new Builder()) {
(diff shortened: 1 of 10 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -259,8 +259,20 @@ public void analyze() throws Exception {
...
+				//
+				// get the packages exported by the annotation. However,
+				// we remove any explicitly explicitly named private package so
+				// that
+				// you can always override an annotation
+				//
+
+				Parameters exportedByAnnotation = getExportedByAnnotation();
+				exportedByAnnotation.keySet()
+					.removeAll(getPrivatePackage().keySet());
```
bjhargrave: I think this is wrong. You need to merge since the Export annotation can contain important information like `uses`, `substitution` as well as attributes and directives.

pkriens: Nope, see the test case. bnd already merges (and properly prioritizes) the attributes from package info like version. I actually considered to remove this because sometime you need to override this but found this too much work. In that rare case you can override the info on the Export-Package. I think you can also remove attrs/dirs.

pkriens:
@cdelg Look at the test case. Export-Package overrides -exportcontents, overrides Export annotation. Somewhere the packageinfo is also in there, I think it comes last.

biz.aQute.bndlib.tests/src/test/ExportAnnotationTest.java
```diff
@@ -12,11 +14,43 @@
...
+	}
+
+	public void testRemoveExportAnnotation() throws Exception {
+		try (Builder builder = new Builder()) {
+			builder.addClasspath(new File("bin"));
+
+			builder.setPrivatePackage("test.export.annotation.testCalculated");
+			builder.build();
+			Packages exports = builder.getExports();
+			assertThat(exports.size()).isEqualTo(0);
```
bjhargrave: This is a problem here. The package is annotated Export and it is not exported. Using Private-Package should not turn off @Export. If you don't want to export a package marked @Export, you should negate it on the Export-Package or -exportcontents instruction.

pkriens: There are valid use cases where you want to ignore the Export annotation. How do you suggest a user can do this?
The -exportcontents is not a good comparison since this is by definition **also** under the control of the bnd file as well. The project owner can do as they see fit but the annotation can come from an external bundle.
So unless there is another mechanism I missed to override the Export annotation than I think this is the right solution.

bjhargrave: `-exportcontents: !the.package.to.not.export, the.package.to.export` would be how to do it.

pkriens: Also, how would you handle slurping a package from the classpath and then not exporting it?

bjhargrave: Private-Package is how you populate the bundle. It is common to PP all the contents of a bundle and then list the packages to export on EP. OSGi's build does this. If you want to convert to use `@Export` to export specific packages instead of using EP in the bnd file, then PP cannot cause `@Export` to be ignored.

pkriens: so how do we then do the original use case?

pkriens: Private-Package is how you populate the **private** part. The @Export was designed for people that did not have any Export-Package nor Private-Package. However, once you start to use these headers you must understand what you do.

pkriens: And in your model the use case to slurp but not export is then impossible?

pkriens: no,. This is a real case where the P-P and E-P dual purpose sucks. The P-P really says **private**. Overriding that with an annotation feels wrong because bnd.bnd file always has priority.

pkriens: you should need something like Auto-Package :-(

bjhargrave: People do

```
Package-Private: ${p}.*
Export-Package: ${p}.api
```
I would like to replace the EP in the bnd file with an `@Export` in the `${p}.api` package. How do I do that?

pkriens: Put nothin in your bnd file?

bjhargrave: Some of `${p}.*` could come from `-buildpath`.

pkriens: I think it is very bad practice to slurp and automatically export actually. That is very error prone. And export is a long term commitment. I can see 'some' utility in the default model of the project (which was the driver, no E-P, no P-P) but I am pretty strong on that if you start using these headers they should be in control

Issue #2545, opened 2018-08-01
Bndlib is generating duplicated Require-Capability for osgi.extender=osgi.component


cvgaviao:
I'm upgrading my project to R7. I've changed the osgi.core, osgi.annotation and osgi.cmpn dependencies to version 7.0.0 and started to play with some annotations DS 1.4.
 
As I use maven-bundle-plugin I needed to add a dependency to it in order to use bndlib 4.0.0 and also tried with the latest 4.1.0-SNAPSHOT.
With both versions I've identified an issue where two Require-Capability declarations are being created for DS: 

`
Require-Capability: osgi.extender;filter:="(&(osgi.extender=osgi.component)(version>=1.3.0)(!(version>=2.0.0)))",osgi.service;filter:="(objectClass=org.osgi.service.component.ComponentFactory)";effective:=active,osgi.extender;filter:="(&(osgi.extender=osgi.component)(&(version>=1.4.0)(!(version>=2.0.0))))",osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"`

bjhargrave:
We don't develop maven-bundle-plugin, so it is unclear what it may be adding and what bndlib may be adding. Can you make a test case which does not require maven-bundle-plugin?

cvgaviao:
@bjhargrave, what I was able to understand from m-b-p code it delegates all manifest generation to bndlib.

bjhargrave:
OK, but we would still need a small example to demonstrate the problem so it can be investigated. Thanks.

cvgaviao:
@bjhargrave, I've created a small project using bndlib-4.0.0.  In order to use 4.1.0-SNAPSHOT you would need to modify and build m-b-p from source since it need a class that was removed from bndlib.
[TestingBND.zip](https://github.com/bndtools/bnd/files/2250486/TestingBND.zip)

I was not able to find anything in m-b-p code that would be responsible for such duplication.
Let me know If you need anything more... ;)

bjhargrave:
I took the source code out  of you zip and created a unit test with it in bndlib.test project. It generated the proper Require-Capability.

[...]

Issue #2544, opened 2018-08-01
aQute.bnd.osgi.Instruction must prevent error when a empty input string is passed


cvgaviao:
While trying to test the bndlib 4.1.0-SNAPSHOT with maven-bundle-plugin I found an error in one of its unit test.
The error is occurring because m-b-p is mounting this string:

     *;classifier=;type=jar;scope=compile,*;classifier=;type=jar;scope=runtime

 and calling Instruction("") constructor with a empty input for the classifier item. 
```
java.lang.StringIndexOutOfBoundsException: String index out of range: 0
	at java.lang.String.charAt(String.java:658)
	at aQute.bnd.osgi.Instruction.<init>(Instruction.java:86)
	at org.apache.felix.bundleplugin.AbstractDependencyFilter$DependencyFilter.<init>(AbstractDependencyFilter.java:86)
	at org.apache.felix.bundleplugin.AbstractDependencyFilter$DependencyFilter.<init>(AbstractDependencyFilter.java:80)
	at org.apache.felix.bundleplugin.AbstractDependencyFilter$7.<init>(AbstractDependencyFilter.java:243)
	at org.apache.felix.bundleplugin.AbstractDependencyFilter.processInstructions(AbstractDependencyFilter.java:243)
	at org.apache.felix.bundleplugin.DependencyEmbedder.processHeaders(DependencyEmbedder.java:92)
	at org.apache.felix.bundleplugin.BundlePlugin.addMavenInstructions(BundlePlugin.java:715)
	at org.apache.felix.bundleplugin.BundlePlugin.buildOSGiBundle(BundlePlugin.java:969)
	at org.apache.felix.bundleplugin.BundlePluginTest.testEmbedDependencyPositiveClauses(BundlePluginTest.java:240)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at junit.framework.TestCase.runTest(TestCase.java:176)
	at junit.framework.TestCase.runBare(TestCase.java:141)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
[...]

PR #2546, opened 2018-08-01
[globs] Documentation and more consistency

pkriens:
Created documentation for the globs and in the
process added testcases and made the globs more
consistent. Mainly had to add a state to the
parsing so escaping and reaction to preceding
groups is more intuitive 


bjhargrave:
The commit comment says `Also fixes #2545` but that bug is to do with duplicate requirements. Is that what you meant?

biz.aQute.bndlib/src/aQute/bnd/osgi/Instruction.java
```diff
@@ -64,6 +64,10 @@ public boolean accept(File pathname) {
 	private boolean					optional;
 
 	public Instruction(String input) {
+
+		if (input == null || input.isEmpty())
```
bjhargrave: I am not sure this is the right fix for #2544 (You referenced a different issue number in the commit comment). I assume when they say `classifier=` they want to match classifier having no value not any value which is what this change would mean. For empty string, I would think the pattern would be `!*`: Match nothing. Passing null should be a programming error, not match all.

aQute.libg/src/aQute/libg/glob/Glob.java
```diff
@@ -40,65 +51,93 @@ private static String convertGlobToRegEx(String line) {
...
 					} else {
 						sb.append(currentChar);
 					}
 					break;
 				case '}' :
-					if (inCurlies > 0) {
+					if (state == State.CURLIES && curlyLevel > 0) {
 						sb.append(')');
-						inCurlies--;
+						currentChar = ')';
```
bjhargrave: Perhaps it would be better to add `}` to `isEnd` than changing `currentChar` to get the same effect? I would prefer that.

pkriens: Nope, then you have to know if the } is for  a quantifier. In this case, it is only for a group.

docs/_chapters/820-instructions.md
```diff
@@ -190,24 +190,130 @@ A PATH defines a number of bundles with its _entries_. Each entry is either a FI
...
+bound can be specified. If no upper bound is specified then the lower bound is also the upper bound. 
+For example, `(?i)([0-9A-F][0-9A-F]){20}` matches a 40 digit SHA digest in hex regardless of case.
 
-A `wildcard` matches any number of characters in the input. A `question` matches a single character. A `subexpr` matches the comma separated set of alternatives. Anything else is matched directly,
+There are some special characters that get special treatment:
 
-To match a literal character that is in `[*?{},]`, prefix it with a backslash ('\');
+    special ::= '{' | '}' | '|' | '+' | '*' | '?' | '(' | ')' | '[' | ']'
+
+These characters can be escaped with a solidus (`\`) to override their special meaning, they will then be matched literally. To
```
bjhargrave: Reverse solidus.

docs/_chapters/820-instructions.md
```diff
@@ -190,24 +190,130 @@ A PATH defines a number of bundles with its _entries_. Each entry is either a FI
...
+is actually non-trivial and takes a number of heuristics that can easily go wrong. Basically, if you need this
+section you're likely doing something that is too complex. 
+
+For backward compatibility reasons, we support the unix like glob style for or'ing like `{a,b,c}`. However, it is
+recommended to use the form using the vertical bar since it is more flexible and closer to regular expressions. 
+
+A glob is mapped to a regular expression character by character. During this traversal the following states are kept:
+
+* `SIMPLE` – Basic simple glob form. This is the default. 
+* `CURLIES`– Inside a curly braces group that is not a cardinality specification. Curly braces can be nested. To s
```
bjhargrave: `To s` end of line trails off

docs/_chapters/820-instructions.md
```diff
@@ -190,24 +190,130 @@ A PATH defines a number of bundles with its _entries_. Each entry is either a FI
...
+* `QUOTED` – Inside a `\Q...\E` block (cannot be nested) 
+* `BRACKETS` – Inside a character group enclosed by square brackets (cannot be nested).
```
bjhargrave: You should `[]` brackets nested in examples below.

biz.aQute.bnd/bnd.bnd
```diff
@@ -38,6 +38,7 @@ Bundle-Description: This command line utility is the Swiss army knife of OSGi. I
 -dependson: biz.aQute.bnd.embedded-repo
 
 -buildpath: \
+    org.apache.felix.framework;version=latest,\
```
bjhargrave: Why is felix framework on buildpath? That  seems odd and potentially dangerous. I don't see anything in this PR that would require this.

PR #2547, opened 2018-08-01
Use org.osgi.framework.Constants

stbischof:
Use ` org.osgi.framework.Constants` when possible
`@Deprecated` this Constants in  `aQute.bnd.osgi.Constants.java`

Signed-off-by: stbischof <stbischof@bipolis.org>

### DIFF
--- a/biz.aQute.bnd/src/aQute/bnd/main/BaselineCommands.java
+++ b/biz.aQute.bnd/src/aQute/bnd/main/BaselineCommands.java
@@ -26,6 +26,7 @@ import javax.xml.transform.TransformerFactory;
 import javax.xml.transform.stream.StreamResult;
 import javax.xml.transform.stream.StreamSource;
 
+import org.osgi.framework.Constants;
 import org.slf4j.Logger;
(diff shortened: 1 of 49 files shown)


bjhargrave:
I am not sure why we would want to do this. It seems a lot of changes for no added value.

stbischof:
For me it was just to learn more about the bnd sources and to know where, and why you handle Bundle-Manifest-Parameters

the only benefit is so see the `@deprected` infos from osgi like at
```
"Bundle-RequiredExecutionEnvironment"
```
Close the pull request if not helfpul.

i learned more about bnd.

pkriens:
On the contrary, I really dislike the long FQN for constants. What is the rational to make the code so unreadable? I could go for making it a static import but this seems to make the code _less_ readable?

bjhargrave:
If we are to do any thing here, I would suggest we have aQute.bnd.osgi.Constants extends org.osgi.framework.Constants. But I am not sure the effects of that. 

I suggest we close this PR.

PR #2551, opened 2018-08-07
Implement meta-annotation for arbitrary manifest header generation

njbartlett:
Example:

```java
@BundleHeader(name = "Bundle-Activator", singleton=true)
public @interface BundleActivatorHeader {}

package org.example;
@BundleActivatorHeader
public class ExampleActivator implements BundleActivator { ... }
```

Output:

```
Bundle-Activator: org.example.ExampleActivator;

### DIFF
--- a/biz.aQute.bnd.annotation/src/aQute/bnd/annotation/headers/BundleHeader.java
+++ b/biz.aQute.bnd.annotation/src/aQute/bnd/annotation/headers/BundleHeader.java
@@ -1,0 +1,68 @@
+package aQute.bnd.annotation.headers;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This meta-annotation allows the developer to define an annotation that
+ * generates an arbitrary manifest header in the output bundle
+ * <p>
+ * For example given the following annotation and its usage:
+ * 
+ * <pre>
+ * &#64;BundleHeader(name = "My-Bundle-Header")
+ * public @interface MyBundleHeader {}
+ * 
+ * package org.example;
+ * &#64;MyBundleHeader
+ * public class Example {}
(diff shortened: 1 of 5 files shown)


timothyjward:
What are the rules for using this, and why is it needed?

There is already a standard OSGi Header annotation, but this one seems to be different...

njbartlett:
This PR is already closed.

timothyjward:
> This PR is already closed.

I can see that, but some work obviously went into this. It isn't clear from the PR closure whether this work is going to continue or not, and if it is then I think it would be good to understand the differences between this annotation and the standard ones. On the other hand if this work is being discarded then it obviously doesn't matter!

njbartlett:
I didn't know about the standard `@Header` annotation, so I did all this work and now I'm embarrassed that somebody noticed... if only PRs could be deleted!

FWIW I think it would be useful if `@Header` could be used as a meta-annotation in the vein of `@RequireCapability`, and the `singleton` attribute is useful for headers like `Bundle-Activator`.

timothyjward:
> FWIW I think it would be useful if @Header could be used as a meta-annotation in the vein of @RequireCapability, and the singleton attribute is useful for headers like Bundle-Activator.

It would also be nice to have some better options for things like classnames (Bundle-Activator is my primary use case too)

timothyjward:
I did not know. That will save me a lot of refactoring pain!

Sent from my iPhone

On 7 Aug 2018, at 19:20, Neil Bartlett <notifications@github.com<mailto:notifications@github.com>> wrote:


You know you can use the macro ${@class} in the value attribute of the @Header annotation, right? So @Header(name = Constants.BUNDLE_ACTIVATOR, value = "${@class}") works for an activator, just not as clean as a plain @BundleActivatorHeader.

njbartlett:
:-)

The other macros are `@class-simple`, `@package` and `@version`. You can
see the code for these in `AnnotationHeaders.add`.

PR #2555, opened 2018-08-10
enhance bnd commandline functionality

stbischof:
- searching files in filetree by using include/exclude pattern,
- using build-order given from Workspace
- index sub…dirs
- TestBndMainBase for a better handling/ observation of testfolders
- some tests 

 @alexgast, Thanks for your great work on index and TestBase!


aQute.libg/src/aQute/lib/io/FileTree.java
```diff
@@ -88,23 +88,20 @@ public void addExcludes(List<String> addExcludes) {
...
-		List<Pattern> defaultPatternList = new ArrayList<>();
-		if (defaultIncludes != null) {
-			for (String defaultInclude : defaultIncludes) {
-				defaultPatternList.add(AntGlob.toPattern(defaultInclude));
-			}
-		}
 
-		List<Pattern> includePatterns = includes.isEmpty() && files.isEmpty() && (!defaultPatternList.isEmpty())
-			? defaultPatternList
+		List<Pattern> includePatterns = includes.isEmpty() && files.isEmpty() && (!defaultIncludes.isEmpty())
```
bjhargrave: You don't support defaultIncludes as null. Just do a null check here. If empty, then includes and files are also empty, so the true path of the expression will make an empty list which is OK.

aQute.libg/src/aQute/lib/io/FileTree.java
```diff
@@ -0,0 +1,124 @@
...
+	 * @param baseDir The base directory for locating files.
+	 * @param defaultIncludes The default include patterns to use if no include
+	 *            patterns were configured.
+	 * @return A list of files.
+	 * @throws IOException If an exception occurs.
+	 */
+	public List<File> getFiles(File baseDir, List<String> defaultIncludes) throws IOException {
+
+
+		List<Pattern> includePatterns = includes.isEmpty() && files.isEmpty() && (!defaultIncludes.isEmpty())
```
bjhargrave: This check needs to be:

```java
List<Pattern> includePatterns = includes.isEmpty() && files.isEmpty() && (defaultIncludes != null)
```

This allows `null` to be passed at the defaultIncludes.  After this PR, I think I would like to add an overload of this method which is varargs:

```java
public List<File> getFiles(File baseDir, String... defaultIncludes) throws IOException {
  return getFiles(baseDir, Arrays.asList(defaultIncludes));
}
```

This would allow someone to simply call `getFiles(baseDir)` when they do not have any default includes.

bjhargrave:
@stbischof It seems tests are failing in the test builds. Can you look into this and get the test builds passing. Also, you need to rebase your commits to remove any merge commits before we can merge this PR.

bjhargrave:
Also, there is a commit in here that does not have the proper Signed-off-by. This should also be fixed when you rebase. Thanks!

stbischof:
@bjhargrave I fixed the failings in the test builds. Was some problems with modificationtime (just in oraclejdk8). So we need to wait a second after copy the workspace. Now its rebased and signed.

bjhargrave:
@pkriens Are you OK with merging this PR?

pkriens:
LGTM

PR #2569, opened 2018-08-20
[xmldtoparser] Improvements

pkriens:
- handles null for collections (creates a collection)
- allows XmlAttributes for enum names
- test case

### DIFF
--- a/aQute.libg/test/aQute/lib/xmldtoparser/ComponentsDTO.java
+++ b/aQute.libg/test/aQute/lib/xmldtoparser/ComponentsDTO.java
@@ -1,0 +1,111 @@
+package aQute.lib.xmldtoparser;
+
+import java.util.List;
+
+import javax.xml.bind.annotation.XmlAttribute;
+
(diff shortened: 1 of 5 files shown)


aQute.libg/test/aQute/lib/xmldtoparser/ComponentsDTO.java
```diff
@@ -0,0 +1,111 @@
+package aQute.lib.xmldtoparser;
+
+import java.util.List;
+
+import javax.xml.bind.annotation.XmlAttribute;
```
bjhargrave: This package is removed from Java SE after Java 8. So I am not sure we should rely upon it in Bnd. Using it may produce deprecation warnings in Java 9 and 10 and may fail in Java 11.

bjhargrave: Actually it fails in the Java 9 test build...

PR #2577, opened 2018-08-23
[properties parser] Quote escapes

pkriens:
- Fixed escaping quotes in quoted strings 
  the properties parser in test mode assumed
  that an unescaped quote always signals the end
  of the quoted string. However, if it is preceded
  by an odd number of backslashes in the output
  then it will be escaped by the OSGi header parser

- Moved test to JUnit 4


pkriens:
Also changed errors to warnings. Probably should always have been warnings to not break existing builds. I think this should be more our strategy since fixupmessages can be used to make them errors.

PR #2580, opened 2018-08-24
REPOSITORY www.jpm4j.org IS DEFUNCT.

mcgitty:


mcgitty:
Agree. See https://github.com/bndtools/bnd/issues/2581

PR #2588, opened 2018-08-27
fixes #2587 - make `-contract` default to `*`

### LINKED ISSUE #2587: make `-contract` default to `*`
With the advent of bundle annotations the `-contract` instruction is one of the last instructions coming up repeatedly in plain old application bundles.

Since we really want to encourage the use of Portable Java Contracts we really should make default of the `-contract` instruction be `*`.

rotty3000:


biz.aQute.bndlib.tests/src/test/ContractTest.java
```diff
@@ -192,6 +192,52 @@ public void testMultiple() throws Exception {
...
+		a.setImportPackage("org.osgi.service.cm,*");
+		a.setProperty("Export-Package", "test.refer");
+		Jar ajar = a.build();
+		assertTrue(a.check());
+		Domain domain = Domain.domain(ajar.getManifest());
+		Parameters rc = domain.getRequireCapability();
+		rc.remove("osgi.ee");
+		System.out.println(rc);
+		assertEquals(1, rc.size());
+		assertNotNull(rc);
```
bjhargrave: You can remove this line and the one that follows. They don't add anything.

biz.aQute.bndlib.tests/src/test/ContractTest.java
```diff
@@ -192,6 +192,52 @@ public void testMultiple() throws Exception {
...
+		a.addClasspath(bjar);
+		a.setImportPackage("org.osgi.service.cm,*");
+		a.setProperty("Export-Package", "test.refer");
+		Jar ajar = a.build();
+		assertTrue(a.check());
+		Domain domain = Domain.domain(ajar.getManifest());
+		Parameters rc = domain.getRequireCapability();
+		rc.remove("osgi.ee");
+		System.out.println(rc);
+		assertEquals(1, rc.size());
```
bjhargrave: Shouldn't you test the `osgi.contract` requirement value?

biz.aQute.bndlib/src/aQute/bnd/osgi/Contracts.java
```diff
@@ -56,7 +56,7 @@ public Contracts(Analyzer analyzer) {
 
 	Instructions getFilter() {
 		if (instructions == null) {
-			String contract = analyzer.getProperty(Constants.CONTRACT);
+			String contract = analyzer.getProperty(Constants.CONTRACT, "*");
```
bjhargrave: This PR should also update https://bnd.bndtools.org/instructions/contract.html to indicate that `*` is now the default.

rotty3000:
@bjhargrave fixed and rebased.

pkriens:
I am not sure contracts should be default on. 

* They are a hack to solve bad software problems. I.e. I think package versions are better than contracts but contracts are required for all that software that has no package versions.
* They cost (as I recall) quite a few CPU cycles.
* It is easy establish a default in build.bnd or the parent pom

Did anybody look at the performance impact?

rotty3000:
> They are a hack to solve bad software problems. I.e. I think package versions are better than contracts but contracts are required for all that software that has no package versions.

No one is disputing that. The problem is there are many important non-OSGI specs that we need to consume and giving them package versions is already and has for a long time been breaking things.

> They cost (as I recall) quite a few CPU cycles.

I'm not really sure this is relevant argument. We're not trying to give contracts to proper OSGi bundles. We're using contracts that come from mostly Java EE world. Fighting the package version battle there has already be lost with countless forks trying to be the best OSGi bundle, and mostly everyone getting it wrong.

> It is easy establish a default in build.bnd or the parent pom

Not when the consumers are all over the world and novice developers need to consume these APIs. Explaining to people that they can otherwise NOT have any bnd files, or NO instructions because of all the fancy new bundle annotations, OH except for `-contract: *` which you have to create the bnd file for, or put that in your pom...

bjhargrave:
> I am not sure contracts should be default on.

I am not sure either. I am not so much worried about performance as the generated bundles. Depending upon by buildpath, my imports may be unversioned if a contract is found. So the change could be unexpected to people upgrading from Bnd 4.0 to 4.1. 

If we ever had a Bndtools call, we could discuss... :-)

rotty3000:
The problem with using the package versions from a bundle that provides a contract is that the package version is probably wrong. Otherwise, why have a contract at all? The ONLY use for contracts is to avoid importing by version. That's the whole point! So given any bundle that provides both exported package versions AND contracts we should ALWAYS use the contract. And we should only use the package versions as a fallback.

Furthermore, opting out of a contract is already well documented so there's already an escape hatch in place.

rotty3000:
> By the same logic, opting into contracts is already well documented :-)

Valid point, however bnd  has been going out of it's way for years to make less work for developers by "doing the right thing" by default. This is one of those cases. Not doing the right thing leads to bad bundles that most developers don't even realize they are creating. The code is less portable than they imagined and we could have easily solved that for them.

pkriens:
Well, there are different customers of bnd. I finally have some customers that want to do it all the right way. One started the right way and the other is having to do a LOT of effort to find the right way. It is kind of frightening how much complexity evaporates when you do it the right way.

And I am worried about performance ...

rotty3000:
I don't understand how this will prevent anyone from doing anything "the right way".

pkriens:
Well, we have to decide the default? Currently it is off by default and you want to make it on. I think contracts are an escape hatch but you want to make it the main exit.

rotty3000:
Contracts are only applied if they are available. If they are available, I think they should be used because that's what the author of that API intended. Otherwise, they would not be available.

rotty3000:
I would like to see one single example of where an API provides both package versions AND a contract that using the package versions is the better long term decision.

pkriens:
that is true. But any idea what the overhead is when no contracts are available on the -buildpath?

rotty3000:
A brief analysis shows that collection is where any variable cost lies and that collection is O(n^2) because:

- iterate through classpath bundles
- iterate through cached capabilities

However, all of the work is performed in memory because all details are already cached in memory. Nothing is additionally processed to accomplish the task.

timothyjward:
I think that O(n^2) is a significant overestimate as it either assumes the number of capabilities per bundle increases with the number of bundles on the classpath, or that each capability needs to be compared with each other capability.

In fact the number of capabilities per bundle on the classpath is unrelated to the size of the classpath, and we’re comparing against a single constant namespace. This is therefore an O(n) operation with respect to the size of the classpath/number of capabilities on the classpath.

Sent from my iPhone

On 29 Aug 2018, at 16:14, Raymond Augé <notifications@github.com<mailto:notifications@github.com>> wrote:


A brief analysis shows that collection is where any variable cost lies and that collection is O(n^2) because:

  *   iterate through classpath bundles
  *   iterate through cached capabilities

However, all of the work is performed in memory because all details are already cached in memory. Nothing is additionally processed to accomplish the task.

pkriens:
Maybe I am too concerned but we seem to be getting slower and slower. When I wrote the code for contracts I was a bit worried about performance but  I guess I am always a bit worried about that :-( If it had been up to me we would still have a button to create the bundle.

I have a pretty big build and will try with contracts on/off

pkriens:
It seems to make very little difference so I am ok.

PR #2589, opened 2018-08-27
Fix 'Update to' repository actions

seanbright:
The <versions> element in Maven metadata is not necessarily sorted high
to low, making the "Update to" action in the Repositories view offer to
upgrade you to a lower version. In the case of mysql-connector-java:

https://repo.maven.apache.org/maven2/mysql/mysql-connector-java/maven-metadata.xml

The "Update to" action suggested updating to 2.0.14, simply because it
is the last version listed.

Instead we find all later revisions and group them by major & minor so
that the user can select an appropriate later version to upgrade to.

I don't know how to exercise this with command line Bnd, so here are before and after images from within Bndtools:

**Before:**

image

**After:**

image


pkriens:
Nice!

PR #2593, opened 2018-08-30
Added functionality to consider Service-Component wildcard entry

amitjoy:
https://groups.google.com/forum/#!topic/bndtools-users/27513Zqaf7M


pkriens:
I guess we could just leave out automatically setting the header completely?

pkriens:
This was done in the wrong place

PR #2598, opened 2018-09-04
Issue #1930: Set refresh to true when refreshing

jellenelis:
The refresh() method in OSGiRepository invoked getIndex with refresh set to false. This meant that the index was not updated even though it was changed. The current test tests whether or not refreshing the repository actually updates the index by checking their last modified times. Furthermore, the added tests checks whether refreshing notifies RepositoryListenerPlugin instances.

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/osgi/OSGiIndex.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/osgi/OSGiIndex.java
@@ -64,6 +64,15 @@ class OSGiIndex {
 		return bridge;
 	}
 
+	/**
+	 * Return the bridge repository in {@link Promise} form.
+	 * 
+	 * @return promise representing the repository
+	 */
+	Promise<BridgeRepository> getBridgeRepository() {
+		return this.repository;
+	}
+
 	private static File checkCache(File cache) throws Exception {
 		IO.mkdirs(cache);
 		if (!cache.isDirectory())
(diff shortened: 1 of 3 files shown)


bjhargrave:
Instead of putting  the text `Fixes #1930` in your commit message, put the full URL to the issue:

` Fixes https://github.com/bndtools/bndtools/issues/1930`

I always use the full URL since it is nicer for people looked at the git repo outside of github. :-)

biz.aQute.repository/src/aQute/bnd/repository/osgi/OSGiIndex.java
```diff
@@ -64,6 +64,15 @@
 		return bridge;
 	}
 
+	/**
+	 * Return the bridge repository in {@link Promise} form.
+	 * 
+	 * @return promise representing the repository
+	 */
+	public Promise<BridgeRepository> getBridgeRepository() {
```
bjhargrave: Does this need to be `public`? I would prefer default access (package private) to avoid creating new API.

jellenelis:
One comment though. I put a TODO in OSGiRepositoryTest line 117 since I had to change a test. I am not completely sure whether or not the changes are warranted, would like some feedback there (to either remove the TODO if my changes are correct or to solve it some other way).

bjhargrave:
> One comment though. I put a TODO in OSGiRepositoryTest line 117 since I had to change a test. I am not completely sure whether or not the changes are warranted, would like some feedback there (to either remove the TODO if my changes are correct or to solve it some other way).

I don't have an issue with the test case change at the TODO. So you can remove the comment.

jellenelis:
> I don't have an issue with the test case change at the TODO. So you can remove the comment.
OK, pushed the change, thanks for the quick review.

bjhargrave:
Thanks for the PR!

jellenelis:
> Thanks for the PR!

No problem, hope it helps people.

PR #2603, opened 2018-09-06
Add maximum version for DS version

njbartlett:
The -dsannotations-options instruction already has a "version" option
that allows a minimum spec version to be stated. This change adds a
maximum version. If any component definition found in the bundle uses
features that require a DS spec version greater than the stipulated
maximum then an error is raised.

### DIFF
--- a/biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
+++ b/biz.aQute.bndlib.tests/src/test/component/DSAnnotationTest.java
@@ -3849,4 +3849,45 @@ public class DSAnnotationTest extends BndTestCase {
 		xt.assertAttribute("activator", "scr:component/@activate");
 	}
 
+	public void testMaximumSpecVersionPassed() throws Exception {
+		Builder b = new Builder();
+		b.setProperty(Constants.DSANNOTATIONS, "test.component.DSAnnotationTest$DS13anno_*");
+		b.setProperty(Constants.DSANNOTATIONS_OPTIONS, "extender, version;minimum=1.2;maximum=1.3");
+		b.setProperty("Private-Package", "test.component");
+		b.addClasspath(new File("bin"));
+
+		Jar jar = b.build();
+		assertOk(b);
+	}
+
+	public void testMaximumSpecVersionExceeded() throws Exception {
+		Builder b = new Builder();
+		b.setProperty(Constants.DSANNOTATIONS, "test.component.DSAnnotationTest$DS13anno_*");
+		b.setProperty(Constants.DSANNOTATIONS_OPTIONS, "extender, version;minimum=1.2;maximum=1.2");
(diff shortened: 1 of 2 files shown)


timothyjward:
@njbartlett - what happens when you set this in conjunction with using the R7 Component annotation? Does the resulting manifest still contain a 1.4 extender requirement, and if so is that a problem?

njbartlett:
@timothyjward If you use R7 features plus `-dsannotations-options: version;maximum=1.3` then you will get a build error for the bundle. It would generate the extender requirement for DS 1.4 but that doesn't seem important.

The following is more of a problem: if you specify `-dsannotations-options: extender, version;maximum=1.2` (or any maximum less than 1.3) then you will get an extender requirement for DS 1.3. This will happen even if your components only use <1.3 features, because there is no extender capability defined before DS 1.3. I would suggest we need to define this as an additional error case.

timothyjward:
> @timothyjward If you use R7 features plus -dsannotations-options: version;maximum=1.3 then you will get a build error for the bundle. It would generate the extender requirement for DS 1.4 but that doesn't seem important.

My concern was more that if you write a simple DS component:

    @Component
    public class MyComponent implements MyService {
        @Reference
        MyOtherService myOtherService;
        
        @Override
        public void doStuff() {
            myOtherService.doOtherStuff();
        }
    }

And then you set:

    -dsannotations-options: version;maximum=1.3 

Then there is no reason that should not work *except* for the 1.4 extender requirement that you pick up from the `@Component`. The resulting bundle will still only work with DS 1.4 - are we ok with this?

njbartlett:
Why would that generate a 1.4 requirement? It only requires 1.3 as far as I can tell (because of the field injection).

timothyjward:
> Why would that generate a 1.4 requirement? It only requires 1.3 as far as I can tell (because of the field injection).

The R7 Component annotation is meta annotated with a `@Requirement` annotation for the 1.4 DS extender. This was done to avoid tools having to use heuristics to work out the DS version (i.e. if you're building against the R7 annotations then you should use R7 at runtime).

timothyjward:
> Doesn't this create a duplicate requirement for both DS 1.3 and DS 1.4??

I’m pretty sure that it does, if you can find a way to fix that it would be awesome 😊

bjhargrave:
I would like to hold off on this for now as I am busy changing the DS code for DS 1.4 (R7) support.

Furthermore, during the Bndtools calls, while discussing #2545, we decided that it would be best if Bnd just generate XML at the version of the org.osgi.service.component package found on the buildpath. Then Bnd would only generate the extender require for 1.3 _if_ it sees the 1.3 version of the package. Otherwise the `@Require` annotation on `@Component` would handle it. This would allow people to control the generated version of XML by the version of the org.osgi.service.component package on the buildpath.

Issue #2605, opened 2018-09-06
`aQute.bnd.osgi.Clazz` looses generic method paramters


rotty3000:

rotty3000:
referenced PR provides a failing test

bjhargrave:
See my comment on the PR.

PR #2606, opened 2018-09-06
demonstrates #2605 `aQute.bnd.osgi.Clazz` looses generic method paramters

rotty3000:


bjhargrave:
You misunderstand how the `ClassDataCollector` works. When you are called at `method`, the class parsing has not processed the attributes of the method which follow the method in the class file. After `method` returns, you will seen be called at `signature` with the generic signature of the last member.

PR #2616, opened 2018-09-09
fixes #2615 Wrap Task for Gradle

### LINKED ISSUE #2615: Wrap Task for Gradle
I guess that would be a nice addition.

jozzan:
This contains an initial implementation of a wrap task for gradle. 
As I found this very useful in one of my projects I thought it might be worth a pull.


bjhargrave:
To be honest, I am not super keen on putting a wrap task in gradle. Wrapping seems something you do occasionally, not every build. Why would you make the same wrapped bundle over and over again each build? You wrap it once, put it in a repo, and then use it from the repo.

Also, the implementation proposed here has issues too. It should use a bnd file (or bnd inline like Bundle task) and not create many task properties to address some subset of bnd properties.

jozzan:
Well I like to do builds with a single command as opposed to a multitude. 
So say I would hand someone only the source code. This would do this for him now.

Also this way its reproducible. So I'd say it enhances maintenance.
Could be helpful, as your first wrap rarely produces exactly what you've been looking for.
In case I forgot an export or something I always have everything at hand to give it another shot.

It gets very easy and quick to run your build against the latest bug fixes.

Also you would have something that helps you with uploading your artifact to publish your efforts in the first place.

You could easily let the library be wrapped and used in ci/cd configurations.

All in all the key point I'm trying to make here is flexibility I guess.

When it comes to implementation details these of course can be changed. 
The idea was maybe to give gradle users something that comes a little more natural to them.

bjhargrave:
See https://github.com/bndtools/bnd/issues/2615#issuecomment-421429781 for why we are not interested in adding this wrap task to the gradle plugin at this time.

Thanks for the proposal though.

PR #2620, opened 2018-09-11
fixes #2619 Bndtools cannot be updated anymore

### LINKED ISSUE #2619: Bndtools cannot be updated anymore
Due to the introduction of OSGi CDI in https://github.com/bndtools/bnd/commit/0a13d0499794d9814e3b0b6b37aabb6178a6f7f1, bnd introduced a dependency on javax.exterprise which cannot be resolved in "Eclipse IDE for Java Developers". It will only work in "Eclipse IDE for Java EE Developers".

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotationReader.java
```diff
@@ -139,9 +134,9 @@
 
 		@Override
 		public void annotation(Annotation annotation) throws Exception {
-			java.lang.annotation.Annotation a = annotation.getAnnotation();
-
-			if (a instanceof Beans) {
+			String fqn = Descriptors.binaryToFQN(annotation.getName()
+				.getBinary());
+			if (fqn.equals("org.osgi.service.cdi.annotations.Beans")) {
```
seanbright: By no means critical, but would it be cleaner to have these as `static final String`s?

bjhargrave: That type name should never change. :-) So parking it in a constant is not so important.

rotty3000: they could...

rotty3000: since there's a regression bug here let's get the fix in and then I'll optimize.

PR #2640, opened 2018-09-18
workspace: Fix ActivelyClosingClassLoader to allow missing files

bjhargrave:
The OSGi build was broken because of the change to the plugin class
loader to use the new ActivelyClosingClassLoader since it failed if
a file did not exist. The OSGi build compiles AnalyzerPlugins into
cnf/bin which, of source, does not exist at the start of a clean build.
This change allows for a file to not exist and keeps checking for the
file upon each load request.

This change also fixes the class loader to respect the ordering of files
to the class loader by using LinkedHashMaps to hold the File/Wrapper
mappings.

There are also some other refactorings.


pkriens:
Not sure if this is sufficient. The Jar object will make a copy of the bin folder file names so it will not be able to see changes. Maybe we should skip the name caching?

bjhargrave:
> Not sure if this is sufficient. The Jar object will make a copy of the bin folder file names so it will not be able to see changes. Maybe we should skip the name caching?

My local test OSGi build seems to have worked. Perhaps it is because we refresh all the projects after cnf builds?

```groovy
/* After building, we need to refresh the bnd Projects to pickup the bnd plugins built in cnf. */
task('refresh') {
  onlyIf {
    compileJava.didWork || processResources.didWork
  }
  doLast {
    logger.info 'Refresh bnd Projects after building bnd plugins in {}', project.name
    bnd.project.refresh()
    parent.subprojects {
      if (project != cnf) {
        bnd.project.refresh()
      }
    }
  }
}

classes {
  finalizedBy refresh
}
```

But I am fine with removing the name caching.

pkriens:
You do it?

PR #2645, opened 2018-09-21
Generating POM dependencies for bundles

### LINKED ISSUE #2644: Generating POM dependencies for bundles
We have the usecase that we develop a couple of OSGi bundles via bnd which we deploy to our maven repository and a couple of these bundles will then be used by other tools and projects (android, standalone internal tools, ...). These projects have nothing to do with OSGi but we want to consume those bundles via out maven repository. Currently the pom.xml which will be generated by BND does not contain any dependency tag. We could create a custom pom.xml for each bundle which then will be used by bnd but maintaining both files (pom.xml and bnd.bnd) is a tedious task and it would be much easier if on default a pom.xml with the dependencies based on e.g. the imported packages or the buildpath would be generated (which still could be overriden by a custom pom.xml). It's just nice to have a proper default (especially for beginners) which only in edge cases have to be overridden.

A pull request will follow which implements this kind of behavior.

swimmesberger:
This pull request adds the ability to automatically generate the dependencies of a bundle in the pom.xml. Currently the only reliable way I have found is to use the buildpath. 
There is possibly a way to do this via requirements (with the advantage of detecting optional dependencies) but the strategies and ways I have tried were not bulletproof so this approach should be sufficient for now.

Fixes #2644

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/maven/PomDependencyFinder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/maven/PomDependencyFinder.java
@@ -1,0 +1,262 @@
+package aQute.bnd.maven;
+
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.jar.Manifest;
+
+import aQute.bnd.build.Container;
+import aQute.bnd.build.Project;
+import aQute.bnd.build.Workspace;
+import aQute.bnd.osgi.Constants;
+import aQute.bnd.osgi.Domain;
+import aQute.bnd.osgi.Jar;
+import aQute.bnd.osgi.Processor;
+import aQute.bnd.osgi.Resource;
+import aQute.bnd.service.RepositoryPlugin;
+import aQute.bnd.service.maven.GetDependencyPom;
+import aQute.bnd.service.maven.PomDependency;
+import aQute.bnd.service.maven.PomDependency.PomRevision;
+import aQute.bnd.util.dto.DTO;
+import aQute.bnd.version.MavenVersion;
+import aQute.bnd.version.Version;
+import aQute.lib.io.IO;
+
+public class PomDependencyFinder {
+	private static final String VERSION = "version";
+
+	public Collection<PomDependency> getMavenDependencies(Manifest manifest, Processor processor) throws Exception {
+		final Domain domain = Domain.domain(manifest);
+		final String bsn = domain.getBundleSymbolicName()
+			.getKey();
+		final Workspace workspace = this.getWorkspace(processor);
+		if (workspace == null)
+			return null;
+		final Project project = workspace.getProject(bsn);
+		if (project == null)
+			return null;
+		return this.getMavenDependenciesByBuildpath(bsn, workspace, project);
+	}
+
+	/**
+	 * First the MATA-INF/.../pom.xml file will be checked then the maven
+	 * repository will be checked if the bundle originated from there the maven
+	 * information of the repository will be used. If the bundle has no pom.xml
+	 * the properties of the bnd project and the MANIFEST.MF will be checked. If
+	 * no properties are set a groupId/artifactId will be generated based on the
+	 * default bnd configurations.
+	 * 
+	 * @param bundle
+	 * @return
+	 * @throws Exception
+	 */
+	private PomDependency toMavenResource(Container bundle) throws Exception {
+		try (Jar binary = new Jar(bundle.getFile())) {
+			PomDependency pomDependency = getDependencyFromFile(binary);
+			if (pomDependency != null) {
+				return pomDependency;
+			}
+
+			final Manifest manifest = this.getManifest(binary);
+			if (manifest != null) {
+				try (Processor projectProcessor = new Processor(bundle.getProject())) {
+					final BundleRevision bundleRev = this.createBundleRevision(bundle, projectProcessor, manifest);
+					pomDependency = this.getDependencyFromRepositories(bundle, bundleRev);
+					if (pomDependency != null) {
+						return pomDependency;
+					}
+					manifest.getMainAttributes()
+						.putValue(Constants.BUNDLE_SYMBOLICNAME, bundleRev.bsn);
+					manifest.getMainAttributes()
+						.putValue(Constants.BUNDLE_VERSION, bundleRev.version.toString());
+					try (PomResource pom = new PomResource(projectProcessor, manifest)) {
+						PomRevision pomRev = createPomRevision(pom);
+						PomDependency pomDep = new PomDependency();
+						pomDep.revision = pomRev;
+						return pomDep;
+					}
+				}
+			}
+		}
+		return null;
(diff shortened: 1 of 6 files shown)


bjhargrave:
I don't think this is a useful implementation. It only works for artifacts in MavenBndRepositoryies. It appears to ignore BndPomRepositories. I also don't see who this will work when I build a set of artifacts all together in the workspace and they depend upon each other?

The dependency find class is decoupled from the workspace in that it must find its way back all the time (getWorkspace). toMavenResource doesn't handle different Container types and assume getFile is always non-null. It also does some odd things like get the pom resource from a Jar, write it to a tmp file, read the tmp file back in (and then delete the tmp file). The getManifest(Jar) method can simply be Jar.getManifest().

There are a number of issue here. But I think the concept is slightly flawed. Bnd workspace builds can use many repository types. So a project can build against a bundle from an OSGiRepository or other types.

I think if you need propertly maintained POMs for others to build against your jars, then you may want to consider moving to a maven based build and using bnd-maven-plugin in that build to make your bundles.

swimmesberger:
First, thanks for taking the time to analyze this pull request.

I agree that the current implementation is lacking in some areas and very specific. It's currently in use for our use case and seems to work good for that but I'm open to make it work for the general public. 

>  I also don't see who this will work when I build a set of artifacts all together in the workspace and they depend upon each other?
The Bundle A depends on Bundle B, Bundle B is built before Bundle A and therefore Bundle B should contain a pom.xml in the jar file already which we can use to retrieve the necessary information for the Bundle A dependency tag in the Bundle A pom.xml.

> The dependency find class is decoupled from the workspace in that it must find its way back all the time (getWorkspace).
I agree that this could be optimized - can do that.

> toMavenResource doesn't handle different Container types and assume getFile is always non-null
What are the possibilities - which Container types exist and in which case getFile does not return a file? I have only tested which with our own workspace in which the current implementation works. I don't know which cases are possible here.

> It also does some odd things like get the pom resource from a Jar, write it to a tmp file, read the tmp file back in (and then delete the tmp file). 
That's needed because the Domain/PomParser API* of bnd is designed in such a way that it does not take a InputStream but only a file.
* https://github.com/bndtools/bnd/blob/79507349a54509fc2804ae085062592d62a0b962/biz.aQute.bndlib/src/aQute/bnd/osgi/Domain.java#L438
* https://github.com/bndtools/bnd/blob/6e1af4c3c15bf4f12d9639193a2ab2cffb3bf95a/biz.aQute.bndlib/src/aQute/bnd/maven/PomParser.java#L63

>  The getManifest(Jar) method can simply be Jar.getManifest()
:+1:

> There are a number of issue here. But I think the concept is slightly flawed. Bnd workspace builds can use many repository types. So a project can build against a bundle from an OSGiRepository or other types.
Yeah you are right but most libraries (nowadays) contain a pom.xml file which can be used for groupid/artifactid retrieval. Each bundle built by bnd automatically contain this file so every bundle built by bnd is automatically compatible with this process.

The think im proposing is a proper default - if a workspace is setup in such a way that we can safely retrieve the correct groupid/artifactid/version for each dependency of a bundle we can safely generate a dependencies tag. If we can't safely determine these information we simply gracefully fail by falling back to the current state - not generating any dependencies tag.

> I think if you need propertly maintained POMs for others to build against your jars, then you may want to consider moving to a maven based build and using bnd-maven-plugin in that build to make your bundles.
We use the bnd-gradle-plugin and I had the impression maven publishing is feature of bnd (e.g. release task, generating pom.xml, ...). If you want to support a feature like publishing on maven or generating pom.xml based on bnd properties then in my opinion there speaks nothing against further enhancing such a feature.
For beginners it's always nice to have something "work out of box" and if I want to create a new OSGi application with gradle and wanted to feel "home" I kind of expect to get pom dependencies based on my OSGi dependencies (especially if I'm working with only a single or two maven dependencies e.g. maven central). When developing with bnd I kind of want to define everything in the bnd.bnd file and everything else managed by bnd itself.

bjhargrave:
> First, thanks for taking the time to analyze this pull request.
> 
> I agree that the current implementation is lacking in some areas and very specific. It's currently in use for our use case and seems to work good for that but I'm open to make it work for the general public.
> 
> > I also don't see who this will work when I build a set of artifacts all together in the workspace and they depend upon each other?
> 
> The Bundle A depends on Bundle B, Bundle B is built before Bundle A and therefore Bundle B should contain a pom.xml in the jar file already which we can use to retrieve the necessary information for the Bundle A dependency tag in the Bundle A pom.xml.
> 
> > The dependency find class is decoupled from the workspace in that it must find its way back all the time (getWorkspace).
> 
> I agree that this could be optimized - can do that.
> 
> > toMavenResource doesn't handle different Container types and assume getFile is always non-null
> 
> What are the possibilities - which Container types exist and in which case getFile does not return a file? I have only tested which with our own workspace in which the current implementation works. I don't know which cases are possible here.

Look in Container. There is an enum which lists the container types.

> 
> > It also does some odd things like get the pom resource from a Jar, write it to a tmp file, read the tmp file back in (and then delete the tmp file).
> 
> That's needed because the Domain/PomParser API* of bnd is designed in such a way that it does not take a InputStream but only a file.
> 

It would be better to refactor PomParser to add `getProperties(InputStream)` and then pass the input stream for the Resouce in the Jar.




> > There are a number of issue here. But I think the concept is slightly flawed. Bnd workspace builds can use many repository types. So a project can build against a bundle from an OSGiRepository or other types.
> 
> Yeah you are right but most libraries (nowadays) contain a pom.xml file which can be used for groupid/artifactid retrieval. Each bundle built by bnd automatically contain this file so every bundle built by bnd is automatically compatible with this process.
> 
> The thing im proposing is a proper default - if a workspace is setup in such a way that we can safely retrieve the correct groupid/artifactid/version for each dependency of a bundle we can safely generate a dependencies tag. If we can't safely determine these information we simply gracefully fail by falling back to the current state - not generating any dependencies tag.
> 

But this is effectively a lie. You are asserting that these jars can be used to build the artifact in a maven build or worse to use when running the jar. In Bnd build, we build with different jars than we run with (-buildpath vs. -runbundles/-run...).

The point of the POMs generated by Bnd is to enable the artifact to be placed in Maven Central (or another Nexus repository). They are not meant to enable the artifact to be used in a maven build UNLESS the user of the artifact sets all necessary dependencies in their POM.

> > I think if you need propertly maintained POMs for others to build against your jars, then you may want to consider moving to a maven based build and using bnd-maven-plugin in that build to make your bundles.
> 
> We use the bnd-gradle-plugin and I had the impression maven publishing is feature of bnd (e.g. release task, generating pom.xml, ...). If you want to support a feature like publishing on maven or generating pom.xml based on bnd properties then in my opinion there speaks nothing against further enhancing such a feature.
> For beginners it's always nice to have something "work out of box" and if I want to create a new OSGi application with gradle and wanted to feel "home" I kind of expect to get pom dependencies based on my OSGi dependencies (especially if I'm working with only a single or two maven repositories e.g. maven central). When developing with bnd I kind of want to define everything in the bnd.bnd file and everything else managed by bnd itself.

For a Bnd non-workspace gradle build, I can see that you could generate the complete POM dependencies from the Gradle build information. But the code doing this would need to interrogate the project's gradle build. For Bnd workspace gradle builds, you are as I mentioned above with -buildpath and -runbundles being disjoint.

PR #2646, opened 2018-09-21
Pkriens/workspace projects cache

pkriens:


pkriens:
@bjhargrave The build time on my changed from 2.10 min to 1.20 ... This was a 5 min build on Windows so I'd really like to get this in. This code also properly removes projects.
The change in bndtools is very small and is already tested.

PR #2648, opened 2018-09-23
Add configurator instruction 

stbischof:
the `-configurator` instruction should help to build configuration bundles

- search  json files from a given directory (default: configurator)
- copy the files to the bundle ( OSGI-INF/configurator/ )
- adds to Require-Capability: osgi.extender=osgi.configurator with version, if necessary
- adds to Require-Capability: configurations Attribute, if necessary

Examples:
**minimal** sets the defaults
```
-configurator:
```
**all**
```
-configurator:\
  from;\
    base=somePath;\                      # default: not set (uses Builder.getBase())
    includes=configurator/*.json;\
    excludes=*_tmpl.json;\  # default: not set
    active=true,\
  target;\
    dir=OSGI-INF/other,\    # default: OSGI-INF/configurator
  requirement;\
    active=true;\
    version=2                      # default: 1
```

-- 
Signed-off-by: stbischof <stbischof@bipolis.org>


stbischof:
What is the right way to use osgi constants

- I copied ConfiguratorConstants.java from osgi.configurator (like in MetaType) (cant sign!)
- use org.osgi:org.osgi.service.configurator:jar:1.0.0
- take own local constants

bjhargrave:
I don't care for the instruction design. It would be better like `-includeresource` with globs for the files to include and exclude (`!`) and then attributes to influence the result.

Also, we have an annotation to add the requirement for configurator. That should be used. In fact, during a recent OSGi meeting, we discussed the desire to instead define annotations to do things and then put those annotation in code instead of instructions in bnd files. This keeps the annotation near the code that care about what it is doing (like the code which cares about the configuration) rather than all alone in the bnd configuration where some developers will ignore it and let it fall into disrepair.

So there will be work in OSGi and Bnd to address this be defining more annotations to "instruct" Bnd.

pkriens:
I agree with BJ that we should not create alternatives to -includeresource. Although complex it is an extremely powerful instruction that I'd rather expand for all purposes than provide a new syntax for one. Also, I find the Maven model of having a resource directory that gets copied in the bundle works quite well.

Since the requirement annotation was the original enRoute model I obvious like it. Maybe we need a way to access the attributes of those annotations in bnd so we can do what this patch attempts to do with the existing code.

           @RequireConfigurator( from = "resources/configuration/*.json" )

           -includeresource ${replace;.*;${annotation;*.RequireConfigurer;from};-$0}
(i think this is possible assuming the code analysis is done ahead of -includeresource?)

In principle instructions in should provide primitives that make it easy to do complex things with the macros. (I agree that I did not always follow this principle :-( ) Second principle is to do things in code as much as possible to get the help of a proper type system.

So I agree with BJ that I prefer not to go in this direction.

stbischof:
Should that be bnd or OSGi Annotations?

Is the goal that bnd.bnd and sub.bnd files are obsolete and I use bnd in this way?

```
@Include(my.domain.project.ProjectGlobals.class)

@Verbose(true)
@IncludeResource(from="${workspace}/cnf/license/LICENSE",to="META-INF/LICENSE")
@RemoveHeaders({"Bnd-LastModified", "Tool", "Created-By", "Private-Package"})
@NoBundles

@Include(my.domain.project.ProjectGlobals.class)
@Bundle(
imports=@Imports({
    @Import("aa.bb.cc"),
    @Import(value="aa.bb.cc",optional=true)})
docUrl=@BundleDocURL("https://sss.aaa.aa")
license=@BundleLicense("No license. All rights reserved ©${now;YYYY}")
copyright=@BundleCopyright("")
vendor=@BundleVendor("Company Name")
contact=@BundleContactAddress("a@@.de")
icons=@BundleIcons(
    @BundleIcon(value="icons/logo_64x64.png",size="size=64"))
)

@OSGiSources(true)
@RequireConfigurator(value="OSGI-INF/conf-files",version="2",optional=true))
```

I see the benefits of the Annotations, because the annotation is near the code , but  if
- you does not have any java file
- want to set an instruction for all projects in the whole workspace

pkriens:
The OSGi meta annotations should support such a scenario, they were modeled after the bnd meta annotations I did for v2Archive OSGi enRoute.

> On 26 Sep 2018, at 14:04, Stefan Bischof <notifications@github.com> wrote:
> Should that be bnd or OSGi Annotations?
> Is the goal that bnd.bnd and sub.bnd files are obsolete and I use bnd in this way?
I do not think the bnd files will become obsolete although there is a drive to make them empty. For normal projects most of the stuff can be defined in build.bnd already.

> @Include(my.domain.project.ProjectGlobals.class)
> 
> @Verbose(true)
> @IncludeResource(from="${workspace}/cnf/license/LICENSE",to="META-INF/LICENSE")
> @RemoveHeaders({"Bnd-LastModified", "Tool", "Created-By", "Private-Package"})
> @NoBundles
> 
> @Include(my.domain.project.ProjectGlobals.class)
> @Bundle(
> imports=@Imports({
>     @Import("aa.bb.cc"),
>     @Import(value="aa.bb.cc",optional=true)})
> docUrl=@BundleDocURL("https://sss.aaa.aa")
> license=@BundleLicense("No license. All rights reserved ©${now;YYYY}")
> copyright=@BundleCopyright("")
> vendor=@BundleVendor("Company Name")
> contact=@BundleContactAddress("a@@.de")
> icons=@BundleIcons(
>     @BundleIcon(value="icons/logo_64x64.png",size="size=64"))
> )
> 
> @OSGiSources(true)
> @RequireConfigurator(value="OSGI-INF/conf-files",version="2",optional=true))
I think that is way overboard ... 

In general you want to require something _logical_ and map it to instructions via macros. I do not think we want to make annotations for the whole build. Especially since with annotations you cannot do any calculations. For many declarative things, the properties file is a lot easier to read.
> I see the benefits of the Annotations, because the annotation is near the code , but if
> 
> you does not have any java file
> want to set an instruction for all projects in the whole workspace
In my experience there are lots of cases where the bnd files are extremely interesting. Mostly globals per project. The annotations are really best used when a _component_ has a requirement or capability. This makes it also work when you copy a competent to another bundle. Having annotations like you suggest would be a nightmare.

Kind regards,

	Peter Kriens

stbischof:
I will close this. I will make an request if i anderstand the bnd annotations more

PR #2649, opened 2018-09-24
Fixes including for new ProjectTracker

bjhargrave:


timothyjward:
> Why do you keep changing to the Path and Files API?

My understanding is that paths have better support for different file system types, are less likely to run into platform specific separator issues, support safer normalisation/relativisation, do better directory listing, support symbolic links, and are generally preferred in modern Java.

See https://docs.oracle.com/javase/tutorial/essential/io/legacy.html

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -1185,7 +1188,9 @@ public Project createProject(String name) throws Exception {
 
 		IO.store("#\n#   " + name.toUpperCase()
 			.replace('.', ' ') + "\n#\n", getFile(pdir, Project.BNDFILE));
-		Project p = new Project(this, pdir);
+
+		projects.refresh();
```
pkriens: I did not do this because they you cannot update many projects. Also, getProject() can return null which is not handled. Last, if you want to do this right we should add add/remove project methods.

biz.aQute.bndlib/src/aQute/bnd/build/ProjectTracker.java
```diff
@@ -1,95 +1,117 @@
...
 	}
 
 	/*
 	 * Sync the directories with the current lists of projects
 	 */
-	private void init() {
-		if (inited)
-			return;
+	private void update() {
+		try {
```
pkriens: I am not happy with this change. 
a) This code is unnecessary for batch  builds so it should be implemented in the driver, the API of Workspace has a refreshProjects() method for this. 
b) I have terrible experiences with WatchService in MacOS and Windows 7. Might be solved now, but still
c) It creates an indeterminism which makes it impossible to rely on since the even is delivered async (and on Mac could take seconds)
d) If we want to support watching directory we should centralize it and not use the raw API in this class. I.e. Workspace could have a watch( File, Callback ) method to subscribe to changes on the file system that can then share the same service.

pkriens:
I did some tests and the File API in this example that you replaced is consistently between 5-10 % faster than the Path API. It is also more readable (it is object oriented style) and quite a bit smaller.

PR #2655, opened 2018-09-26
git alias commit example - Signed-off-by

stbischof:


bjhargrave:
Generally you do not want to encourage things like this which automates the signed off by process. While people will themselves do it, we should not promote nor encourage it.

PR #2656, opened 2018-09-26
bndlib: Replace URLConnection usage with Resource.fromURL w/ HttpClient

bjhargrave:
This allows -pluginpath URLs and -includeresource URLs to be handled by
HttpClient which utilizes connection settings to handle authentication
and proxies.

### DIFF
--- a/biz.aQute.bndlib.comm.tests/test/aQute/bnd/comm/tests/HttpClientTest.java
+++ b/biz.aQute.bndlib.comm.tests/test/aQute/bnd/comm/tests/HttpClientTest.java
@@ -1,10 +1,14 @@
 package aQute.bnd.comm.tests;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.assertj.core.api.Assertions.assertThat;
+
 import java.io.File;
 import java.net.HttpURLConnection;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
+import java.nio.ByteBuffer;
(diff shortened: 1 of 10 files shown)


pkriens:
How did you get the removal of a public constructor passed the baseliner??

image

bjhargrave:
URLResource is a package private class and the constructor is also package private. So there was nothing to get past the baseliner :-)

bjhargrave:
You should use `Resource.fromURL(URL)`.

pkriens:
Yes, but it broke an old project I had.  I thought we tried to be backward compatible? This was a major API change? 

I could fix it but I am curious how you got this past the baseliner? Are there more places were we made these kind of backward incompatible changes?

bjhargrave:
But URLResource is not a public class. That is, it is _not_ API. So there is no backwards incompatible API change here. If your project uses non-public classes, it can be affected by changes to the non-public classes. That is the risk.

Baselining does not baseline non-API. So nothing got "past" the baseliner.

The only way your code could have use the URLResource constructor (which is package private) was for your code to be in the same package. That is you joined the existing package to use non-public classes (non-API).

pkriens:
URLResource was a public class with a public constructor as far as I can see?

image

bjhargrave:
That was 2 years ago when it was changed for Bnd 4.0 with a major version change on the package to signal some breaking changes.

bjhargrave:
This change went into Bnd 4.1.

PR #2658, opened 2018-09-27
Shell and graph command in bnd, fixes, and convenience methods

pkriens:
See commits


biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
```diff
@@ -2148,4 +2151,57 @@ public Object get(String key) throws Exception {
...
+
+		try {
+			Class<?> c = m.getDeclaringClass();
+			Field field = c.getField(m.getName() + "Help");
+			help = (String) field.get(null);
+		} catch (Exception e) {
+			help = "";
+		}
+		String h = help;
+		return new Map.Entry<String, String>() {
```
bjhargrave: You know about https://docs.oracle.com/javase/8/docs/api/java/util/AbstractMap.SimpleImmutableEntry.html ? You don't need to write your own Map.Entry implementations.

PR #2662, opened 2018-10-01
bnd cli improvements

stbischof:
- use getSourcePath (not getAllsourcepath) when compiling
- use include/exclude globs for parallel build.
- test parallel build
- removed single project itest in par (@pkriens )
- test transitive dependencys
- remove full option in buildoptions
- forker print infos to out not to error stream
- formatter


biz.aQute.bnd/test/aQute/bnd/main/TestBndMain.java
```diff
@@ -218,7 +259,7 @@ public void testBuild() throws Exception {
 
 		capturedStdIO.getSystemErrContent()
 			.contains(
-			"000: p: The JAR is empty: The instructions for the JAR named p did not cause any content to be included, this is likely wrong");
+			EMPTY_JAR_MSG);
```
pkriens: in general we can do `assertTrue(processor.check("The JAR is empty:"));` The given strings in the parameters are filtered from the errors and warnings. If anything remains or a string is not found false is returned.

bjhargrave: We also need to get rid of test which assert that out/err is empty. When run on Java 9+, we get can get warnings from the JVM about illegal access issue which have nothing to do with the test but the test fails anyway.

stbischof:
updated!
- fix windows linebreak issues
- can ignore warnings in System.err
- tests expected errors

PR #2668, opened 2018-10-05
[wablib] fallback to includeresource for classpath searching

rotty3000:


rotty3000:
I actually got `bnd-maven-plugin` `-wab`s to work with the `maven-war-plugin`.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -283,7 +278,7 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 				Jar bndJar = builder.build();
 
 				// Expand Jar into target/classes
-				expandJar(bndJar, outputDirectory);
+				expandJar(bndJar, outputDir);
```
bjhargrave: There are now 3 output dir variables: classedDir, ouputDir and the local outputDirectory. This is confusing.  Why are there so many. Can we collapse outputDirectory and outputDir together?

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -176,14 +179,6 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
 			}
 
-			// Reject wab projects
-			if (builder.getProperty(Constants.WAB) != null) {
-				throw new MojoExecutionException(Constants.WAB + " not supported in a maven build");
-			}
-			if (builder.getProperty(Constants.WABLIB) != null) {
-				throw new MojoExecutionException(Constants.WABLIB + " not supported in a maven build");
-			}
-
```
bjhargrave: I don't see how this will work since the wab instruction fundamental reorganize the arrangement of the bundle. Classes move from the root to web-inf, bundle class path is set. How will the bnd-maven-plugin do that given the the files already in place in  target/classes? I don't see any code here to rearrange target/classes into the wab shape.

bjhargrave: I see your comment about maven-war-plugin now. Perhaps then the check here should confirm those plugins are in use when a wab instruction is in effect?

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
rotty3000: "naming is hard..."

There's a logical explanation for why we need all 3, distinctly, even though 99% of the time they resolve to the same default value (i.e. `${project.build.outputDirectory}`).

1. the `bnd-maven-plugin` must know where the `maven-compiler-plugin` has placed it's output, which is defined by the property `${project.build.outputDirectory}` so that it can be placed on the builder's classpath. Therefore we need to read this property directly.
2. in case we want to _filter_ the contents of the resulting bundle (like the bnd workspace; where the bnd file must specify an instruction that causes packages to be included in the bundle, e.g. `-privatepackage`, `Export-Package`, etc.) we need to prevent the `bnd-maven-plugin` from ingesting the whole of `${project.build.outputDirectory}` and so we have the parameter `classesDir` which can point to different directory (maybe this _could_ be a `boolean` flag instead of a `File`, but I didn't want to break backward compatibility as someone might actually be using this).
3. in case we want to _redirect_ where the builder will write it's contents (normally back into `${project.build.outputDirectory}`) we need a _third_ parameter. Particularly in the case of building WARs we need to redirect the output to `${project.build.directory}/${project.build.finalName}` which is the assembly directory of the `maven-war-plugin`.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
rotty3000: To my knowledge there isn't a reliable way to know this in maven.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
rotty3000: If you want to suggest better names, then by all means please do. However at least `classesDir` is legacy and we shouldn't change it. The local var `outputDirectory` could be changed but it's litterally that so not sure what would be better. That leaves `outputDir`...

Note I do still have to document the scenarios above in the README, which I planned to do.

PR #2670, opened 2018-10-05
[wablib] automate -wablib in bnd-maven-plugin

rotty3000:


rotty3000:
this was too easy not to implement.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -226,8 +232,21 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 					}
 					builder.updateModified(cpe.lastModified(), cpe.getPath());
 					buildpath.add(cpe);
+
+					if ((isWab && !hasWablibs) && !Artifact.SCOPE_PROVIDED.equals(artifact.getScope())
+						&& !Artifact.SCOPE_TEST.equals(artifact.getScope()) && !artifact.isOptional()) {
+
+						String fileNameMapping = MappingUtils
+							.evaluateFileNameMapping(MappingUtils.DEFAULT_FILE_NAME_MAPPING, artifact);
+						wablibs.add("WEB-INF/lib/" + fileNameMapping + "=" + cpe.getName() + ";lib:=true");
```
bjhargrave: Should `-wablib` be set to just the paths to the jar files? And Bnd will put them in WEB-INF/lib? For example, see https://github.com/bndtools/bnd/blob/61ed372bad81a7564dc1aa0195edd37502e8d601/biz.aQute.bndlib.tests/test/test/BuilderTest.java#L1628-L1650

rotty3000: we can't, the names are different on the classpath than what the artifacts will be in the WEB-INF/lib, we need to use a resolver to do the mapping, just like I did.

bjhargrave: Well, it does not really matter what the names are inside the bundle :-) Just that they are there and on the BCP. But this is not important.

rotty3000: sorry, what I mean to say is that the maven-jar-plugin will assign names to artifacts it places in the WAR, if we try to play with those names, we may end up with two jars of of the same thing in the WAR, which we don't want. So it's best to make sure we preserve these names and do the proper mapping.

rotty3000: for instance, if you have a dependency which is locally built in the same project, artifact.getFile() will return a file who's name doesn't contain the version, BUT maven-war-plugin will shove a jar into the WEB-INF/lib/ that DOES have a version, which means if we also write a jar with no version into WEB-INF/lib then you'll have 

```
WEB-INF/lib/foo.jar
WEB-INF/lib/foo-3.4.6-SNAPSHOT.jar
```

rotty3000:
@bjhargrave updated!

maven/bnd-maven-plugin/README.md
```diff
@@ -167,17 +148,15 @@ The plugin has 3 distinct usage scenarios:
...
+3. The third case is when the project packaging is `war`, for instance when the plugin is paired with the `maven-war-plugin` instead of the `maven-jar-plugin`. The output of the plugin re-directed to the default assembly directory of the `maven-war-plugin`, which is `${project.build.directory}/${project.build.finalName}`:
 
    ```xml
    <plugin>
        <configuration>
-           <outputDir>${project.build.directory}/${project.build.finalName}</outputDir>
            <bnd><![CDATA[
                Web-ContextPath: /${project.build.finalName}
```
bjhargrave: It is wrong to show the `<bnd>` example lines having leading spaces. Bnd will interpret this as a line continuation.

maven/bnd-maven-plugin/README.md
```diff
@@ -61,7 +52,7 @@ pom. Bnd instructions in the pom are not used if the project has a bnd file.
  org.example.api,\
  org.example.types
 -sources: true
-]]></bnd>
+        ]]></bnd>
```
bjhargrave: It is wrong to show the `<bnd>` example lines having leading spaces. Bnd will interpret this as a line continuation.

maven/bnd-maven-plugin/README.md
```diff
@@ -137,7 +116,9 @@ The plugin has 3 distinct usage scenarios:
    </plugin>
    ```
 
-2. The second case, let's call it the _filtering_ use case, is a more advanced usage emulating the behavior found in the Bnd _Workspace_ model where instructions are used to include contents in the resulting jar. Note that controlling the actual contents of the jar is handled through configuration of the `maven-jar-plugin`:
+2. The second case, let's call it the _filtering_ case, is a more advanced usage emulating the behavior found in the Bnd _Workspace_ model where instructions are used to include (i.e. *filter*) contents in the jar.
+
+   **Note** that controlling the actual contents of the jar is handled through configuration of the `maven-jar-plugin`:
 
    ```xml
    <plugin>
```
bjhargrave: It is wrong to show the `<bnd>` example lines having leading spaces. Bnd will interpret this as a line continuation.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -85,15 +88,18 @@
 	@Parameter(defaultValue = "${project.build.resources}", readonly = true)
 	private List<org.apache.maven.model.Resource>	resources;
 
-	@Parameter(defaultValue = "${project.build.outputDirectory}")
+	@Parameter(defaultValue = "${project.build.outputDirectory}", readonly = true)
```
bjhargrave: Why are these back to readonly=true? I thought we expect people to be able to configure this?

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -194,9 +200,24 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 				}
 			}
 
+			boolean isWab = PACKAGING_WAR.equals(project.getPackaging());
```
bjhargrave: We are now conflating `-wab` with packaging type = war. What if someone uses `-wab` with packaging type = jar?

maven/bnd-maven-plugin/README.md
rotty3000: it works perfectly in practice!

maven/bnd-maven-plugin/README.md
rotty3000: works perfectly in practice...

maven/bnd-maven-plugin/README.md
rotty3000: works perfectly in practice...

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
rotty3000: .. I thought we just wanted to use the flag to say don't add this to the jar automatically, but I can make it configurable if you like.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
rotty3000: if someone uses -wab with jar they'll get messed up jar. Should I block the combination with an error?

maven/bnd-maven-plugin/README.md
rotty3000: do you really think this needs to be changed? It's really ugly in the POMs when you do that... meanwhile there is no evidence that it causes any problem.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
rotty3000: my last update solves this concern by throwing an error (like it previously did) when pairing `-wab` or `-wablib` with packaging `jar`.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -194,9 +200,28 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+				if (wabProperty == null) {
+					builder.setProperty(Constants.WAB, "");
+				}
+				outputDir = warOutputDir;
+				logger.info(
+					"WAB mode enabled. Bnd output will be expanded into the 'maven-war-plugin' default <webappDirectory>:"
+						+ outputDir);
+			}
+			else if ((wabProperty != null) || hasWablibs) {
+				throw new MojoExecutionException(
```
bjhargrave: I think this should be `MojoFailureException`.

rotty3000:
updated!

PR #2678, opened 2018-10-16
Filter simplification

bjhargrave:


bjhargrave:
@rotty3000 Can you test the perf of this change? I think it is comparable to your prior change and removes special casing for Version classes. It also fixes some other bugs in the Filter code.

PR #2694, opened 2018-10-22
[libg] relocate Bnd home (~/.bnd)

### LINKED ISSUE #2661: Provide a way fo relocate ~/.bnd if  env prop. $BND_USER_HOME is set
in some cases it might be necessary to relocate `~/.bnd` by setting a env. property.

e.g. gitlab-ci could not cache directories in`~/`
https://gitlab.com/gitlab-org/gitlab-runner/issues/327

in most cases like mvn and gradle they relocate the dir.
```
export GRADLE_USER_HOME=`pwd`/.gradle
MAVEN_OPTS: "-Dmaven.repo.local=.m2"
```

Is there a way in bnd to do this at the moment.
Is there a better way than using a SystemProperty?

```
	String	BND_USER_HOME_DEFAULT	= "~/.bnd";
	String	BND_USER_HOME_ENV		= "BND_USER_HOME";

	default String BND_USER_HOME() {
		return System.getProperty(BND_USER_HOME_ENV, BND_USER_HOME_DEFAULT);
	}
```

stbischof:
**Status**
- [ ] IN DEVELOPMENT
- [x] READY
- [ ] HOLD

**Description**
in some cases it might be necessary to relocate ~/.bnd by setting a env. property.

e.g. gitlab-ci could not cache directories in~/
https://gitlab.com/gitlab-org/gitlab-runner/issues/327
in most cases like mvn and gradle they relocate the dir.
```
export GRADLE_USER_HOME=`pwd`/.gradle
MAVEN_OPTS: "-Dmaven.repo.local=.m2"
```

bnd should also be able to to relocate the ~/.bnd folder.

**Checklist**
- [ ] You have added unit tests
- [ ] Documentation `please paste url`
- [x] Squash changes, related to a single issue, into a separate commit `git rebase -i` and `git push -f`
- [x] Tag commit-message [exporter,wab,cli,libg,...]
- [x] Sign-off every commit message `Signed-off-by: Joe Smith <joe.smith@email.com>` or use `git commit -s`
- [ ] Related Spec:`none`
- [ ] Reference Issues: Fixes #2661
- [ ] Backward compatibility conflicts:
 * none
- [ ] Known open Issues:
 * none

**Questions**

none

### DIFF
--- a/biz.aQute.bndlib/bnd.bnd
+++ b/biz.aQute.bndlib/bnd.bnd
@@ -23,10 +23,11 @@ Export-Package: \
 	aQute.bnd.service.*;-noimport:=true,\
 	aQute.bnd.help.*;-noimport:=true,\
 	aQute.bnd.header;-noimport:=true,\
+	aQute.bnd.home;-noimport:=true,\
 	aQute.bnd.version;-noimport:=true,\
 	aQute.bnd.filerepo;-noimport:=true,\
 	aQute.lib.deployer;-noimport:=true,\
-	aQute.bnd.junit,\
+	aQute.bnd.junit;-noimport:=true,\
 	aQute.bnd.properties;-noimport:=true,\
 	aQute.bnd.build.model;-noimport:=true,\
 	aQute.bnd.build.model.clauses;-noimport:=true,\
@@ -40,7 +41,7 @@ Export-Package: \
 	aQute.bnd.component.error;-noimport:=true,\
 	org.osgi.service.log;-split-package:=first,\
 	org.osgi.service.repository;-split-package:=first,\
-	org.osgi.util.promise;-split-package:=first, \
+	org.osgi.util.promise;-split-package:=first,\
 	org.osgi.util.function;-split-package:=first
 
 -conditionalpackage: \
--- a/biz.aQute.bnd/src/aQute/bnd/main/bnd.java
+++ b/biz.aQute.bnd/src/aQute/bnd/main/bnd.java
@@ -83,6 +83,7 @@ import aQute.bnd.header.Attrs;
 import aQute.bnd.header.OSGiHeader;
 import aQute.bnd.header.Parameters;
 import aQute.bnd.help.Syntax;
+import aQute.bnd.home.Home;
 import aQute.bnd.main.BaselineCommands.baseLineOptions;
 import aQute.bnd.main.BaselineCommands.schemaOptions;
 import aQute.bnd.main.DiffCommand.diffOptions;
@@ -155,7 +156,8 @@ public class bnd extends Processor {
 	private static Logger						logger					= LoggerFactory.getLogger(bnd.class);
 	static Pattern								ASSIGNMENT				= Pattern.compile(															//
 		"([^=]+) (= ( ?: (\"|'|) (.+) \\3 )? ) ?", Pattern.COMMENTS);
-	Settings									settings				= new Settings();
+	Settings									settings				= new Settings(
+		Home.getUserHomeBnd() + "/settings.json");
 	final PrintStream							err						= System.err;
 	final public PrintStream					out						= System.out;
 	Justif										justif					= new Justif(80, 40, 42, 70);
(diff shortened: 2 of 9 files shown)


aQute.libg/src/aQute/lib/home/Home.java
```diff
@@ -0,0 +1,21 @@
+package aQute.lib.home;
+
```
bjhargrave: This class seems overkill and no cohesive (Bnd things, maven things...)

aQute.libg/src/aQute/lib/home/Home.java
```diff
@@ -0,0 +1,21 @@
...
+public class Home {
+	public static final String	USER_HOME_BND_DEFAULT	= "~/.bnd";
+	public static final String	USER_HOME_BND_ENV		= "bnd.home.dir";
```
bjhargrave: This is a system property not an ENV var as the name might suggest.

aQute.libg/src/aQute/lib/home/Home.java
```diff
@@ -0,0 +1,21 @@
...
+
+	public static final String BND_USER_HOME() {
```
bjhargrave: The all caps method name is not typical. All caps is generally only used to constants.

aQute.libg/src/aQute/lib/home/Home.java
```diff
@@ -0,0 +1,21 @@
...
+		return USER_HOME(USER_HOME_BND_ENV, USER_HOME_BND_DEFAULT);
+	}
+
+	public static final String	USER_HOME_MVN_DEFAULT	= "~/.m2";
+	public static final String	USER_HOME_MVN_ENV		= "maven.home.dir";
```
bjhargrave: I cannot find any reference to this in the maven documentation. And I don't think Bnd should be creating system property names in the `maven` namespace. That namespace belongs to maven.

bjhargrave: Also, since this would only be a Bnd thing, we cannot influence maven itself. 

You only seem to use it for maven settings and not or the maven local repository. So the current PR is only partial.

In any case, I think if maven defines a standard system property (like `maven.repo.local` which we already use to find the `~/.m2/repository` folder), then we must use the maven standard way and not define a bnd way.

Perhaps you should limit this PR to just defining a way to specific the location of the bnd folder and leave maven things for another PR?

bjhargrave: `git grep '~/\.bnd'` finds a number of places which reference the default folder which this PR does not seem to address.

stbischof: `git grep '~/\.bnd' ` does only reference documentations, cli output, and readme files

aQute.libg/src/aQute/lib/home/Home.java
pkriens: This seems to be more a bndlib than a aQute.lib thing.

bjhargrave: It is used by `aQute.libg/src/aQute/lib/settings/Settings.java` which is not in bndlib.

pkriens:
Hmm. It does not feel in the right place. It is kind of not what aQute.lib is about. But I understand that it is rather convenient since it is at the top of all dependencies.

So I am ok reluctantly.

stbischof:
I put it there because aQute/lib/settings/Settings.java use it.

bjhargrave:
> Hmm. It does not feel in the right place. It is kind of not what aQute.lib is about.

Agree, BUT Settings is in aQute.libq and it uses the `~/.bnd` folder. Perhaps we move Home into bndlib, and just make Settings do the same thing itself.

@stbischof Can you try that change?

biz.aQute.bndlib/bnd.bnd
```diff
@@ -23,6 +23,7 @@ Export-Package: \
 	aQute.bnd.service.*;-noimport:=true,\
 	aQute.bnd.help.*;-noimport:=true,\
 	aQute.bnd.header;-noimport:=true,\
+	aQute.bnd.home,\
```
bjhargrave: I think this should be `-noimport:=true` like the other exports.

bjhargrave: Yes, I think that is a good idea.

stbischof: done

bjhargrave:
Thanks!

PR #2702, opened 2018-10-28
consistent formatting only, because I was annoyed

rotty3000:


stbischof:
Thank you for formatting! 


There are also two kinds of formatting before the linebreak. (Whitespace After separator)
,  \
, \

rotty3000:
@stbischof I split the formatting out into a standalone PR because it was unrelated to the POC going on.

PR #2703, opened 2018-10-28
consistent formatting only, because I was annoyed

rotty3000:


rotty3000:
fix stray asm reference

pkriens:
Wouldn't it be easier and better for the future to invest this time in a formatter? A formatter should not be hard to make looking at the BndEditModel? Imho would also be more fun?

Then we could also sort he headers, instructions, and macros ...

rotty3000:
I do agree it would be nice. @pkriens  you ok with this small change however?

pkriens:
Yup, no problem ... just find it not so interesting since the first edit usually mucks it up again.

rotty3000:
@pkriens I tend to use a small eclipse plugin called AnyEdit which does do these types of whitespace corrections for me automatically. What usually ends up happening in projects where I collaborate with others is that I back whitespace changes out of commits so that real code changes are always the centre point. Or I end up disabling this whitespace plugin in specific projects :) (like bnd)

pkriens:
Ah well, just need a volunteer. I think it is actually surprisingly easy since the BndEditModel already contains all the info.

PR #2706, opened 2018-10-30
[bnd] Add exportreport command

### LINKED ISSUE #2138: Generate a file that reports the bundle metadata

cdelg:
Closes #2138

### DIFF
--- a/biz.aQute.bnd/bnd.bnd
+++ b/biz.aQute.bnd/bnd.bnd
@@ -24,6 +24,7 @@
 	@${repo;biz.aQute.repository;latest}!/!META-INF/*, \
 	@${repo;biz.aQute.resolve;latest}!/!META-INF/*, \
 	@${repo;biz.aQute.bnd.exporters;latest}!/!META-INF/*, \
+	@${repo;biz.aQute.bnd.reporter;latest}!/!META-INF/*, \
 	@${repo;slf4j.api;latest}!/!META-INF/*, \
 	@${repo;slf4j.simple;latest}!/!META-INF/*, \
 	@${repo;jline}!/!META-INF/*
@@ -53,6 +54,7 @@ Bundle-Description: This command line utility is the Swiss army knife of OSGi. I
 	biz.aQute.resolve;version=latest, \
 	biz.aQute.repository;version=latest, \
 	biz.aQute.bnd.exporters;version=latest, \
+ 	biz.aQute.bnd.reporter;version=latest,\
 	biz.aQute.bnd.annotation;version=project,\
 	org.apache.ant:ant;version=latest,\
 	biz.aQute.remote.api;version=latest,\
--- a/biz.aQute.bnd/src/aQute/bnd/main/ExportReportCommand.java
+++ b/biz.aQute.bnd/src/aQute/bnd/main/ExportReportCommand.java
@@ -1,0 +1,265 @@
+package aQute.bnd.main;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.util.List;
+import java.util.Map;
+
+import aQute.bnd.build.Project;
+import aQute.bnd.build.Workspace;
+import aQute.bnd.main.bnd.HandledProjectWorkspaceOptions;
+import aQute.bnd.main.bnd.ProjectWorkspaceOptions;
+import aQute.bnd.osgi.Jar;
+import aQute.bnd.osgi.Processor;
+import aQute.bnd.osgi.Resource;
+import aQute.lib.getopt.Arguments;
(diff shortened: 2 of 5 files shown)


bjhargrave:
I think you need to add some test cases. bnd now has test cases for a number of the commands and as we add new commands, we need to add tests with them.

cdelg:
Ok, I will look how other commands are tested.

cdelg:
There is few tests compared to the number of commands, no? Are they all in the test folder of `biz.aQute.bnd`? Is there some utility functions to run commands programmatically and other to create temporary workspace and projects?

stbischof:
Here are 2 examples

https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd/test/aQute/bnd/main/TestBndMain.java

https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd/test/aQute/bnd/main/TestBndResolve.java

cdelg:
Thanks!

bjhargrave:
> There is few tests compared to the number of commands, no?

Yes, sadly. So we need to be better as we add new features. They should come with some tests.

stbischof:
There is the `ProjectWorkspaceOptions` 
https://github.com/bndtools/bnd/blob/765a44435ad5247f406955de61eec62824632a61/biz.aQute.bnd/src/aQute/bnd/main/bnd.java#L213

This helps you to set the the Workspace or filter projects using includes and excludes (glob pattern) and to iterate over the filtered project by using the perProject() method.
https://github.com/bndtools/bnd/blob/765a44435ad5247f406955de61eec62824632a61/biz.aQute.bnd/src/aQute/bnd/main/bnd.java#L4349

It would be nice if most commands uses the same pattern.

cdelg:
Thanks, I will integrate this!
With `ProjectWorkspaceOptions` and `perProject` how do I distinguish between a workspace with only one project and the user selecting one project? For example:
```
$ pwd
~/com.domain.product
$ java -jar biz.aQute.bnd.jar exportreport export
## this include action on the workspace

$ pwd
~/com.domain.product/com.domain.product.api
$ java -jar biz.aQute.bnd.jar exportreport export
## this should not invoke action on the workspace 
``` 
I have to implement this myself?

stbischof:
give me some minutes have to understand more what exporter does

cdelg:
Also the `handleOptions` method throws a `NullPointerException` if the workspace cannot be found, maybe we should throw a better exception with more context.

stbischof:
do you have some test-data? a small workspace with some projects?

it would be greate i we can handle multiple jars
https://github.com/stbischof/bnd/tree/2138-reporter-command

cdelg:
The `jarexport` sub command is not intended to be used for Jars in `generated/*`. This sub command is for "external" jar such as in your .m2 folder that you don't build. The `export` sub command works on projects and so jar in `generated` are handled by this sub command.

So I think the use case for multiple jars can be handled by an external script (a bash script that iterates in a folder for example) since multiple jars in the generated folder is already handled by the `export` sub command in case of sub-project.

There is a sample repo that you can use to test this: [https://github.com/cdelg/org.cdecode.docdemo](https://github.com/cdelg/org.cdecode.docdemo)

stbischof:
what does i have to do if i want to get an separate report for every sub-bundle in my project?

cdelg:
You can look at [https://github.com/cdelg/org.cdecode.docdemo/blob/master/org.cdecode.scan/bnd.bnd](https://github.com/cdelg/org.cdecode.docdemo/blob/master/org.cdecode.scan/bnd.bnd).

You can extract a report of a project, if this project build multiple jar, the report will contain an entry such as `bundles:[{bundle1...}, {bundle2...}]`, if you want to transform this metadata or to get a file for each bundles you can use a template and provide a parameter to the template in order to create a file for a specific bundle (like it is done in the scan project)

cdelg:
Or instead, the export command can be modified to also call the extractor on Builder object with a bundle scope for example. (if you have a specific use case for this I can do this in this PR)

cdelg:
I add some tests, now I use the `ProjectWorkspaceOptions` interface with a modified version of `perProject` which does not handle *dependson*.

bjhargrave:
Thanks for all your work on Bnd!

PR #2708, opened 2018-10-31
Fixes #1946 - On Java 11 bnd issues many HttpClient processes do download JRE classes such as `Download jrt:/java.base/java/lang/Iterable.class`

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/osgi/URLResource.java
```diff
@@ -55,7 +55,7 @@ private ByteBuffer getBuffer() throws Exception {
 	private InputStream open() throws Exception {
 		URLConnection conn;
 		InputStream in;
-		if (client != null) {
+		if (client != null && !"jrt".equals(url.getProtocol())) {
```
bjhargrave: I would instead suggest moving the test to the constructor:

```java
this.client = "jrt".equals(url.getProtocol()) ? null : client;
```

Then the test is not done each open call and the object does not hold a HttpClient reference it will never use.

rotty3000: yes of course

rotty3000:
updated

PR #2717, opened 2018-11-09
Clazz: Rewrite using a new classfile object model and parsing package

bjhargrave:
A new `classfile` package as added which defines a class file object model and a parser which parses a class file into the object model.

The `Clazz` class is completely rewritten using the new `classfile` package. Parsing a class file into the object model and visiting the parsed object model are now cleanly separated. A Clazz object only needs to parse the class file once and then it can be visited multiple time. Given that most classes are now visiting multiple times (DS annotation processing, CDI annotation processing, annotation header processing, etc.) it should improve performance (perhaps at some expense of holding the object model for class files in memory).

Next steps would include adding new API to Clazz to enable access to all the things (Annotations, etc.) currently available in the visitor model so that users can avoid the visitor model if they desire.


pkriens:
What is the runtime memory impact?

bjhargrave:
> What is the runtime memory impact?

I don't have any numbers. I tried to be careful to minimize the memory footprint of the object model.

PR #2729, opened 2018-11-14
[gradle] Add exportReports and listReports Gradle tasks

cdelg:
I will need some help if tests are needed. Also I don't know the Groovy language so maybe my implementation is not perfect :/


cdelg:
Hi, thanks, no problem. I push the code because it was there from the previous implementation of the feature. I didn't need this at this time, but with maven, if create another maven plugin in Bnd is it fine? Or you prefer to stay only with the bnd command line?

bjhargrave:
Another maven plugin would be fine. Please do make sure to include tests for it though.

Another gradle plugin would be fine too. But I don't want all this added to the current gradle plugin which is already too large.

cdelg:
Perfect, I will proceed with maven first. I close this PR.

PR #2731, opened 2018-11-15
[cdi] filter out non-bean classes early, and limit to bean archives

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
```diff
@@ -62,11 +69,56 @@ public boolean analyzeJar(Analyzer analyzer) throws Exception {
...
+				}
+			}
+		}
+
 		List<String> names = new ArrayList<>();
 		TreeSet<String> provides = new TreeSet<>();
 		TreeSet<String> requires = new TreeSet<>();
 
 		for (Clazz c : list) {
+			if (c.isEnum() || c.isInterface() || c.isAnnotation() || c.isInnerClass() || c.isSynthetic()
```
bjhargrave: Since you exclude interfaces, you don't need to also exclude annotations since an annotation is an interface.

rotty3000:
@bjhargrave updated with suggestions and switched to new clazz annotation lookup

biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
```diff
@@ -1,55 +1,71 @@
...
-				negation = true;
-				s = s.substring(1);
-			}
-			Discover option = Discover.valueOf(s);
-			if (negation) {
-				options.remove(option);
-			} else {
-				options.add(option);
-			}
+	static final DocumentBuilderFactory	dbf	= DocumentBuilderFactory.newInstance();
```
bjhargrave: You can move `dbf` to be a local var in the static initialized since you don't need it any more after the static initializer completes. Best to let it be GC'd.

rotty3000:
@bjhargrave fixed.

PR #2737, opened 2018-11-20
Use pom.properties found on -buildpath to generate dependencies section in pom

bjhargrave:
This is a replacement for and closes https://github.com/bndtools/bnd/pull/2645.

### DIFF
--- a/aQute.libg/src/aQute/libg/glob/PathSet.java
+++ b/aQute.libg/src/aQute/libg/glob/PathSet.java
@@ -182,14 +182,22 @@ public class PathSet {
 			return path -> false;
 		}
 		if (excludePatterns.isEmpty()) {
-			return path -> includePatterns.stream()
+			if (includePatterns.size() == 1) {
+				Pattern include = includePatterns.get(0);
+				return path -> include.matcher(path)
+					.matches();
+			}
+			List<Pattern> includes = new ArrayList<>(includePatterns);
+			return path -> includes.stream()
 				.anyMatch(include -> include.matcher(path)
 					.matches());
 		}
-		return path -> includePatterns.stream()
+		List<Pattern> includes = new ArrayList<>(includePatterns);
+		List<Pattern> excludes = new ArrayList<>(excludePatterns);
+		return path -> includes.stream()
 			.anyMatch(include -> include.matcher(path)
 				.matches())
-			&& !excludePatterns.stream()
+			&& !excludes.stream()
 				.anyMatch(exclude -> exclude.matcher(path)
 					.matches());
 	}
--- a/biz.aQute.bndlib/src/aQute/bnd/build/packageinfo
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/packageinfo
@@ -1,1 +1,1 @@
-version 3.0
+version 3.1
--- a/biz.aQute.bndlib/src/aQute/bnd/junit/package-info.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/junit/package-info.java
@@ -1,2 +1,2 @@
-@org.osgi.annotation.versioning.Version("1.2.0")
+@org.osgi.annotation.versioning.Version("1.3.0")
 package aQute.bnd.junit;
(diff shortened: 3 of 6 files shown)


bjhargrave:
@swimmesberger Here is an alternate implementation for adding a `<dependencies>` section to the generated pom. 

Note, since the MavenBndRepository release process, by default, generates a new pom file during release that newly generated pom will not contain a `<dependencies>` section since the MavenBndRepository is disconnected from the project which built the bundle. So you must use `-maven-release: pom;path=JAR` configuration to tell MavenBndRepository to use the pom built in the jar when releasing instead of making a new pom.

(I would like to change the default behavior of MavenBndRepository to use the pom built in the jar by default, but I will need to convince @pkriens.)

swimmesberger:
@bjhargrave Thanks for looking into that. I have looked over the implementation. So what this does is checking all jars from the classpath for a pom.properties in the META-INF folder, getting the groupdid/artifactid/version and using that for the dependencies tag in the bnd generated pom.xml?

The implementation looks much cleaner than mine so it looks good. I haven't tried this version in our internal tool (which is based on bnd) yet but on the paper it should work for us.

The current default for MavenBndRepository seems new because in bnd 4.0.0 the behaviour was like that:
```java
				Resource pomResource;

				if (instructions.pom.path != null) {
					File f = options.context.getFile(instructions.pom.path);
					if (!f.isFile())
						throw new IllegalArgumentException(
							"-maven-release specifies " + f + " as pom file but this file is not found");

					pomResource = new FileResource(f);
				} else {

					pomResource = getPomResource(binary);
					if (pomResource == null) {
						pomResource = createPomResource(binary, options.context);
						if (pomResource == null)
							throw new IllegalArgumentException(
								"No POM resource in META-INF/maven/... The Maven Bnd Repository requires this pom.");
					}
				}
```
So it first checked the pom path instruction for a pom**.xml** file if no path is set it checked for a pom.xml file in the META-INF/maven folder (getPomResource) and if there was no such file it generated a new pom.xml file.

The new behavior seems to be that the pom path instruction is checked if that equals to "JAR" the pom.xml in the META_INF/maven folder is used otherwise the pom.xml of the path which is set is used.
If no property is set it will check the META-INF/maven folder for a pom**.properties** file (which does not contain any pom dependencies) first and if no such file exists a pom.xml based on the MANIFEST will be generated. So in the case that no pom.xml path is defined the pom**.xml** will not be checked.

That behavior sounds strange to me because most likely you want the more sophisticated information contained in the pom.xml file of the jar instead of the basic information which is contained in the pom.properties file.

P.S.: Thanks for the great talks at eclipsecon Ludwigsburg :smile:

bjhargrave:
> So what this does is checking all jars from the classpath for a pom.properties in the META-INF folder, getting the groupdid/artifactid/version and using that for the dependencies tag in the bnd generated pom.xml?

Yes. 

> 
> The implementation looks much cleaner than mine so it looks good. I haven't tried this version in our internal tool (which is based on bnd) yet but on the paper it should work for us.
> 
> The current default for MavenBndRepository seems new because in bnd 4.0.0 the behaviour was like that:

Yes, in 4.1 Peter changed it to regenerate the pom file during release. This change caused Bnd's build itself to have pom files which maven central rejected, so I had to make a last minute fix to solve that before Bnd 4.1 could be finalized and released to maven central.

> The new behavior seems to be that the pom path instruction is checked if that equals to "JAR" the pom.xml in the META_INF/maven folder is used otherwise the pom.xml of the path which is set is used.

This is behavior I *just* added in 4.2 so that you can release with the pom built in the jar instead of always regenerating the pom during release.

> If no property is set it will check the META-INF/maven folder for a pom.**properties** file (which does not contain any pom dependencies) first and if no such file exists a pom.xml based on the MANIFEST will be generated. So in the case that no pom.xml path is defined the pom.**xml** will not be checked.
> 
> That behavior sounds strange to me because most likely you want the more sophisticated information contained in the pom.xml file of the jar instead of the basic information which is contained in the pom.properties file.

I don't fully understand why Peter made that change.

PR #2747, opened 2018-11-27
[jar] get module name/version

rotty3000:


bjhargrave:
This doesn't compile.

biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java
```diff
@@ -784,6 +786,11 @@ public Clazz(Analyzer analyzer, String path, Resource resource) {
 		classDef = new ClassDef(classFile);
 		constantPool = classFile.constant_pool;
 		referred = new HashMap<>(constantPool.size());
+		moduleAttribute = Arrays.stream(classFile.attributes)
```
bjhargrave: I think this should be in ClassDef. Also, we really should not proactively get things and put the in fields. They should be obtained and returned upon demand.

bjhargrave: Also, this should then use the attribute methods from ElementDef.

biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java
```diff
@@ -2027,4 +2034,22 @@ public Resource getResource() {
 		return resource;
 	}
 
+	public int getModuleAccess() {
```
bjhargrave: These methods should all delegate to the equivalent methods in ClassDef.

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -1007,6 +1007,75 @@ public String getBsn() throws Exception {
...
+	 * @return null when the jar has no module-info.class, manifest, when the
+	 *         manifest has no {@link Constants#AUTOMATIC_MODULE_NAME} header
+	 *         and no {@link Constants#BUNDLE_SYMBOLICNAME} header, or when
+	 *         either header value is not a valid bsn according to {@link #BSN}.
+	 * @throws Exception when the jar is closed or when the manifest could not
+	 *             be retrieved.
+	 */
+	public String getModuleName(Analyzer analyzer) throws Exception {
+		check();
+		Resource resource = getResource(Constants.MODULE_INFO_CLASS);
```
bjhargrave: Do we want to create a new Clazz here? Hmm, this seems odd since you need to pass an Analyzer in to this method. This is the first time Jar touches Analyzer which seems inverted.

I am not sure what the right solution here is but this feels odd. I need to think more. It also most seems like Analyzer needs a method which returns the module name of the bundle being analyzed.

rotty3000:
@bjhargrave  .. just fixed it

rotty3000:
@bjhargrave wdyt?

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -3708,4 +3708,63 @@ public boolean assignable(Clazz annoServiceClazz, Clazz inferredServiceClazz) {
...
+	 * c) the {@link Constants#BUNDLE_SYMBOLICNAME} manifest header.
+	 *
+	 * @return null when the jar has no module-info.class, manifest, when the
+	 *         manifest has no {@link Constants#AUTOMATIC_MODULE_NAME} header
+	 *         and no {@link Constants#BUNDLE_SYMBOLICNAME} header, or when
+	 *         either header value is not a valid bsn according to {@link #BSN}.
+	 * @throws Exception when the jar is closed or when the manifest could not
+	 *             be retrieved.
+	 */
+	public String getModuleName(Jar jar) throws Exception {
```
bjhargrave: I guess I would ask why the method does not just return the module name for the jar being analyzed? That is, why to you need to pass a Jar object? Create an Analyzer with a Jar object (or setJar after construction) and then the getModuleName method takes no args. It just uses the set jar as the receiver.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -3708,4 +3708,63 @@ public boolean assignable(Clazz annoServiceClazz, Clazz inferredServiceClazz) {
...
+	 * a) the module-info.class or,
+	 * <p>
+	 * b) the {@link Constants#BUNDLE_VERSION} manifest header.
+	 *
+	 * @return null when the jar has no module-info.class, manifest, when the
+	 *         manifest has no {@link Constants#BUNDLE_VERSION} header.
+	 * @throws Exception when the jar is closed or when the manifest could not
+	 *             be retrieved.
+	 */
+	public String getModuleVersion(Jar jar) throws Exception {
```
bjhargrave: Same here. This should be no args.

biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java
```diff
@@ -479,6 +480,21 @@ public TypeRef getType() {
 			return type;
 		}
 
+		public int getModuleAccess() {
+			return attribute(ModuleAttribute.class).map(attr -> attr.module_flags)
```
bjhargrave: Use `mapToInt` instead of `map`. To do that, you will need to use `attributes`.

```java
			return attributes(ModuleAttribute.class).mapToInt(attr -> attr.module_flags)
				.findFirst()
				.orElse(0);
```

biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java
```diff
@@ -479,6 +480,21 @@ public TypeRef getType() {
 			return type;
 		}
 
+		public int getModuleAccess() {
```
bjhargrave: These methods in ClassDef shoud be package private to allow future freedom to refactor.

biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
```diff
@@ -476,6 +477,14 @@
 	 */
 	String								TSTAMP										= "_@tstamp";
 
+	/*
+	 * Module Section
+	 */
+
+	String								MODULE_INFO_CLASS							= "module-info.class";
+	String								MODULE_INFO_JAVA							= "module-info.java";
```
bjhargrave: These two constants (_JAVA and _NAME) are unused by this PR. Perhaps it is better to define them later when actually needed?

biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java
bjhargrave: Also, this method and the one on Clazz are better named `getModuleFlags`. Since this not the access https://docs.oracle.com/javase/specs/jvms/se11/html/jvms-4.html#jvms-4.7.25

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
rotty3000: because I need to inspect every jar in the classpath to associate it's module name with a package it contains

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
rotty3000: no, I need to call this on a slew of jars

biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java
rotty3000: kk, will do

biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java
rotty3000: right! will do

biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
rotty3000: will remove

rotty3000:
@bjhargrave updated.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
bjhargrave: Hmm. This code parses the module-info.class file each time and then discards it. I suggested putting these methods in Analyzer since I had mistakenly assumed that the analyzer would be making the bundle for which you wanted this information and therefore the analyzer would already have the Clazz object (findClass). 

Given this information, we may then needs to have Jar do this which custom code which directly uses the ClassFIle class on module-info.class. It could then dig this information out when the Jar object is  built like we do for the manifest. WDYT? I could take a stab at that if you want.

rotty3000: Sounds good!

PR #2749, opened 2018-11-27
[verifier] don't complain about the specified `register:` directive

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/osgi/Verifier.java
```diff
@@ -673,7 +673,7 @@ public void verify() throws Exception {
 		verifyDirectives(Constants.IMPORT_PACKAGE, "resolution:", PACKAGEPATTERN, "package");
 		verifyDirectives(Constants.REQUIRE_BUNDLE, "visibility:|resolution:", SYMBOLICNAME, "bsn");
 		verifyDirectives(Constants.FRAGMENT_HOST, "extension:", SYMBOLICNAME, "bsn");
-		verifyDirectives(Constants.PROVIDE_CAPABILITY, "effective:|uses:", null, null);
+		verifyDirectives(Constants.PROVIDE_CAPABILITY, "effective:|register:|uses:", null, null);
```
bjhargrave: `register` is not a listed directive for all capabilities. See https://osgi.org/specification/osgi.core/7.0.0/framework.module.html#d0e2821. So it not quite correct here to allow in all cases. It is only specified for the "osgi.serviceloader" namespace: https://osgi.org/specification/osgi.cmpn/7.0.0/service.loader.html#org.osgi.service.serviceloader.ServiceLoaderNamespace.CAPABILITY_REGISTER_DIRECTIVE

So the allowance of register should be more specific.

rotty3000: But we don't have a more specific verification for this header. the possible permutations are huge.

bjhargrave: There is one known capability namespace which uses the register directive. That does not seems like a large problem space.

rotty3000: I'm talking about opening the can of worms of validating specific namespaces.

bjhargrave: True, but that is why things are the way they are. This change just suggests opening register up to every capability which is part way to simply allowing all possible directives. 

You could also use a -fixupmessage to elide the register warning.

rotty3000: Imagine the developer uses a nice annotation `@ServiceProvider` You mean that he has to then write a fixupmessages to prevent the build from failing?

biz.aQute.bndlib/src/aQute/bnd/osgi/Verifier.java
```diff
@@ -673,7 +674,12 @@ public void verify() throws Exception {
 		verifyDirectives(Constants.IMPORT_PACKAGE, "resolution:", PACKAGEPATTERN, "package");
 		verifyDirectives(Constants.REQUIRE_BUNDLE, "visibility:|resolution:", SYMBOLICNAME, "bsn");
 		verifyDirectives(Constants.FRAGMENT_HOST, "extension:", SYMBOLICNAME, "bsn");
-		verifyDirectives(Constants.PROVIDE_CAPABILITY, "effective:|uses:", null, null);
+		verifyDirectives(Constants.PROVIDE_CAPABILITY, "effective:|uses:", null, null, (namespace, directive) -> {
```
bjhargrave: Perhaps a better way, is for the lamba return of the directives string used to make the pattern? Then the default is to return "effective:|uses:" but the lambda could return "effective:|uses:|register:" for the osgi.serviceloader namespace? This way we don't need special code, just customized directives strings based upon the namespace.

BTW, great idea for the solution!

rotty3000: ok, you mean a simple `Function<String, String>`  input is the namespace, return is the directives pattern?

rotty3000:
is this what you meant @bjhargrave ?

biz.aQute.bndlib/src/aQute/bnd/osgi/Verifier.java
```diff
@@ -673,7 +675,12 @@ public void verify() throws Exception {
 		verifyDirectives(Constants.IMPORT_PACKAGE, "resolution:", PACKAGEPATTERN, "package");
 		verifyDirectives(Constants.REQUIRE_BUNDLE, "visibility:|resolution:", SYMBOLICNAME, "bsn");
 		verifyDirectives(Constants.FRAGMENT_HOST, "extension:", SYMBOLICNAME, "bsn");
-		verifyDirectives(Constants.PROVIDE_CAPABILITY, "effective:|uses:", null, null);
+		verifyDirectives(Constants.PROVIDE_CAPABILITY, "effective:|uses:", null, null, (namespace, pattern) -> {
```
bjhargrave: I was thinking of something more like:

```java
verifyDirectives(Constants.PROVIDE_CAPABILITY,
   namespace -> "osgi.serviceloader".equals(namespace) ? "effective:|uses:|register:" : "effective:|uses:",
   null, null);
```
The namespace check is simple today. When we need more, we can convert to a switch statement

```java
verifyDirectives(Constants.PROVIDE_CAPABILITY,
	namespace -> {
		switch (namespace) {
			case "osgi.serviceloader" :
				return "effective:|uses:|register:";
			default :
				return "effective:|uses:";
		}
	},
	null, null);
```

biz.aQute.bndlib/src/aQute/bnd/osgi/Verifier.java
```diff
@@ -1061,15 +1068,36 @@ private void verifyType(@SuppressWarnings("unused") Attrs.Type type, @SuppressWa
...
+	 */
+	private void verifyDirectives(String header, String directives, Pattern namePattern, String type,
+		BiFunction<String, Pattern, Pattern> namespaceDirectives) throws Exception {
 		Pattern pattern = Pattern.compile(directives);
 		Parameters map = parseHeader(manifest.getMainAttributes()
 			.getValue(header));
 		for (Entry<String, Attrs> entry : map.entrySet()) {
 			String pname = removeDuplicateMarker(entry.getKey());
 
+			pattern = namespaceDirectives.apply(pname, pattern);
```
bjhargrave:
```java
Pattern pattern = Pattern.compile(directives.apply(pname));
```

biz.aQute.bndlib/src/aQute/bnd/osgi/Verifier.java
```diff
@@ -1061,15 +1068,36 @@ private void verifyType(@SuppressWarnings("unused") Attrs.Type type, @SuppressWa
 	 * 
 	 * @param header
 	 * @param directives
+	 * @param namePattern
+	 * @param type
 	 * @throws Exception
 	 */
 	private void verifyDirectives(String header, String directives, Pattern namePattern, String type) throws Exception {
+		verifyDirectives(header, directives, namePattern, type, (namespace, pattern) -> pattern);
```
bjhargrave:
```java
private void verifyDirectives(String header, String directives, Pattern namePattern, String type) throws Exception {
	verifyDirectives(header, namespace -> directives, namePattern, type);
}

private void verifyDirectives(String header, Function<String, String> directives, Pattern namePattern, String type)
	throws Exception {
```

rotty3000:
updated.

PR #2773, opened 2018-12-05
Add the bnd-reporter-maven-plugin

cdelg:
This PR contains the implementation of the new bnd-reporter-maven-plugin project. It provides the reporter feature for maven users. 

I have tested this plugin on some of my projects and it works fine, but as I am not an experimented maven user it could be nice if someone can test this plugin (or check the implementation) and provide feedback before I write unit tests?!

Thanks


rotty3000:
Thank you! I'll take a closer look post 4.2 release.

cdelg:
@bjhargrave When I compile the reporter project, the resulting bundle has now a maven dependency to org.eclipse.orbit.bundles:org.objectweb.asm:6.0.0-SNAPSHOT. This dependency is not found when I compile the maven plugin, do you know from were it come? 
I solved it by appending `maven-scope=provided` to all the bundles on the buildpath except for slf4j and bndlib since there is no calculated imports for them but I am not sure it is the right thing to do.

cdelg:
The merge with the bndtools repo brings an updated version of asm, solved with a version filter. But I don't understand why maven dependencies are generated differently with different versions of bundles on the buildpath...

cdelg:
Can I have help for the tests? How they work?

bjhargrave:
> @bjhargrave When I compile the reporter project, the resulting bundle has now a maven dependency to org.eclipse.orbit.bundles:org.objectweb.asm:6.0.0-SNAPSHOT.

Well this is the first asm on the project's `-buildpath` and we map the `-buildpath` entries with maven poms to the maven dependencies in the project. If you do not need a maven compile scope dependency to a `-buildpath` entry, add `maven-scope=provided` to the `-buildpath` entry.

> I solved it by appending `maven-scope=provided` to all the bundles on the buildpath except for slf4j and bndlib since there is no calculated imports for them but I am not sure it is the right thing to do.

This is the answer since you include the packages as private packages. But this will not be helpful at all in the maven world since the reporter jar will not be used in an OSGi class loader and thus these package will not truly be private and thus can conflict with other maven plugins used by the developer.

It would be far better for the reporter jar to have proper external dependencies (maven-scope=compile) (and you probably should use the canonical asm deps and not Eclipse Orbit). When bnd jar is build and included reporter, then you will need to include the deps as private packages since bnd jar is a standalone app. 

If this is not properly handled, then a maven plugin for reporting will be dangerous to use given the high opportunity for duplicate copies of packages (you private copies and other plugins expressed dependencies).

bjhargrave:
> Can I have help for the tests? How they work?

I am not sure what help to offer here. Our other maven projects have test cases which can be used to model your test cases. We use the maven-invoker-plugin to run the tests in the integration test task.

cdelg:
> It would be far better for the reporter jar to have proper external dependencies...

You mean: In the reporter project, I should keep `org.jtwig.*` and `org.parboiled.*` as private package (since they are not bundle) and remove `com.google.common.*`, `com.googlecode.concurrentlinkedhashmap`, `org.apache.commons.lang3.*` and `org.objectweb.asm.*` from `-conditionalpackage` (since there are bundles) to have proper deps. Then I should add those later packages as private in the bnd project. I am correct?

Thanks for the tests, I will try and ask specific questions in case.

bjhargrave:
> You mean: In the reporter project, I should keep `org.jtwig.*` and `org.parboiled.*` as private package (since they are not bundle) and remove `com.google.common.*`, `com.googlecode.concurrentlinkedhashmap`, `org.apache.commons.lang3.*` and `org.objectweb.asm.*` from `-conditionalpackage` (since there are bundles) to have proper deps.

Think of the project as making a jar rather than a bundle. Since the jar will be used in maven as a jar and thus all packages in the jar which are third party package (obtained from maven central) may conflict with another maven plugin using the same/similar dependencies from maven central.

The reporter jar can be used in 3 ways: (1) packaged into bnd jar for standalone use, (2) used as a plain jar in gradle or maven scenarios, (3) as an OSGi bundle perhaps one day in Eclipse. For (1), the build of bnd jar must take care to include all necessary packages since it must run standalone. For (2), you must be careful with including any 3rd party packages in the jar since they may conflict with other users of those 3rd party packages since maven and gradle will lump all the plugins and their dependencies into the same class loader. For (3), private packages are OK, but since we must also support (primarily) (2), we should be super careful with private packages.

So for the reporter jar, you probably should not private package any 3rd party dependency packages.

cdelg:
Ok so, I removed all 3rd party private package, but now with Bndtools: 4.3.0.DEV-201904222201-g890b6b7, the generated maven does not contain a deps on jtwig, the calculated imports have no version, is that bad? Is it related?

bjhargrave:
> the generated maven does not contain a deps on jtwig

This is because, for some odd reason, jtwig-core does not include pom metadata in the jar like most jars in maven central. So Bnd cannot figure out the maven dependency information for that buildpath entry. So we will need to manually add that. But for now Bnd cannot do that, I will make a PR to allow that.

bjhargrave:
> But for now Bnd cannot do that, I will make a PR to allow that #3171.

I merged #3171. Once it is built, you can add the jtwig maven deps to the jar's pom by adding the following to the bnd file:

```
-maven-dependencies.jtwig: jtwig-core;groupId=org.jtwig;artifactId=jtwig-core;version=5.86.1.RELEASE;scope=compile, jtwig-reflection;groupId=org.jtwig;artifactId=jtwig-reflection;version=5.86.1.RELEASE;scope=compile
```

This will supplement the computed dependency information since the jtwig jars do not include pom metadata.

cdelg:
This PR can be reviewed now.

biz.aQute.bnd.reporter/bnd.bnd
```diff
@@ -1,23 +1,36 @@
...
+	groupId=org.jtwig;\
+	artifactId=jtwig-core;\
+	version=5.86.1.RELEASE;\
+	scope=compile,\
+	\
+	jtwig-reflection;\
+	groupId=org.jtwig;\
+	artifactId=jtwig-reflection;\
+	version=5.86.1.RELEASE;\
+	scope=compile,\
```
bjhargrave: Line should not end in `,\`

biz.aQute.bnd.reporter/bnd.bnd
```diff
@@ -30,29 +43,16 @@ Export-Package: \
...
 	
 Private-Package: \
-	org.parboiled.*;-split-package:=merge-first,\
-	org.objectweb.asm.*,\
-	org.jtwig.*,\
 	biz.aQute.bnd.reporter.plugins.*,\
 	biz.aQute.bnd.reporter.generator,\
 	biz.aQute.bnd.reporter.helpers,\
-	biz.aQute.bnd.reporter.component.dto,\
-	biz.aQute.bnd.reporter.manifest.dto
```
bjhargrave: Perhaps this whole thing should be `-includepackage: biz.aQute.bnd.reporter.*`?

biz.aQute.bnd.reporter/bnd.bnd
```diff
@@ -30,29 +43,16 @@ Export-Package: \
...
+	biz.aQute.bnd.reporter.component.dto
 
 -builderignore: testresources
 
--baseline: *
+-baseline: *
```
bjhargrave: Not sure why you removed the EOL here.

maven/bnd-reporter-maven-plugin/README.md
```diff
@@ -0,0 +1,212 @@
...
+            </report>
+        </reports>
+    </configuration>
+```
+
+The resulting `metadata.json` file could be:
+
+```json
+{
+  "configuration":{<the configuration from the configuration.json file>}
```
bjhargrave: I think your json is problematic. The `<>` are flagged and there is no comma between the map entries.

cdelg: This was a comment, I placed it with sample values.

biz.aQute.bnd.reporter/bnd.bnd
cdelg: Mistake

biz.aQute.bnd.reporter/bnd.bnd
cdelg: Yes, I will replace it.

biz.aQute.bnd.reporter/bnd.bnd
cdelg: ok

maven/bnd-reporter-maven-plugin/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+
+	<build>
+		<plugins>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-invoker-plugin</artifactId>
+				<dependencies>
+					<dependency>
+						<groupId>biz.aQute.bnd</groupId>
+						<artifactId>aQute.libg</artifactId>
```
bjhargrave: This runtime dep on aQute.libg seems a risky thing to do in the tests. I think it can mask an error in packaging the plugin where the plugin failed to properly conditional package aQute.lib* that could creep into a release. I suspect this is because you need to use this in verify.groovy. I would like to remove this runtime dep here and, if verify.groovy uses aQute.lib* packages not already in bndlib (or any other bnd jar on the runtime classpath), then change verify.groovy to avoid using the package.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-multi-project/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>biz.aQute.bnd-test</groupId>
+		<artifactId>reporter-test</artifactId>
+		<version>0.0.1</version>
+	</parent>
+
+	<artifactId>report-multi-project</artifactId>
```
bjhargrave: You should add `<groupId>biz.aQute.bnd-test</groupId>` to the coordinates.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-multi-project/projectA/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>biz.aQute.bnd-test</groupId>
+		<artifactId>report-multi-project</artifactId>
+		<version>0.0.1</version>
+	</parent>
+
+	<artifactId>projectA</artifactId>
```
bjhargrave: You should add `<groupId>biz.aQute.bnd-test</groupId>` to the coordinates.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-multi-project/projectA/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+	<version>0.0.1</version>
+	<packaging>pom</packaging>
```
bjhargrave: I suspect this should really be `jar` packaging.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-multi-project-scoped/projectC/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+
+	<parent>
+		<groupId>biz.aQute.bnd-test</groupId>
+		<artifactId>report-multi-project-scoped</artifactId>
+		<version>0.0.1</version>
+	</parent>
+
+	<artifactId>projectC</artifactId>
+	<version>0.0.1</version>
+	<packaging>pom</packaging>
```
bjhargrave: `jar` packaging.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-multi-project-scoped/projectC/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>biz.aQute.bnd-test</groupId>
+		<artifactId>report-multi-project-scoped</artifactId>
+		<version>0.0.1</version>
+	</parent>
+
+	<artifactId>projectC</artifactId>
```
bjhargrave: You should add `<groupId>biz.aQute.bnd-test</groupId>` to the coordinates.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-multi-project-scoped/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>biz.aQute.bnd-test</groupId>
+		<artifactId>reporter-test</artifactId>
+		<version>0.0.1</version>
+	</parent>
+
+	<artifactId>report-multi-project-scoped</artifactId>
```
bjhargrave: You should add `<groupId>biz.aQute.bnd-test</groupId>` to the coordinates.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-multi-project-scoped/projectD/pom.xml
```diff
@@ -0,0 +1,16 @@
...
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>biz.aQute.bnd-test</groupId>
+		<artifactId>report-multi-project-scoped</artifactId>
+		<version>0.0.1</version>
+	</parent>
+
+	<artifactId>projectD</artifactId>
```
bjhargrave: You should add `<groupId>biz.aQute.bnd-test</groupId>` to the coordinates.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-multi-project-scoped/projectD/pom.xml
```diff
@@ -0,0 +1,16 @@
...
+	<version>0.0.1</version>
+	<packaging>pom</packaging>
```
bjhargrave: `jar` packaging.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-multi-project/projectB/pom.xml
```diff
@@ -0,0 +1,16 @@
...
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>biz.aQute.bnd-test</groupId>
+		<artifactId>report-multi-project</artifactId>
+		<version>0.0.1</version>
+	</parent>
+
+	<artifactId>projectB</artifactId>
```
bjhargrave: You should add `<groupId>biz.aQute.bnd-test</groupId>` to the coordinates.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-multi-project/projectB/pom.xml
```diff
@@ -0,0 +1,16 @@
...
+	<version>0.0.1</version>
+	<packaging>pom</packaging>
```
bjhargrave: `jar` packaging.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-single-project-no-report/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>biz.aQute.bnd-test</groupId>
+		<artifactId>reporter-test</artifactId>
+		<version>0.0.1</version>
+	</parent>
+
+	<artifactId>report-single-project-no-report</artifactId>
```
bjhargrave: You should add `<groupId>biz.aQute.bnd-test</groupId>` to the coordinates.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-single-project-no-report/pom.xml
```diff
@@ -0,0 +1,35 @@
...
+	<version>0.0.1</version>
+	<packaging>pom</packaging>
```
bjhargrave: `jar` packaging.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-single-project/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>biz.aQute.bnd-test</groupId>
+		<artifactId>reporter-test</artifactId>
+		<version>0.0.1</version>
+	</parent>
+
+	<artifactId>report-single-project</artifactId>
```
bjhargrave: You should add `<groupId>biz.aQute.bnd-test</groupId>` to the coordinates.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-single-project/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+	<version>0.0.1</version>
+	<packaging>pom</packaging>
```
bjhargrave: `jar` packaging.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-multi-project/projectA/pom.xml
cdelg: Indeed! This is why my bundles wasn't built, thanks.

maven/bnd-reporter-maven-plugin/src/test/resources/integration-test/test/report-multi-project/pom.xml
cdelg: Why should I explicitly add  the GroupId, it is not inherited from the parent?

bjhargrave: For consistency with all other poms and to avoid defaulting. Lets be explicit.

cdelg:
Is this PR ok?

bjhargrave:
Thanks. I merged.

PR #2823, opened 2018-12-15
[maven] update docs

rotty3000:


maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
```diff
@@ -90,6 +92,12 @@
 	@Parameter(defaultValue = "${session}", readonly = true)
 	private MavenSession				session;
 
+	@Parameter
+	private List<String>				scopes	= Arrays.asList("compile", "runtime");
```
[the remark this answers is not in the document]
rotty3000: Good idea! I'll check that.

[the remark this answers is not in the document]
rotty3000: so trying:
```
@Parameter
private EnumSet<ResolutionScope>	scopes	= EnumSet.of(ResolutionScope.COMPILE, ResolutionScope.RUNTIME);
```
at runtime I get:
```
[INFO] [ERROR] Failed to execute goal biz.aQute.bnd:bnd-export-maven-plugin:4.2.0-SNAPSHOT:export (default) on project export: Execution default of goal biz.aQute.bnd:bnd-export-maven-plugin:4.2.0-SNAPSHOT:export failed: A required class was missing while executing biz.aQute.bnd:bnd-export-maven-plugin:4.2.0-SNAPSHOT:export: org/apache/maven/plugins/annotations/ResolutionScope
```
So seems the annotations are not available at runtime since they are part of the tooling API. I can implement the checking with our own wrapper?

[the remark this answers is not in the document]
bjhargrave: OK, I guess Maven did the samething OSGi did with the annotations package by only using CLASS retention annotations and since the enum is in that package, it is not available at runtime.

So the alternate would be for us to create our own enum type:

```java
public enum Scope {
  compile, runtime, ...
}
```

which we use:

```java
@Parameter(defaultValue="compile,scope")
private Set<Scope> scopes;
```

maven/bnd-resolver-maven-plugin/src/main/java/aQute/bnd/maven/resolver/plugin/ResolverMojo.java
```diff
@@ -69,6 +72,18 @@
 	@Parameter(defaultValue = "true")
 	private boolean						reportOptional;
 
+	@Parameter
+	@SuppressWarnings("serial")
+	private HashSet<Scope>				scopes	= new HashSet<Scope>() {
```
bjhargrave: What is this crazy code? Surely the parameter type should be `Set<Scope>`. And why do you need a HashSet anonymous subclass when you could simply use EnumSet.of?

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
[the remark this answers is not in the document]
rotty3000: It's not that the annotations are CLASS retention, the entire annotations artifact isn't available at runtime. Do we really want to make that a dependency?

Secondly, the @Parameter annotation cannot be used to set multiple values. That's why we set a default map. And I already made an enum type.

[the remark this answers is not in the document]
bjhargrave: Well it is not available at runtime since the contents are CLASS retention annotations and supporting enums. Just like org.osgi.service.component.annotations.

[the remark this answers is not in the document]
bjhargrave: > the @parameter annotation cannot be used to set multiple values.

OK. So:

```java
@Parameter
private Set<Scope> scopes = EnumSet.of(Scope.compile, Scope.runtime);
```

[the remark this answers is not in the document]
bjhargrave: Actually you can't intialize to an enum set since it is not writeable. So you could use a HashSet. But you still cannot populate with values here since there is no way to remove runtime if the config only has compile.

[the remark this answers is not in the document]
bjhargrave: We need some test cases for the scope config since the propose initialization appears to be a bug.

maven/bnd-resolver-maven-plugin/src/main/java/aQute/bnd/maven/resolver/plugin/ResolverMojo.java
bjhargrave: As I mentioned, EnumSet wont work. But neither will initialization with the default values since there would be no way to not have the default values in the set.

rotty3000:
Yup, that's really what I wanted to do. Thanks for the insight.

maven/bnd-resolver-maven-plugin/src/main/java/aQute/bnd/maven/resolver/plugin/ResolverMojo.java
rotty3000: Can you look at the updated code? Also, maven is really simple, It doesn't do ADDs to the preconfigured value, it just does a set, so if you have a default of:
```
	@Parameter
	private Set<Scope>					scopes	= EnumSet.of(Scope.compile, Scope.runtime);
```
and then the plugin is configured as:
```
<configuration>
	<scopes><scope>compile</scope></scopes>
</configuration>
```
it just does a `set` on the field with the assembled set. It does not try to rationalize the inputs against the defaults.

rotty3000: it DOES however coerce the values into the member type. Which is where the validation will occur.

So
```
<configuration>
	<scopes><scope>conpile</scope></scopes>
</configuration>
```
will fail.

maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java
```diff
@@ -91,6 +95,9 @@
 	@Parameter(defaultValue = "${session}", readonly = true)
 	private MavenSession				session;
 
+	@Parameter
+	private Set<Scope>					scopes	= EnumSet.of(Scope.compile, Scope.runtime, Scope.test);
```
bjhargrave: Isn't adding `test` a change from before when it was only compile and runtime? Does this present any backwards compatibility issues?

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
```diff
@@ -100,9 +110,16 @@
...
+			logger.debug("skip project as configured");
+			return;
+		}
+
 		try {
 			DependencyResolver dependencyResolver = new DependencyResolver(project, repositorySession, resolver,
-				system);
+				system, scopes.stream()
+					.map(Scope::name)
+					.collect(Collectors.toList()));
```
bjhargrave: Perhaps this logic should be centralized in DependencyResolver with a constructor that takes a `Set<Scope>` argument?

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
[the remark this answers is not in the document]
bjhargrave: I still think we need integration tests for the scope configuration.

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
rotty3000: sure!

rotty3000:
Updated, and I added a test that proves the parameter value is not ADD, but REPLACE.

maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java
bjhargrave: What about this question?

rotty3000: I suppose I could unset `test` from the defaults here.

rotty3000: fixed!

maven/bnd-testing-maven-plugin/README.md
```diff
@@ -59,6 +59,7 @@ Here's an example setting the `bundles` used for resolution.
...
+|`bundles`                       | This is the collection of files to use for locating bundles during the bndrun resolution. Can contain `bundle` child elements specifying the path to a bundle. These can be absolute paths. You can also specify `include` and `exclude` child elements using Ant-style globs to specify bundles. These are relative to the `${project.basedir}` directory. _Defaults to dependencies in the scopes specified by the `scopes` property, plus the current artifact (if any and `useMavenDependencies` is `true`)._ |
 |`useMavenDependencies`          | If `true`, adds the project's compile and runtime dependencies to the collection of files to use for locating bundles during the bndrun resolution. _Defaults to `true`._|
 |`reportOptional`                | If `true`, resolution failure reports (see `resolve`) will include optional requirements. _Defaults to `true`._|
+|`scopes`               | Specify from which scopes to collect dependencies. _Defaults to `compile, runtime, test`._|
```
bjhargrave: This doc will need updating to remove test.

rotty3000: fixed!

Issue #2800, opened 2018-11-01
updated to latest m2e framework results in endlessly building project


rotty3000:
The latest m2e is slightly more pedantic about propagating warnings.

You may end up in a scenario where you have a build _warning_, which `bndtools.m2e` ignores and allows the `jar:jar` task to be executed, finalized by a project refresh, which results in the build firing the _warning_, and so on in an endless loop.

PR #2826, opened 2018-12-17
[maven] prevent m2e infinitely building project that doesn't include the classesDir

rotty3000:
Fixes https://github.com/bndtools/bnd/issues/2800


rotty3000:
Nice! This is working @bjhargrave

Issue #2769, opened 2018-12-04
[bnd-maven-plugin] Add an <instructions> section


bosschaert:
The maven-bundle-plugin allows projects to customize bnd behaviour via an <instructions> section where configuration can be specified as XML which is passed to bnd for processing.

It would be nice if the bnd-maven-plugin would support a similar mechanism to specify bnd instructions to make it easier for people to migrate between the maven-bundle-plugin and the bnd-maven plugin.

As this now provides an additional mechanism to specify configuration (the others being a separate bnd.bnd file or a <bnd> section) only one mechanism should be used. 

In order to specify bnd keys that start with a dash, like `-sources`, the keys need to transformed a little as in XML a key is not allowed to start with a `-`. We can support these cases by allowing the keys to start with an underscore `_` and converting these into a dash. This is also what the maven-bundle-plugin does.

If a developer specifies more than one mechanism of configuring the plugin only one will be used and a warning should be emitted to indicate dubious plugin configuration.

bjhargrave:
Closing as we decided not to make this change

PR #2829, opened 2018-12-18
Initial support for bnd instructions via mvn <instructions> tag

bosschaert:
Instructions that start with a dash '-' can be specified in the
<instructions> XML as starting with an underscore '_'.
This addresses #2769

Signed-off-by: David Bosschaert <david.bosschaert@gmail.com>

### DIFF
--- a/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
+++ b/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
@@ -17,10 +17,12 @@ package aQute.bnd.maven.plugin;
  */
 
 import java.io.File;
+import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -64,6 +66,7 @@ import aQute.bnd.osgi.Builder;
 import aQute.bnd.osgi.Constants;
 import aQute.bnd.osgi.FileResource;
 import aQute.bnd.osgi.Jar;
+import aQute.bnd.osgi.Macro;
 import aQute.bnd.osgi.Processor;
 import aQute.bnd.osgi.Resource;
 import aQute.bnd.version.MavenVersion;
@@ -72,6 +75,7 @@ import aQute.lib.io.IO;
 import aQute.lib.strings.Strings;
 import aQute.lib.utf8properties.UTF8Properties;
 import aQute.service.reporter.Report.Location;
+import aQute.service.reporter.Reporter;
 
 @Mojo(name = "bnd-process", defaultPhase = LifecyclePhase.PROCESS_CLASSES, requiresDependencyResolution = ResolutionScope.COMPILE)
 public class BndMavenPlugin extends AbstractMojo {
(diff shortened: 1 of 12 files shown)


maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -523,15 +535,122 @@ private File loadProjectProperties(Builder builder, MavenProject bndProject, Mav
...
+	 * @param d1 The first XML object.
+	 * @param d2 The seconds XML object.
+	 * @param projectFile A properties file used by the properties loader for
+	 *            error reporting.
+	 * @param reporter A reporter for logging.
+	 * @return {@code true} if the two XML objects are equal after placeholder
+	 *         substitution.
+	 * @throws IOException If an IO exception occurs.
+	 */
+	private boolean xppDomEqualsWithSubtitution(Xpp3Dom d1, Xpp3Dom d2, File projectFile, Reporter reporter)
```
kwin: xppDomEqualsWithSubtitution -> xppDomEqualsWithSubstitution

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -523,15 +535,122 @@ private File loadProjectProperties(Builder builder, MavenProject bndProject, Mav
...
+					// we use setProperties to handle -include
+					builder.setProperties(baseDir, transformUnderscoreKeys(properties.replaceHere(baseDir)));
+				}
+			} else {
+				logger.warn("Pom defines both bndfile and bnd configuration. Ignoring the bnd configuration in pom: {}",
+					pomProject);
+			}
+		}
+
+		if (instructions != null && instructions.size() > 0 && configuration.getChild("instructions") != null) {
```
kwin: This is called once per each parent pom, but the evaluated properties field `instruction` is always the same (i.e. already considering the inheritance for this configuration parameter alone). Are you sure that the inheritance behaviour is the same for `bnd` and `instructions` (according to https://blog.sonatype.com/2011/01/maven-how-to-merging-plugin-configuration-in-complex-projects/)?

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
bosschaert: Thanks for spotting! I will update this.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
bosschaert: I put tests in place that ensure the correct inheritance behaviour is happening. They are part of the pull request. If you have a more specific concert, please let me know.

bosschaert:
Thanks for the comments @kwin - yes I'll update the documentation, but would like to do that in a separate pull request.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
kwin: I would stop recursion in case you found instructions, cause the additional calls of `loadProjectProperties` mean additional effort but will not lead to different results IIUC.

bosschaert: I think I know what you mean but I think for that loadProjectProperties would need to return some stopcondition which may make the code a little more complicated as it already returns a File. Is the slight extra time spent during the build worth the extra complexity or do you have a simple solution in mind?

bjhargrave:
After careful consideration, we decided that this is not something we should add to Bnd. Thanks for the hard work and sorry, that in the end, we did not want to merge it.

PR #2839, opened 2018-12-21
use cast method

gamerson:


bjhargrave:
Why is this necessary? It adds a method invoke when there was none. Generally the calls sites of these methods will already perform a cast to the generic type the caller expects. So then you end up with 2 casts.

gamerson:
I wanted to remove the unnecessary cast warning suppressions and use the cast method.  I am reprised that the `type<T>.cast(...)` would add an extra bytecode instruction where `(T) ...` would not have the exact same (or very similar) instruction.

gamerson:
But i agreed, if this adds an extra cast, it isn't worth it to avoid a compiler warning.

bjhargrave:
>  I am surprised that the `type<T>.cast(...)` would add an extra cast instruction where `(T) ...` would not have the exact same (or very similar) instruction.

Lets look at this simple class:

```java
public class CastTest {
	public CastTest() throws Throwable {
		String s1 = newInstance1(String.class);
		String s2 = newInstance2(String.class);
	}
	private static final MethodType defaultConstructor = methodType(void.class);
	public static <T> T newInstance1(Class<T> rawClass) throws Throwable {
		return (T) publicLookup().findConstructor(rawClass, defaultConstructor)
			.invoke();
	}
	public static <T> T newInstance2(Class<T> rawClass) throws Throwable {
		return rawClass.cast(publicLookup().findConstructor(rawClass, defaultConstructor)
			.invoke());
	}
}
```

`newInstance1` uses a `(T)` cast which is a noop since `T` is erased to `Object` while `newInstance2` uses `Class.cast`. This generates the following bytecode:

```
public static <T extends java.lang.Object> T newInstance1(java.lang.Class<T>) throws java.lang.Throwable;
    Code:
      stack=3, locals=1, args_size=1
         0: invokestatic  #48                 // Method java/lang/invoke/MethodHandles.publicLookup:()Ljava/lang/invoke/MethodHandles$Lookup;
         3: aload_0
         4: getstatic     #22                 // Field defaultConstructor:Ljava/lang/invoke/MethodType;
         7: invokevirtual #54                 // Method java/lang/invoke/MethodHandles$Lookup.findConstructor:(Ljava/lang/Class;Ljava/lang/invoke/MethodType;)Ljava/lang/invoke/MethodHandle;
        10: invokevirtual #60                 // Method java/lang/invoke/MethodHandle.invoke:()Ljava/lang/Object;
        13: areturn

public static <T extends java.lang.Object> T newInstance2(java.lang.Class<T>) throws java.lang.Throwable;
    Code:
      stack=4, locals=1, args_size=1
         0: aload_0
         1: invokestatic  #48                 // Method java/lang/invoke/MethodHandles.publicLookup:()Ljava/lang/invoke/MethodHandles$Lookup;
         4: aload_0
         5: getstatic     #22                 // Field defaultConstructor:Ljava/lang/invoke/MethodType;
         8: invokevirtual #54                 // Method java/lang/invoke/MethodHandles$Lookup.findConstructor:(Ljava/lang/Class;Ljava/lang/invoke/MethodType;)Ljava/lang/invoke/MethodHandle;
        11: invokevirtual #60                 // Method java/lang/invoke/MethodHandle.invoke:()Ljava/lang/Object;
        14: invokevirtual #69                 // Method java/lang/Class.cast:(Ljava/lang/Object;)Ljava/lang/Object;
        17: areturn
```

The `newInstance2` variation adds an additional `invokevirtual` method call to Class.cast.

When we look at the call sites for both methods:

```
public aQute.configurable.CastTest() throws java.lang.Throwable;
    Code:
      stack=1, locals=3, args_size=1
         0: aload_0
         1: invokespecial #30                 // Method java/lang/Object."<init>":()V
         4: ldc           #32                 // class java/lang/String
         6: invokestatic  #34                 // Method newInstance1:(Ljava/lang/Class;)Ljava/lang/Object;
         9: checkcast     #32                 // class java/lang/String
        12: astore_1
        13: ldc           #32                 // class java/lang/String
        15: invokestatic  #38                 // Method newInstance2:(Ljava/lang/Class;)Ljava/lang/Object;
        18: checkcast     #32                 // class java/lang/String
        21: astore_2
        22: return
```

We see that the call site for both methods calls `checkcast String.class`  on the return value where `String` is the generic type expected by the caller. So adding a `Class.cast` call in the method instead of the simple `(T)` cast does not add any additional value since the call site will always cast and the call site knows the actual type to use and thus does not need the dynamic cast of `Class.cast`.

gamerson:
Thanks for taking time to explain in detail with the relevant byte code, every enlightening.  What happens if we weren't returning from a function, and just coming two method bodies one with a `(T)` type cast and one with `type.cast()`?

bjhargrave:
Function returns here don't change things. The cast`(T)` is always a `checkcast` to the type T erases to (unless `T` erases to `Object` in which point a `checkcast` is pointless). 

For example: 

```java
public <T extends java.lang.annotation.Annotation> T getAnnotation(Class<T> c) throws Exception {
	if (elements == null) {
		elements = new LinkedHashMap<>();
	}
	return CONVERTER.convert(c, elements);
}
```

`T` erases to `java.lang.annotation.Annotation`, so the result of the `CONVERTER.convert` call must be cast to `java.lang.annotation.Annotation` before returning it to ensure the generic contract of the method's return type is met.

```
public <T extends java.lang.annotation.Annotation> T getAnnotation(java.lang.Class<T>) throws java.lang.Exception;
    Code:
      stack=3, locals=2, args_size=2
         0: aload_0
         1: getfield      #5                  // Field elements:Ljava/util/Map;
         4: ifnonnull     18
         7: aload_0
         8: new           #41                 // class java/util/LinkedHashMap
        11: dup
        12: invokespecial #42                 // Method java/util/LinkedHashMap."<init>":()V
        15: putfield      #5                  // Field elements:Ljava/util/Map;
        18: getstatic     #54                 // Field CONVERTER:LaQute/lib/converter/Converter;
        21: aload_1
        22: aload_0
        23: getfield      #5                  // Field elements:Ljava/util/Map;
        26: invokevirtual #55                 // Method aQute/lib/converter/Converter.convert:(Ljava/lang/Class;Ljava/lang/Object;)Ljava/lang/Object;
        29: checkcast     #56                 // class java/lang/annotation/Annotation
        32: areturn
```

And we see that is indeed the case here. A call site of this method which may expect a subclass of Annotation:

```java
public void annotation(Annotation annotation) throws Exception {
	String fqn = annotation.getName()
		.getFQN();
	if (fqn.equals("aQute.bnd.annotation.xml.XMLAttribute")) {
		xmlAttr = annotation.getAnnotation(XMLAttribute.class);
	}
}
```

will do a `checkcast` on the result ensure the result is of the expected class:

```
public void annotation(aQute.bnd.osgi.Annotation) throws java.lang.Exception;
    Code:
      stack=3, locals=3, args_size=2
         0: aload_1
         1: invokevirtual #7                  // Method aQute/bnd/osgi/Annotation.getName:()LaQute/bnd/osgi/Descriptors$TypeRef;
         4: invokeinterface #15,  1           // InterfaceMethod aQute/bnd/osgi/Descriptors$TypeRef.getFQN:()Ljava/lang/String;
         9: astore_2
        10: aload_2
        11: ldc           #16                 // String aQute.bnd.annotation.xml.XMLAttribute
        13: invokevirtual #17                 // Method java/lang/String.equals:(Ljava/lang/Object;)Z
        16: ifeq          32
        19: aload_0
        20: aload_1
        21: ldc           #10                 // class aQute/bnd/annotation/xml/XMLAttribute
        23: invokevirtual #18                 // Method aQute/bnd/osgi/Annotation.getAnnotation:(Ljava/lang/Class;)Ljava/lang/annotation/Annotation;
        26: checkcast     #10                 // class aQute/bnd/annotation/xml/XMLAttribute
        29: putfield      #12                 // Field xmlAttr:LaQute/bnd/annotation/xml/XMLAttribute;
        32: return
```

These `checkcast`s at the call sites of generic methods are in important part of the type safety of the generic type system in Java. So for the method returning the object, we really only need to use something like `(T)` to tell the compiler we know what we are doing and sometimes we have to further tell the compiler with a suppress warnings. A call to `Class.cast` can "help" here at the expense of a method call and that method does the '(T)' cast with a suppress warnings.

```java
    public T cast(Object obj) {
        if (obj != null && !isInstance(obj))
            throw new ClassCastException(cannotCastMsg(obj));
        return (T) obj;
    }
```

So using `Class.cast` in these circumstances just pushes the buck to another method :-)

PR #2842, opened 2018-12-25
[maven] unify bundles handling

rotty3000:


timothyjward:
This looks reasonable

maven/bnd-shared-maven-lib/src/main/java/aQute/bnd/maven/lib/resolve/ResolveSupport.java
```diff
@@ -58,4 +58,14 @@ public void setAutoRunEE(Bndrun run) {
...
 			.getEEName());
 	}
+
+	public void setAutoRunRequires(Bndrun run, MavenProject project) {
+		if (run.getProperty(Constants.RUNREQUIRES) == null && (project.getPackaging()
+			.equals("jar")
+			|| project.getPackaging()
+				.equals("war"))) {
+			run.setProperty(Constants.RUNREQUIRES,
+				String.format("osgi.identity;filter:='(osgi.identity=%s)'", project.getArtifactId()));
```
bjhargrave: Isn't this assuming that the generated bundle always uses the artifactId as the BSN? While that may be the default, it can be overridden easily and can make this auto setting quite wrong.

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
```diff
@@ -47,6 +49,7 @@
 import biz.aQute.resolve.ResolveProcess;
 
 @Mojo(name = "export", defaultPhase = PACKAGE, requiresDependencyResolution = ResolutionScope.COMPILE_PLUS_RUNTIME)
+@SuppressWarnings("deprecation")
```
bjhargrave: This seems horribly wrong to suppress deprecation at the class level. What is deprecated and why would we add new code which uses deprecated things? I can see that over time, things we use can become deprecated and we decide to continue to use the newly deprectated thing. But I don't think we should add new function which starts off using deprecated things. We should use the replacement for the deprecated thing or find another way which is not deprecated.

maven/bnd-resolver-maven-plugin/src/main/java/aQute/bnd/maven/resolver/plugin/ResolverMojo.java
```diff
@@ -38,6 +40,7 @@
  * Resolves the <code>-runbundles</code> for the given bndrun file.
  */
 @Mojo(name = "resolve", requiresDependencyResolution = ResolutionScope.COMPILE_PLUS_RUNTIME)
+@SuppressWarnings("deprecation")
```
bjhargrave: Same deprecated comment here.

maven/bnd-shared-maven-lib/src/main/java/aQute/bnd/maven/lib/resolve/ResolveSupport.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Set;
+
+import org.apache.maven.artifact.Artifact;
+import org.apache.maven.artifact.factory.ArtifactFactory;
+import org.apache.maven.model.Dependency;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.project.MavenProject;
+import org.apache.maven.project.artifact.InvalidDependencyVersionException;
+
+@SuppressWarnings("deprecation")
```
bjhargrave: Same deprecation comment here.

maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java
```diff
@@ -39,6 +41,7 @@
 import biz.aQute.resolve.ResolveProcess;
 
 @Mojo(name = "testing", defaultPhase = LifecyclePhase.INTEGRATION_TEST, requiresDependencyResolution = ResolutionScope.COMPILE_PLUS_RUNTIME)
+@SuppressWarnings("deprecation")
```
bjhargrave: Same deprecation comment here.

rotty3000:
Thanks for reviewing Tim. However, I am still playing with the implementation to sync up with m2e. It won't be nice if this breaks in m2e. I'll ping back here again when I think I'm done with that work.

rotty3000:
... and of course all @bjhargrave comments need resolving ...

bjhargrave:
The `unifiy bundles handling` commit is a fine commit on its own and could be a nice little PR while other issues are sorted out.

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
rotty3000: I'll have to investigate more deeply, but it appears the Maven project deprecates things before there is a replacement. I searched the Maven codebase high and low and there is no alternative for this deprecated API. Even Maven itself still uses it.

maven/bnd-shared-maven-lib/src/main/java/aQute/bnd/maven/lib/resolve/ResolveSupport.java
rotty3000: This is bnd's default. But yes I do agree and loading bnd-maven-plugin configuration is part of the ongoing work. Thx

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
bjhargrave: If we must use the deprecated API, then we need to minimize the scope of the deprecation suppression and not use it on a broad scope such as the class. Using it on a broad scope will hide other deprecation warnings we might want to do something about.

maven/bnd-shared-maven-lib/src/main/java/aQute/bnd/maven/lib/resolve/ResolveSupport.java
```diff
@@ -32,4 +37,25 @@ public void includeDependencyManagement(boolean includeDependencyManagement, Mav
...
+		if (run.getProperty(Constants.RUNEE) != null) {
+			return;
+		}
+
+		String javaSpecificationVersion = System.getProperty("java.specification.version");
+		Version javaVersion = new Version(javaSpecificationVersion);
+
+		// The maven-compiler-plugin supports both `1.8` and `8` for
+		// this range only.
+		String mavenCompilerTarget = run.getProperty("maven.compiler.target");
```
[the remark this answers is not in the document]
rotty3000: Yup. Will do.

rotty3000:
Yeah I can do that @bjhargrave

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
rotty3000: I can do that.

PR #2851, opened 2019-01-08
[builder] only remove subdirectories from META-INF/maven dir during `-pom`

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -1210,4 +1210,14 @@ public void removePrefix(String prefixLow) {
...
 			.clear();
 	}
+
+	public void removeSubDirs(String prefixLow) {
+		String prefixHigh = prefixLow + "\uFFFF";
+		if (prefixLow.endsWith("/")) {
+			prefixLow = prefixLow.substring(0, prefixLow.length() - 1);
+			prefixHigh = prefixLow + "\uFFFF";
+		}
+		directories.subMap(prefixLow, prefixHigh)
```
bjhargrave: This seems wrong. You always must modify both resources and directories. It is wrong to only clear things out of directories and not resources. You also need to clear the resources in those directories from resources.

rotty3000: yeah, you're right. My previous logic was better and did `removePrefx()` on each subdir dir. I'll re-revert to that. thx

rotty3000: Updated

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -1210,4 +1210,24 @@ public void removePrefix(String prefixLow) {
...
+		}
+
+		Map<String, Resource> dir = directories.get(path);
+
+		if ((dir == null) || dir.isEmpty()) {
+			return;
+		}
+
+		dir.keySet()
+			.forEach(s -> {
```
bjhargrave: Can't this be simplified to 

```java
dir.keySet().forEach(this::removePrefix);
```

I cannot imagine the null check is needed and emptiness would be handled anyway.

rotty3000: the `isEmpty` is not needed, but the null check is.

bjhargrave: Why would the keySet have null values in it?

rotty3000:
After much turbulence I've got a fix and some tests.

rotty3000:
ping @bjhargrave

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -1210,4 +1210,28 @@ public void removePrefix(String prefixLow) {
 		directories.subMap(prefixLow, prefixHigh)
 			.clear();
 	}
+
+	public void removeSubDirs(String prefixLow) {
```
bjhargrave: Here is a much simpler implementation which I tested using the test cases here.

```java
	public void removeSubDirs(String prefixLow) {
		if (prefixLow.endsWith("/")) {
			prefixLow = prefixLow.substring(0, prefixLow.length() - 1);
		}
		Set<String> subDirs = new HashSet<>(directories.subMap(prefixLow, prefixLow + "\uFFFF")
			.keySet());
		subDirs.remove(prefixLow);
		subDirs.forEach(subDir -> removePrefix(subDir + "/"));
	}
```

bjhargrave:
There is a better fix in my edited comment:

```java
	public void removeSubDirs(String dir) {
		if (!dir.endsWith("/")) {
			dir = dir + "/";
		}
		List<String> subDirs = new ArrayList<>(directories.subMap(dir, dir + "\uFFFF")
			.keySet());
		subDirs.forEach(subDir -> removePrefix(subDir + "/"));
	}
```

Please use it.

rotty3000:
re-updated...

PR #2852, opened 2019-01-08
[reporter] Fixes

cdelg:
* Removes unused optional imports.
* Some fixes in the "helpers" private package.


biz.aQute.bnd.reporter/bnd.bnd
```diff
@@ -31,8 +31,6 @@ Export-Package: \
 
 Import-Package: \
     sun.*; resolution:=optional,\
-    com.google.appengine.*; resolution:=optional,\
-    com.google.apphosting.*; resolution:=optional,\
```
bjhargrave: Removing these lines results in a mandatory import on these packages which is bad since the packages are only used by protected Class.forName references.

cdelg: Hmm, I didn't see the mandatory import. I removed these lines because Eclipse generated a warning "package unused"? Do you have this warning too?

cdelg: Here is the Import-Package clause of the resulting jar compiled with Eclipse

Import-Package: 
sun.misc;resolution:=optional,
aQute.bnd.annotation.plugin;version="[1.0,2)",
aQute.bnd.build;version="[3.1,4)",
aQute.bnd.header;version="[2.1,3)",
aQute.bnd.help;version="[1.2,2)",
aQute.bnd.osgi;version="[4.2,5)",
aQute.bnd.service;version="[4.3,5)",
aQute.bnd.service.reporter;version="[1.0,1.1)",
aQute.bnd.version;version="[2.1,3)",
aQute.service.reporter;version="[1.0,2)",
biz.aQute.bnd.reporter.service.resource.converter;version="[1.0,1.1)",
javax.annotation,javax.xml.parsers,
javax.xml.transform,
javax.xml.transform.stream,
org.osgi.dto;version="[1.0,2)",
org.slf4j;version="[1.7,2)",
org.w3c.dom

Effectively when I compile the project with gradle I get the import! Do you know why the build are different?

bjhargrave: You must be using an older version of Bndtools which includes the corresponding older version of Bnd. The latest Bnd (4.2.0-SNAPSHOT) will find the type references and add the Import-Package. That is why I added those lines to the bnd file. The gradle build uses the 4.2.0-SNAPSHOT version of Bnd.

cdelg: I have removed the commit.

PR #2857, opened 2019-01-12
[launcher] embed the single class JUL bridge, launcher already depends on slf4j api

rotty3000:


rotty3000:
@pkriens I added this because **a)** JUL logging is painful and this makes it super easy **b)** I couldn't decide a good use case for when using the launcher you wouldn't want records logged by JUL to be routed through slf4j if you could get it for free.

Thoughts?

rotty3000:
hmm, I must have misread the imports and how they affect the runtime. thanks @bjhargrave

bjhargrave:
> I must have misread the imports

The jar is built as a bundle and thus imports the slf4j package. But this bundle is used as a plain jar on the classpath of the launched VM and then the plugin package is not used. This bundle is used on both sides of the launch: by the launcher and by the launchee.

rotty3000:
yeah, just got bamboozled by my own setup. This JUL stuff is just such a PITA that I'm trying to make it as seamless as possible, but no solution seems to be able to acomplish that.

PR #2858, opened 2019-01-12
[m2e] use existing maven logic to setup implicit repo, support dependencyManagement

rotty3000:


rotty3000:
@timothyjward, @gamerson if you wouldn't mind taking a look at this change (building it might be the best way to do that).

The summary is as follows:

- deduplicated all the boiler plate code from the 3 maven plugins that did resolve by creating a super type (instantiable class, not abstract, but also not a plugin on it's own)
- implemented support for implicit FileSetRepository to index `<dependencyManagement>` (via new `includeDependencyManagement` parameter)
- elevate the `bnd-shared-maven-lib` to the bnd workspace so it is consumable at that level
- re-use the `bnd-shared-maven-lib` logic to simplify bndtools.m2e integration and also gain `<dependencyManagement>` visibility in the bndrun repo view and allowing bndrun editor resolve to work against it.

bnd-shared-maven-lib/bnd.bnd
```diff
@@ -0,0 +1,59 @@
...
+group=biz.aQute.bnd
+
+Bundle-Description: Shared logic used by bnd maven plugins.
+Bundle-SymbolicName: bnd-shared-maven-lib
+Export-Package: \
+	aQute.bnd.maven.lib.configuration,\
+	aQute.bnd.maven.lib.resolve
+
+Import-Package: \
+	org.apache.maven;provider=m2e,\
```
timothyjward: This is a bad way to write this `Import-Package` instruction as it is fragile with respect to future changes. It should definitely be something more like

    Import-Package: \
        org.apache.maven.*;provider=m2e,\
        org.eclipse.aether.*;provider=m2e,\
        org.codehaus.plexus.*;provider=m2e,\
        *

bndtools.m2e/bnd.bnd
```diff
@@ -33,14 +36,31 @@ Bundle-SymbolicName: bndtools.m2e; singleton:=true
 Bundle-ActivationPolicy: lazy
 
 Import-Package: \
+	org.apache.maven;provider=m2e,\
```
timothyjward: As per the shared maven lib, this is a poor way to do this (look how many extra things had to be added!). It should definitely be something more like

    Import-Package: \
        org.apache.maven.*;provider=m2e,\
        org.eclipse.aether.*;provider=m2e,\
        org.codehaus.plexus.*;provider=m2e,\
        *

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
```diff
@@ -4,82 +4,41 @@
...
-import aQute.bnd.repository.fileset.FileSetRepository;
-import aQute.bnd.service.RepositoryPlugin;
 import aQute.lib.io.IO;
 import biz.aQute.resolve.Bndrun;
 import biz.aQute.resolve.ResolveProcess;
 
 @Mojo(name = "export", defaultPhase = PACKAGE, requiresDependencyResolution = ResolutionScope.COMPILE_PLUS_RUNTIME)
-public class ExportMojo extends AbstractMojo {
+@SuppressWarnings("deprecation")
+public class ExportMojo extends ResolveMojo {
```
timothyjward: I'm somewhat concerned by this - "has a" relationships are almost always better than "is a" relationships. In this case an Export task isn't a resolve task, it just might need one.

maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java
```diff
@@ -1,64 +1,41 @@
...
-import aQute.lib.io.IO;
 import aQute.lib.strings.Strings;
 import aQute.libg.glob.Glob;
 import biz.aQute.resolve.Bndrun;
 import biz.aQute.resolve.ResolveProcess;
 
 @Mojo(name = "testing", defaultPhase = LifecyclePhase.INTEGRATION_TEST, requiresDependencyResolution = ResolutionScope.COMPILE_PLUS_RUNTIME)
-public class TestingMojo extends AbstractMojo {
+@SuppressWarnings("deprecation")
+public class TestingMojo extends ResolveMojo {
```
timothyjward: As with Export - Testing may do a resolve, but that doesn't make it a resolve.

timothyjward:
> * elevate the bnd-shared-maven-lib to the bnd workspace so it is consumable at that level

We need to be really careful about doing this. Maven can be pretty picky about where the dependencies of a plugin come from - it would be very easy for this change to break all of the maven plugins (as consumed by users from Maven Central) that we provide if the dependency information of the released artifacts isn't 100% correct. 

We have had this sort of issue before when bndlib wasn't properly available in a way that Maven found palatable. Unfortunately the issues don't tend to show up if you have built the artefacts locally because they end up in your local repo.

bnd-shared-maven-lib/.classpath
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bjhargrave: I thought we had discussed that bnd-shared-maven-lib would remain part of the maven build, but some of the classes would be moved down to another project where the classes could be conditional packaged in m2e plugins and maven-dependency-plugin included in bnd-shared-maven-lib.

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
bjhargrave: And again with the SuppressWarnings on the class declaration. I really dislike this since it will hide future warning we should take action on.

maven/bnd-resolver-maven-plugin/src/main/java/aQute/bnd/maven/resolver/plugin/ResolverMojo.java
```diff
@@ -1,173 +1,13 @@
...
-		for (String warning : run.getWarnings()) {
-			logger.warn("Warning : {}", warning);
-		}
-		for (String error : run.getErrors()) {
-			logger.error("Error   : {}", error);
-			errors++;
-		}
-	}
-
+public class ResolverMojo extends ResolveMojo {
```
bjhargrave: An imperceptibly subtle class name difference!

maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java
bjhargrave: And again with the SuppressWarnings on the class declaration. I really dislike this since it will hide future warning we should take action on.

bnd-shared-maven-lib/bnd.bnd
rotty3000: got it!

bndtools.m2e/bnd.bnd
rotty3000: can do!

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
rotty3000: There's no other way in Maven plugin model to support inheritance. What do you propose? To keep the over 80% duplication?

bjhargrave: Aggregation?

bjhargrave: For example, see how maven-jar-plugin uses maven-archiver. That is aggregation.

bnd-shared-maven-lib/.classpath
bjhargrave: The naming suggests a maven lib but you are building in a bundle which one could assume will be installed in Eclipse with the Bndtools.m2e feature. But instead you are conditional packaging it into the bndtools m2e bundle. So why make a bundle that is never used as a bundle? bndlib _is_ used as a bundle (and a jar).

bjhargrave: Also this change converts bnd-shared-maven-lib to no longer have bnd-plugin-parent as its parent. Not sure if this is a real issue. But it is a change that is observable.

rotty3000:
updated!

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
rotty3000: hmm... I could look into this but it will not be possible to maintain compatibility as we'll have to move all shared parameters to a sub element of the configuration.

bjhargrave: I don't think you need to do that. Leave the configuration on the mojo and pass the fields to the aggregation which has the shared logic. This is what maven-jar-plugin does.

bnd-shared-maven-lib/.classpath
rotty3000: The fact that bnd-shared-maven-lib is a bundle (now) is just a side effect of being built by bnd workspace. It is not the intention to use it that way (this is true of many projects inside the bnd repo, not sure how this would be any different.)

bjhargrave: In order for bnd-shared-maven-lib to be used by the maven plugins, it must be released to maven central. That means you are making a bundle available to the world. That it, it becomes API. Currently, it is essentially impl detail to the maven plugins which I think it should remain.

rotty3000: I'm still not sure I understand. How is it any different than what we have today?

https://search.maven.org/artifact/biz.aQute.bnd/bnd-shared-maven-lib/4.1.0/jar

bjhargrave: That is a jar with a pom. This PR makes it a bundle with exported packages (API) which, of course, should be baselined since it is API. I don't think this jar is API. It is shared impl detail because we cannot easily do conditionalpackage in a maven build. I will note you clearly see this code as impl detail since you conditionalpackage it into the m2e bundle vs. importing the package and adding the bundle to the m2e feature.

maven/bnd-export-maven-plugin/src/main/java/aQute/bnd/maven/export/plugin/ExportMojo.java
rotty3000: You mean leave the 13 common parameters and 3 common components still spread across each Mojo? What does that buy us?

rotty3000: The whole point started because all 3 mojos were constantly out of sync as far as their parameters. By using a super type we consolidate all of that into one place.

bjhargrave: It buys us aggregation instead of inheritance. Aggregation is always the first best choice since it provides maximum freedom to the change the implementation.

rotty3000: So, I don't think that what you guys are suggesting is possible, not while also maintaining backward compatibility. Please point me at what you think is an implementation of this because so far I can't find one. The examples pointed out don't appear to be solutions.

bjhargrave: For example, Define a DTO class which has all the common fields. The execute can copy the mojo fields into a DTO object and pass the DTO object to the common resolve operation.

Or, for example, define a mojo base class which has getters for all the common fields and then pass the mojo (`this`) to the common resolve operation.

rotty3000: @bjhargrave I'm well aware of how that could be designed. What I'm suggesting is that does NOT solve the repeated 13+3 fields that are at issue here.

rotty3000: so you're suggesting throwing out the Parameter annotations...

bjhargrave: Agree, but I don't care about repeating the fields. I care about shared logic. The DTO/base getters class can enforce the commonality in the fields.

bjhargrave: Repeat the fields. Or maybe this is telling us, we need to common configuration type for resolving to transition to.

rotty3000: The repeating fields is where we constantly mess up!

bjhargrave: Which is why maybe we need to introduce a common resolving configuration type and transition to use it in the various plugins which support resolving.

rotty3000:
What about now?

bnd-shared-maven-lib/.classpath
rotty3000: I'm not sure how to address this particular issue.

bjhargrave: My suggestion is to leave bnd-shared-maven-lib project in the maven build. Move any packages that must be shared to bndtools.m2e project (so bndtools.m2e has use of the packages). Then change bnd-shared-maven-lib to pull those packages into it using maven-dependency-plugin on the bndtools.m2e jar which will be in the local repo.

The we have the shared package source code in one place (bndtools.m2e project) and copies of the shared package class files in 2 places (bndtools.m2e jar and bnd-shared-maven-lib jar).

rotty3000: If i do that, there will be zero code in the `bnd-shared-maven-lib` project!

rotty3000:
updated!

rotty3000:
@bjhargrave you good with this now?

maven/bnd-export-maven-plugin/pom.xml
```diff
@@ -26,6 +26,7 @@
 		<dependency>
 			<groupId>org.apache.maven</groupId>
 			<artifactId>maven-plugin-api</artifactId>
+			<scope>compile</scope>
```
bjhargrave: I would have thought the scope should have been set in the dependenciesManagement section in the parent.  I would also think this should not need to be set here?

maven/bnd-resolver-maven-plugin/pom.xml
```diff
@@ -26,6 +26,7 @@
 		<dependency>
 			<groupId>org.apache.maven</groupId>
 			<artifactId>maven-plugin-api</artifactId>
+			<scope>compile</scope>
```
bjhargrave: Same here.

maven/bnd-testing-maven-plugin/pom.xml
```diff
@@ -26,6 +26,7 @@
 		<dependency>
 			<groupId>org.apache.maven</groupId>
 			<artifactId>maven-plugin-api</artifactId>
+			<scope>compile</scope>
```
bjhargrave: And here

maven/bnd-export-maven-plugin/pom.xml
rotty3000: So, this is needed when you have Mojo inheritance using maven annotations (look it up). But since we don't have that anymore, I can remove it.

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/configuration/FileTree.java
```diff
@@ -0,0 +1,28 @@
+package aQute.bnd.maven.lib.configuration;
+
+@SuppressWarnings("restriction")
```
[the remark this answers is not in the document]
rotty3000: No, these stem from the fact that `aQute.lib.io.FileTree` is not in an exported package, but is super type of `aQute.bnd.maven.lib.configuration.FileTree` which is exported. However, `aQute.lib.io` is conditionally packaged into `biz.aQute.bnd.maven` BUT bndtools isn't smart enough to recognize that.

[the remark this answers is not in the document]
rotty3000: I'm just going to remove all of those for now.

biz.aQute.bnd.maven/bnd.bnd
```diff
@@ -0,0 +1,38 @@
...
+	aQute.lib*
+
+-testpath: \
+	${junit},\
+	${mockito},\
+	org.codehaus.plexus.classworlds,\
+	org.codehaus.plexus:plexus-utils,\
+	slf4j.simple;version=latest
+
+-fixupmessages.maven: "Export aQute.bnd.maven.lib.configuration...";is:=ignore
```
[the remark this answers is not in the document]
rotty3000: Otherwise I get:
```
> Task :biz.aQute.bnd.maven:jar
warning: Export aQute.bnd.maven.lib.configuration,  has 1,  private references [aQute.lib.io]
```
which I believe is inert since `aQute.lib.io` is conditionally packaged into the bundle.

rotty3000:
updated!

rotty3000:
@bjhargrave the failed build is just a p2/network download issue. I can't restart the build with my azure credentials.

rotty3000:
updated!

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/configuration/FileTree.java
```diff
@@ -0,0 +1,111 @@
+package aQute.bnd.maven.lib.configuration;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.List;
+
+public class FileTree {
+
+    private class _FileTree extends aQute.lib.io.FileTree {
```
bjhargrave: This should be a static class to avoid passing `this`.

bjhargrave: Or just changed `aQute.lib.io.FileTree.addFile` from protected to public to avoid the needs to subclass.

rotty3000:
updated!

rotty3000:
@timothyjward are you ok with this at this point?

timothyjward:
> @timothyjward are you ok with this at this point?

LGTM

PR #2863, opened 2019-01-16
Runtime support

pkriens:
Require-Bundle support and runtime project (which is in progress)


biz.aQute.bnd.runtime/bnd.bnd
```diff
@@ -0,0 +1,29 @@
+# Set javac settings from JDT prefs
+-include: ${workspace}/cnf/includes/jdt.bnd
+
+-buildpath: \
+    osgi.annotation;version=latest, \
+    aQute.libg;version=latest;maven-scope=provided;packages=*,\
+    osgi.core;version=@6;maven-scope=provided,\
+    osgi.cmpn;version=@6;maven-scope=provided,\
```
bjhargrave: Please DO NOT USE osgi.cmpn. I put some effort into ridding the build of osgi.cmpn. Use the individual jars so we have proper control over the versions of individual components used.

Thanks.

cnf/build.bnd
```diff
@@ -57,3 +57,5 @@ Bundle-Developers: \
 -runproperties.testlauncher0: in.workspace=workspace
 -runproperties.testlauncher1: in.project=workspace
 -runproperties.testlauncher2: in.bndrun=workspace
+
+-remoteworkspace: true
```
bjhargrave: It already says remoteworkspace true 4 lines above.

cnf/ext/central.mvn
```diff
@@ -1,6 +1,7 @@
 org.osgi:osgi.annotation:7.0.0
 org.osgi:org.osgi.core:4.2.0
 org.osgi:osgi.core:6.0.0
+org.osgi:osgi.cmpn:6.0.0
```
bjhargrave: Please DO NOT USE osgi.cmpn. I put some effort into ridding the build of osgi.cmpn. Use the individual jars so we have proper control over the versions of individual components used.

Thanks.

PR #2874, opened 2019-01-18
Configurable packaging types (Fixes #2870)

### LINKED ISSUE #2870: bnd-maven-plugin: packaging type could be other than jar and war
Unfortunatly with 4.1.0 the packaging types `jar` and `war` are hardcoded. While it makes sense in most cases to limit it to those two types, there are situations where Manifest generation is needed for other types. For example, we have some legacy modules that are packaged as `ejb` and can also be used as bundles (the packaging type is needed for Eclipse WTP). This used to work fine before, but is now impossible. At least there should be a way to suppress this check or to specify a custom packaging type if needed.

erwint:
By default, only jar and war packaging types are being processed. If you
want to run bnd for other packaging types, you can define the list of
packaging types bnd will be run for. All types will be treated like
'jar', only 'war' has a special codepath.

```
<configuration>
  <packagingTypes>
    <packagingType>jar</packagingType>
    <packagingType>war</packagingType>
    <packagingType>xxx</packagingType>
  </packagingTypes>
</configuration>
```

Alternatively, you can use -Dbnd.packagingTypes=jar,war,xxx to specify
the types to use

### DIFF
--- a/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
+++ b/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
@@ -119,6 +119,9 @@ public class BndMavenPlugin extends AbstractMojo {
 	@Parameter(property = "bnd.skip", defaultValue = "false")
 	private boolean									skip;
 
+	@Parameter(property = "bnd.packagingTypes", defaultValue = PACKAGING_JAR + "," + PACKAGING_WAR)
+	private List<String>							packagingTypes;
+
 	/**
 	 * File path to a bnd file containing bnd instructions for this project.
 	 * Defaults to {@code bnd.bnd}. The file path can be an absolute or relative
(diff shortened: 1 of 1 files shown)


maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
```diff
@@ -119,6 +120,9 @@
 	@Parameter(property = "bnd.skip", defaultValue = "false")
 	private boolean									skip;
 
+	@Parameter(property = "bnd.packagingTypes", defaultValue = PACKAGING_JAR + "," + PACKAGING_WAR)
+	private String[]								packagingTypes;
```
bjhargrave: Can we not make this `List<String>` to avoid the need for `Arrays.asList` below?

erwint:
I'm not sure why it was changed in the first place, that's why I asked before on #2870 if you want it configurable. I could see reasons why it might make sense to make it a very narrow whitelist instead of just excluding `pom` packaging. I would also agree, that _normally_ an `ejb` packaged module is not a bundle.

So just let me know which way you actually prefer, and then I'll either perform the request change or remove the parameter and restore the old "ignore poms only" behavior.

kwin:
Regarding

> Perhaps a better idea may be to revert to how is was before. If the packaging is pom, then exit. Otherwise proceed. Then we don't need to configure a list of packaging types.

As long as Maven does not support mixins  let us please stick to whitelisting instead of blacklisting. It is very common to have the same parent pom for different packagings. Not all of them might want to leverage bnd, even though you might want to configure this plugin in the parent.

rotty3000:
@kwin so you are inclined to have the behavior suggested in the PR?

kwin:
Yes, I actually am in favor of merging this PR to make the whitelisting configurable.

rotty3000:
I also agree, but I would like the change that @bjhargrave suggested to be made first :)

bjhargrave:
Thanks!

PR #2878, opened 2019-01-21
[junit] make sure dependencies are not transitive

rotty3000:


rotty3000:
@bjhargrave updated!

biz.aQute.junit/bnd.bnd
```diff
@@ -6,7 +6,8 @@
 	osgi.cmpn;version=4.2;maven-scope=provided,\
 	aQute.libg;version=project,\
 	biz.aQute.bndlib;version=latest;maven-scope=provided,\
-	${junit}
```
bjhargrave: With the new list macro I just added, you should be able to change `${junit}` to 

```java
${replace;${list;junit};$;\\;maven-scope=provided}
```

bjhargrave: It doesn't work. The comma in the version range, cause the splitting logic to split in the middle of the version.
More work required...

PR #2880, opened 2019-01-22
[m2e] link resolve configuration by maven execution id

rotty3000:


biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
```diff
@@ -176,7 +176,9 @@ public FileSetRepository getFileSetRepository() throws Exception {
 
 		DependencyResolver dependencyResolver = new DependencyResolver(project, repositorySession, resolver, system, scopes);
 
-		return fileSetRepository = dependencyResolver.getFileSetRepository(project.getName(), bundles, useMavenDependencies);
+		String name = (project.getName() != "") ? project.getName() : project.getArtifactId();
```
bjhargrave: I think this is better as:

```java
String name = project.getName().isEmpty() ? project.getArtifactId() : project.getName();
```

rotty3000: sure

bjhargrave: `!= ""` is a pretty specific test since not all empty strings are the interned empty string.

rotty3000: got it, already updated.

rotty3000: note that I also pushed a formatting only commit in here to sync up settings with other workspace projects, so be advised. I could pull it out if it confuses the reset.

PR #2885, opened 2019-01-23
Reporter enhancements

cdelg:
Some enhancements for the reporter project:
* Add methods to setup a generator for a specific use case: The generator is mostly used to extract data from a bundle, a project or a workspace/aggregator project. A method is added to the builder to customize the generator for each of this use case. This prevents to duplicate code when used from maven, bnd CLI, gradle,...
* Replace reflection with a DTO in the manifest extractor

### DIFF
--- a/biz.aQute.bnd.reporter/src/biz/aQute/bnd/reporter/generator/ReportGeneratorBuilder.java
+++ b/biz.aQute.bnd.reporter/src/biz/aQute/bnd/reporter/generator/ReportGeneratorBuilder.java
@@ -105,6 +105,47 @@ public class ReportGeneratorBuilder {
   }
 
   /**
+   * Add the default plugins used to extract bundle data.
+   * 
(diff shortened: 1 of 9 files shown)


biz.aQute.bnd.reporter/src/biz/aQute/bnd/reporter/manifest/dto/OSGIHeadersDTO.java
```diff
@@ -0,0 +1,37 @@
+package biz.aQute.bnd.reporter.manifest.dto;
+
+import java.util.List;
+import org.osgi.dto.DTO;
+
+public class OSGIHeadersDTO extends DTO {
```
bjhargrave: Please spell `OSGi` correctly with a lower case `i`. Thanks.

bjhargrave:
Thanks!

PR #2892, opened 2019-01-29
Support Maven parallel builds (mvn -T XXX)

erwint:
Mark as threadSafe = true and add static synchronization according to
guide on:
https://cwiki.apache.org/confluence/display/MAVEN/Parallel+builds+in+Maven+3#ParallelbuildsinMaven3-Mojothreadsafetyassertionchecklist

### DIFF
--- a/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
+++ b/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java
@@ -73,7 +73,7 @@ import aQute.lib.strings.Strings;
 import aQute.lib.utf8properties.UTF8Properties;
 import aQute.service.reporter.Report.Location;
 
-@Mojo(name = "bnd-process", defaultPhase = LifecyclePhase.PROCESS_CLASSES, requiresDependencyResolution = ResolutionScope.COMPILE)
+@Mojo(name = "bnd-process", defaultPhase = LifecyclePhase.PROCESS_CLASSES, requiresDependencyResolution = ResolutionScope.COMPILE, threadSafe = true)
 public class BndMavenPlugin extends AbstractMojo {
 	private static final Logger						logger					= LoggerFactory
 		.getLogger(BndMavenPlugin.class);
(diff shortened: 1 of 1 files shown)


erwint:
Maven assumes all plugins non thread safe, unless declared otherwise. This results in lots of warnings during parallel builds about non thread safe plugins being in use and possibly messing up your build. So you need to either be sure you're threadsafe (which bnd is not), or you need to add synchronization (which I did). Then add the annotation to declare it's not okay to be used in parallel builds.

You could of course add this to all other plugins, but this is the major one in use.

bjhargrave:
So then all this does is remove a warning when running maven in parallel mode? Seems the like warning is accurate since bnd code is not known to be thread-safe.

Still would rather waiting on exploring this until after 4.2.

erwint:
The annotation removes the warning, the lock makes it threadsafe

bjhargrave:
Defer til after 4.2. The we should just declare them threadsafe and skip a global lock. The Builder, which is all that this plugin uses, should be thread safe.

erwint:
I can assure you it's not threadsafe and will produce wrong manifests sometimes without the lock! Since we added the lock it works fine

bjhargrave:
OK, 4.2 is in the can. 

If we want to do this for 4.3, then I would think we would want to do it for all of the Bnd maven plugins.

And if we do it for all of the Bnd maven plugins, then we probably need a central lock object that all plugins can use since all plugins can call into bndlib and potentially use non-thread safe code. So we would need to locate the lock object in biz.aQute.bnd.maven. This jar is used by most (but not currently all) Bnd maven plugins. It is also used by the Bndtools m2e plugins. Do the Bndtools m2e plugins also need to use this lock?

Seems like lots of thought is required here. I don't think it is enough for just bnd-maven-plugin to have its own lock.

bjhargrave:
@erwint Please check out https://github.com/bndtools/bnd/pull/3250. We have made some fixes to Bnd which should address some thread safety issues for the bnd maven plugin. The above PR now marks the bnd maven plugins as thread safe. It would be great if you could try this PR out in your environment?

bjhargrave:
I am going to close this since #3250 has now marked the bnd maven plugins as thread safe.

PR #2893, opened 2019-01-29
Actually use -donotcopy when unrolling jar

erwint:
While the documentation states `-donotcopy` will be taken into account
when unrolling a JAR, this has not yet been implemented.

see https://github.com/bndtools/bnd/blob/master/docs/_instructions/includeresource.md

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
@@ -52,6 +52,7 @@ import aQute.bnd.version.Version;
 import aQute.lib.collections.MultiMap;
 import aQute.lib.hex.Hex;
 import aQute.lib.io.IO;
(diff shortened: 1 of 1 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
```diff
@@ -1273,6 +1273,10 @@ public boolean addAll(Jar to, Jar sub, Instruction filter, String destination) {
 			if ("META-INF/MANIFEST.MF".equals(name))
 				continue;
 
+			String parts[] = name.split("/");
```
bjhargrave: This is an expensive operation to get the last segment of the path. Instead do

```java
if (doNotCopy(Strings.getLastSegment(name, '/'))
```

erwint: changed as suggested

PR #2907, opened 2019-02-01
Fix docs manual for -pom instruction

georgievst:
Fix Bundle-Developer to Bundle-Developers manifest header.
Fix warning on diffignore.md when processed by the jekyll.
Make -pom instruction more readable by introducing given parameters in a table view.


bjhargrave:
Thanks!

PR #2913, opened 2019-02-02
Reintroduce ability to launch Bndtools through eclipse

seanbright:
With the upgrade to Eclipse Oxygen and the merging of the Bnd and
Bndtools repository, we lost the ability to debug Bndtools by running it
from within Eclipse.


cnf/ext/central.mvn
```diff
@@ -70,6 +70,7 @@ org.littleshoot:littleproxy:1.1.2
 io.netty:netty-all:4.1.0.Final
 
 com.google.guava:guava:19.0
+com.google.guava:guava:21.0
```
bjhargrave: I think you also need to modify the bnd.bnd file in biz.aQute.bnd.reporter to add a version range

    com.google.guava;version="[18.0,19)",\

PR #2914, opened 2019-02-02
TemplateLoader: Race condition when loading templates

seanbright:
Because TemplateLoader.findTemplates() returns a collection of Promises, we don't want to release our service reference until after those Promises have finished, as they may still be using internal service state.

This fixes an issue where the list of available templates is sometimes not populated when trying to create a new project.


pkriens:
Great to have you working on this! Thx!

PR #2920, opened 2019-02-05
fixes #2792 Unable to Run As OSGi JUnit Test when using Maven

### LINKED ISSUE #2792: Unable to Run As OSGi JUnit Test when using Maven
If I'm using the bnd-testing-maven-plugin then I don't need to use a standalone index. If the plugin is configured for it then all of my maven dependencies are made available for resolve and run by the testing plugin, so no indexes are actually needed in that mode

This is great from the command line, but the resulting bndrun cannot be run from Bndtools. The main error is about a missing launcher plugin, but additional failures occur due to the lack of a Bndtools project nature. This also causes the Run As OSGi JUnit Test... option to disappear from the right click menu of the test class.

rotty3000:


bndtools.m2e/bnd.bnd
```diff
@@ -36,10 +37,11 @@ Bundle-SymbolicName: bndtools.m2e; singleton:=true
...
 
 Import-Package: \
- org.apache.maven.*;bundle-symbolic-name="${@bundlesymbolicname}";bundle-version="${range;[==,+);${@bundleversion}}";version=!,\
- org.codehaus.plexus.*;bundle-symbolic-name="${@bundlesymbolicname}";bundle-version="${range;[==,+);${@bundleversion}}";version=!,\
- ${eclipse.importpackage},\
- *
+	org.apache.maven.*;bundle-symbolic-name="${@bundlesymbolicname}";bundle-version="${range;[==,+);${@bundleversion}}";version=!,\
+	org.codehaus.plexus.*;bundle-symbolic-name="${@bundlesymbolicname}";bundle-version="${range;[==,+);${@bundleversion}}";version=!,\
+	${eclipse.importpackage},\
+	bndtools.launch,\
```
bjhargrave: Why isn't `bndtools.launch` found by the `*` entry? It seems wrong to hard code this. Is this because it is only referenced in the plugin.xml? If so, then a comment is needed here so someone tidying up later does not delete the line assuming it is covered by `*`. Perhaps a better idea is to put some reference (along with a comment why) to bndtools.launch in an m2e class so that Bnd will import bndtools.launch based upon package use.

bndtools.m2e/src/bndtools/m2e/LaunchPropertyTester.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    public boolean test(Object receiver, String property, Object[] args, Object expectedValue) {
+        IJavaElement elem = (IJavaElement) receiver;
+        IJavaProject javaProject = elem.getJavaProject();
+
+        if (javaProject == null) {
+            return false;
+        }
+
+        if (PROP_IS_IN_BND_MAVEN_PROJECT.equals(property)) {
```
bjhargrave: How about a switch statement? I much prefer them to if/then/else on a series of strings.

bndtools.m2e/bnd.bnd
rotty3000: you're assumption is correct, I'll add a reference and comment.

bndtools.m2e/src/bndtools/m2e/LaunchPropertyTester.java
rotty3000: sure

rotty3000:
updated!

PR #2921, opened 2019-02-05
[launcher] support eager start option

rotty3000:


rotty3000:
Woo, with this we no longer need a caffeine bundle to run the CT's against lazy equinox bundles!

biz.aQute.launcher/src/aQute/launcher/constants/LauncherConstants.java
```diff
@@ -44,11 +44,12 @@
 	final static String				LAUNCH_NAME					= "launch.name";
 	final static String				LAUNCH_NOREFERENCES			= "launch.noreferences";
 	final static String				LAUNCH_NOTIFICATION_PORT	= "launch.notificationPort";
+	final static String				LAUNCH_EAGER_ACTIVATION		= "launch.eager.activation";
```
seanbright: Would `launch.activation.eager` make more sense? General to specific.

```suggestion
	final static String				LAUNCH_EAGER_ACTIVATION		= "launch.activation.eager";
```

rotty3000: I was used the exact property name specified by @bjhargrave!?!

seanbright: Ah. My suggestion is better though!

rotty3000: :) but I'm open for naming options, as long as this passes review before release ;)

bjhargrave: I like Sean's naming suggestion.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -514,11 +514,16 @@ void startBundles(List<Bundle> tobestarted) throws Exception {
 		// Add all bundles that we've tried to start but failed
 		all.addAll(wantsToBeStarted);
 
+		int startOptions = Bundle.START_ACTIVATION_POLICY;
+		if (parms.eagerActivation) {
+			startOptions = 0;
```
bjhargrave: I guess instead of setting to 0, it may be more future-proof to AND-off the bit.

rotty3000:
updated.

biz.aQute.launcher/src/aQute/launcher/constants/LauncherConstants.java
```diff
@@ -94,6 +96,7 @@
 			p.setProperty(LAUNCH_NAME, name);
 
 		p.setProperty(LAUNCH_NOTIFICATION_PORT, String.valueOf(notificationPort));
+		p.setProperty(LAUNCH_ACTIVATION_EAGER, activationEager + "");
```
seanbright: Sorry... one more thing - can we not keep kicking this can down the road?
```suggestion
		p.setProperty(LAUNCH_ACTIVATION_EAGER, String.valueOf(activationEager));
```

PR #2922, opened 2019-02-05
Fix biz.aQute.junit build file during Gradle evaluation. Use replacel…

georgievst:
…ist instead of list macro.

An error occurs during Gradle evaluation with the following message:
> Failed to apply plugin [id 'biz.aQute.bnd.workspace']
   > Invalid syntax for version: [4.11;maven-scope=provided,5)

Signed-off-by: Stefan Georgiev <stefangeorgiev@outlook.com>


bjhargrave:
It already is `replacelist`. See https://github.com/bndtools/bnd/commit/ee30a2758a6b331dbc87767b116d3accd234d91f#diff-fe78bba5a123f099bb08cd61da0fc1c4R9.

PR #2924, opened 2019-02-06
[m2e] fix duplicated launch shortcuts

rotty3000:


biz.aQute.junit/src/aQute/junit/JUnitEclipseReport.java
```diff
@@ -103,8 +103,8 @@ public void startTest(Test test) {
 		this.current = test;
 		message("%TESTS  ", test);
 
-		BundleContext context = targetBundle.getBundleContext();
 		try {
+			BundleContext context = targetBundle.getBundleContext();
```
bjhargrave: I have a more robust fix for this I will PR. So you can remove this commit from this PR.

rotty3000:
updated.

biz.aQute.junit/src/aQute/junit/JUnitEclipseReport.java
bjhargrave: See #2925

PR #2929, opened 2019-02-06
Clazz: A non-final field in a final class is not final.

### LINKED ISSUE #2928: final Service Component class with @Reference field injection mistakenly yields (wrong) error message
Given following code:
```
import org.osgi.service.component.annotations.Reference;
import org.osgi.service.log.LogService;

@Component
public final class MyServiceComponent {
    @Reference
    private LogService logService;
}
```
yields following error in Eclipse with bnd version 4.0.0.201805111645 / bndtools 4.2.0.DEV-20190129-224025-gc9c90cf:
`In component com.product.MyServiceComponent field logService is final and fieldOption is not 'update'.`
Without the `final` modifier of the class, the error is not present anymore.

The error is wrong, because the field to be injected is not `final`

Might be related to changes of https://github.com/bndtools/bnd/pull/2861/files

seanbright:
Final final final.

Fixes #2928

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/component/DSAnnotationTest.java
+++ b/biz.aQute.bndlib.tests/test/test/component/DSAnnotationTest.java
@@ -3573,6 +3573,24 @@ public class DSAnnotationTest extends BndTestCase {
 	}
 
 	@Component
+	public static final class FinalClassNonFinalField {
+		@Reference
+		private LogService logService;
+	}
+
+	// A field in a final class is not final:
+	// https://github.com/bndtools/bnd/issues/2928
+	public void testFinalFieldReference() throws Exception {
+		try (Builder b = new Builder()) {
+			b.setProperty(Constants.DSANNOTATIONS, "test.component.*FinalClassNonFinalField*");
+			b.setProperty("Private-Package", "test.component");
+			b.addClasspath(new File("bin_test"));
+			b.build();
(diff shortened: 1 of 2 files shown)


seanbright:
This is wrong

bjhargrave:
> This needs to be moved to `MethodRef`

Yes, the override needs to be moved to MethodDef. It is unfortunate that MethodDef extends FieldDef. But since they are public API, it is hard to change now.

bjhargrave:
We should also modify/create a DS test to make sure non-final fields in final classes is not treated as final.

biz.aQute.bndlib.tests/test/test/component/DSAnnotationTest.java
```diff
@@ -3569,6 +3569,23 @@ public void testFinalDynamicCollectionField() throws Exception {
...
+	@Component
+	public static final class FinalClassNonFinalField {
+		@Reference
+		private LogService logService;
+	}
+
+	// A field in a final class is not final:
+	// https://github.com/bndtools/bnd/issues/2928
+	public void testFinalFieldReference() throws Exception {
+		Builder b = new Builder();
```
bjhargrave: To clean up properly, please place the Builder in a try-with-resource block

```java
try (Builder b = new Builder()) {
   // test
}
```

PR #2931, opened 2019-02-06
[m2e] finally fix shortcuts

rotty3000:


bndtools.m2e/bnd.bnd
```diff
@@ -48,4 +48,6 @@ Import-Package: \
 
 -includeresource: \
 	lifecycle-mapping-metadata.xml,\
-	plugin.xml=_plugin.xml
+	plugin.xml=_plugin.xml,\
+	icons/bricks_junit.png=../bndtools.core/resources/unprocessed/icons/bricks_junit.png,\
```
bjhargrave: Instead of using `..`, we should use the `${workspace}` macro to reference the workspace folder.

rotty3000: ah yes of course!

rotty3000:
updated!

PR #2933, opened 2019-02-07
[m2e] create a post processor that can deference artifacts from workspace projects (instead of from local .m2)

rotty3000:


rotty3000:
If you have a running launch, changes form the workspace are applied without having to do a `mvn install`.

rotty3000:
@bjhargrave you ok with this logic?

PR #2938, opened 2019-02-08
[cdi annotations] properly handle BCP bean archives

rotty3000:


rotty3000:
@bjhargrave I had to tighten up `isInnerClass()` check because some classes were getting through. Check the test case I added if you want to see how.

biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java
```diff
@@ -467,7 +468,9 @@ String getSourceFile() {
 		boolean isInnerClass() {
 			String binary = type.getBinary();
 			return attributes(InnerClassesAttribute.class).flatMap(a -> Arrays.stream(a.classes))
-				.anyMatch(inner -> !Modifier.isStatic(inner.inner_access) && inner.inner_class.equals(binary));
+				.filter(inner -> Objects.nonNull(inner.inner_name))
```
bjhargrave: Instead of adding a filter operation, could we just AND the non-null check into the anyMatch test?

```java
.anyMatch(inner -> (inner.inner_name != null)
  && !Modifier.isStatic(inner.inner_access)
  && inner.inner_class.equals(binary));
```

rotty3000: Done!

PR #2939, opened 2019-02-08
fixes #2935 Directive "-runrequires" is missing for completion in source tab view

### LINKED ISSUE #2935: Directive "-runrequires" is missing for completion in source tab view
We can define "-runrequires" in Run tab of bnd and bndrun editors.
But in source tab view we see when entering "-run" some completions like "-runbundles" and some more. But "-runrequires" is missing in proposal list.

(just a minor issue for completeness and consistency between UIs)

rotty3000:


JochenHiller:
Wow, that was REALLY fast. Thx a lot for your help

PR #2948, opened 2019-02-10
[m2e] don't fail when used with older bnd-resolver-maven-plugin.

rotty3000:


bndtools.m2e/src/bndtools/m2e/MavenImplicitProjectRepository.java
```diff
@@ -141,12 +141,23 @@ private void createRepo(IMavenProjectFacade projectFacade, IProgressMonitor moni
 
                 containerBuilder.setBundles(bundles.getFiles(mavenProject.getBasedir()));
 
-                containerBuilder.setIncludeDependencyManagement(maven.getMojoParameterValue(mavenProject, mojoExecution, "includeDependencyManagement", Boolean.class, monitor));
+                try {
+                    containerBuilder.setIncludeDependencyManagement(maven.getMojoParameterValue(mavenProject, mojoExecution, "includeDependencyManagement", Boolean.class, monitor));
+                } catch (Exception e) {
```
bjhargrave: This is BAD. It catches all Exceptions so it can wallpaper over real programming errors. The code should either figure out the version of the bnd-resolver-maven-plugin in use and then call the method if safe; or it should use reflection to confirm the method exists before calling it directly. We should not use exception-based control flow.

rotty3000: a'right, I can't use reflection because I can't see the classes. I'll do versions.

rotty3000:
updated!

PR #2968, opened 2019-02-12
Configure gradle build scan plugin

aalmiray:
This small PR configures the `build-scan` Gradle plugin (https://guides.gradle.org/creating-build-scans/) which yields further insight into the build, allowing the team to make informed decisions on where and when the build may be tweaked to get faster, more performant builds.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,6 +7,13 @@
 
 import aQute.lib.io.IO
 
+apply plugin: 'com.gradle.build-scan'
+
+buildScan {
+    termsOfServiceUrl   = 'https://gradle.com/terms-of-service'
+    termsOfServiceAgree = 'yes'
+}
+
 if (JavaVersion.current().isJava9Compatible()) {
   ext.java9options = [
     '--add-opens=java.base/java.lang=ALL-UNNAMED', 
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -7,6 +7,13 @@
 
 import aQute.lib.io.IO
 
+apply plugin: 'com.gradle.build-scan'
+
+buildScan {
+    termsOfServiceUrl   = 'https://gradle.com/terms-of-service'
+    termsOfServiceAgree = 'yes'
```
bjhargrave: This also make a blanket assumption the everyone accepts the terms of service. A developer should really make this decision for them self.

bjhargrave:
I don't think we should accept this change.

build.gradle
aalmiray: I've added a check that solves this problem. Scans are enabled by default when run on CI, but the terms of service are checked if that's not the case. Developers have the choice to accept or refuse the terms.

aalmiray:
Understood. In which case the objection is to enable scans on CI. I can remove the automatic assignment when run on CI as well as the `--scan` flag found in the Travis settings.

PR #2971, opened 2019-02-13
[bndtools-runtime] Removed transitive dependencies from run/debug

pkriens:
The default in Eclipse seems to be transitive dependencies
for all projects reachable from the class path container.

This patch uses the bnd project buildpath and test path
to set up the list of dependencies. This should match the
Gradle setup for test/run

### DIFF
--- a/bndtools.core/src/bndtools/launch/BndContainerRuntimeClasspathEntryResolver.java
+++ b/bndtools.core/src/bndtools/launch/BndContainerRuntimeClasspathEntryResolver.java
@@ -1,0 +1,98 @@
+package bndtools.launch;
+
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.debug.core.ILaunchConfiguration;
+import org.eclipse.jdt.core.IClasspathEntry;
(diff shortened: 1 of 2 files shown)


bndtools.core/src/bndtools/launch/BndContainerRuntimeClasspathEntryResolver.java
```diff
@@ -0,0 +1,98 @@
...
+
+        try {
+            return Central.bndCall(() -> {
+                List<IRuntimeClasspathEntry> resolved = new ArrayList<>();
+                Set<File> classpath = new HashSet<>();
+
+                Project project = Central.getProject(entryProject.getProject());
+
+                List<Container> containers = new ArrayList<>(project.getBuildpath());
+                containers.addAll(project.getTestpath());
```
bjhargrave: There seems to be a number of issues here. 

First, this work is already been done by the BndContainerInitializer and held in the BndContainer object:

```
IClasspathEntry[] entries = JavaModelManager.getJavaModelManager()
            .containerGet(javaProject, BndtoolsConstants.BND_CLASSPATH_ID).getClasspathEntries();
```

So we should not recompute it here since this code and BndContainerInitializer will diverge over time.

Second, the test path entries should only be added when it is a test launch, no? This code adds them all the time. We can use the new "test" attribute (`JavaCore.newClasspathAttribute("test", Boolean.TRUE.toString());` to identify entries which are testpath entries.

pkriens: These are IRuntimeClasspathEntry, not IClasspathEntry objects

bjhargrave:
Replaced by #2973.

PR #2977, opened 2019-02-14
this is just a question

StinkyLord123:
this works like a charm on windows..
but it failed with this : (?:.*[/\\]|)[^/\\]*\.
pattern at my customer's linux machine... some character class error ,
it thinks we are skipping brackets or something..
can you make it work for linux too?

### DIFF
--- a/aQute.libg/src/aQute/libg/glob/AntGlob.java
+++ b/aQute.libg/src/aQute/libg/glob/AntGlob.java
@@ -14,7 +14,7 @@ public class AntGlob extends Glob {
 		super(globString, toPattern(globString, flags));
 	}
 
-	// match forward slash or back slash (windows)
+	// match forward slash or back slash (windows)? what about linux?
 	private static final String	SLASHY		= "[/\\\\]";
 	private static final String	NOT_SLASHY	= "[^/\\\\]";
 


bjhargrave:
It works on Linux and macOS. I use it on my macOS. The CI builds/test run on Linux. @rotty3000 uses it on his Linux workstation. 

Linux uses forward slash which is in the character class.

This is not an actual PR, so I am closing.

bjhargrave:
If you have an actual bug with some information on how to reproduce it, then please open an issue with the information. Thanks.

StinkyLord123:
hi , i found the issue and fixed it,
in AntGlob.java,
the Compile pattern of regex on some Mac/Linux machines is different, it
looks if there's a double slash "\\" before "]", and then it thinks ] is
being skipped, which will give u a character class error that the first
bracket "[" is not closed.
it doesnt happen on my machine not even on my linux ubunto.. yet it still
happened on 2 customers which have linux and mac.

private static final String SLASHY    = "[\\\\/]";
private static final String    NOT_SLASHY = "[^\\\\/]";

instead of

private static final String SLASHY    = "[/\\\\]";
private static final String    NOT_SLASHY = "[^/\\\\]";

bjhargrave:
That sounds like a buggy regex library in whatever version of java your customers are using. The Java source `"[/\\\\]"` is the string `[/\\]`. If the regex library does not understand `\\` as `\` and not escaping the `]`, then why would it not also mistake `\\/` to be `/` instead of `\/`. 

See https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html#cc where it is clear that, within a character class, escaping has precedence over grouping.

The strings used in the code are legal and valid Java regular expressions. Your customers should report this as a defect to their Java vendor and get a fix.

StinkyLord123:
i saw some code where this happens, according to the error line, i checked
the class + the error,
it gives a character class error in this case.. so i changed it.. i asked
customers to test it.. and it worked..
i understand you, but this was a good solution as of now.. and its the same
result.. so its a safe change.. i suggest u change it hehe

and btw thanx for the AntGlob class ... really helped me in project. :D

PR #2981, opened 2019-02-15
[Bndtools Core] Fixed issue with Private Packages View

### LINKED ISSUE #2937: Directive "-privatepackages" is not visible in Contents view
We define "-privatepackages: ..." in a bnd.bnd file. The packages will NOT be shown in Contents view of editor.
If we define "Private-Package: ..." in bnd.bnd file, it will be shown.

amitjoy:
Packages specified with -privatepackages weren't listed in the view. This fix solves the problem.

Fixes #2937

Signed-off-by: Amit Kumar Mondal <admin@amitinside.com>

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
@@ -1,6 +1,7 @@
 package aQute.bnd.build.model;
 
 import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toSet;
 
 import java.beans.PropertyChangeListener;
 import java.beans.PropertyChangeSupport;
@@ -19,7 +20,9 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
+import java.util.Set;
 import java.util.TreeMap;
+import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
 import org.osgi.resource.Requirement;
@@ -77,7 +80,8 @@ public class BndEditModel {
 		Constants.BUNDLE_COPYRIGHT, Constants.BUNDLE_UPDATELOCATION, Constants.BUNDLE_VENDOR,
 		Constants.BUNDLE_CONTACTADDRESS, Constants.BUNDLE_DOCURL, Constants.BUNDLE_SYMBOLICNAME,
 		Constants.BUNDLE_VERSION, Constants.BUNDLE_ACTIVATOR, Constants.EXPORT_PACKAGE, Constants.IMPORT_PACKAGE,
-		aQute.bnd.osgi.Constants.PRIVATE_PACKAGE, aQute.bnd.osgi.Constants.SOURCES,
+		Constants.PRIVATE_PACKAGE, Constants.PRIVATEPACKAGE,
+		aQute.bnd.osgi.Constants.SOURCES,
 		aQute.bnd.osgi.Constants.SERVICE_COMPONENT, aQute.bnd.osgi.Constants.CLASSPATH,
 		aQute.bnd.osgi.Constants.BUILDPATH, aQute.bnd.osgi.Constants.RUNBUNDLES, aQute.bnd.osgi.Constants.RUNPROPERTIES,
 		aQute.bnd.osgi.Constants.SUB, aQute.bnd.osgi.Constants.RUNFRAMEWORK, aQute.bnd.osgi.Constants.RUNFW,
@@ -303,6 +307,7 @@ public class BndEditModel {
 		converters.put(aQute.bnd.osgi.Constants.OUTPUT, stringConverter);
 		converters.put(aQute.bnd.osgi.Constants.SOURCES, includedSourcesConverter);
 		converters.put(aQute.bnd.osgi.Constants.PRIVATE_PACKAGE, listConverter);
+		converters.put(aQute.bnd.osgi.Constants.PRIVATEPACKAGE, listConverter);
 		converters.put(aQute.bnd.osgi.Constants.CLASSPATH, listConverter);
 		converters.put(Constants.EXPORT_PACKAGE, exportPackageConverter);
 		converters.put(aQute.bnd.osgi.Constants.SERVICE_COMPONENT, serviceComponentConverter);
@@ -323,6 +328,7 @@ public class BndEditModel {
 		// converters.put(BndConstants.RESOLVE_MODE, resolveModeConverter);
 		converters.put(Constants.BUNDLE_BLUEPRINT, headerClauseListConverter);
 		converters.put(Constants.INCLUDE_RESOURCE, listConverter);
+		converters.put(Constants.INCLUDERESOURCE, listConverter);
 		converters.put(Constants.STANDALONE, headerClauseListConverter);
 
 		formatters.put(aQute.bnd.osgi.Constants.BUNDLE_LICENSE, newlineEscapeFormatter);
@@ -344,6 +350,7 @@ public class BndEditModel {
 		formatters.put(aQute.bnd.osgi.Constants.OUTPUT, newlineEscapeFormatter);
 		formatters.put(aQute.bnd.osgi.Constants.SOURCES, defaultFalseBoolFormatter);
 		formatters.put(aQute.bnd.osgi.Constants.PRIVATE_PACKAGE, stringListFormatter);
+		formatters.put(aQute.bnd.osgi.Constants.PRIVATEPACKAGE, stringListFormatter);
 		formatters.put(aQute.bnd.osgi.Constants.CLASSPATH, stringListFormatter);
 		formatters.put(Constants.EXPORT_PACKAGE, headerClauseListFormatter);
 		formatters.put(aQute.bnd.osgi.Constants.SERVICE_COMPONENT, headerClauseListFormatter);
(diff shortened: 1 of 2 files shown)


biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
```diff
@@ -676,13 +676,23 @@ public void setIncludeSources(boolean includeSources) {
...
 
 	public void setPrivatePackages(List<? extends String> packages) {
 		List<String> oldPackages = getPrivatePackages();
-		doSetObject(aQute.bnd.osgi.Constants.PRIVATE_PACKAGE, oldPackages, packages, stringListFormatter);
+		String key = getPrivatePackageKey();
+		doSetObject(key, oldPackages, packages, stringListFormatter);
+	}
+
+	private String getPrivatePackageKey() {
+		return hasPrivatePackageInstruction() ? Constants.PRIVATEPACKAGE : Constants.PRIVATE_PACKAGE;
```
bjhargrave: It is permissible to use both `-privatepackage` and `Private-Package`. This change seems to ignore `Private-Package` if you have also used `-privatepackage`.

There are many other places where BndEditModel is sort of broken when we have dual instructions such as `-includeresource` and `Include-Resource` :-(

amitjoy: Yeah, I agree. I have modified the implementation a bit to return a merged list of entries specified in `<Private-Package, -privatepackage>` and `<Include-Resource, -includeresource>` groups. This will make the UI more consistent. The problem with #2937 will be solved since all the private packages (irrespective of where they are mentioned) will be listed on the rich editor view. However, while adding the packages from the rich editor, I add the new entries to the existing one, that is, either `Private-Package` (if present) or 
`-privatepackage`. Similarly for `Include-Resource` and `-includeresource`.

amitjoy:
This PR is ready for review

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
```diff
@@ -677,12 +685,90 @@ public void setIncludeSources(boolean includeSources) {
...
+	public void removeEntries(String key) {
+		List<String> oldPackages = getEntries(key, listConverter);
+		doRemoveObject(key, oldPackages, null, stringListFormatter);
 	}
 
-	public void setPrivatePackages(List<? extends String> packages) {
-		List<String> oldPackages = getPrivatePackages();
-		doSetObject(aQute.bnd.osgi.Constants.PRIVATE_PACKAGE, oldPackages, packages, stringListFormatter);
+	public void addPrivatePackage(String packageName) {
+		String key = hasPrivatePackageHeader() ? Constants.PRIVATE_PACKAGE : Constants.PRIVATEPACKAGE;
```
bjhargrave: In setPrivatePackages, you prefer the instruction. Here you prefer the header. I think we should prefer the newer instruction. If you use the instruction, we add new things to it.

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
```diff
@@ -1111,22 +1196,73 @@ public void addBundleBlueprint(String location) {
...
 
 	public void addIncludeResource(String resource) {
-		List<String> includeResource = getIncludeResource();
-		if (includeResource == null)
-			includeResource = new ArrayList<>();
-		else
-			includeResource = new ArrayList<>(includeResource);
-		includeResource.add(resource);
-		setIncludeResource(includeResource);
+		String key = hasIncludeResourceHeader() ? Constants.INCLUDE_RESOURCE : Constants.INCLUDERESOURCE;
```
bjhargrave: In setIncludeResource, you prefer the instruction. Here you prefer the header. I think we should prefer the newer instruction. If you use the instruction, we add new things to it.

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
```diff
@@ -1111,22 +1196,73 @@ public void addBundleBlueprint(String location) {
...
+			setEntries(resourceEntries1, Constants.INCLUDERESOURCE);
+		}
+		if (resourceEntries2.contains(resource)) {
+			resourceEntries2.remove(resource);
+			setEntries(resourceEntries2, Constants.INCLUDE_RESOURCE);
+		}
+	}
+
+	private boolean hasIncludeResourceHeader() {
+		return objectProperties.containsKey(Constants.INCLUDE_RESOURCE);
```
bjhargrave: Shouldn't this be `properties` like hasIncludeResourceInstruction?

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
amitjoy: Agreed and done

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
amitjoy: Agreed and done

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
amitjoy: No need for this method anymore. Removed.

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
```diff
@@ -677,12 +685,86 @@ public void setIncludeSources(boolean includeSources) {
...
+		} else {
+			privatePackagesEntries2.addAll(addedEntries);
+			setEntries(privatePackagesEntries2, Constants.PRIVATE_PACKAGE);
+		}
 	}
 
-	public void setPrivatePackages(List<? extends String> packages) {
-		List<String> oldPackages = getPrivatePackages();
-		doSetObject(aQute.bnd.osgi.Constants.PRIVATE_PACKAGE, oldPackages, packages, stringListFormatter);
+	public void setEntries(List<? extends String> packages, String key) {
```
bjhargrave: We should make all the new methods you have added `private` instead of `public`. Unless there is a real reason for the new methods to be API, lets not make them API.

pkriens: Baselining does tend you to make a lot more conservative regarding public! :-)

bjhargrave: It also preserves your future freedom to not make things public unless necessary.

amitjoy: Agreed and fixed 👍

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
```diff
@@ -677,12 +685,86 @@ public void setIncludeSources(boolean includeSources) {
...
+	}
+
+	public void addPrivatePackage(String packageName) {
+		String key = hasPrivatePackageInstruction() ? Constants.PRIVATEPACKAGE : Constants.PRIVATE_PACKAGE;
+		List<String> packages = getEntries(key, listConverter);
+		packages.add(packageName);
+		setEntries(packages, key);
+	}
+
+	public void removePrivatePackage(String packageName) {
```
bjhargrave: This new method should also be private.

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
```diff
@@ -1111,22 +1192,69 @@ public void addBundleBlueprint(String location) {
...
-			includeResource = new ArrayList<>(includeResource);
-		includeResource.add(resource);
-		setIncludeResource(includeResource);
+		String key = hasIncludeResourceInstruction() ? Constants.INCLUDERESOURCE : Constants.INCLUDE_RESOURCE;
+		List<String> entries = getEntries(key, listConverter);
+		entries.add(resource);
+		setEntries(entries, key);
+	}
+
+	public void removeIncludeResource(String resource) {
```
bjhargrave: This new method should also be private.

bjhargrave: Actually, why is this method here? It is not used anywhere in this PR.

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
bjhargrave: Actually, why is this method here? It is not used anywhere in this PR. Just adding it for test case use is not sufficient justification, really.

amitjoy: fixed

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
amitjoy: fixed

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
```diff
@@ -677,12 +685,72 @@ public void setIncludeSources(boolean includeSources) {
 	}
 
 	public List<String> getPrivatePackages() {
-		return doGetObject(aQute.bnd.osgi.Constants.PRIVATE_PACKAGE, listConverter);
+		List<String> privatePackagesEntries1 = getEntries(Constants.PRIVATEPACKAGE, listConverter);
+		List<String> privatePackagesEntries2 = getEntries(Constants.PRIVATE_PACKAGE, listConverter);
+
+		Set<String> privatePackages = Stream.concat(privatePackagesEntries1.stream(), privatePackagesEntries2.stream())
+			.collect(toSet());
```
bjhargrave: Since you make a stream, you don't need to collect into a set and then make a list from the set. Just use `distinct()` before the collect into list to remove duplicates.

amitjoy: You are right. Fixed 👍

bjhargrave:
Thanks!

Issue #2957, opened 2019-02-12
remote.junit is a misnomer


kriegfrj:
Having gone through the documentation for `biz.aQute.bnd.remote.junit`, I can see that it is a very useful set of libraries for programmatically creating and launching OSGi environments. This would be very useful for testing bundle implementations (and in hindsight, I could have used it for some of my own tests rather than at least partially re-inventing the wheel).

However, I feel that the name is rather unfortunate and misleading. As the documentation itself notes, the code doesn't have any dependencies on JUnit. Really, it is a kind of generic mocking/faking test harness for OSGi. Naturally, it has some features that make it useful for testing OSGi bundles (including possibly within a JUnit context), but really its scope isn't limited to unit testing (much less to JUnit specifically) - you could potentially use it to dynamically spawn standalone OSGi environments in a live application.

Given the above, I suggest that it would be good to rename this bundle and its classes to something more accurate. For example:
* Bundle `biz.aQute.bnd.remote.junit` -> `biz.aQute.bnd.remote.harness` or maybe `biz.Qute.bnd.remote.framework`
* Package `aQute.bnd.remote.junit` ->`aQute.bnd.remote.harness` or `aQute.bnd.remote.framework`
* Class `JUnitFramework` -> `OSGiFramework` or `OSGiHarness`
* Class `JUnitFrameworkBuilder` -> `OSGiFrameworkBuilder` or `OSGiHarnessBuilder`

Of course, depending on where this bundle is in the public release cycle, such a name change may not be feasible. I'll leave that judgement to other more experienced heads to consider.

pkriens:
Well, I am a bit like humpty dumpty :-)

This is quite new and still in progress. I am working closely with a customer to iron out the bugs. Although the harness  is super easy to use, the techniques to make this so easy are surprisingly complex.

Since you're not the only one making comments about the name I might consider renaming it. Though part of the reason is that I've not yet found a really good name. I like the JUnit aspect since it indicates java unit testing. After I've written documentation it should become more clear that you can also use it with other unit test frameworks. Since I expect it will become quite popular, I am not that worried that people will miss it once the documentation is in place.

But thanks for the feedback. Kind regards,

	Peter Kriens

kriegfrj:
> Well, I am a bit like humpty dumpty :-)

In that you're sitting on a "wall" dividing the code-under-test from the test code? And if your code falls off either side of the wall, it's hard to put it back together again? :smile:

[...]

PR #2982, opened 2019-02-15
Fix misspellings of "Launch"

kriegfrj:
Started trying to work with Peter Kriens' newly-renamed "Launchpad" (see #2957). Wondered why the compiler couldn't find LaunchpadBuilder even after I added it to the build path and imported it.

Turns out it was misspelled as "Lauchpad". The joys of automated refactoring means that this typo propagated through the whole repository while still compiling. This patch fixes the spelling.

I also had the thought to search the workspace for other similar misspellings of "launch", which is an easy enough mistake to make. Found one in a message in biz.aQute.bndlib/src/aQute/bnd/osgi/About.java which I have also fixed in this patch.


bjhargrave:
Due to time zone difference and that we should get this fix in ASAP, I manually accepted the DCO for this PR. Please fix your user.name for future PRs.

PR #2986, opened 2019-02-15
[maven] bnd-run-maven-plugin

rotty3000:


maven/bnd-run-maven-plugin/.classpath
```diff
@@ -0,0 +1,38 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bjhargrave: We don't commit the .classpath file for other maven projects since it is generate by the m2e plugins.

rotty3000: any suggestions on how I might test this?

rotty3000: oops, ok will remove, but question still stands for the other comment.

bjhargrave: I assume you would have a simple bndrun which starts a framework which then ends. We have some tests like that for the command line that could perhaps be copied and adapted? 

https://github.com/bndtools/bnd/blob/a3e9bdb338c8f7c406f7e98219d83f3f268b92df/biz.aQute.bnd/test/aQute/bnd/main/TestBndMain.java#L13-L31

rotty3000: So the thing is that the maven build is actually halted when the run is executing...

bjhargrave: > So the thing is that the maven build is actually halted when the run is executing...

I am not sure what that means. 

I assume the test can start a run, the runs will start and then exit with some return code which the test case can confirm, right?

rotty3000: No... the RUN runs... forever... until `^C` to kill it. That's kind of the point.

bjhargrave: > No... the RUN runs... forever... until `^C` to kill it. That's kind of the point.

That does not make any sense. You mean the run code cannot even return or System.exit?

If it is the case that any code run from this plugin can never end, then I don't see any point in this plugin. It is a trap and we should not have it.

bjhargrave: The code run by the tester plugin somehow ends. So I don't see why a run cannot end.

rotty3000: This plugin is intended to start a long lived process that remains running until killed by a sigkill. I'm not sure how to gauge the success of this type of operation. What should a test for this do? Start the process... and then what? We already know runs can run as this is very well tested by all other plugins. This one just remains open until terminated by the user.

rotty3000: I guess I could implement some code that runs inside said framework and responds via socket indicating it's alive and finally terminate the process. The maven invoker plugin won't be able to handle this case I don't think.

bjhargrave: This is exactly how the testing framework works. It processed the Junit test in the bundles and when that is done, the the testing code System.exits the VM to return control to  the invoker. 

For run, we just need a bndrun like this. The code in the bundle can just do something to verify it is happy that the framework was properly launched and then System.exit.

rotty3000:
updated including a test

rotty3000:
Hey @bjhargrave about getting this into 4.2, do you think it's a problem? Unless you already created the branch... This is very isolated and affects nothing else, even if there are bugs, what's the issue? I know it's late, but this will help the Intellij folks feel like they're being heard and possibly get a little more involved with the bnd story. Anyway, that's all I have to say about it. I'll let you decide.

PR #2988, opened 2019-02-16
CDIAnnotations fails to properly detect the discover mode… 

rotty3000:


rotty3000:
Yeah, I'm working on this.

rotty3000:
@bjhargrave updated.

biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
```diff
@@ -90,8 +89,7 @@ public boolean analyzeJar(Analyzer analyzer) throws Exception {
 				Discover discover = findDiscoveryMode(beansResource);
 
 				jar.getResources()
-					.values()
-					.forEach(r -> discoverPerResource.put(r, discover));
+					.forEach((k, r) -> resourceDiscovers.add(new Tuple(k, r, discover)));
```
bjhargrave: Since you only care about class resources (endsWith(".class")), perhaps you can filter non-class resources out here? To make the list shorter when processing below?

biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
```diff
@@ -90,8 +89,7 @@ public boolean analyzeJar(Analyzer analyzer) throws Exception {
 				Discover discover = findDiscoveryMode(beansResource);
 
 				jar.getResources()
```
bjhargrave: You need to take care to close certain Jar objects which may be created above by Jar.fromResource. You may want to close them here instead of delegating to the Processor with addClose. See

https://github.com/bndtools/bnd/blob/a3e9bdb338c8f7c406f7e98219d83f3f268b92df/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L2540-L2544

biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
```diff
@@ -304,4 +304,28 @@ private Document readXMLResource(Resource resource) {
...
+		final Discover	discover;
+
+		public Tuple(String resourcePath, Resource resource, Discover discover) {
+			this.resourcePath = resourcePath;
+			this.resource = resource;
+			this.discover = discover;
+		}
+
+		boolean sameOrigin(Clazz c) {
+			if (!resource.getClass()
```
bjhargrave: It seems kind of kludgy to check the resource objects to be of the same type. I don't see that it does much since you use Jar.fromResource just like Analyzer.analyzeBundleClassPath which calls analyzeJar which creates the Clazz objects and puts them in classspace. I guess I am not sure under what conditions they would not be the same type?

biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
```diff
@@ -304,4 +304,28 @@ private Document readXMLResource(Resource resource) {
...
+				.equals(c.getResource()
+					.getClass())) {
+				return false;
+			}
+			if (resourcePath.contains(c.getClassName()
```
bjhargrave: The contains test seems too weak. It would seem that `com/foo/bar/Outer$StaticInner` contains `com/foo/bar/Outer` but they are not the same class.

biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
bjhargrave: You have a hard problem given the current information held by Analyzer. Perhaps the answer here is a change to Analyzer so that it maintains information on the BCP entry from which a Clazz object comes. Analyzer.analyzeBundleClassPath could be updated to pass the BCP entry information to analyzeJar which could update some new data structure when it creates a new Clazz object. Then the code here could simply consult the data structure to determine which BCP entry owns a Clazz object (which is really what you need to know here.)

Your current attempts to reverse engineer this information are merely guess and not reliable.

rotty3000: .. yeah. honestly I was thinking the same :(

rotty3000:
redone!

rotty3000:
updated again based on discussion

PR #2992, opened 2019-02-18
fixes #2959 [maven] bnd-run-maven-plugin

### LINKED ISSUE #2959: [Feature Proposal] Create a maven bnd:run plugin
Hi,

it would be great if bnd provided a maven plugin that offers a goal similar to the jetty:run goal for Jetty, i.e., a goal that mimics the behavior of pressing the "Run/Debug OSGi" buttons in the "run" tab of a bndrun file in bndTools.

Why:
Right now I'm tied to Eclipse when I want really good OSGi development support, especially when I want the hot reloading of bundles into a running OSGi framework during development time. If there were a Maven plugin as mentioned above, I would be able to have a similar development experience as with bndTools, but on the commandline which would be IDE independent.  I know that some features (jar viewer, repository browsing, etc.) would still be missing, but I think the hot redeployment of bundles would tremendously easen the development of OSGi applications in non-Eclipse IDEs.

Kind regards,
Thomas

rotty3000:


rotty3000:
had closed the PR by accident.

PR #2994, opened 2019-02-18
[Bndtools Core] Fixed Problem with Instruction Highlighting

### LINKED ISSUE #2990: Some bnd instructions and headers ain't colorized
Not all bnd instrctions and headers ain't colored `blue` in Source View.

amitjoy:
Added missing syntaxes to Syntax class

Fixes #2990


pkriens:
Nice work!

PR #2997, opened 2019-02-19
[plugins] show how to use plugins from snapshot repo

rotty3000:


rotty3000:
alright @bjhargrave this is ready I think.

PR #2998, opened 2019-02-19
index - maven plugin: allow local indices without relativize

maggu2810:
The maven local-index goal allows to set a baseFile that is used to
relativize the found file locations URLs.
If the baseFile property is missing the directory of the output XML file
is used.
There is currently no option (using the Maven indexer plugin) to disable
the relativize step -- which is supported by the used SimpleIndexer
implementation.

This change allows to use a special baseFile property to disable the
relativize.

This way it is possible to create indices that are using e.g. abolute
paths to the local JAR files.

To create nicer file URLs the canonical representation is used
(`foo/../bar` can be replaced by `bar` without changing the location
itself).

Signed-off-by: Markus Rathgeb <maggu2810@gmail.com>

### DIFF
--- a/maven/bnd-indexer-maven-plugin/src/main/java/aQute/bnd/maven/indexer/plugin/LocalIndexerMojo.java
+++ b/maven/bnd-indexer-maven-plugin/src/main/java/aQute/bnd/maven/indexer/plugin/LocalIndexerMojo.java
@@ -18,6 +18,8 @@ import org.slf4j.LoggerFactory;
 import aQute.bnd.maven.lib.configuration.FileTree;
 import aQute.bnd.osgi.repository.SimpleIndexer;
 import aQute.lib.io.IO;
+import java.io.IOException;
+import java.util.stream.Collectors;
 
 /**
  * Exports project dependencies to OSGi R5 index format.
@@ -43,6 +45,9 @@ public class LocalIndexerMojo extends AbstractMojo {
 
 	@Parameter(property = "bnd.indexer.skip", defaultValue = "false")
 	private boolean				skip;
+
+	@Parameter(property = "bnd.indexer.absolute", defaultValue = "false")
+	private boolean				absolute;
 
 	/**
 	 * This configuration parameter is used to set the name of the repository in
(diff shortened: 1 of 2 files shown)


maggu2810:
I hope it is okay to not create an issue first, because this is no "fundamental" change but a very little one.

maven/bnd-indexer-maven-plugin/src/main/java/aQute/bnd/maven/indexer/plugin/LocalIndexerMojo.java
```diff
@@ -26,14 +26,23 @@
...
+	/**
+	 * This configuration parameter is used to relativize the path of the index
+	 * URIs stored in the index. By default these relative paths use the index
+	 * output location as their base directory, but a different base directory
+	 * can be supplied if needed.
+	 * The special string "///" can be used to disable the relativize process.
+	 */
 	@Parameter(property = "bnd.indexer.base.file")
-	private File				baseFile;
+	private String				baseFile;
```
rotty3000: I think it would be better to have a new parameter `boolean relativize` who's default is `true` to match the existing default. Then we don't need a special marker value.

maggu2810: Okay for me.
I was not sure, which is the preferred approach.

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/SimpleIndexer.java
```diff
@@ -202,7 +202,11 @@ private Resource indexFile(File file) {
 
 	private URI relativize(File file) {
 		if (base == null) {
-			return file.toURI();
```
bjhargrave: I am not sure this is a good idea. The caller supplies the File objects and can supply canonical/absolute/relative File objects if they choose.

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java
```diff
@@ -715,7 +715,7 @@ public void addContentCapability(URI uri, String sha256, long length, String mim
 
 	public boolean addFile(File file, URI uri) throws Exception {
 		if (uri == null)
-			uri = file.toURI();
+			uri = file.getCanonicalFile().toURI();
```
bjhargrave: I am not sure this is a good idea. The caller supplies the File objects and can supply canonical/absolute/relative File objects if they choose.

maven/bnd-indexer-maven-plugin/src/main/java/aQute/bnd/maven/indexer/plugin/LocalIndexerMojo.java
```diff
@@ -26,14 +26,23 @@
 public class LocalIndexerMojo extends AbstractMojo {
 	private static final Logger	logger		= LoggerFactory.getLogger(LocalIndexerMojo.class);
 
+	private static final String ABS_URL_IDENTIFIER	= "///";
```
bjhargrave: Instead of some magic file name, it would be better to specify a configuration setting to use absolute Files and then when set you can process `toIndex` to convert the Files to absolute/canonical before passing to `SimpleIndexer`

maggu2810:
I incorporated your review comments.

I have added a boolean parameter `absolute`.
This way only a small change of the Maven plugin has been necessary without touching the other files.

Should I rebase, squash and change the commit message?

maven/bnd-indexer-maven-plugin/src/main/java/aQute/bnd/maven/indexer/plugin/LocalIndexerMojo.java
maggu2810: Would be an option. Can you check my latest commit and comment if it should be changed.

bjhargrave:
Please also update the README for the plugin to document the new configuration parameter.

bjhargrave:
Thanks!

PR #3012, opened 2019-02-25
[launchpad] Launchpad imported too much

pkriens:
Included OSGiHeaders but this included too many
dependencies and these were not even properly 
conditionally included.

Created a simple Parameters class that assumes the
header is correct and does not try to analyze 
what might be wrong.

### DIFF
--- a/biz.aQute.launchpad/src/aQute/launchpad/Launchpad.java
+++ b/biz.aQute.launchpad/src/aQute/launchpad/Launchpad.java
@@ -50,8 +50,6 @@ import org.osgi.framework.wiring.BundleCapability;
 import org.osgi.framework.wiring.BundleWiring;
 import org.osgi.util.tracker.ServiceTracker;
 
-import aQute.bnd.header.Parameters;
-import aQute.bnd.osgi.Builder;
 import aQute.lib.converter.Converter;
 import aQute.lib.exceptions.Exceptions;
 import aQute.lib.inject.Injector;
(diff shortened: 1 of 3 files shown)


pkriens:
@bjhargrave any change you can cherry pick this one on 4.2?

bjhargrave:
> @bjhargrave any change you can cherry pick this one on 4.2?

Seems like this change is still not complete/finished? You have an empty enum class Mode. Why?

pkriens:
That is a left over, I can remove that.

bjhargrave:
Also, you add the `~` for duplicate keys but don't provide the related helper method to check if a key is duplicate and to remove the duplicate marker.

bjhargrave:
Since launchpad is still basically in beta, how important is it to make changes for 4.2?

pkriens:
Not crucial but extremely convenient. Both my customers start to use it in anger. Due to always having bndlib on the class path I'd not noticed the bndlib imports. Fr Krieg is also using it so it would be nice to have something at least working in 4.2. Since it is so far out of the main bnd code it looks pretty harmless.

bjhargrave:
ok, I can cherry-pick over if you can please hurry up and add timeout/retry support to HttpClient? :-) The build failures due to failed Eclipse downloads is insufferable.

pkriens:
LOL ok , that is a good excuse to not work on the documentation :-)

PR #3016, opened 2019-02-25
[httpclient] Added retries

pkriens:
Under threat of BJ going insane added a retry to
the http client.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
@@ -1,7 +1,6 @@
 package aQute.bnd.http;
 
 import static java.net.HttpURLConnection.HTTP_CREATED;
-import static java.net.HttpURLConnection.HTTP_GATEWAY_TIMEOUT;
 import static java.net.HttpURLConnection.HTTP_MOVED_PERM;
 import static java.net.HttpURLConnection.HTTP_MOVED_TEMP;
 import static java.net.HttpURLConnection.HTTP_NOT_MODIFIED;
@@ -36,6 +35,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.TimeZone;
 import java.util.concurrent.Callable;
+import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.zip.GZIPInputStream;
 import java.util.zip.InflaterInputStream;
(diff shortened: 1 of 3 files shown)


biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -561,8 +582,7 @@ private TaggedData doConnect(Object put, Type ref, final URLConnection con, fina
 			// origin server presented.
 			return new TaggedData(request.url.toURI(), 526, request.useCacheFile);
 		} catch (SocketTimeoutException ste) {
-			task.done(ste.toString(), null);
-			return new TaggedData(request.url.toURI(), HTTP_GATEWAY_TIMEOUT, request.useCacheFile);
+			throw new TimeoutException();
```
bjhargrave: It may be useful to make `ste` be the cause of the TimeoutException. Also, since you convert the TimeoutException to a string in the catch clause above, a message for the exception may be useful here.

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -300,42 +300,61 @@ TaggedData doCached0(final HttpRequest<?> request) throws Exception, IOException
...
-		} catch (Throwable t) {
-			task.done("Failed " + t, t);
-			throw t;
+				TaggedData td = connectWithProxy(proxy,
+					() -> doConnect(request.upload, request.download, con, hcon, request, task));
+				logger.debug("result {}", td);
+				return td;
+			} catch (TimeoutException toe) {
+				if (trial++ < retries) {
+					logger.debug("connection timed out. url={}, timout={}, trial={}", request.url, request.timeout,
```
bjhargrave: Can we make the log text indicate we are retrying? Otherwise it is the same text as when we are not retrying.

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
pkriens: There is nothing interesting the exception

pkriens:
LOL someone was eager ...

bjhargrave:
Yes! Next I would like to see if we can use retries for non-timeout failures like 5xx response codes. curl --retry will retry on 5xx response codes.

We may also want to have a expanding back off time before starting each retry. e.g. trial * 1 second.

pkriens:
well, that would now be trivial. Maybe throw a dedicated RetryException?

bjhargrave:
I think, in `doConnect`, after we check for the redirect response codes, we need to check for the 5xx response codes and, if we have retries left, we retry (after some increasing delay). I don't think we need an exception for this flow control, since the check and action are all local to the method.

PR #3024, opened 2019-02-27
[launchpad] Fix sanity checking for bundleSymbolicName()

### LINKED ISSUE #3019: [launchpad] bundleSymbolicName() cannot be invoked manually
I may have this wrong, but...

It seems to me the intended use of `BundleSpecBuilder.bundleSymbolicName()` is to be able to manually set the BSN of the synthetic bundle you are creating.

The default implementation of `BundleSpecBuilder.bundleSymbolicName()` doesn't permit you to set the BSN twice, and will throw an `IllegalArgumentException` if you attempt it. This kinda makes sense and guards against accidentally trying to set it twice.

However, the constructor in `BundleBuilder` calls `bundleSymbolicName()` as a matter of course with unique, synthetic name. This is useful in that it relieves the programmer of the burden of setting it in most cases where the BSN is not important. However, it also means that if you want to set it manually, by the time you actually get a chance to set it it's already been set, and you get the `IllegalArgumentException`.

It's possible that I've not understood how to use this properly, in which case I apologise. However, it seems to me that the synthetic BSN feature should be added at the latest possible opportunity, if the BSN has not been manually set (ie, during `BundleSpecBuilder.install()` rather than in the constructor).

kriegfrj:
Instead of throwing an IllegalArgumentException if you try to set the BSN twice, this PR changes it so that the second set will simply override the first. This fixes #3019.

Also added some sanity checking using aQute.bnd.osgi.Verifier to make sure that you can't accidentally set an invalid BSN.

### DIFF
--- a/aQute.libg/src/aQute/lib/io/packageinfo
+++ b/aQute.libg/src/aQute/lib/io/packageinfo
@@ -1,1 +1,1 @@
-version 3.2
+version 3.3
--- a/biz.aQute.launchpad/src/aQute/launchpad/BundleSpecBuilder.java
+++ b/biz.aQute.launchpad/src/aQute/launchpad/BundleSpecBuilder.java
@@ -5,6 +5,7 @@ import java.io.File;
 import java.net.URL;
 import java.nio.file.Files;
 import java.util.LinkedHashMap;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -21,7 +22,8 @@ import aQute.lib.io.IO;
  * 
  */
 public interface BundleSpecBuilder {
-	final static String CONFIGURATION_JSON = "configuration/configuration.json";
+	final static String		CONFIGURATION_JSON	= "configuration/configuration.json";
+	final static Pattern	SYMBOLICNAME		= Pattern.compile("[a-zA-Z0-9_-]+(\\.[a-zA-Z0-9_-]+)*");
 
 	BundleBuilder x();
 
@@ -30,9 +32,15 @@ public interface BundleSpecBuilder {
 	 */
 	default BundleSpecBsn bundleSymbolicName(String name) {
 		BundleBuilder x = x();
+		if (name == null) {
+			throw new IllegalArgumentException("bsn cannot be null");
+		}
+		if (!SYMBOLICNAME.matcher(name)
+			.matches()) {
+			throw new IllegalArgumentException("invalid bsn: " + name);
+		}
+		x.spec.bundleSymbolicName.clear();
 		x.spec.bundleSymbolicName.put(name, new LinkedHashMap<>());
-		if (x.spec.bundleSymbolicName.size() > 1)
-			throw new IllegalArgumentException("You can only have one Bundle-SymbolicName");
 
 		return new BundleSpecBsn() {
 
--- a/biz.aQute.launchpad/test/aQute/launchpad/BundleBuilderTest.java
+++ b/biz.aQute.launchpad/test/aQute/launchpad/BundleBuilderTest.java
@@ -1,0 +1,80 @@
+package aQute.launchpad;
+
+import org.assertj.core.api.JUnitSoftAssertions;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+import org.junit.Test;
+import org.osgi.framework.Bundle;
+
+import aQute.bnd.osgi.Verifier;
+import aQute.lib.io.IO;
+
+public class BundleBuilderTest {
+	LaunchpadBuilder	builder;
+	Launchpad			lp;
+	BundleBuilder 		bb;
+	
+	@Rule
+	public JUnitSoftAssertions	softly	= new JUnitSoftAssertions();
+
+
+	@Before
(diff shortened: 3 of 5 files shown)


biz.aQute.launchpad/src/aQute/launchpad/BundleSpecBuilder.java
```diff
@@ -12,6 +12,7 @@
 import org.osgi.framework.BundleActivator;
 import org.osgi.framework.Version;
 
+import aQute.bnd.osgi.Verifier;
```
bjhargrave: @pkriens May not want this tight linkage to bndlib. He recently removed a reference to Parameters to avoid such a linkage.

biz.aQute.launchpad/test/aQute/launchpad/BundleBuilderTest.java
```diff
@@ -0,0 +1,88 @@
...
+		} catch (Exception e) {
+			builder.close();
+		}
+		bb = lp.bundle();
+	}
+
+	@After
+	public void after() throws Exception {
+		try {
+			bb.close();
```
bjhargrave: You can also use `IO.close(bb)` as `IO.close` will not throw any exception.

biz.aQute.launchpad/src/aQute/launchpad/BundleSpecBuilder.java
kriegfrj: Thanks for the pointer.

I have changed it to use `Verifier.SYMOBLICNAME_STRING` intead of `Verifier.SYMOBLICNAME`. Because this dependency is a static final String it is resolved at compile time and doesn't create any runtime dependency on bndlib. Hopefully this is sufficient to address @pkriens ' concerns. Otherwise, I'll have to duplicate the string.

biz.aQute.launchpad/test/aQute/launchpad/BundleBuilderTest.java
kriegfrj: Such a simple suggestion which led me down a small rabbit warren...
* First problem: `BundleBuilder` does not implement `AutoCloseable`. In fact, its `close()` method was package private, and wasn't even called anywhere, ever. I've now made it public and made it implement `AutoCloseable` in this PR.
* Second problem: `IO.close()` only takes a `Closeable`, whereas `Launchpad` and `LaunchpadBuilder` (and now `BundleBuilder`) all implement `AutoCloseable` rather than `Closeable`. I tried to change the method signature, but Bnd complained that I had changed the API. Instead, I added IO.close(AutoCloseable) and changed IO.close(Closeable) to simply call the former. Had to bump the version of the package - wasn't sure whether to bump minor or micro; I bumped micro but that is easily fixed if it would be better to bump minor instead - please let me know (not that experienced with versioning, yet).

If necessary/appropriate, I can split both of these changes out into one or two separate PR.

biz.aQute.launchpad/src/aQute/launchpad/BundleSpecBuilder.java
pkriens: Please just copy the string, most of the work was make this work without bnd dependencies

pkriens:
I think we do not need extensive comments s in the code on these change. Only extremely anal people want to 'clean' these kind of things  up and they will find out why they should not touch a running program. A tiny comment like `// don't change`, should warn them. Same for the version changes they will get a warning when they try to remove it. Large comment sections imply they contain something really relevant for the reviewer and in this case there is nothing really there.

aQute.libg/src/aQute/lib/io/packageinfo
```diff
@@ -1 +1 @@
-version 3.2
+version 3.2.1
```
bjhargrave: This should be `3.3`. You are adding a new method. 

Because we have not released 4.2, we are still baselining 4.3 against 4.1. This will be changed to baseline against 4.2 once it is released and then not making it `3.3` will be an error.

bjhargrave:
As a best practice, this PR should really have 3 commits. One to change IO, one to use the new IO method, and a 3rd to do the fix for BSN. They are separate things where each builds upon the work of the previous. One reason to make them separate commits, is cherry-picking. Since this is now one big commit, I could not, for example, cherry-pick just the IO change onto another branch.

No need to change this PR now. I am just providing advice for future PRs. I do very much appreciate your contributions to Bnd. It is because of you that I am now an AssertJ fan :-)

bjhargrave:
Thanks!

kriegfrj:
> As a best practice, this PR should really have 3 commits.

Yes, my apologies. A combination of being too ~~busy~~lazy and mis-remembering the developer instructions - I thought it was recommended to squash each PR into a single commit (and I was too lazy for two more PRs!), but I just re-read the instructions and realise you wrote one commit per "logical unit[] of work", which is a lot easier than one PR per logical unit of work. I'll try and stick to that for the future.

>It is because of you that I am now an AssertJ fan :-)

Heh, I did notice that AssertJ managed to get embedded into `biz.aQute.junit`, and a recent version at that. Given that it's because of you and the rest of the team that I'm a Bnd/Bndtools fan, I think that's a fair trade. 😉

PR #3027, opened 2019-02-27
launchpad: Deep clone runspec in LaunchpadBuilder()

### LINKED ISSUE #3026: [launchpad] LaunchpadBuilder instance settings not isolated from each other
I was using Launchpad (which seems like a great tool) but was perplexed that my builder settings seemed to be "sticky" between tests and not isolated as I had hoped. This in spite of the fact that I was creating a fresh `LaunchpadBuilder` instance on every test execution. This made my test results dependent on test execution order, which makes it unmaintainable.

A bit of digging turned up the following:

1. `LaunchpadBuilder.set(String, String)` is putting the key/value pair into the static `projectTestSetup`'s properties instance instead of the instance variable `local`'s properties map.
2. The constructor clones `projectTestSetup` and assigns it to `local` at the start of every test. However, `RunSpecification`'s `clone()` implementation is simply a wrapper around `Object.clone()` to narrow the return type. This means that the returned value is a shallow clone rather than a deep clone. This means that every time you edit any of the container values in the builder (`runpath`, `properties`, `runfw`, `runbundles`, `extraSystemPackages`, `extraSystemCapabilities`, `errors`), you are updating the values in `projectTestSetup`.

I'm pretty sure that this is not the intended behaviour - the intended behaviour (it would seem to me) is for individual builder instances to be isolated from each other's changes, and that the constructor should deep clone `projectTestSetup`. However, I thought I'd give @pkriens the opportunity to comment before I submitted a patch in case there was something I hadn't understood.

kriegfrj:
Changes constructor to ensure that it deep clones projectTestSetup, to avoid making accidental changes to the master copy.

Defensive guarding of projectTestSetup to make it harder (though not impossible) for code to manipulate it accidentally.

Also ensures that LaunchpadBuilder.set() updates the instance runspec and not the static one.

Fixes #3026.

### DIFF
--- a/biz.aQute.launchpad/src/aQute/launchpad/LaunchpadBuilder.java
+++ b/biz.aQute.launchpad/src/aQute/launchpad/LaunchpadBuilder.java
@@ -7,6 +7,7 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.ServiceLoader;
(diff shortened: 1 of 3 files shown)


biz.aQute.launchpad/src/aQute/launchpad/LaunchpadBuilder.java
```diff
@@ -45,13 +46,18 @@
...
-		// makes it impossible to start a clean framework
+		// makes it impossible to start a clean framework. Make them
+		// immutable so that they can't be modified accidentally.
 		//
 
-		projectTestSetup.runbundles.clear();
-		projectTestSetup.runpath.clear();
-		projectTestSetup.properties.clear();
-		projectTestSetup.runfw.clear();
+		projectTestSetup.runbundles = Collections.EMPTY_LIST;
```
bjhargrave: Please use `Collections.emptyList()` and `Collections.emptyMap()` to avoid warnings in Eclipse.

kriegfrj: Odd, I didn't get a warning and I'm using Eclipse 2018-12.

bjhargrave:
Thanks!

PR #3034, opened 2019-02-28
bndcontainer: Always include project source for runtime classpath

bjhargrave:
Otherwise, when debugging you can end up viewing un-editable source from
a project's jar rather than the editable source of the project.


bjhargrave:
I want to cherry-pick this back for 4.2 as it is a large annoyance.

Issue #3036, opened 2019-03-01
Multiple NPEs when opening/using bndrun file in Maven Project


timothyjward:
Seen in 4.2.0.DEV-20190212-123646-gdeab1ca:

```
java.lang.NullPointerException
	at bndtools.m2e.MavenImplicitProjectRepository.versions(MavenImplicitProjectRepository.java:108)
	at bndtools.editor.project.OSGiFrameworkContentProvider.refreshProviders(OSGiFrameworkContentProvider.java:90)
	at bndtools.editor.project.OSGiFrameworkContentProvider.access$100(OSGiFrameworkContentProvider.java:36)
	at bndtools.editor.project.OSGiFrameworkContentProvider$2.run(OSGiFrameworkContentProvider.java:136)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:60)
```

and 

```
java.lang.NullPointerException
	at bndtools.m2e.MavenImplicitProjectRepository.list(MavenImplicitProjectRepository.java:80)
	at bndtools.model.repo.RepositoryTreeContentProvider$1.run(RepositoryTreeContentProvider.java:294)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:60)
```

Bndrun file:

```
-standalone: target/index.xml

-resolve.effective: active

-runfw: org.apache.felix.framework
-runee: JavaSE-1.8

-runvm: -ea,\
 -Dorg.slf4j.simpleLogger.defaultLogLevel=debug,\
 -Dlogback.configurationFile=${.}/src/test/resources/logback/logback-test.xml

-runproperties: \
    posh.home=${.}/src/test/resources/,\
    test.resources=${.}/src/test/resources/

-runsystempackages: sun.reflect

-runrequires: bnd.identity;id='com.paremus.ui.rest.net.test'
```

I will now try updating to the latest RC, but I wanted to capture the stack traces while I had them.

timothyjward:
> I will now try updating to the latest RC, but I wanted to capture the stack traces while I had them.

Still seen in 4.2.0 RC3

timothyjward:
Another similar NPE occurs when attempting to launch:

```
java.lang.NullPointerException
	at bndtools.m2e.MavenImplicitProjectRepository.versions(MavenImplicitProjectRepository.java:108)
	at aQute.bnd.build.Project.getBundle(Project.java:1268)
	at aQute.bnd.build.Project.getBundles(Project.java:648)
[...]

PR #3038, opened 2019-03-01
attempt to handle #3036

rotty3000:


bndtools.m2e/src/bndtools/m2e/MavenImplicitProjectRepository.java
```diff
@@ -172,6 +173,20 @@ private void createRepo(IMavenProjectFacade projectFacade, IProgressMonitor moni
...
+
+            try {
+                String name = mavenProject.getName()
+                    .isEmpty() ? mavenProject.getArtifactId() : mavenProject.getName();
+
+                fileSetRepository = new FileSetRepository(name, Collections.emptyList());
+
+                Central.refreshPlugin(this);
+            } catch (Exception e2) {
+                // we're really in trouble if we get here...
```
bjhargrave: Yes, but we still just eat the exception which means some NPE later. Why not just throw the (ducked) exception here? Then at least we have a more useful stack trace?

bndtools.m2e/src/bndtools/m2e/MavenImplicitProjectRepository.java
```diff
@@ -172,6 +173,20 @@ private void createRepo(IMavenProjectFacade projectFacade, IProgressMonitor moni
             if (logger.isErrorEnabled()) {
                 logger.error("Failed to create implicit repository for m2e project {}", getName(), e);
             }
```
bjhargrave: I think you should do Eclipse logging of the exception here so it shows up in the Eclipse log. I have no idea where slf4j logging goes. But people will see and can report the exception if it goes into the Eclipse log.

https://github.com/bndtools/bnd/blob/e7f814b8d335a6d1efc5241b7433852742ff956c/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java#L60

rotty3000:
updated!

PR #3041, opened 2019-03-01
Add -maven-scope instruction to set default scope for deps

bjhargrave:
This will allow a project to change the default scope of "compile" for
dependencies in the generated POM. For example: `-maven-scope=provided`.


pkriens:
How fast you are ...

PR #3043, opened 2019-03-02
Use -maven-scope instruction to simplify some -buildpaths

bjhargrave:
Some projects have many -buildpath entries with maven-scope=provided
attributes. For those projects, we set the instruction
-maven-scope=provided so the default scope for dependencies in generated
POMs is provided rather than the default of compile. This means that
some -buildpath entries will then need to specify maven-scope=compile
to make sure they are not provided.


pkriens:
Thanks!

PR #3050, opened 2019-03-06
Launchpad & runtime

pkriens:


pkriens:
@bjhargrave I switched the version of scr and log. Can you see anything wrong? The new log cratered a lot of code because for some (imho misguided) reason LogService extends LogFactory. (I think it should have been deprecated, what is the benefit of extending it for old code?) SCR was tricky with all the version constraints.

bjhargrave:
> I switched the version of scr and log. Can you see anything wrong? The new log cratered a lot of code because for some (imho misguided) reason LogService extends LogFactory. (I think it should have been deprecated, what is the benefit of extending it for old code?) SCR was tricky with all the version constraints.

I am not sure why you went backward in felix log and scr? I can imagine that may break other parts of the repo which care about the later versions.

The LogService 1.4 changes were discussed at length at OSGi. I am sorry you don't like them.

You may want to try the most recent version of Felix SCR which has addressed some odd package exports/imports.

pkriens:
I tried to find a matching configuration and this was the best I got :-( Problem is that the new SCR requires framework R7. Probably should talk when you got a moment.

PR #3063, opened 2019-03-13
baseline: Use MavenBndRepository support for macros

bjhargrave:
We centralize the baseline version to a Workspace property.


pkriens:
I guess this requires everybody to update their bndtools?

bjhargrave:
> I guess this requires everybody to update their bndtools?

Yes if they are building Bnd. Like many other times in the past when we use new features added to Bnd. Eating our own food...

pkriens:
Might be nice to get a warning on Slack or mailing list this is happening. It can take some time before you figure this out.

bjhargrave:
Sure, I haven't merged this PR yet. But, in the past when such changes have happened, I have mentioned it on slack.

PR #3077, opened 2019-03-19
p2 repo: Remove unused class

bjhargrave:
It looks like P2Indexer serves targets.


pkriens:
You're sure? I thought there were subtle differences?

bjhargrave:
It is a package private class and no one in the repo referenced the class. I think you addressed this via P2Impl and TargetImpl. TargetIndexer looked like some left over from a code refactoring.

PR #3080, opened 2019-03-20
fixes #3045 bnd launcher doesn't support framework extensions on Java 9+

### LINKED ISSUE #3045: bnd launcher doesn't support framework extensions on Java 9+
```
[1f-a299a1de9ee9] ERROR Events.Framework.org.eclipse.osgi:? - FrameworkEvent ERROR
org.osgi.framework.BundleException: Cannot support framework extension bundles without a public addURL(URL) method on the framework class loader: org.apache.aries.spifly.dynamic.framework.extension_1.2.0 [6]
	at org.eclipse.osgi.storage.FrameworkExtensionInstaller.addExtensionContent0(FrameworkExtensionInstaller.java:119)
	at org.eclipse.osgi.storage.FrameworkExtensionInstaller.addExtensionContent(FrameworkExtensionInstaller.java:94)
	at org.eclipse.osgi.internal.loader.SystemBundleLoader$SystemModuleClassLoader.loadFragments(SystemBundleLoader.java:134)
	at org.eclipse.osgi.internal.loader.SystemBundleLoader.loadClassLoaderFragments(SystemBundleLoader.java:116)
	at org.eclipse.osgi.internal.loader.BundleLoader.loadFragments(BundleLoader.java:309)
	at org.eclipse.osgi.container.ModuleWiring.loadFragments(ModuleWiring.java:276)
	at org.eclipse.osgi.container.ModuleContainer.applyDelta(ModuleContainer.java:835)
	at org.eclipse.osgi.container.ModuleContainer.resolveAndApply(ModuleContainer.java:551)
	at org.eclipse.osgi.container.ModuleContainer.resolve(ModuleContainer.java:494)
	at org.eclipse.osgi.container.ModuleContainer.refresh(ModuleContainer.java:1129)
	at org.eclipse.osgi.container.ModuleContainer$ContainerWiring.dispatchEvent(ModuleContainer.java:1510)
	at org.eclipse.osgi.container.ModuleContainer$ContainerWiring.dispatchEvent(ModuleContainer.java:1)
	at org.eclipse.osgi.framework.eventmgr.EventManager.dispatchEvent(EventManager.java:233)
	at org.eclipse.osgi.framework.eventmgr.EventManager$EventThread.run(EventManager.java:343)
```

rotty3000:


bjhargrave:
Looks like you have a test failure on windows. Good luck debugging!

rotty3000:
nice....!

aQute.libg/src/aQute/libg/classloaders/ModifiableURLClassLoader.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.util.Enumeration;
+
+public class ModifiableURLClassLoader extends URLClassLoader {
+
+	private final ClassLoader lastLoader;
```
bjhargrave: This is a big behavior change from before. The current code using URLClassLoader uses the argument class loader as the parent (parent-first). You have changed it here to last. Not sure that is wise? And it requires odd behavior when used below.

aQute.libg/src/aQute/libg/runtime/JavaRuntime.java
```diff
@@ -0,0 +1,42 @@
...
+import java.io.File;
+import java.lang.management.ManagementFactory;
+import java.lang.management.RuntimeMXBean;
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.stream.Stream;
+
+public class JavaRuntime {
```
bjhargrave: This class now requires java.lang.management for the launcher. This may or may not be an issue for some users. We have tried to make sure the launcher side runs on older systems. This is why, for example, we limit OSGi to 4.2 and not later. I suspect that some older systems may have issues with the JMX usages in the launcher. Perhaps @pkriens can comment on his customer's use.

biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
```diff
@@ -55,7 +55,9 @@
 public class ProjectLauncherImpl extends ProjectLauncher {
 	private final static Logger		logger					= LoggerFactory.getLogger(ProjectLauncherImpl.class);
 	private static final String		EMBEDDED_LAUNCHER_FQN	= "aQute.launcher.pre.EmbeddedLauncher";
-	private static final String		EMBEDDED_LAUNCHER		= "aQute/launcher/pre/EmbeddedLauncher.class";
+	private static final String[]	EMBEDDED_CLASSES		= new String[] {
+		"aQute/launcher/pre/EmbeddedLauncher.class", "aQute/libg/classloaders/ModifiableURLClassLoader.class"
```
bjhargrave: This pulls in the the complete `aQute/launcher/pre` package which is part of the launcher project. It now pulls in part of the `aQute/libg/classloaders` package from another project. Bit worried about the partial package thing...

biz.aQute.launcher/src/aQute/launcher/pre/EmbeddedLauncher.java
```diff
@@ -49,18 +47,19 @@ public static void main(String... args) throws Exception {
...
+					for (String path : runpath.split("\\s*,\\s*")) {
+						URL url = toFileURL(cl.getResource(path));
+						ml.addURL(url);
+					}
 
-				for (String path : runpath.split("\\s*,\\s*")) {
-					URL url = toFileURL(cl.getResource(path));
-					classpath.add(url);
-				}
+					Class<?> embeddedLauncher = ml.loadClass("aQute.launcher.pre.EmbeddedLauncher");
```
bjhargrave: You are loading this class again from the new class loader and setting MANIFEST? That seems only necessary because of my earlier concern about `ModifiableURLClassLoader` not delegating to parent first, so `ModifiableURLClassLoader` loads a new EmbeddedLauncher when Launcher want to access it's MANIFEST field. So why not parent first?

rotty3000:
@bjhargrave let me start by explaining and, then with your help, I can formulate what that comment you requested might be.

Pre Java 9, the AppClassLoader inherited from URLClassLoader which meant it was by nature modifiable (even if you had to use reflection). This enabled frameworks to add extension bundles to it's classloader.

Since Java 9, the AppClassLoader is no longer modifiable, which means that starting a framework directly from it means framework extensions cannot be added to its classloader.

aQute.libg/src/aQute/libg/classloaders/ModifiableURLClassLoader.java
rotty3000: the point is that the parent classloader MUST NOT load the launcher class. Therefore making it parent first would break the fix entirely.

biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
rotty3000: Not sure how else to solve this.

aQute.libg/src/aQute/libg/classloaders/ModifiableURLClassLoader.java
bjhargrave: Then it would probably be better to configure the class loader which a set of class names which must not be delegated to the parent class loader RATHER than parent-last behavior which causes the other weird issues below.

biz.aQute.launcher/src/aQute/launcher/pre/EmbeddedLauncher.java
rotty3000: This is the only solution. The Launcher class MUST NOT be on the parent classloader because it's not modifiable.

biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
bjhargrave: Well the new Modifiable class loader could be part of the pre package in the launcher since it is a special-for-bnd-launcher class loader, no?

aQute.libg/src/aQute/libg/classloaders/ModifiableURLClassLoader.java
rotty3000: perhaps.. I can try that.

biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
rotty3000: It's also used in bndlib for launching in other similar scenarios.

bjhargrave: But you made that change in this PR and I would submit it is wrong to change the expected parent-first semantics there.

rotty3000:
@bjhargrave trying a different approach here... this is just now a test asserting the behaviour. It should in theory only fail in Java >8.

biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
```diff
@@ -326,10 +325,12 @@ public Jar executable() throws Exception {
...
 		m.getMainAttributes()
-			.putValue(EmbeddedLauncher.EMBEDDED_RUNPATH, Processor.join(classpath));
-		Resource embeddedLauncher = Resource.fromURL(this.getClass()
-			.getResource("/" + EMBEDDED_LAUNCHER));
-		jar.putResource(EMBEDDED_LAUNCHER, embeddedLauncher);
+			.putValue(EMBEDDED_RUNPATH, Processor.join(classpath));
+
+		Resource preJar = Resource.fromURL(this.getClass()
+			.getResource("/pre.jar"));
+		jar.addAll(new Jar("pre", preJar.openInputStream(), preJar.lastModified()));
```
bjhargrave: You need to close the newly created Jar object after the `addAll` call.

Issue #3081, opened 2019-03-21
[Launchpad] Add a way to prevent packages from being exported from the framework bundle


brampouwelse:
Launchpad is exporting all exported packages from the framework, it would be nice if there is a way to filter these exports. 

@pkriens has some idea on this: https://groups.google.com/forum/?fromgroups=#!topic/bndtools-users/ER9KKcmTDXo

kriegfrj:
+1 I was thinking the same thing just recently.

One possible implementation would be if you could specify a list of package prefixes to exclude on the LaunchpadBuilder. The Similar to Mockito's internal `ExcludingClassLoader` (https://github.com/mockito/mockito/blob/release/2.x/src/test/java/org/mockitoutil/ClassLoaders.java). Using an excluding classloader to bootstrap the framework should do part of the trick.

brampouwelse:
I don't think special class loaders are required, it's just the framework packages extra if Launchpad has a way to filter that before is used that would do. 

For now I've managed to find a place to where I can change the list just before it's used. It's a hack but it makes my logging work :D 

```java
package aQute.launchpad;

import java.util.regex.Pattern;
import java.util.stream.Collectors;

import org.osgi.framework.Constants;
import org.osgi.framework.launch.Framework;

/**
 * {@link LaunchpadBuilder} that removes the slf4j packages from the FRAMEWORK_SYSTEMPACKAGES_EXTRA property value
 *
 * Workaround for: https://github.com/bndtools/bnd/issues/3081
 */
public class LaunchpadBuilderExt extends LaunchpadBuilder {

    @Override
    Framework getFramework() {

        String systemPackagesExtra = local.properties.get(Constants.FRAMEWORK_SYSTEMPACKAGES_EXTRA);
        String filteredSystemPackagesExtra = Pattern.compile(",").splitAsStream(systemPackagesExtra)
                .map(s -> s.trim().startsWith("org.slf4j") ? s.endsWith("\"") ? "\"" : "" : s)
                .filter( s-> !s.equals(""))
                .collect(Collectors.joining(","));

        local.properties.put(Constants.FRAMEWORK_SYSTEMPACKAGES_EXTRA, filteredSystemPackagesExtra);

        return super.getFramework();
    }

}
```

kriegfrj:
You're right, no extra classloader needed - just need to add additional options for configuring the classloader that is already set up.

[...]

PR #3084, opened 2019-03-21
[Launchpad] Add ability to exclude system packages

kriegfrj:
Proposed implementation of #3081.

Adds ability to filter the package list that is exported into the `SYSTEMPACKAGES_EXTRA` property, before it is actually used to generate the framework.

Can filter either by `Predicate<String>`(s), or by a plain string(s). If plain strings are used, ending with a ".*" indicates that it will exclude all subpackages also, whereas without this on the end an exact match is required. Hopefully this is a pretty good combination of flexibility and simplicity to cover most situations.

Feedback welcome.

### DIFF
--- a/biz.aQute.launchpad.tests/test/aQute/launchpad/LaunchpadBuilderTest.java
+++ b/biz.aQute.launchpad.tests/test/aQute/launchpad/LaunchpadBuilderTest.java
@@ -2,11 +2,15 @@ package aQute.launchpad;
 
 import static aQute.launchpad.LaunchpadBuilder.projectTestSetup;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import org.assertj.core.api.JUnitSoftAssertions;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
+import org.osgi.framework.Constants;
 
 import aQute.bnd.service.specifications.RunSpecification;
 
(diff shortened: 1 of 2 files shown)


pkriens:
You beat me to it ... A few comments.

- I'd like to use Glob from aQute.lib, it is easier on the user instead of patterns.
- I do not like this to be done in the toString, it goes against cohesion. Printing and filtering should not be the same. I created a method to filter the Map giving a lost of Glob
- We should add a header in the bndrun files that sets this also so all settings can be in the bndrun file

kriegfrj:
Sounds like you already have an implementation ready - I'm happy for it to supersede mine if you think it's better.

Re: the bndrun header: is there a use case for it other than for Launchpad? If not, IMHO, I don't think it's worth the extra complexity of adding a new header. But I'm happy to be overruled.

kriegfrj:
Also, just to explain the toString thing - I did it this way for efficiency reasons, as it avoids the creation of a temporary intermediate object and the need to iterate over the map twice. If you'd like to sacrifice efficiency for readability, that's fine.

brampouwelse:
You guys are quick! 

Even if it's just for launchpad I'd like to be able to configure it in the bndrun as that would allow me to "fix it" on in the (Amdatu Blueprint) workspace template we use. 

Does it make sense to use the existing `-runsystempackages` instruction and allow package exclusions there similar to how one can override package imports for an `Import-Package` header?

kriegfrj:
> You guys are quick!
> 

Heh, it only seems like it. I had already encountered a similar situation to yours and had started work on it slightly before you raised the issue.

> Even if it's just for launchpad I'd like to be able to configure it in the bndrun as that would allow me to "fix it" on in the (Amdatu Blueprint) workspace template we use.
> 
> Does it make sense to use the existing `-runsystempackages` instruction and allow package exclusions there similar to how one can override package imports for an `Import-Package` header?

This I think makes sense. In that case, the syntax would be something like:

`runsystempackages("!org.slf4j.*")`

Is that the sort of thing you had in mind by using globs, @pkriens ?

pkriens:
Fixed by #3095

PR #3092, opened 2019-03-25
Handle symlinks to non-existent target folders on Windows

bjhargrave:
A Windows file system oddity means that when a symlink is created, the
file system has to know if the target is a directory. So when we make
a directory via a symlink, we must recreate the symlink so that Windows
knows that the symlink points to a directory and not a file.

See dwFlags parameter:
https://docs.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-createsymboliclinkw#parameters

### DIFF
--- a/aQute.libg/src/aQute/lib/io/IO.java
+++ b/aQute.libg/src/aQute/lib/io/IO.java
@@ -503,11 +503,11 @@ public class IO {
 		return copy(in, new ByteBufferOutputStream()).toByteArray();
 	}
 
-	public static void write(byte[] data, OutputStream out) throws Exception {
+	public static void write(byte[] data, OutputStream out) throws IOException {
 		copy(data, out);
 	}
 
-	public static void write(byte[] data, File file) throws Exception {
+	public static void write(byte[] data, File file) throws IOException {
 		copy(data, file);
 	}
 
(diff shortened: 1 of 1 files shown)


aQute.libg/src/aQute/lib/io/IO.java
```diff
@@ -862,7 +863,14 @@ public static File mkdirs(File dir) throws IOException {
 
 	public static Path mkdirs(Path dir) throws IOException {
 		if (Files.isSymbolicLink(dir)) {
-			return mkdirs(Files.readSymbolicLink(dir));
+			Path target = Files.readSymbolicLink(dir);
+			boolean recreateSymlink = isWindows() && !Files.exists(target, LinkOption.NOFOLLOW_LINKS);
```
seanbright: This could use a comment I think

bjhargrave: I think it is handled in detail in the commit message, as well as a comment below. Using git blame, you can easily find the commit which modified this line and its commit message. I find that is often more powerful and useful the source code comments. YMMV. :-)

seanbright: Cool. Thanks for nothing 👍

PR #3108, opened 2019-03-28
/docs update. typo. A fix for wrapping a non-osgi jar into a bundle using bnd.

Med116:
I was trying this out and noticed that this text had a typo. I replaced "bnd" with "jar"

### DIFF
--- a/docs/_chapters/125-tour-features.md
+++ b/docs/_chapters/125-tour-features.md
@@ -66,7 +66,7 @@ Since bnd files are property files, you cannot repeat a property. Later properti
 
 We can now wrap the source JAR with the following command:
 
-	$ bnd javax.activation.bnd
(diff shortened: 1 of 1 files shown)


docs/_chapters/125-tour-features.md
```diff
@@ -66,7 +66,7 @@ Since bnd files are property files, you cannot repeat a property. Later properti
 
 We can now wrap the source JAR with the following command:
 
-	$ bnd javax.activation.bnd
+	$ bnd javax.activation.jar
```
bjhargrave: This change is incorrect. The line was correct before your change. The bnd file was created at the end of `Simplistic` above.

bjhargrave:
Closing since the suggested change makes the currently correct text wrong.

Issue #2643, opened 2018-09-20
Add support of maximum version for SCR and Metatype Annotations


amitjoy:
`-dsannotations-options` and `-metatypeannotations-options` instructions do support the minimum version but if the developer mistakenly uses any functionality which is introduced in higher versions, the DS or Metatype XML gets generated without any errors or warning but it gets really difficult to find out the problem in runtime since the XML namespace uses higher versions. To make it simple and usable, a maximum version would really help users to encounter such problems beforehand.

amitjoy:
Found the discussion and related PR: https://github.com/bndtools/bnd/pull/2603

bjhargrave:
You should control the maximum version of DS xml by the version of the component annotations on your build path. If you use DS 1.3 component annotations, then you wont use DS 1.4 annotation features.

kwin:
Just for the record: A maximum can now be given: https://github.com/bndtools/bnd/commit/1482d0734c388e1ed9217bd97b6a5538450af2c5

pkriens:
Fixed in #3121

PR #3113, opened 2019-03-29
[dsannotations] Check ds annotations against max version

pkriens:
- Provide reason for selected version


bjhargrave:
We rejected doing this in #2603 and #2643. What is different now that we should change our minds?

biz.aQute.bndlib/src/aQute/bnd/component/ComponentDef.java
```diff
@@ -254,8 +254,9 @@ Tag getTag() {
 		return "$".equals(v) ? name : v;
 	}
 
-	void updateVersion(Version version) {
+	void updateVersion(Version version, String reason) {
 		this.version = max(this.version, version);
+		this.versionReason = reason;
```
bjhargrave: This is wrong since it always changes the reason string even when the supplied version is less than the old version!  The reason should only be changed if the supplied version is greater than the old version.

biz.aQute.bndlib/src/aQute/bnd/component/DSAnnotationReader.java
```diff
@@ -514,7 +515,7 @@ private void processMethodActivationArgs(String propertyDefKeyFormat, String mem
 				String propertyDefKey = String.format(propertyDefKeyFormat, arg);
 				processActivationObject(propertyDefKey, param, memberDescriptor, details, deactivate);
 			} else if (deactivate && (type == BaseType.I)) {
-				component.updateVersion(V1_1);
+				component.updateVersion(V1_1, "processMethodActivationArgs deactivate && type==BaseType.I");
```
bjhargrave: An more meaningful reason here is that the deactivate method takes an `int reason` argument.

pkriens:
I guess you threw me off on the #2643 bug. In that bug you seem to imply we get the maximum version of the org.osgi.service.component.annotations, which I liked a lot. However, I realized this week that this is not implemented yet as you imply that you lacked time at the time of the discussion.

We currently have a problem that currently Gradle builds sometimes seems to randomly use 1.3 while the component is absolutely 1.2 and is as such generated in Eclipse. This is only detected late when the generated requirement is resolved. So the primary driver is to find out why the gradle build does this, this is why the reason for the version increase is added in `updateVersion()`.

Reading these bugs I do not think we rejected the idea of a maximum version, we decided that the best solution is to use the annotation version set for the maximum version.

I really do like the option and we have the place but I could make the default for the minimum and maximum version the package of the annotation.

biz.aQute.bndlib/src/aQute/bnd/component/DSAnnotationReader.java
pkriens: Could you help with the reasons? For me many of the reasons are rather hard to guess from the code.

pkriens: In a way I would be ok with a number that we then describe in more detail somewhere

bjhargrave:
> We currently have a problem that currently Gradle builds sometimes seems to randomly use 1.3 while the component is absolutely 1.2 and is as such generated in Eclipse.

This I would like to understand better. You are saying the project is compiling against component.annotations;version-1.2 and in Eclipse the generated XML is <= 1.2 but in gradle the generated XML is sometimes > 1.2? That seems like a huge heisenbug. Can we reproduce this?

pkriens:
Well, this PR is an attempt to nail it down when it happens ...
However, as far as I can see we ignore the package version today? Do I miss something?

pkriens:
If I could easily reproduce it the PR would look quite different :-(

bjhargrave:
If we do this max version, then I would submit the max version must be set from the version of component.annotations on the `-buildpath` not from the `-dsannotation-options`. It seems wrong I could put component.annotations;version=1.4 on -buildpath and -dsannotation-options=max;version=1.2 in bnd file.

biz.aQute.bndlib/src/aQute/bnd/component/DSAnnotationReader.java
bjhargrave: I think that level of indirection will be a source of errors for us to maintain. I can help with the reason strings.

pkriens:
I think the defaults for minimum and maximum should come from the buildpath but being able to override it is often a godsend. (With a warning if they seem to mismatch)

djencks:
I don't see any need to define a maximum and enforce it.  If you compile against the version of annotations for your desired maximum, won't you get a compile error if you use any newer features?  So, I think even inferring the maximum from the annotation bundle on the build path is unnecessary.
I think recording the reason for the version selected will be very helpful, and perhaps it could be logged whether or not greater than some maximum. Your reason strings could use a lot of work, I can perhaps help if BJ runs out of time.  To solve the bug you mentioned I'd expect logging every call that might change the version will be more helpful than only the reason for the maximum.

biz.aQute.bndlib/src/aQute/bnd/component/DSAnnotations.java
```diff
@@ -222,6 +239,24 @@ public boolean analyzeJar(Analyzer analyzer) throws Exception {
 		return false;
 	}
 
+	/*
+	 * Check for any version conflicts and report them as errors
+	 */
+	private void checkVersionConflicts(Analyzer analyzer, ComponentDef definition) {
+		DeclarativeServicesAnnotationError dse = new DeclarativeServicesAnnotationError(
```
djencks: Only create dse if we use it?

biz.aQute.bndlib/src/aQute/bnd/component/DSAnnotations.java
```diff
@@ -134,7 +146,7 @@ public boolean analyzeJar(Analyzer analyzer) throws Exception {
 
 		TreeSet<String> provides = new TreeSet<>();
 		TreeSet<String> requires = new TreeSet<>();
-		Version maxVersion = V1_0;
+		Version maxVersionUsedByAnyComponent = V1_0;
```
djencks: This name grates on my sensibilities.  Perhaps maxComponentVersion? maxVersionEncountered? maxVersionUsed?

bjhargrave: I like `highestVersionUsed`.

djencks:
On further thought, enforcing a maximum is useful as it takes care of including components compiled elsewhere, possibly with a later annotation bundle.
I don't see a need to allow setting the maximum in -dsannotation-options.
I think determining the "extender" setting from the build path annotation bundle version and deprecating explicit use in -dsannotation-options would be a good idea.

bjhargrave:
>  If you compile against the version of annotations for your desired maximum, won't you get a compile error if you use any newer features?

I originally held this same position, but I think there are cases where you can use features from later DS versions than what you may compile against. Since much of DS is reflective, you could use a method signature shape that is supported by a later version of DS than you desire to run against. So, for example, even if you have DS 1.2 annotations on your buildpath, you could use a bind method signature which is only supported by DS 1.3. Bnd would see this and emit a 1.3 xml. Then your bundle does not work at runtime with SCR 1.2.

rotty3000:
What's about `highestVersionObserved`?

bjhargrave:
> What's about `highestVersionObserved`?

I think the important concept is that you _used_ the version not just observed it :-)

pkriens:
Bike shedding anyone?

biz.aQute.bndlib/src/aQute/bnd/component/DSAnnotations.java
pkriens: important to show that it is over all components and not per component

pkriens:
Starting over in #3121

Issue #19, opened 2010-08-03
RFE: Backslashes in ${somePath} causes problems on Windows


ddekany:
For example assume you have this in the bnd file, that you load with the bnd Ant task:

  Include-Resource: META-INF=${basedir}/something/META-INF

This works fine on UN*X, but on Windows Ant will use the native Windows path format for basedir, and it seems bnd will try to interpret the backslashes in that as escapes, corrupting the path. This certainly occurs because you are using the Properties reader, which will interpret backslashes as escapes.

To solve this, interpolation could support something like ${var|antFiles}, where "|antFiles" indicates the the value is in the format that Ant uses for file lists, rather than in the format that bnd natively expects. This could escape backslashes, so that C:\temp becomes to C:\temp, etc.

PR #3128, opened 2019-04-03
[bundle annotations] add a opt-in model for bundle meta-annotations to inherit values from their annotation host

rotty3000:
This proposes to allow a bundle annotation, _used in the meta-annotation scenario_, to inherit from it's host. The mechanism is opt-in, and __NO automatic inheritance is provided__.

The model is simply that any property of the host is accessible using the name of the field prefixed with `#` inside bnd's macro language (__and thus is only supported on targets that accept `String`__).

Consider the following code:
```java
enum RGB {
    RED, GREEN, BLUE
}
@interface RGBLight { 
     RGB value() default RED;
}
```
Adding the following bundle meta-annotation means the `value` field of `@RGBLight` is slurped into the `name` field of `@Requirement` using the macro `${#value}`:
```java
@Requirement(
    name = "${#value}",
    namespace = "rgb.light"
)
@interface RGBLight { 
    RGB value() default RED;
}
```
And finally the usage looks like:
```java
@RGBLight(BLUE)
class Some {...}
```

If you wish to view more complex examples please review the test cases as this simple model proves powerful enough to solve a wide variety of cases.


cdelg:
In addition to inheritance, does something can be done specifically for filters? Something like the macro feature from the RequireCapability annotation or may be a bnd @filter annotation (with the same behavior as the OSGi @Attribute annotation but for filters)?

rotty3000:
That problem cannot occur with this proposal.

bjhargrave:
This PR is devoid of a meaningful title and any design description of the change. A reviewer must read the test case to glean the design of the change. Perhaps some design description in this PR's openinig comment is in order?

rotty3000:
@bjhargrave @timothyjward fair enough. This proposal is not well described. I'm fixing that now.

rotty3000:
Please see the PR comment and if you need further info don't hesitate to ask.

rotty3000:
@cdelg this proposal does work with filters, please see the test cases.

rotty3000:
Thanks @djencks , fixed (and also squashed)!

timothyjward:
> That problem cannot occur with this proposal.

Now that I've had an opportunity to reacquaint myself with what's going on here I'm still quite concerned about three things:

1. The large number of "magic strings" involved. I see references to member names in meta annotated annotations, macro string processing, and all sorts of other things going on.

2. How far down the rabbit hole does this processing go? If I apply a Capability Annotation with a placeholder `${#name}` to an annotation which specifies a `name` member then it's clear how that works, but what if that is applied to an annotation which supplies a `name` member? This would be a repeat of the undefined behaviour we had previously.

3. The introduction of full bnd macro processing is a **very bad thing** it should only be possible for these annotations to include things which are 100% compile time constants derivable from the class files. If we do anything else then a build project which slurps in code from a different build project may get completely incorrect metadata due to it having different bnd properties set. The output of the annotations should not be determined by what properties I set in my bnd file - this is why we separated out the small number of static properties, such as the name of the class you were annotating.

In summary, if we're going to solve this problem, which I think we probably should, then the correct way to do it is to solve the actual problem, not to try to work around it. In this case that would mean something like:

	@Capability(namespace = SLM_NAMESPACE, version = SLM_SPEC_VERSION)
	public @interface ServiceProvider {
		/**
		 * The <em>type</em> of the service.
		 */
                @OverrideMember("name")
		Class<?> value();
	
		/**
		 * The <em>implementation type</em> of the service. When not specified, uses
		 * the type on which the {@code @ServiceProvider} annotation is placed.
		 */
                @Directive
		Class<?> register() default Target.class;
	
		/**
		 * A list of service property names and values.
		 * <p>
		 * Each string should be specified in the form:
		 * <ul>
		 * <li>{@code "name=value"} for attributes.</li>
		 * <li>{@code "name:type=value"} for typed attributes.</li>
		 * <li>{@code "name:=value"} for directives.</li>
		 * </ul>
		 * These are added, separated by semicolons, to the export package clause.
		 */
                @OverrideMember("attribute")
		String[] serviceProperty() default {};
	}
	
	@Requirement(name = SLM_EXTENDER_REGISTRAR, namespace = EXTENDER_NAMESPACE, resolution = Resolution.OPTIONAL, version = SLM_SPEC_VERSION)
	@Requirement(effective = "active", name = SLM_EXTENDER_REGISTRAR, namespace = EXTENDER_NAMESPACE, version = SLM_SPEC_VERSION)
	public @interface RequireServiceLoaderRegistrar {

	}

	@ServiceProvider(MyService.class)
	@RequireServiceLoaderRegistrar
	public class MyServiceProvider implements MyService {
		...
	}

rotty3000:
@timothyjward can you clarify the last part of `2.` I don't quite get it. Maybe an example would help.

> The introduction of full bnd macro processing is a very bad thing
I didn't add that in this PR... I'm just using it  (see [1]).

[1] https://github.com/bndtools/bnd/pull/3128/files#diff-7f871e4610c2ca503dc5fefd82be96eaR878

rotty3000:
The macro processing and it's visibility into the current analyzer has been there for 5 years and was added by @pkriens https://github.com/bndtools/bnd/blame/450c90cd3ff01b1113ed290c2e3b0d9a67d4ceb4/biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java#L877

biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
```diff
@@ -419,6 +455,29 @@ private void handleAttributeOrDirective(Annotation a) {
...
+
+			if (object == null) {
+				try {
+					object = c.getDefaults().get(name);
+				} catch (Exception e) {
+					throw Exceptions.duck(e);
+				}
+			}
+
+			if ((object instanceof TypeRef) && Target.class.getName()
```
bjhargrave: I think we will need a doc page in the PR to cover all the features like this little bit of magic. I am not sure I like reusing `Target` for this. Perhaps we should define a new (annotation?) type for this purpose.

biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
```diff
@@ -138,6 +155,14 @@
 	// we parse the annotations separately at the end
 	boolean							finalizing;
 
+	static String convert(Object value) {
+		try {
+			return CONVERTER.convert(String.class, value);
```
bjhargrave: The doc will also need to say how each element type is converted. E.g. Enum to a String of the enum name. (Hence the need to `${tolower}` in the test case above). It is unclear how arrays are handled here, or annotation values. We probably will need tests for the handling of all the element types (not just string and enum).

biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
rotty3000: yeah, this was just a lazy solution. On the other hand it's never referenced by the user, so it could be an internal marker type.

biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
rotty3000: yup, good point

biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
rotty3000: Should I just use `Object` instead?

rotty3000: I was looking for a common type that would not be accidentally mistaken for some other usable type. Object is sometimes used in real scenarios. Also, I didn't want to force a type dependency for this feature, more of a convention (since this whole feature is just about convention).

bjhargrave: Don't use Object. I think we need to define a specific type which can be used here and has no other meaning. Classes are cheap.

rotty3000: So which jar are we gonna force users to depend on when they code up their annotations?

bjhargrave: That is a fair question. But I am not a fan of reusing a class for a different purpose. I think `Target` is very close to the meaning here. Maybe we use it to avoid a new compile dependency.

timothyjward:
> @timothyjward can you clarify the last part of 2. I don't quite get it. Maybe an example would help.

Lets write a flexible whiteboard requirement:
    
    @Requirement(namespace="osgi.implementation", name="${#value}", version="${#version}")
    public @interface @RequireWhiteboard {
        /**
         * Set the name of the whiteboard you require
         * /
        String value();

        String version() default "1.0.0";
    }

Now lets use it to build the JAX-RS Name ComponentPropertyType

    @RequireWhiteboard("osgi.jaxrs")
    @ComponentPropertyType
    public @interface JaxrsName {
        String value();
    }
        
When I apply this annotation to my component:

    @JaxrsName("foo")
    @Component
    public MyMessageBodyReader implements MessageBodyReader {
        ...
    }

Which of the `value` members should be pulled in by `${#value}`? There's one defined in the `@RequireWhiteboard` applied to the `@JaxrsName` annotation, and another defined in the `@JaxrsName` applied to `MyMessageBodyReader`.

This is why it's the overriding location that needs to opt in, not the source location. When you write a meta annotation you can **never** know how many levels will end up below it, or what their members will be. On the other hand you **always** know what meta annotations are applied to you and so what you can safely override. In this case it could easily be fixed by not using a replacement template adding the proposed `@OverrideMember` to the `value` member of the `@RequireWhiteboard` annotation. This would unambiguously state that the value `osgi.jaxrs` is intended to be used as an override to the `value` member in the applied `@Requirement` meta-annotation, but that `foo` is just the value of the member in the `@JaxrsName` annotation.

rotty3000:
> Which of the `value` members should be pulled in by `${#value}`? There's one defined in the `@RequireWhiteboard` applied to the `@JaxrsName` annotation, 

So the inheritance only goes down exactly one level. How is this done? The inherited members, which come exactly from the host, are pushed into a macro processor instance used only for the current meta-annotation. Then they are thrown way.

> and another defined in the `@JaxrsName` applied to `MyMessageBodyReader`.

This one is not visible in the proposed inheritance model.
 
> This is why it's the overriding location that needs to opt in, not the source location.

Agreed! and this is the case. I will write some unit tests specifically to assert this (if I don't already have some).

> When you write a meta annotation you can **never** know how many levels will end up below it, or what their members will be.

Which is why the inheritance is limited to exactly one level down and only visible to the immediate macro processor instance.

rotty3000:
Please re-review.

* added docs
* at the request of @timothyjward I've omitted the annotations from the accessor properties entirely (this is documented)

rotty3000:
I had missed the details about `java.lang.Target`.

biz.aQute.bnd.annotation/bnd.bnd
```diff
@@ -2,5 +2,14 @@
 -include: ${workspace}/cnf/includes/jdt.bnd
 
 Bundle-Description:     bnd Annotations Library
--exportcontents:        aQute.bnd.annotation.*
 
+Export-Package: \
+	aQute.bnd.annotation.*,\
+	org.osgi.annotation.bundle,\
```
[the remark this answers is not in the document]
rotty3000: So it's ok for `aQute.bnd.annotation` to depend on `osgi.annotation` at compile time? I wasn't sure.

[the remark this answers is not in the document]
rotty3000: furthermore, this means that bndlib will also have an import for the package and need it at runtime to resolve in bndtools... because it embeds `aQute.bnd.annotations`.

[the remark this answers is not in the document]
bjhargrave: Of course, since the project is using the CLASS retention annotations. Not exporting them from a jar that will be going into maven central.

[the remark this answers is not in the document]
bjhargrave: Hmm, are saying the OSGi annotations appear in the type signatures (compile dependency) or the OSGI annotations are used at runtime? Bnd does not do the latter since it has its own private copies to use at runtime.

biz.aQute.bnd.annotation/src/aQute/bnd/annotation/spi/SPIConstants.java
```diff
@@ -0,0 +1,38 @@
+package aQute.bnd.annotation.spi;
+
+public @interface SPIConstants {
```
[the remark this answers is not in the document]
rotty3000: typo

biz.aQute.bnd.annotation/src/aQute/bnd/annotation/headers/About.java
```diff
@@ -1,69 +1,7 @@
...
- * component now wants to depend this resource, it can declare its component as
- * follows:
- * 
- * <pre>
- * &#64;Webserver.Require
- * public class MyResource {
- *   ...
- * }
- * </pre>
+ * See {@link aQute.bnd.annotation.headers}
```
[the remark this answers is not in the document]
rotty3000: this package was already at `2`.

biz.aQute.bnd.annotation/src/aQute/bnd/annotation/spi/ServiceProvider.java
```diff
@@ -30,6 +53,19 @@
...
+	 * <p>
+	 * Specifies the time the {@code osgi.extender} requirements are available. The
+	 * OSGi framework resolver only considers requirements without an effective
+	 * directive or {@code effective:=resolve}. Requirements with other values for the
+	 * effective directive can be considered by an external agent.
+	 * <p>
+	 * If not specified, the {@code effective} directive is omitted from the
+	 * requirement clause.
+	 */
+	String effective() default "resolve";
```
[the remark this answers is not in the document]
rotty3000: yup

biz.aQute.bnd.annotation/src/aQute/bnd/annotation/headers/About.java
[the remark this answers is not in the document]
rotty3000: sorry, I'm wrong... it was 1.1. So removing the About class caused the MAJOR version change for the bundle. Was that acceptable? wouldn't that mean we need to change all to 5.0.0?

biz.aQute.bnd.annotation/src/aQute/bnd/annotation/headers/ProvideCapability.java
```diff
@@ -10,20 +10,26 @@
...
+ *
  * <pre>
  * interface Webserver {
  * 	&#64;ProvideCapability(ns = "osgi.extender", name = "aQute.webserver", version = "${&#64;version}")
- * 	&#64;interface Provide {}
- * 
+ * 	&#64;interface Provide {
+ * 	}
+ *
  * 	&#64;RequireCapability(ns = "osgi.extender", filter = "(&(osgi.extender=aQute.webserver)${frange;${&#64;version}})")
```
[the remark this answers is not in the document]
rotty3000: sure

biz.aQute.bnd.annotation/bnd.bnd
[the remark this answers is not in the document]
rotty3000: Note the annotations, but the Enums in the package.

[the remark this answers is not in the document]
rotty3000: it's in reference to this usage: https://github.com/bndtools/bnd/pull/3128/files#diff-25fa004735a759b33dfd907e560379adR104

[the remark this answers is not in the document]
rotty3000: and https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/bnd.bnd#L11

rotty3000:
@bjhargrave I addressed all suggestions except the last comment on convert. I'm not really sure what to do about that one.

biz.aQute.bndlib/bnd.bnd
```diff
@@ -8,7 +8,6 @@ Bundle-Description: bndlib: A Swiss Army Knife for OSGi
 	aQute.lib.spring
 
 Export-Package: \
-	aQute.bnd.annotation.*;-noimport:=true,\
```
bjhargrave: bndlib has long contained these packages. I don't think it is wise to remove them now. Perhaps on 5.0.

biz.aQute.bndlib/bnd.bnd
```diff
@@ -38,6 +37,7 @@ Export-Package: \
 	aQute.bnd.util.home;-noimport:=true,\
 	aQute.bnd.util.repository;-noimport:=true,\
 	aQute.bnd.component.error;-noimport:=true,\
+	org.osgi.annotation.bundle,\
```
bjhargrave: bndlib should not contain or export this package. bndlib should not have any runtime dependency on this package.

biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
```diff
@@ -828,11 +929,16 @@ private Attrs getAttributes(Annotation a, String... ignores) {
...
 		if (value == null)
 			return;
 
-		Processor next = new Processor(analyzer);
+		Processor next = new Processor();
+		next.addProperties(annotation.entrySet()
+			.stream()
+			.filter(entry -> entry.getKey()
+				.startsWith("#"))
+			.collect(Collectors.toMap(Entry::getKey, entry -> convert(entry.getValue()))));
```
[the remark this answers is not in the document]
bjhargrave: The question here is: Are the values to be treated as plain text (in which case the macro characters must be escaped so they are not processed as macro characters macro processor) or they are treated as possible macro values (in which case no escaping by us is needed but the documentation must make clear that the values are processed by the macro processor and if the developer intends to use macro characters as plain text, then the developer must properly escape them.)

biz.aQute.bndlib/src/aQute/bnd/plugin/spi/SPIAnnotations.java
```diff
@@ -1,238 +1,52 @@
...
 
-public class SPIAnnotations implements AnalyzerPlugin {
-
-	private static final Version	SLM_VERSION			= new Version(1);
-
-	private static final String		SERVICE_CONSUMER	= "aQute/bnd/annotation/spi/ServiceConsumer";
-	private static final String		SERVICE_CONSUMERS	= "aQute/bnd/annotation/spi/ServiceConsumers";
-	private static final String		SERVICE_PROVIDER	= "aQute/bnd/annotation/spi/ServiceProvider";
-	private static final String		SERVICE_PROVIDERS	= "aQute/bnd/annotation/spi/ServiceProviders";
+public class SPIAnnotations implements VerifierPlugin {
```
bjhargrave: Is this a good idea to run as a Verifier and not an Analyzer? I would like to get @pkriens perspective on this. I assume you do this to avoid complaints about the `register:` directive.

biz.aQute.bndlib/bnd.bnd
rotty3000: It's still in there! This is just to avoid double exportation warning. The packages are still contained and exported. This is a side effect of the new `package-info.java` usage in biz.aQute.bnd.annotation`

biz.aQute.bndlib/bnd.bnd
rotty3000: so should it forcibly do `!org.osgi.annotation.bundle` on import? The annotations packages USE the enums in `org.osgi.annotation.bundle.Requirement`! Which makes them part of the required API.

biz.aQute.bndlib/src/aQute/bnd/plugin/spi/SPIAnnotations.java
rotty3000: I guess I can try escaping macros to see what happens.

biz.aQute.bndlib/bnd.bnd
bjhargrave: We should not even put that package in the jar. This is for the same reason we no longer put the OSGi ds and metatype annotation packages in the jar. It created problem in non-OSGi scenarios where bndlib is on the classpath with the OSGi annotation jars and people end up with the wrong classes. Since we don't want to end up with similar issues with another OSGi annotation package, we must not put this package in the bnd jars. The answer may be to not use the OSGi enum types in the Bnd annotations but add specific support, like we do for the OSGi annotations, to map the UPPERCASE enum names to lowercase for the generated manifest headers.

bjhargrave: See https://github.com/bndtools/bnd/pull/2637/commits/e6470d48133a0f3305347e52f77149686fe98325 where we changed to stop using the real OSGi bundle annotations in bndlib.

biz.aQute.bnd.annotation/src/aQute/bnd/annotation/Constants.java
```diff
@@ -0,0 +1,25 @@
+package aQute.bnd.annotation;
+
+public interface Constants {
```
bjhargrave: I prefer we use a class (not an interface) for constant definition. See Effective Java, 2nd Ed `#19`.

biz.aQute.bnd.annotation/src/aQute/bnd/annotation/spi/Constants.java
```diff
@@ -0,0 +1,18 @@
+package aQute.bnd.annotation.spi;
+
+public interface Constants {
```
bjhargrave: I prefer we use a class (not an interface) for constant definition. See Effective Java, 2nd Ed `#19`.

PR #3137, opened 2019-04-08
[launchpad] Add methods for copying classes into the framework

kriegfrj:
I had a use case for Launchpad where it was necessary for my class-under-test to be loaded from within a bundle, instead of the usual Launchpad way of being loaded by the system classloader and then imported by the synthetic test bundle. However, the test harness still needed access to it to verify the behavior.

This PR gives a mechanism for this. 
* `BundleBuilder.addResourceWithCopy(Class)` actually loads the class file into the synthetic bundle, instead of importing it.
* `Launchpad.instantiateInFramework(Class)` will create a synthetic test bundle, copy the class into it, load it using the bundle classloader, then instantiate it using the no-args constructor. The result will not be directly assignable to a variable of the same type - it will need to be a superclass or superinterface that is shared across the framework boundary.


bjhargrave:
This PR needs a rebase on master.

kriegfrj:
> This PR needs a rebase on master.

Done.

biz.aQute.launchpad/src/aQute/launchpad/Launchpad.java
```diff
@@ -870,6 +870,37 @@ public Framework getFramework() {
...
+			clazz.getConstructor();
+		} catch (NoSuchMethodException e) {
+			Exceptions.duck(e);
+		}
+		Bundle b = bundle().addResourceWithCopy(clazz)
+			.start();
+		try {
+			@SuppressWarnings("unchecked")
+			Class<? extends T> insideClass = (Class<? extends T>) b.loadClass(clazz.getName());
+			return insideClass.newInstance();
```
bjhargrave: This should be `return insideClass.getConstructor().newInstance();`

biz.aQute.launchpad/src/aQute/launchpad/BundleSpecBuilder.java
```diff
@@ -871,6 +871,28 @@ default BundleSpecBuilder addResource(Class<?> class1) {
...
+	 * (typically from the system classloader). Classes added using this method
+	 * will be loaded from within the framework, rather than by the system
+	 * framework.
+	 * 
+	 * @param class1 the class to add as a resource
+	 */
+	default BundleSpecBuilder addResourceWithCopy(Class<?> class1) {
+		String className = class1.getName();
+		String path = className.replace('.', '/') + ".class";
+		URL url = BundleSpecBuilder.class.getClassLoader()
```
bjhargrave: Shouldn't this use the specified class' class loader to load the resource? For example, `URL url = class1.getClassLoader().getResource(path);`? Or even `URL url = class1.getResource(class1.getSimpleName()+".class");` since the class file resource is in the package of the class.

biz.aQute.launchpad/src/aQute/launchpad/Launchpad.java
kriegfrj: Can you explain your reasoning a bit? Is it that you would prefer instantiateInFramework() to throw an InvocationTargetException rather than throwing the raw exception?

biz.aQute.launchpad/src/aQute/launchpad/BundleSpecBuilder.java
kriegfrj: I think I tried this and it didn't work. I'll try again.

kriegfrj: Ok:
- `getSimpleName()` doesn't work with nested classes. Eg, for `my.pkg.Clazz.InnerClass` it simply gives you `InnerClass`, but the class file on disk will be `Clazz$InnerClass.class`.
- Using class1's classloader works. I can't remember why I used `BundleSpecBuilder`'s classloader, but I think it must have been either a misguided micro-optimisation, or else a miscalculation as to how class visibility issues might play out. I'll change it to use class1's classloader.

biz.aQute.launchpad/src/aQute/launchpad/Launchpad.java
bjhargrave: See https://github.com/bndtools/bnd/commit/8ae6af3bb8d252463d9f6e5cbca46044e15251b4

bjhargrave:
Thanks!

biz.aQute.launchpad/src/aQute/launchpad/Launchpad.java
kriegfrj: Thanks for educating me! Also ironic that the reason `Class.newInstance()` has been deprecated is the same reason that I originally preferred it over `getConstructor()`, and that throughout Launchpad we have gone to efforts to bypass checked exceptions...

I'm wondering if perhaps `insideClass.getDeclaredConstructor().newInstance()` might not be better (as `getConstructor()` will only work with public constructors). Your thoughts?

bjhargrave: It is best to limit to the default (public) constructor. Then people can control whether they want the constructor called. And we avoid needing setAccessible and all that baggage. The OSGi specs (e.g.) only use public constructors.

kriegfrj: I mostly agree, but one pushback for consideration: Restricting it to public definitely makes sense for production code; however this is code that is typically used for testing, and often in testing you want to do things and exercise code paths that you would usually avoid in a production environment. Hence the thinking that it might be nice to make it easier to access them.

bjhargrave: Perhaps, but it seems a small chance that there is a non-public no-param constructor (that is meaningful) and not a public no-param constructor. The more interesting case is that there are constructors which take arguments and that is not useful here since this code only deals with no-param constructor

kriegfrj: Fair enough. Adding support for constructors with args was something I considered, and I think it would be a worthwhile addition, but as my immediate need was only for a no-arg constructor I tried to limit the scope to correspond to my limited time. :smile:

PR #3139, opened 2019-04-08
[cdi annotations] support pre-existing beans.xml file but only if -cdiannotations default is not overridden

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
```diff
@@ -83,6 +83,13 @@ public boolean analyzeJar(Analyzer analyzer) throws Exception {
 
 		Jar currentJar = analyzer.getJar();
 
+		Resource beansXml = currentJar.getResource("META-INF/beans.xml");
+
+		if ((beansXml != null) && "*".equals(header.toString())) {
+			Discover discover = findDiscoveryMode(beansXml);
+			header = new Parameters("*;discover=" + discover);
```
bjhargrave: Why does this only happen for `*`?

Also, you could do `header.get("*").put("discover", discover.toString());` to set the discovery value rather than build and parse a new string.

rotty3000: so yeah, the comment should have explained it. We only do it when there's no explicitly modified default. Otherwise use the instruction. This is simply to enabled zero config mode where the dev already has a beans.xml file in their project.

rotty3000:
updated!

biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
rotty3000: i.e. if the developer added a mode complex instruction, like:

`-cdiannotations: *, some.other.package;discover=none` then they already are doing what you'd have achieved from inside the beans.xml, so the instruction is overridden as a whole.

Does that clarify things?

bjhargrave: I guess another approach might be 

```java
if (beansXml != null) {
  Attrs attrs = header.get("*");
  if ((attrs != null) && !attrs.containsKey("discover")) {
    Discover discover = findDiscoveryMode(beansXml);
    attrs.put("discover", discover.toString());
  }
}
```

That is, `META-INF/beans.xml` is consulted to find the default `discover` value for `*`, if `*` is present in the instruction.

rotty3000: I suppose there is no harm in doing this.

rotty3000:
updated!

biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
rotty3000: btw, I did change this

PR #3144, opened 2019-04-10
[bundle annotations] docs

rotty3000:


timothyjward:
@rotty3000 - my understanding that not putting the default value of members marked `@Attribute` or `@Directive` into the manifest was deliberate. Otherwise we will end up with things like `resolution:=mandatory` everywhere, which is much harder to read.

@bjhargrave - do you agree?

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,284 @@
...
+```
 
-* `value` – (String) Anything in the value field is appended after the calculated header.
-* `ns` – (String) The namespace of the requirement
-* `effective` – (String default `resolve`) The effective time
-* `name` – (String) This creates a `<namespace>=<name>` attribute, the common convention to set the primary name attribute of a capability in most OSGi namespaces.
-* `version` – (String) Set the version of the capability.
-* `uses` – (String[]) Package names that are used by this capability and require the same class loader
-* `mandatory` – (String[]) List of mandatory attributes
+When you write a DS component using `@Component`
```
timothyjward: I think this is missing a step - the doc doesn't say that the `@Component` annotation is meta-annotated with `@RequireServiceComponentRuntime` which means it is unclear why `@Component` causes the header generation

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,284 @@
...
+
+// usage
+
+@Extended
+class Foo {}
+```
+
+which results in the manifest header:
+
+`Provide-Capability: foo.namespace;foo.attribute=bar`
```
timothyjward: I think this example needs to be changed because I don't think defaults are supposed to be output into the manifest.

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,284 @@
...
+
+// usage
+
+@Extended
+class Foo {}
+```
+
+results in the manifest header:
+
+`Provide-Capability: foo.namespace;foo.directive:=bar`
```
timothyjward: Same as above - I don't think that defaults should be going into the manifest

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,284 @@
...
+
+**However**, _specified namespaces_ will be validated and will not accept directives which are not part of the spec _unless_ they are prefixed with `x-`.
```
timothyjward: Change "_specified namespaces_" to "Namespaces defined in the OSGi specifications" because otherwise it's not clear whether it's just a list of namespaces we declare somewhere.

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,284 @@
...
+
+```java
+@Capability(namespace = "osgi.extender", name = "bar", version = "1.0.0")
+@interface Extended {
+	@Directive("foo")
+	String value() default "bar";
```
timothyjward: More use of defaults. Even if we do decide to write out default values I would avoid this in the examples here and have a section explaining what we do with defaults.

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,284 @@
...
+
+// usage
+
+@Extended
+public class Foo {}
+```
+
+results in the manifest header:
+
+`Provide-Capability: osgi.extender;osgi.extender=bar;version:Version="1.0.0";x-foo:=bar`
```
timothyjward: More use of defaults

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,284 @@
...
+
+### Where to find Bundle Annotations
+
+**OSGi** *bundle annotations* can be found in the `osgi.annotation` bundle.
+
+```xml
+<dependency>
+  <groupId>org.osgi</groupId>
+  <artifactId>org.osgi.annotation.bundle</artifactId>
```
timothyjward: This isn't the maven artifact for `osgi.annotation` is it meant to be?

docs/_chapters/230-manifest-annotations.md
rotty3000: I tried to imply this via the code comment, but I can make it more explicit.

docs/_chapters/230-manifest-annotations.md
rotty3000: will do.

docs/_chapters/230-manifest-annotations.md
rotty3000: hmm, I'm always torn with the various annotation bundles. In general I feel that using the most specific dependency is best. Other times I use the `osgi.annotation` aggregate out of convenience. I'll follow consensus on this though.

rotty3000:
Regarding the defaults, I found that not having had the defaults was prety unexpected. It's the least intuitive option and I think will trip up developers. Without my OSGi/BND hat on, if the default is NEVER used anywhere (and since these are CLASS retention and I can't load the annotation to see what the default value is), then it feels like a bug to me.

On the other hand I'm not very strongly tied to this change itself. However, I do still feel strongly that it's not intuitive and feels buggy.

docs/_chapters/230-manifest-annotations.md
timothyjward: > In general I feel that using the most specific dependency is best.

I'm fine with that approach, it's just that line 213 explicitly states `osgi.annotation`. Whichever we use these two places should be consistent.

timothyjward:
> Regarding the defaults, I found that not having had the defaults was pretty unexpected. It's the least intuitive option and I think will trip up developers.

The problem is that the opposite was also found to be true. When the defaults were applied we ended up with large numbers of entries like:

    Require-Capability: osgi.implementation; filter:="(&(osgi.implementation=osgi.jaxrs)(version=1))";
        resolution:="mandatory"; cardinality:="single"; effective:="resolve"

Which became totally unreadable and made the manifest huge.

bjhargrave:
> Regarding the defaults, I found that not having had the defaults was pretty unexpected.

See the spec: https://osgi.org/specification/osgi.core/7.0.0/framework.api.html#org.osgi.annotation.bundle.Attribute

`If the element annotated with Attribute is unspecified when applied, then the attribute must not be generated in the generated capability or requirement clause.`

This means that defaults are not to be processed since they can only apply when the element is unspecified. And if the element is unspecified, then it must not be emitted.

So your commit to process the defaults violates the spec.

biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java
```diff
@@ -398,7 +400,7 @@ private void mergeAttributesAndDirectives(Annotation a) {
 		}
 
 		private void handleAttributeOrDirective(Annotation a) {
-			Object o = annotation.get(lastMethodSeen);
+			Object o = getOrDefault(lastMethodSeen);
```
bjhargrave: This violates the spec: https://osgi.org/specification/osgi.core/7.0.0/framework.api.html#org.osgi.annotation.bundle.Attribute.

bjhargrave:
Basically, defaults in bundle annotations are documentary (e.g. resolution defaults to mandatory) and otherwise ignored.

rotty3000:
I believe I've made all the suggested updates.

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,286 @@
 ---
 order: 230
-title: Manifest Annotations
+title: Bundle Annotations
 layout: default
 ---
 
-Entering manifest headers is error prone for a number of reasons since these headers are often complex, have different often hard to remember fields, and are singletons. Because they are singletons, there is only one place where they can be entered: the manifest. This is in contrast with the promoted component model where a component is stand-alone. Components should be cheap and easy to rename or move between bundles. This works fine for its Java imports but if you have an associated header in the manifest then it is easy to forget to also move its corresponding headers from the manifest. This can cause orphaned headers or missing headers either in the old bundle or the new bundle. The old bundle can miss the header because multiple components were depending on that header but it was mistakingly removed.
+Manifest headers are notoriously hard to write. It often takes several attempts to get all the details correct.
```
bjhargrave: I think "notoriously hard to write" is over the top. Perhaps that they are challenging to keep in sync with the code in the bundle.

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,286 @@
...
 
-* `value` – (String) Anything in the value field is appended after the calculated header.
-* `ns` – (String) The namespace of the requirement
-* `effective` – (String default `resolve`) The effective time
-* `filter` – The filter directive. There is a handy macro [${frange;<version>][1] that turns a version into a filter expression for a filter range.  
-* `resolution` – (`mandatory` | `optional`) The resolution of the requirement.
+```
+Provide-Capability: osgi.extender;
+    osgi.extender="osgi.component";
+    version:Version="x";
```
bjhargrave: `x` is not a real version number and the example should be use a proper version number.

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,286 @@
...
-The `@BundleLicense` annotation creates entries in the Bundle-License header. The annotation has the following fields:
+You'll note the string `${@class}` used in the above example. String fields in *bundle annotations* are processed through bnd's macro processor. This macro processor provides access to all default and builder macros. *More info on bnd macros can be found in the [macros chapter](/chapters/850-macros.html).* 
 
-* `name` – The name of the license, should preferably the URI to the corresponding [Open Source Initiative][2] page about this license.
-* `description` – A short description of the license.
-* `url` – The URL to further information about the license
+Bnd also provides access to certain key properties of the current processing state.
 
-A number of licenses have gotten their own custom annotation:
+- ***@class*** - gives the fully qualified name of the class where the annotation hierarchy began
```
bjhargrave: It is the annotated class. I am not entirely sure what "the class where the annotation hierarchy began" means.

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,286 @@
...
-	Bundle-License: \
-		http://www.opensource.org/licenses/apache2.0.php; \
-		description='Apache Software License 2.0';\
-		link='http://www.apache.org/licenses/LICENSE-2.0.html'
+```java
+@BundleLicense(
+    name = "http://www.opensource.org/licenses/apache2.0.php", 
+    link = "http://www.apache.org/licenses/LICENSE-2.0.html", 
+    description = "Apache Software License 2.0")
+@interface ASL_2_0 {}
```
bjhargrave: We need to fix these annotations for licensing. (1) we should use the names recommended by the OSGi spec and (2) ASL_2 is not a license. It is the Apache License 2. (no S).

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,286 @@
...
+```
+
+which results in the manifest header:
+
+`Provide-Capability: foo.namespace;foo.attribute=bar`
+
+##### `@Directive`
+
+`@Directive` behaves similarly; with some caveats. You can add new or update existing directives for _non-specified namespaces_.
+
```
bjhargrave: What do you mean by "non-specified namespaces"? Is this because the Verifier will complain? People can elide the error with `-fixupmessages`.

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,286 @@
...
+```xml
+<dependency>
+  <groupId>org.osgi</groupId>
+  <artifactId>osgi.annotation</artifactId>
+  <version>7.0.0</version>
+</dependency>
+```
+
+**Bnd** *bundle annotations* can be found in `biz.aQute.bndlib` or `biz.aQute.bnd.annotations` bundles.
```
bjhargrave: I think we should ONLY direct people to use `biz.aQute.bnd.annotations` for Bnd annotations. bndlib is too much other stuff.

docs/_chapters/230-manifest-annotations.md
```diff
@@ -1,136 +1,286 @@
...
+  * `@CPL_1_0`
+  * `@EPL_1_0`
+  * `@GPL_2_0`
+  * `@GPL_3_0`
+  * `@LGPL_2_1`
+  * `@MIT_1_0`
+  * `@MPL_2_0`
+
+* `@ServiceConsumer` - Generates requirements in support of the consumer side of the [Service Loader Mediator](https://osgi.org/specification/osgi.cmpn/7.0.0/service.loader.html) specification.
+* ###### `@ServiceProvider` - Generates requirements and capabilities in support of the provider side of the [Service Loader Mediator](https://osgi.org/specification/osgi.cmpn/7.0.0/service.loader.html) specification. Also generates `META-INF/service` descriptors.
```
bjhargrave: Why all the `####` here?

docs/_chapters/230-manifest-annotations.md
rotty3000: does this really need to be part of this PR?

docs/_chapters/230-manifest-annotations.md
rotty3000: FWIW `-fixupmessages` is not a viable solution for a reusable annotation. You expect people to first use the annotation, and then have to cleanup errors with `-fixupmessages`?

djencks:
Grrrr! Please fix per bj’s comment, it’s the Apache License, or AL, the Apache Software License or ASL is not a thing.

rotty3000:
@djencks this is not new code so I'm not fixing it right now. Please send a PR if it's really urgent for you to have this fix.

rotty3000:
we can create an issue to track it if necessary.

PR #3152, opened 2019-04-18
WIP: Reporter - FileResourceLoader loads local files, default templates

stbischof:


bjhargrave:
@stbischof  Do you plan on fixing the failed DCO check?

stbischof:
oh sorry. its WIP.

Will fix some more.

bjhargrave:
You need to rebase on master and remove all the merge commits.

cdelg:
Concerning the templates:

- Maybe a bundle with no category is not a "fail".
- You should remove the tips, some of them are wrong and, for example, if a bundle has an "api" category, the readme file could provide wrong information if the bundle does not follow what is said in the tip. Maybe we could define, for each bundle usage, a classifier with a "bnd-" prefix.
- Some variables in the template (eg; currentBsn) are provided by the user in the bnd.bnd file, if you use them it should be mentioned in the doc.
- The embedded templates will be more useful if the user can easily extend them. Between each blocks you can add an empty block (eg; afterTitle, afterOverview, etc), then the user will only have to "extend" those block to customize their readmes. 
- Note that the workspace template will only work if the project is built with the Bnd workspace model. You can consider to have a more generic "workspace" template that will also work with a top reactor maven project. In this case, more information can be provided (eg; version, description, developers, scm, repositories, ...). Those information can be collected from the pom for maven projects and for Bnd workspace we can either collect them from the OSGi headers defined in the build.bnd file or from documented specific properties (eg;ws-description,ws-title, ws-licenses, etc). In the same way, the project templates will be better if they can handle multiple build tools.
- Also, I didn't test if the templates are working well, if they are now embedded in the Jar, maybe they should be unit tested.

cdelg:
I am currently working on the maven plugin, I can add an EntryPlugin for the top reactor maven project that collect information in a way that is compatible with the Bnd worksapce model. In this way the workspace readme template will be easier to write.

cdelg:
If you wait #3169 for being merged, this PR contains a new plugin which adds a `commonInfo` entry to a workspace report. If you rework the workspace template around it, the template will also be usable from maven.

bjhargrave:
Can we close this?

cdelg:
@stbischof if you need help, I can make an other PR in a couple of weeks with new templates and use the work you already done in this PR or I can send you a demo repo with the templates and let you integrate the all. Let me know if you already rework the templates to not duplicate the efforts.

cdelg:
@bjhargrave , @stbischof I have updated the templates, you can find them here: https://github.com/cdelg/bnd-doc-demo 
There is now only one readme template that handle any data source (bundle, simple project, multi-project) from maven or from a workspace model.

If you look at the repo, there is also some example projects that could be used to unit test the templates then.

stbischof:
#3287 Does the Job better! 
Thank you @cdelg  for finishing

PR #3160, opened 2019-04-23
[launchpad] BundleSpecBuilder to allow multiple versions of a bundle

kriegfrj:
Original implementation used the BSN as the bundle key ("location") when installing the bundle. This means you can't install two bundles in the same Launchpad with the same BSN, even if they have different versions, which means you can't use Launchpad to test the full gamut of OSGi situations.

This patch changes `BundleSpecBuilder` so that it uses BSN + bundle version as the bundle key rather than just the BSN, which allows the Launchpad to install multiple versions of the same bundle with the same BSN.

### DIFF
--- a/biz.aQute.launchpad/src/aQute/launchpad/BundleSpecBuilder.java
+++ b/biz.aQute.launchpad/src/aQute/launchpad/BundleSpecBuilder.java
@@ -917,11 +917,21 @@ public interface BundleSpecBuilder {
 	}
 
 	default Bundle install() throws Exception {
-		byte[] build = LaunchpadBuilder.workspace.build(LaunchpadBuilder.projectDir.getAbsolutePath(), x().spec);
-		String name = x().spec.bundleSymbolicName.toString();
+		BuilderSpecification spec = x().spec;
+		byte[] build = LaunchpadBuilder.workspace.build(LaunchpadBuilder.projectDir.getAbsolutePath(), spec);
+		String location;
+		if (spec.location == null) {
+			String name = spec.bundleSymbolicName.toString();
+			String version = spec.bundleVersion;
+			version = (version == null) ? "0" : version;
+			location = name + '-' + version;
+		} else {
+			location = spec.location;
+		}
+
 		ByteArrayInputStream bin = new ByteArrayInputStream(build);
 		return x().ws.getBundleContext()
-			.installBundle(name, bin);
+			.installBundle(location, bin);
 	}
 
 	default Bundle start() {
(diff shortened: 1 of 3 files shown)


biz.aQute.launchpad/test/aQute/launchpad/BundleBuilderTest.java
```diff
@@ -175,4 +175,22 @@ public void testOrder() throws Exception {
 
 		}
 	}
+
+	@Test
+	public void differentBundles_withSameSymbolicName_shouldBeTheSameBundle() throws Exception {
```
bjhargrave: Shouldn't this method name end in "should _Not_ BeTheSameBundle"?

biz.aQute.launchpad/src/aQute/launchpad/BundleSpecBuilder.java
```diff
@@ -919,9 +919,11 @@ default BundleSpecBuilder addConfiguration(URL url) {
 	default Bundle install() throws Exception {
 		byte[] build = LaunchpadBuilder.workspace.build(LaunchpadBuilder.projectDir.getAbsolutePath(), x().spec);
 		String name = x().spec.bundleSymbolicName.toString();
+		String version = x().spec.bundleVersion;
+		version = (version == null) ? "0" : version;
 		ByteArrayInputStream bin = new ByteArrayInputStream(build);
 		return x().ws.getBundleContext()
-			.installBundle(name, bin);
+			.installBundle(name + ';' + version, bin);
```
bjhargrave: I would suggest using `-` or `_` instead of `;` to join the bsn and version;

biz.aQute.launchpad/test/aQute/launchpad/BundleBuilderTest.java
kriegfrj: Yes... I thought I had fixed that... oops

biz.aQute.launchpad/src/aQute/launchpad/BundleSpecBuilder.java
kriegfrj: Ok, makes sense as that's how they normally are concatenated in urls.

kriegfrj:
> It seems like BundleSpecBuilder.install should take a String location argument since I should be able to install the exact same bundle (bsn+version) multiple times. And the default value for location (no args install method) can default to bsn+version.

I have also implemented this (via an additional builder property rather than as an arg to `install()`). However, I'm not quite sure I understand the first statement - the OSGi framework prevents you from attempting to install two bundles with the same BSN and version into the same framework. If you want to re-install a bundle, you either have to uninstall the first one first, or else you have to use a different code path than the one that `BundleBuilder` uses for installing a brand new bundle. To prevent accidental re-installing of a bundle (and producing an accidentally-undesired test situation), it would probably be better (IMO) if Launchpad also required you to use a different code path to re-install a bundle than for installing one from scratch. This is doable but probably best done in a different issue/PR.

bjhargrave:
> the OSGi framework prevents you from attempting to install two bundles with the same BSN and version into the same framework.

No. The framework requires each location string to be unique. But it is possible to install the same bsn+version using different location strings. See https://osgi.org/specification/osgi.core/7.0.0/framework.lifecycle.html#framework.lifecycle.bundleidentifiers.

bjhargrave:
Thanks!

kriegfrj:
> No. The framework requires each location string to be unique. But it is possible to install the same bsn+version using different location strings. See https://osgi.org/specification/osgi.core/7.0.0/framework.lifecycle.html#framework.lifecycle.bundleidentifiers.

Got it. My comment was based on an empirical test, but I was unaware of the existence and function of the `org.osgi.framework.bsnversion` property.

PR #3163, opened 2019-04-25
WIP: [CLI] Missing non .java files in bundle

stbischof:
The command "**bnd.jar compile**" compiles all *.java files into .class files in the **bin** folder.
The command "**bnd.jar build**" builds a jar file using the **bin** folder.

If there is a non .java file (e.g. some.sql) in a package it will not be copied into the **bin** and not be build into the jar.

What is the best way to handle this issue?

- copy as separate command
- copy while compile
- copy while build
- copy with include-resources Instruction

### DIFF
--- a/biz.aQute.bnd/test/aQute/bnd/main/TestBndMain.java
+++ b/biz.aQute.bnd/test/aQute/bnd/main/TestBndMain.java
@@ -1,9 +1,12 @@
 package aQute.bnd.main;
 
+import java.io.File;
+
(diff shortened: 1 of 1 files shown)


pkriens:
I think this depends on the compiler in use. The standard java compiler does not copy but the Eclipse ecj does. I think the gradle plugin does the copying on the fly. So I think you should do it on the fly before/after compilation.

bjhargrave:
Can we close this?

PR #3167, opened 2019-04-28
[gradle] non-workspace gradle Export/Resolve tasks don't have access to project properties

rotty3000:


biz.aQute.bnd.gradle/src/aQute/bnd/gradle/Bndrun.groovy
```diff
@@ -144,7 +145,21 @@ public class Bndrun extends DefaultTask {
...
+          Properties gradleProperties = new PropertiesWrapper()
+          gradleProperties.put('task', this)
+          gradleProperties.put('project', project)
+          Processor processor = new ProcessorWrapper(gradleProperties)
+          processor.setParent(parent)
+          run.setParent(processor)
+        }
+
+        worker(run)
+      } finally {
```
bjhargrave: Not sure I see the utility in restoring the parent just before the run is closed. I would just avoid this.

biz.aQute.bnd.gradle/src/aQute/bnd/gradle/Export.groovy
```diff
@@ -226,6 +236,7 @@ public class Export extends DefaultTask {
         }
       } finally {
         logReport(run, logger)
+        run.setParent(parent)
```
bjhargrave: Not sure I see the utility in restoring the parent just before the run is closed. I would just avoid this.

biz.aQute.bnd.gradle/src/aQute/bnd/gradle/Resolve.groovy
```diff
@@ -145,6 +155,7 @@ public class Resolve extends DefaultTask {
         throw new GradleException("${run.getPropertiesFile()} resolution exception", e)
       } finally {
         logReport(run, logger)
+        run.setParent(parent)
```
bjhargrave: Not sure I see the utility in restoring the parent just before the run is closed. I would just avoid this.

biz.aQute.bnd.gradle/src/aQute/bnd/gradle/Export.groovy
```diff
@@ -208,7 +209,16 @@ public class Export extends DefaultTask {
         throw new GradleException("${run.getPropertiesFile()} workspace errors")
       }
 
+      def parent = run.parent;
       try {
+        if (parent instanceof Processor) {
+          Properties gradleProperties = new PropertiesWrapper()
+          gradleProperties.put('task', this)
+          gradleProperties.put('project', project)
+          Processor processor = new ProcessorWrapper(gradleProperties)
```
bjhargrave: Why do you need ProcessorWrapper? Isn't `new Processor(gradleProperties, false)` enough?

biz.aQute.bnd.gradle/src/aQute/bnd/gradle/Resolve.groovy
```diff
@@ -136,7 +137,16 @@ public class Resolve extends DefaultTask {
         throw new GradleException("${run.getPropertiesFile()} workspace errors")
       }
 
+      def parent = run.parent;
       try {
+        if (parent instanceof Processor) {
+          Properties gradleProperties = new PropertiesWrapper()
+          gradleProperties.put('task', this)
+          gradleProperties.put('project', project)
+          Processor processor = new ProcessorWrapper(gradleProperties)
```
bjhargrave: Why do you need ProcessorWrapper? Isn't `new Processor(gradleProperties, false)` enough?

biz.aQute.bnd.gradle/src/aQute/bnd/gradle/Bndrun.groovy
```diff
@@ -144,7 +145,21 @@ public class Bndrun extends DefaultTask {
...
       }
 
-      worker(run)
+      def parent = run.parent;
+      try {
+        if (parent instanceof Processor) {
+          Properties gradleProperties = new PropertiesWrapper()
+          gradleProperties.put('task', this)
+          gradleProperties.put('project', project)
+          Processor processor = new ProcessorWrapper(gradleProperties)
```
bjhargrave: Why do you need ProcessorWrapper? Isn't `new Processor(gradleProperties, false)` enough?

biz.aQute.bnd.gradle/src/aQute/bnd/gradle/Bndrun.groovy
rotty3000: You may be right here since this may have been a side effect of earlier failed attempts. I'll try without this restoration.

biz.aQute.bnd.gradle/src/aQute/bnd/gradle/Export.groovy
rotty3000: The default `setParent` is not without side effects. In that implementation the properties of the `parent` are merged with those of the child. This breaks the nature of `PropertiesWrapper` which, has no real properties, is designed to only have effect when called directly. So this change is absolutely necessary to preserve that behavior.

rotty3000:
@bjhargrave can I get your blessing on this?

PR #3184, opened 2019-05-11
reporter: Add codeSnippet EntryPlugin

### LINKED ISSUE #3165: [reporter] Add an EntryPlugin to collect code snippets
We can collect code snippets from a project and use them to generate documentation. Code snippets could be localized in the test class path and extracted either with a specific annotation or a specific comment syntax (as maven does with their reports). 
The annotation or the comment should allow the user to provide the following information:

1. A title.
2. A short description.
3. If imports should be included or not.
4. If apply on a method, if the method signature should be included or not (or only the method content)
5. An ID. By default, it could be the class name or the method name. This id could then be used in a template to allow the user to attach further information for a specific code snippet.
6. A rank. By default it could be the derived from the order of the code snippet in a project.

Another annotation or comment type could be used to group several code snippets. A template could then be used to render examples with multiple steps.

cdelg:
Closes #3165 

This PR provide a new EntryPlugin to extract code snippets. It contains documentation too.

The implementation is based on the `JavaParser` core library (which has no transitive deps) and code snippets are extracted thanks to a special tag in comment blocks (more information in docs).


bjhargrave:
Your PR failed tests on Windows. Perhaps the code is generated with CRLF line endings and your tests expect LF line endings? Perhaps you can configure the code generator to always use LF line endings. This would be best so the output is the same regarding of the host OS.

cdelg:
Indeed, just one line to change, phew!

cdelg:
Now it's travis local rebuild.

biz.aQute.bnd.reporter/test/biz/aQute/bnd/reporter/codesnippet/CodeSnippetExtractorTest.java
```diff
@@ -0,0 +1,34 @@
+package biz.aQute.bnd.reporter.codesnippet;
+
+import biz.aQute.bnd.reporter.codesnippet.dto.CodeSnippetDTO;
+import biz.aQute.bnd.reporter.codesnippet.dto.CodeSnippetGroupDTO;
+import biz.aQute.bnd.reporter.codesnippet.dto.CodeSnippetProgramDTO;
+import java.io.File;
+import java.util.List;
+import junit.framework.TestCase;
+
+public class CodeSnippetExtractorTest extends TestCase {
```
bjhargrave: I think we need a test cases added to the reporter maven plugin as well.

biz.aQute.bnd.reporter/bnd.bnd
```diff
@@ -30,7 +30,8 @@
 	org.objectweb.asm;version="[5.0.3,6.0.0)",\
 	org.objectweb.asm.analysis;version="[5.0.3,6.0.0)",\
 	org.objectweb.asm.tree;version="[5.0.3,6.0.0)",\
-	org.objectweb.asm.util;version="[5.0.3,6.0.0)"
+	org.objectweb.asm.util;version="[5.0.3,6.0.0)",\
+	com.github.javaparser.javaparser-core;version=3.13
```
bjhargrave: I am getting concerned that the reporter is dragging in more and more 3rd party dependencies. Since all of these are included in biz.aQute.bnd.jar, it gets larger and larger. The reporter is by far the biggest adder of new 3rd party dependencies to the Bnd project.

jtwig, parboiled, apache commons, guava, asm and now javaparser.

rotty3000: I have the same concerns.

cdelg: Do you think it is possible to move plugins that bring those dependencies (I think of the JtwigTransformer and now the CodeSnippet plugin) into a dedicated project such as `biz.aQute.bnd.reporter.extra` and let user add this bundle to the class path when he uses the bnd CLI (to get the extra features). For maven is not a problem no? As it is a separate plugin. Or is it an issue at the workspace level too?

bjhargrave: I don't see that this helps bnd as a command line tool. It is currently installable from brew as a standalone jar.

We could make the reporter jar its own command line tool (have a main), but it would then need to download and class load its own deps.

Also, the deps in the maven plugin are not without risk of conflict with the deps of other maven plugins in a user's build.

rotty3000: Hmm, deps in maven can be applied in isolation like so:

```xml
<plugin>
    ...
    <dependencies>
       ...
    </dependencies>
</plugin>
```
so the risk is low to collide with other plugin dependencies

rotty3000: but in gradle there is no such magic, there you could be in trouble.

bjhargrave: So maven gives each plugin its own class loader? Did not realize that.

rotty3000: correct!

cdelg: Should I still work on this PR? I think after this one I will not add more features that need 3rd party dependencies, I could add a comment in the bnd.bnd to recall to not add extra deps and thus postpone this issue in the future.
If it is already an issue, I can:
* In bnd CLI, check if it is possible to precisely import what we need (`-conditionalpackage`) for Jtwig and JavaParser.
* In Maven, apparently it is not a problem.
* In Gradle, the plugin does not exist yet.
* Is there a benefice for maintenance and clarity to create a dedicated repository (cnf/ext/reporter.mvn)?

bjhargrave: Lets leave this PR chill for a little while until the Bnd maintainers can discuss at our next call.

bjhargrave:
This PR will need a rebase and merge conflict fixes due to changes in https://github.com/bndtools/bnd/pull/3192 where I removed reporter dependencies from `-buildpath` in `biz.aQute.bnd/bnd.bnd`.

cdelg:
I have rebased and solved the merge conflict. Does this change help with your concern about 3rd party dependencies?

Should I continue working on this PR?

bjhargrave:
> I have rebased and solved the merge conflict. 

Thanks.

> Does this change help with your concern about 3rd party dependencies?
> 
> Should I continue working on this PR?

Not really. But we still need to discuss on a Bnd maintainers call. So lets continue to keep the PR open.

bjhargrave:
Please fix the merge problem and rebase. 

We agreed to take this PR. But we do not want to continue to drag in more 3rd party dependencies. So any future PRs which do so will be not well received.

cdelg:
resolved and rebased. Thanks!

biz.aQute.bnd.reporter/test/biz/aQute/bnd/reporter/codesnippet/CodeSnippetExtractorTest.java
cdelg: I have added junit tests for the maven plugins.

cdelg:
Fine for me if you want to merge.

cdelg:
I think the build should be restarted.

maven/bnd-reporter-maven-plugin/pom.xml
```diff
@@ -60,6 +60,11 @@
 			<groupId>biz.aQute.bnd</groupId>
 			<artifactId>biz.aQute.bnd.reporter</artifactId>
 		</dependency>
+		
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
```
bjhargrave: This dependency needs to be test scoped. We do not want junit in the compile scope.

bjhargrave: Perhaps that is already handled by the parent pom's dependencyManagement section? @rotty3000

rotty3000: correct!

PR #3189, opened 2019-05-14
Added functionality to perform remote bundle operations

amitjoy:
Bnd agent has been enhanced to perform remote bundle operations (install, update and list)

### DIFF
--- a/biz.aQute.bnd/src/aQute/bnd/main/RemoteCommand.java
+++ b/biz.aQute.bnd/src/aQute/bnd/main/RemoteCommand.java
@@ -3,7 +3,10 @@ package aQute.bnd.main;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.io.OutputStreamWriter;
+import java.net.URL;
+import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashSet;
@@ -13,6 +16,7 @@ import java.util.Set;
 import java.util.jar.Attributes;
 import java.util.jar.Manifest;
 
+import org.osgi.framework.dto.BundleDTO;
 import org.osgi.framework.namespace.NativeNamespace;
 import org.osgi.framework.namespace.PackageNamespace;
 import org.osgi.framework.wiring.dto.BundleRevisionDTO;
@@ -22,6 +26,8 @@ import org.osgi.service.repository.ContentNamespace;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.yaml.snakeyaml.Yaml;
+
+import com.google.common.io.ByteStreams;
 
 import aQute.bnd.header.Attrs;
 import aQute.bnd.header.Parameters;
@@ -39,6 +45,7 @@ import aQute.lib.converter.TypeReference;
 import aQute.lib.getopt.Arguments;
 import aQute.lib.getopt.Description;
 import aQute.lib.getopt.Options;
+import aQute.lib.json.JSONCodec;
 import aQute.remote.api.Agent;
 import aQute.remote.api.Event;
 import aQute.remote.api.Supervisor;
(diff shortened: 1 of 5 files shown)


biz.aQute.remote/src/aQute/remote/api/packageinfo
```diff
@@ -1 +1 @@
-version 1.1.0
\ No newline at end of file
+version 2.0.0
```
bjhargrave: Should this really be a version 2 bump?

pkriens: Nope, we can make it a ProviderType. All known implementers are in our build.

amitjoy:
@pkriens I have applied all you requested changes. Please feel free to have another look.

biz.aQute.bnd/src/aQute/bnd/main/RemoteCommand.java
```diff
@@ -109,6 +111,57 @@ public void event(Event e) throws Exception {
...
+	}
+
+	public void _list(ListOptions options) throws Exception {
+		List<BundleDTO> installedBundles = agent.getBundles();
+		installedBundles.forEach(this::prettyPrintBundleDTO);
+	}
+
+	@Description("Communicate with the remote agent to install or update bundle")
+	interface InstallOptions extends Options {
+		@Description("Specify the location identifier of the bundle")
```
pkriens: I do not think the location and file should be options. Just make them the first and second argument?

It is generally also nice if the second argument (the file) can also be a URL

biz.aQute.bnd/src/aQute/bnd/main/RemoteCommand.java
```diff
@@ -109,6 +111,57 @@ public void event(Event e) throws Exception {
...
+				return f;
+			}
+		} catch (Exception e) {
+			error("Error occurred while parsing the specified file");
+		}
+		return null;
+	}
+
+	private void prettyPrintBundleDTO(BundleDTO dto) {
+		bnd.out.println("==================================================");
```
pkriens: this is very verbose since you also use it to print out the bundles in list. In the list command, I think a table is more appropriate with one bundle per line? I think a simple one-line string would suffice. For example in the standard syntax: <bsn>-<version>[<id>]?

In the list version I think it would be nice to see:

* bsn
* version
* location
* startlevel
* last update
* status

biz.aQute.bnd/src/aQute/bnd/main/RemoteCommand.java
```diff
@@ -109,6 +111,57 @@ public void event(Event e) throws Exception {
 		agent = launcher.getAgent();
 	}
 
+	@Description("List the bundles already installed in remote framework")
+	@Arguments(arg = {})
+	interface ListOptions extends Options {
```
pkriens: Could you add an option to get output in JSON?

amitjoy: Done

biz.aQute.bnd/src/aQute/bnd/main/RemoteCommand.java
amitjoy: Done

biz.aQute.bnd/src/aQute/bnd/main/RemoteCommand.java
amitjoy: We cannot do for now since `BundleDTO` doesn't comprise `location`. To include start level, we need to use `BundleStartLevelDTO` too. As discussed earlier, we will proceed without these properties now. Later on, we could try to use `aQute.bnd.runtime.facade.FrameworkFacade.XBundleDTO`.

biz.aQute.bnd/src/aQute/bnd/main/RemoteCommand.java
amitjoy: Done

biz.aQute.bnd/src/aQute/bnd/main/RemoteCommand.java
```diff
@@ -109,6 +116,120 @@ public void event(Event e) throws Exception {
 		agent = launcher.getAgent();
 	}
 
+	@Description("Communicate with the remote framework to list the installed bundles")
+	interface ListOptions extends Options {
+		@Description("Specify the type of output - currently supported option is 'json'")
+		String output();
```
pkriens: In the sake of type safety, this is better a flag: -j/--json. I.e. boolean json()

pkriens:
Just make the json option a flag and I am ok with this

biz.aQute.bnd/src/aQute/bnd/main/RemoteCommand.java
amitjoy: Done

PR #3191, opened 2019-05-15
remote: Use IO class for I/O

bjhargrave:
Also, avoid use of Google classes.


pkriens:
How on earth did we get a google dependency in our repo?

seanbright:
https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.reporter/bnd.bnd

bjhargrave:
> How on earth did we get a google dependency in our repo?

It is used by biz.aQute.bnd.reporter jar and thus the reporting sub-commands in the bnd command. When the bnd jar is built, it includes, among other jars, the biz.aQute.bnd.reporter jar and thus the google jars are on the -buildpath to include the needed packages to make bnd jar standalone.

PR #3210, opened 2019-05-24
[bndrun] Order the bundles of a resolution in dependency order

pkriens:
Although OSGi does not have ordering because you can
dynamically install/uninstall bundles this patch orders
the -runbundles out of a resolution by looking at their dependencies.
If a bundle depends on another bundle then it is tried to start
after it. If there are circular dependencies then of course
this not work well.

To activate:

   -runorder true

If not set we get a more or less random order

### DIFF
--- a/biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java
+++ b/biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java
@@ -5,9 +5,11 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.osgi.resource.Resource;
 import org.osgi.resource.Wire;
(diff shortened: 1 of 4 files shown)


biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java
```diff
@@ -96,7 +98,7 @@ public String resolve(boolean failOnChanges, boolean writeOnChanges) throws Exce
 				if (!projectResolver.isOk()) {
 					return runbundlesFormatter.convert(Collections.<VersionedClause> emptyList());
 				}
-				Set<Resource> resources = resolution.keySet();
+				Collection<Resource> resources = getResources(resolution);
```
bjhargrave: Since the point of this PR is that the resources be ordered, this should be a `List<Resource>`, no?

biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java
```diff
@@ -157,4 +159,54 @@ public int compare(VersionedClause a, VersionedClause b) {
...
+
+			// if we have cycles, break them
+			if (resources.contains(this))
+				return;
+
+			resources.add(this);
+		}
+	}
+
+	public Collection<Resource> getResources(Map<Resource, List<Wire>> resolution) {
```
bjhargrave: Return `List<Resource>` per comment above.

bjhargrave: Perhaps this method should not be `public` to avoid creating API where we don't really need the method to be API?

pkriens:
I am rewriting it, I found a much better way to handle this but I needed to extract the code

PR #3213, opened 2019-05-27
[repository] add FileSetRepository test

rotty3000:


rotty3000:
@bjhargrave you ok with this?

biz.aQute.repository/src/aQute/bnd/repository/fileset/FileSetRepository.java
```diff
@@ -87,11 +92,45 @@ private BridgeRepository getBridge() throws Exception {
...
 				if (!hasIdentity) {
-					return null;
+					String name;
+					String version;
+					Optional<POM> pom;
+
+					// Check if there's a module name
+					if (jar.getModuleName() != null) {
+						name = jar.getModuleName();
+						version = new MavenVersion(jar.getModuleVersion()).getOSGiVersion()
```
bjhargrave: This is a bit of a risk since Module Versions, https://docs.oracle.com/javase/9/docs/api/java/lang/module/ModuleDescriptor.Version.html, do not conform to maven version rules or OSGi version rules. So it is possible to take an exception here. We may need to add a ModuleVersion class to deal with the rules of module versions and how we want to map them to OSGi versions.

biz.aQute.repository/src/aQute/bnd/repository/fileset/FileSetRepository.java
```diff
@@ -87,11 +92,45 @@ private BridgeRepository getBridge() throws Exception {
...
 				Domain manifest = Domain.domain(jar.getManifest());
 				boolean hasIdentity = rb.addManifest(manifest);
 				if (!hasIdentity) {
-					return null;
+					String name;
+					String version;
+					Optional<POM> pom;
+
+					// Check if there's a module name
+					if (jar.getModuleName() != null) {
```
bjhargrave: I also think you should use the version from maven before falling back to any module version. That is, module version is the last hope.

rotty3000: so do you mean to turn this around? maven before JPMS? I'm not sure how you want me to interleave the module info and the maven info here with this comment.

bjhargrave: To get the version of an artifact, use this order: Bundle-Version, maven version, module version.
All bundles have a bundle version, all maven artifacts have a maven version, not all modules have a module version. For module version, you may need to handle that the module version cannot be coerced into a bundle version (perhaps via a maven version).

rotty3000: I updated the logic.

biz.aQute.junit/bnd.bnd
```diff
@@ -19,6 +19,7 @@ Tester-Plugin: aQute.junit.plugin.ProjectTesterImpl
 -includeresource.perm: OSGI-INF/permissions.perm;literal="(java.security.AllPermission)"
 
 Export-Package: \
+	aQute.bnd.rule,\
```
bjhargrave: This should be `aQute.junit.rule`, no?

biz.aQute.repository/bnd.bnd
```diff
@@ -20,8 +20,11 @@
 -testpath: \
 	${junit},\
 	${mockito},\
+	biz.aQute.junit;version=latest;packages=aQute.junit.rule,\
```
bjhargrave: Once you fix the export package above, the `packages` is not necessary since you export the right package.

bjhargrave: But I think this sort of violates the "rule" were _we_ do not depend upon `biz.aQute.junit` in any of our projects. We discussed this on a call with Peter and asserted it was the case the no Bnd project depends upon `biz.aQute.junit` for testing. This PR will change things, so perhaps we should not define such a rule here. Just get the classpath you need in the actual test case.

biz.aQute.repository/src/aQute/bnd/repository/fileset/FileSetRepository.java
```diff
@@ -87,11 +92,49 @@ private BridgeRepository getBridge() throws Exception {
...
+
+					String version = null;
+
+					if (pom.isPresent()) {
+						version = pom.get()
+							.getRevision().version.getOSGiVersion()
+							.toString();
+					}
+					else if (jar.getModuleVersion() != null) {
+						version = new MavenVersion(jar.getModuleVersion()).getOSGiVersion()
```
bjhargrave: The `new MavenVersion` should be in a try/catch since the module version may not conform to maven/OSGi version rules.

biz.aQute.repository/bnd.bnd
rotty3000: right.. and I do remember this conversation. This position however leaves no place to add convenient junit rules to bnd. This particular rule has been laying around my branches for months. I've been moving around as needed in several branches because I can't find a home for it. Furthermore, I'd like to add even more rules.

pkriens: Make a separate project? I also have a runtime project for interesting bundles from the bnd world. A testutil project would be nice as well.

bjhargrave: > I've been moving around as needed in several branches because I can't find a home for it. Furthermore, I'd like to add even more rules.

Why not put the junit.rules package in the aQute.libg project? Then it would be easily accessible for our test cases from the -buildpath/-testpath. We don't need to put the package in a jar.

biz.aQute.repository/bnd.bnd
```diff
@@ -21,7 +21,9 @@
 	${junit},\
 	${mockito},\
 	biz.aQute.http.testservers;version=latest,\
-	slf4j.simple;version=latest
+	slf4j.simple;version=latest,\
+	org.nanohttpd:nanohttpd;version=latest,\
+	javax.annotation:jsr250-api;version=latest
```
bjhargrave: We should not add test data here. See comment on the test case below.

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -1246,4 +1247,15 @@ public void removeSubDirs(String dir) {
 			.keySet());
 		subDirs.forEach(subDir -> removePrefix(subDir + "/"));
 	}
+
+	private static final Predicate<String> pomXmlFilter = new PathSet("META-INF/maven/*/*/pom.xml").matches();
+
+	public Stream<Resource> getPomResources() {
```
bjhargrave: Since this method is now API on Jar, perhaps a more precise method name is `getPomXmlResources`? This is because a jar may also have pom properties resources and the name `getPomResources` is a bit ambiguous in that regard.

biz.aQute.repository/test/aQute/bnd/repository/fileset/FileSetRepositoryTest.java
```diff
@@ -0,0 +1,51 @@
...
+		List<File> files = classpath();
+
+		assertTrue(files.size() > 0);
+
+		FileSetRepository repository = new FileSetRepository("test", files);
+
+		List<String> list = repository.list(null);
+
+		assertTrue(list.contains("biz.aQute.http.testservers"));
+		assertFalse(list.contains("javax.annotation:jsr250-api"));
```
bjhargrave: Can't we just use jars from the `testdata` or `testresources` folders rather than add some to the testpath just for this test method? For example, use `testresources/mavenrepo/commons-cli/commons-cli/1.0/commons-cli-1.0.jar` as a jar without a BSN or pom xml resource. It seems contrived for the purposes of testing FileSetRepository that we must use the actual class path of the test case. It is too risky should we change the build later. The testpath should be about test execution and not testdata.

rotty3000: I really hate adding binaries for test data.

rotty3000: .. but fine.

bjhargrave: The alternate is to create a MavenBndRepository object, like we do in some tests, and have it download the binaries you need for testing. But adding jars to testpath when the test cases do not need to compile or run against them is wrong.

PR #3218, opened 2019-05-29
[maven] make it possible to omit the -runee and infer it from maven-compiler-plugin/configuration/target

### LINKED ISSUE #2783: [m2e integration] compute -runee in m2e projects from maven-compiler-plugin/configuration/target
When not explicitly set, it would be great if execution environment could be derived from settings of the maven project.

rotty3000:
fixes #2783

### DIFF
--- a/biz.aQute.bnd.maven/bnd.bnd
+++ b/biz.aQute.bnd.maven/bnd.bnd
@@ -23,6 +23,7 @@ Import-Package: \
 	org.apache.maven:maven-plugin-api,\
 	org.apache.maven:maven-repository-metadata,\
 	org.apache.maven:maven-settings,\
+	org.codehaus.plexus:plexus-utils,\
 	org.eclipse.aether.api,\
 	org.slf4j.api;version=latest
 
--- a/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
+++ b/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
@@ -6,20 +6,25 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Dependency;
+import org.apache.maven.model.Plugin;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.project.ProjectDependenciesResolver;
+import org.codehaus.plexus.util.xml.Xpp3Dom;
 import org.eclipse.aether.RepositorySystem;
 import org.eclipse.aether.RepositorySystemSession;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import aQute.bnd.build.Workspace;
+import aQute.bnd.build.model.EE;
 import aQute.bnd.maven.lib.configuration.BeanProperties;
+import aQute.bnd.osgi.Constants;
 import aQute.bnd.osgi.Processor;
 import aQute.bnd.repository.fileset.FileSetRepository;
 import aQute.bnd.service.RepositoryPlugin;
@@ -182,6 +187,7 @@ public class BndrunContainer {
 			for (RepositoryPlugin repo : workspace.getRepositories()) {
 				repo.list(null);
 			}
+			setEEfromBuild(run);
 			run.getInfo(workspace);
 			int errors = report(run);
 			if (!run.isOk()) {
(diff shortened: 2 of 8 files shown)


biz.aQute.bnd.maven/bnd.bnd
```diff
@@ -24,7 +24,8 @@ Import-Package: \
 	org.apache.maven:maven-repository-metadata,\
 	org.apache.maven:maven-settings,\
 	org.eclipse.aether.api,\
-	org.slf4j.api;version=latest
+	org.slf4j.api;version=latest,\
+	org.codehaus.plexus:plexus-utils
```
bjhargrave: I would put thus up near the other maven deps.

biz.aQute.bndlib/src/aQute/bnd/build/model/EE.java
```diff
@@ -99,6 +101,15 @@ public Version getCapabilityVersion() {
 		return capabilityVersion;
 	}
 
+	public static EE highestFromVersion(String versionString) {
```
bjhargrave: I would suggest this API return `Optional<EE>`. This lets the caller decide how to deal with optionality vs. returning null.

bjhargrave: The _version_ referred to here is unclear. It is the compiler _target version_. So perhaps a name of `highestFromTargetVersion` is a more descriptive name?

biz.aQute.bndlib/src/aQute/bnd/build/model/EE.java
```diff
@@ -99,6 +101,15 @@ public Version getCapabilityVersion() {
...
+		Version version = new Version(versionString);
+		return Arrays.stream(values())
+			.sorted(Collections.reverseOrder())
+			.filter(ee -> ee.capabilityVersion.compareTo(version) == 0)
```
bjhargrave: You should probably filter first before you sort as the filter may reduce to a single element which is easy to sort :-)

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
```diff
@@ -209,6 +215,20 @@ public FileSetRepository getFileSetRepository() throws Exception {
...
 	}
 
+	private void getEEfromBuild(Bndrun run) {
+		String runee = run.getProperty(Constants.RUNEE);
+
+		if (runee == null) {
+			Optional.ofNullable(project.getPlugin("org.apache.maven.plugins:maven-compiler-plugin"))
+				.map(Plugin::getConfiguration)
+				.map(Xpp3Dom.class::cast)
+				.map(dom -> dom.getChild("target"))
```
bjhargrave: I assume the target element is set to the actual value event when using maven properties or letting things default to the current JDK being run?

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
```diff
@@ -209,6 +215,20 @@ public FileSetRepository getFileSetRepository() throws Exception {
 		return fileSetRepository = dependencyResolver.getFileSetRepository(name, bundles, useMavenDependencies);
 	}
 
+	private void getEEfromBuild(Bndrun run) {
```
bjhargrave: `setEEfromBuild` may be more descriptive a name.

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
rotty3000: yes, that's why I get it from here rather than the property, it may be set either way, but this holds the final result.

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
```diff
@@ -209,6 +215,21 @@ public FileSetRepository getFileSetRepository() throws Exception {
...
+	private void setEEfromBuild(Bndrun run) {
+		String runee = run.getProperty(Constants.RUNEE);
+
+		if (runee == null) {
+			Optional.ofNullable(project.getPlugin("org.apache.maven.plugins:maven-compiler-plugin"))
+				.map(Plugin::getConfiguration)
+				.map(Xpp3Dom.class::cast)
+				.map(dom -> dom.getChild("target"))
+				.map(Xpp3Dom::getValue)
+				.map(EE::highestFromTargetVersion)
```
bjhargrave: This needs to be a `flatMap` now since the method returns an Optional. Then you don't need the `.map(Optional::get)` :-)

bjhargrave: Did your recent push force intend to address the `flatMap` comment above?

PR #3223, opened 2019-06-04
Pkriens/ordering

pkriens:


.gitignore
```diff
@@ -4,7 +4,6 @@
 Archive
 .DS_Store
 .*.swp
-buildfiles
```
bjhargrave: This seems wrong. I know you may need to commit a buildfiles for a test case data. But in general we need to have this here.

pkriens: Why? They are always in the target/generated directory and already ignored?

bjhargrave: True. It was added long before generated was added.

PR #3235, opened 2019-06-11
[jpms] change title

rotty3000:
The menu doesn't look nice with such a long title.


bjhargrave:
The failed travis build is actually fine. So merge when ready.

PR #3241, opened 2019-06-12
[spi] avoid duplicating providers in descriptor files

rotty3000:


rotty3000:
@bjhargrave this solves the serviceloader CT test failures

biz.aQute.bndlib/src/aQute/bnd/plugin/spi/SPIDescriptorGenerator.java
```diff
@@ -56,18 +58,40 @@ public void verify(Analyzer analyzer) throws Exception {
...
+						}
+						list.remove(line);
+					}
+				}
 			}
 
-			analyzer.getJar()
-				.putResource(key, new EmbeddedResource(value, analyzer.lastModified()));
+			// Only write if we have something to add
+			if (!list.isEmpty()) {
```
bjhargrave: Shouldn't the BufferedReader loop also be in this if? If there is nothing to add, why do anything here?

rotty3000: The Map.Entry won't exist if there's nothing in this list (since an entry is only added when there is a list to add).

Later on, however, we may have removed all entries from the list if they already existed in the descriptor.

biz.aQute.bndlib/src/aQute/bnd/plugin/spi/SPIDescriptorGenerator.java
```diff
@@ -56,18 +58,40 @@ public void verify(Analyzer analyzer) throws Exception {
...
+				list.add(0, "# Generated by bnd (modified: " + new Date(analyzer.lastModified()) + ")");
```
bjhargrave: Don't write the date in here. It breaks reproducibility.

biz.aQute.bndlib/src/aQute/bnd/plugin/spi/SPIDescriptorGenerator.java
rotty3000: does this explanation make sense?

rotty3000:
updated!

Issue #3209, opened 2019-05-24
Gradle parallel builds and Analyzer Plugins


pkriens:
I suspect that a DS bug that caused random version assignments is a concurrency bug. Now bnd is not ready to run parallel but the client indicated they run it in parallel. Need to discuss how we protect against this

ViToni:
Might be related to #2801

bjhargrave:
Bnd plugins are singleton objects in the Workspace. They need to be careful to avoid instance variables which change based upon the project being built. 

We need to spend some time to analyze the plugins for this and fix.

bjhargrave:
MetaypeAnnotations needs to be fixed.

bjhargrave:
I looked through the Analyzer and Verifier plugins and only found that MetatypeAnnotations needed fixing.

bjhargrave:
The DS annotations processing was fixed by https://github.com/bndtools/bnd/pull/3208.

I think we are done with this issue, so I will close.

PR #3248, opened 2019-06-13
metatype: Avoid instance variable for parallelism

bjhargrave:
Since the plugin instance is a singleton, we must avoid instance
variables that mutate for each invocation.

See #3209


rotty3000:
ah, thanks for taking this one of my todo list

PR #3253, opened 2019-06-14
header: Change duplicate key handling code to avoid O(N^2) containsKey calls

bjhargrave:
If a key is used N times in a header, the code to deduplicate the key
name used O(N^2) containsKey calls to find a deduplicated name. When
large headers are parsed, this can consume considerable time. One
scenario called containsKey 265M times! This fix uses a separate map to
hold the last used deduplicated key name so the next deduplicated key
name can be computed and confirmed available in a single containsKey
call.


pkriens:
If you called this 256M times then it sounds more someone higher is calling the same code too often.  It is hard to see in what use case you would need to call this that often? Except for capabilities and requirements duplication is rare?

PR #3254, opened 2019-06-14
[resource] use hashset instead of linkedlist

rotty3000:


rotty3000:
@bjhargrave I'm not really sure if this is the correct approach, but it is way faster...

rotty3000:
nah, it's not right...

pkriens:
This is a bad idea since you get random order.

PR #3257, opened 2019-06-17
[launcher] Moved to 4.2 and changed startlevel handling

pkriens:
My customer does not need 4.2 anymore so moved the launcher 
to 5.0. Primarily driven by the ability to use the
new startlevel facilities. It was kind of inconvenient
to only be able to change start levels after the
framework was launched.

### DIFF
--- a/cnf/ext/central.mvn
+++ b/cnf/ext/central.mvn
@@ -1,5 +1,4 @@
 org.osgi:osgi.annotation:7.0.0
-org.osgi:org.osgi.core:4.2.0
 org.osgi:osgi.core:6.0.0
 org.osgi:org.osgi.compendium:4.2.0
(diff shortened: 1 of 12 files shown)


cnf/ext/central.mvn
```diff
@@ -1,5 +1,4 @@
 org.osgi:osgi.annotation:7.0.0
-org.osgi:org.osgi.core:4.2.0
 org.osgi:osgi.core:6.0.0
 org.osgi:org.osgi.compendium:4.2.0
```
bjhargrave: Can we also remove compendium 4.2?

pkriens: guess so

pkriens: will do

PR #3258, opened 2019-06-17
[startlevel] Start level for resolve ordering support

pkriens:
- Changed the instruction from -runorder to -runstartlevel
- Order bundles now also with a start level attribute
- Support a begin and step for the auto numbering in resolve
- Added tests
- Provide documentation

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/help/instructions/ResolutionInstructions.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/help/instructions/ResolutionInstructions.java
@@ -1,7 +1,5 @@
 package aQute.bnd.help.instructions;
 
-import java.util.Optional;
-
 import aQute.bnd.help.SyntaxAnnotation;
 
 /**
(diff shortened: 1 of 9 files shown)


biz.aQute.bndlib/src/aQute/bnd/help/instructions/ResolutionInstructions.java
```diff
@@ -34,18 +32,18 @@
...
-	};
-
-		public boolean isMerge() {
-			return false;
-		}
+	MERGESORTBYNAMEVERSION;
+	}
+
+	@interface RunStartLevel {
+		Runorder order() default Runorder.MERGESORTBYNAMEVERSION;
```
bjhargrave: Looks like you still have a `Runorder` type around even after renaming the instruction to runstartlevel. Perhaps Runorder is a subtype of Runstartlevel? Not sure of proper naming here.

pkriens: It is the enum for the run order.

PR #3259, opened 2019-06-17
bndlib: Use DUPLICATE_MARKER constant

bjhargrave:
Use the defined constant rather than local string constants.


pkriens:
Doesn't this drag in aQute.bnd.osgi? The idea to use the string was to NOT let that happen!

bjhargrave:
It references a static final char from Constants. So that will not drag in the package. I specifically avoided the use of a method on Processor.

pkriens:
ok, I trust you ....

bjhargrave:
BTW, the desired coupling/noncoupling of packages in bndlib is not documented anywhere. I think it exists only your mind :-/

pkriens:
it would be nice to do some splitting in different projects ... although now the maven stuff depends on the workspace it is probably moot

PR #3260, opened 2019-06-17
resource: Recompute ResourceImpl.hashCode/equals after capabilities set

bjhargrave:
If hashCode/equals was called on a ResourceImpl before the capabilities
were set, the empty location information would be cached.


rotty3000:
This worked well.

PR #3261, opened 2019-06-18
[resource] Use LinkedHashMap instead of LinkedList

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceImpl.java
```diff
@@ -51,7 +51,8 @@ void setCapabilities(List<Capability> capabilities) {
 
 	@Override
 	public List<Capability> getCapabilities(String namespace) {
```
pkriens: We seem to basically reimplement MultiMap? Shouldn't we use that one and do any optimizations there?

bjhargrave: Maybe a separate PR? But best not to conflate with this PR.

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java
```diff
@@ -124,17 +123,17 @@ public Resource build() {
 			throw new IllegalStateException("Resource already built");
 		built = true;
 
-		resource.setCapabilities(capabilities);
-		resource.setRequirements(requirements);
+		resource.setCapabilities(getCapabilities());
```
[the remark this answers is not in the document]
bjhargrave: Yes, I would just change the package private set methods in ResourceImpl to take the value `capabilities.values()`/`requirements.values()`.

pkriens:
LGTM

PR #3278, opened 2019-06-25
[tester] Use static import rather than inheritance for constants

kriegfrj:
Update from pre JDK 1.5 pattern for constants.


bjhargrave:
Thanks

PR #3284, opened 2019-06-25
do not quote unstructured bundle headers

### LINKED ISSUE #3282: bnd-maven-plugin: Quoted string used on BundleDescription header values
Although quoted strings are valid according to https://osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.common.header.syntax and https://osgi.org/specification/osgi.core/7.0.0/ch01.html#framework.general.syntax, they seem not to be unquoted for e.g. the `BundleDescription` values. Once I have a value like

```
BundleDescription: "Some Description"
```

the value will be exposed in the e.g. Felix Web Console with the quotes `"Some Description"` instead of just the value `Some Description`.

kwin:
This fixes #3282


bjhargrave:
Thanks

PR #3290, opened 2019-06-26
Fix small typo

kwin:


bjhargrave:

bjhargrave:
Thanks!

PR #3293, opened 2019-06-26
Bndrun refresh

bjhargrave:
Change ProjectLauncher.update to call getProject().refresh() and make FileSetRepository implement Refreshable. Also cleaned up code in FileSetRepository.


pkriens:
Did you miss OSGiRunLaunchDelegate for a reason? There is also a refresh + update call in there?

pkriens:
Overall looks good to me.

bjhargrave:
> Did you miss OSGiRunLaunchDelegate for a reason? There is also a refresh + update call in there?

That code:

https://github.com/bndtools/bnd/blob/ff43712bc034e339f3aa17e5966a4db8d8b7d3de/bndtools.core/src/bndtools/launch/OSGiRunLaunchDelegate.java#L312-L314

did not call Processor.refresh(). It called other methods which do slightly different things. So I was afraid to remove these additional calls as I was not sure of the effect.

pkriens:
Think we should also remove them and then see what happens

pkriens:
It sucks we have no bndtools tests :-(

bjhargrave:
> Think we should also remove them and then see what happens

Done in https://github.com/bndtools/bnd/pull/3297.

kriegfrj:
> Think we should also remove them and then see what happens

I think this may have caused #3360.

PR #3299, opened 2019-07-01
ci: Run all gradle jobs --parallel

bjhargrave:
Parallel gradle builds seem to be very stable. So this moves us to the next level by running all CI gradle builds in parallel mode. This will now include a parallel build on Windows.


pkriens:
I've never seen you this brave :-)

PR #3309, opened 2019-07-04
Launcher fixes and getJavaExecutable improvements

bjhargrave:


biz.aQute.bndlib/src/aQute/bnd/build/ProjectLauncher.java
```diff
@@ -143,13 +110,57 @@ public ProjectLauncher(Project project) throws Exception {
...
+		setCwd(getProject().getBase());
 
 		// pkr: could not use this because this is killing the runtests.
-		// project.refresh();
+		// getProject().refresh();
 		runbundles.clear();
 		startlevels.clear();
-		Collection<Container> run = project.getRunbundles();
+
+		getProject().getBundles(Strategy.HIGHEST, Constants.DEFAULT_LAUNCHER_BSN, null)
```
pkriens: This must go to the ProjectLauncherImpl since it is only valid there. This Project Launcher is abstract. I also think it needs its own method and use other methods. I actually worked on the same thing :-(

pkriens:
Nice work, I'd done 80% of the same things but not yet committed.

What is left imho is that the ProjectLauncher became dependent on ProjectLauncherImpl details. I.e. it knows the name of the Launcher class and the pre jar stuff is an implementation detail in the ProjectLauncherImpl, it is not abstract behavior. I.e. it is of no use for the remote launcher.  This breaks the distinction between the abstract launcher and the implementation in biz.aQute.launcher as was described in the [Launcher Architecture](https://bnd.bndtools.org/chapters/300-launching.html#launcher-architecture)

For the biz.aQute.launcher project, I think we need a separate starting class for the Java 9 case which uses the -Dlauncher.runpath and the embedded case. It is currently confusing.

I'll work on that unless you want to do it.

biz.aQute.bndlib/src/aQute/bnd/build/ProjectLauncher.java
bjhargrave: For the record, and as we discussed, this was fixed by https://github.com/bndtools/bnd/pull/3318!

PR #3331, opened 2019-07-12
[dates/formatter] Made formatter more complete and better parsing

pkriens:
The code did not handle a number of the more 
complex formatter rules. Added:

- Indexing
- Last 
- Handles now all standardized formats and many more for dates
- Figured out how to use the Temporal Adjuster class :-(


aQute.libg/src/aQute/lib/date/Dates.java
```diff
@@ -0,0 +1,133 @@
...
+	public static ZonedDateTime parse(String dateString) {
+
+		int n = 0;
+		for (DateTimeFormatter df : DATE_TIME_FORMATTERS) {
+
+			try {
+				return toZonedDateTime(df.parse(dateString));
+			} catch (DateTimeParseException dte) {
+				// we ignore wrong formats
+				n++;
```
bjhargrave: What does the `n` variable do here?

aQute.libg/src/aQute/lib/formatter/Formatters.java
```diff
@@ -0,0 +1,150 @@
...
+import java.util.Locale;
+import java.util.function.Function;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import aQute.lib.date.Dates;
+
+public class Formatters {
+	private final static Pattern PRINTF_P = Pattern.compile(
+		"%(?:(?<index>\\d+)\\$)?(?<previous><)?(-|\\+|0|\\(|,|\\^|#| )*(\\d*)?(?:\\.(\\d+))?(?<conversion>a|A|b|B|c|C|d|e|E|f|g|G|h|H|n|o|s|S|x|X|(?:[tT][HIklMSLNpzZsQBbhAaCYyjmdeRTrDFc])|%)");
```
bjhargrave: Argument index can be `<` as an alternate to `1$`. The regex implies you can say both, e.g. `1$<`, which is not the case.

`(?:(?<index>\\d+)\\$)?(?<previous><)?`

I think this should be 

`(?:(?:(?<index>\\d+)\\$)|(?<previous><))?`

```
Argument Index

The argument index is a decimal integer indicating the position of the argument in the argument list. The first argument is referenced by "1$", the second by "2$", etc.

Another way to reference arguments by position is to use the '<' ('\u003c') flag, which causes the argument for the previous format specifier to be re-used. For example, the following two statements would produce identical strings:

       Calendar c = ...;
       String s1 = String.format("Duke's Birthday: %1$tm %1$te,%1$tY", c);

       String s2 = String.format("Duke's Birthday: %1$tm %<te,%<tY", c);
 ```

File: biz.aQute.bndlib.tests/test/test/MacroTest.java [hunk 3]
```diff
@@ -1604,20 +1603,40 @@ public void testNonStringFlattenedValue() throws Exception {
...
+				.process("${format;%2$tY/%2$tm %<tH %2$tZ;X;1970-01-01T00:00:00+08:00}"));
+
+			assertEquals("UTC", processor.getReplacer()
+				.process("${format;%TZ;190704}"));
 
 			assertEquals("201907", processor.getReplacer()
-				.process("${format;%TY%tm;190704;190704}"));
+				.process("${format;%TY%<tm;20190704}"));
+			assertEquals("2019", processor.getReplacer()
+				.process("${format;%tY;${now}}"));
```
bjhargrave: This seems like a time bomb that will start failing next year!

aQute.libg/src/aQute/lib/date/Dates.java
pkriens: Diagnostic when I need it

aQute.libg/src/aQute/lib/formatter/Formatters.java
pkriens: That will get caught on the actual formatting and it makes the regex even more complicated than it already is.

pkriens: a) it does not matter since the error will be found by the real formatter, we do not need fidelity. We just need to be able to parse anything that is legal for the formatter but it is not a problem if we parse wrong strings.
b) you do not need a group 2
c) you do not need ?: when you use named groups

PR #3334, opened 2019-07-15
[reporter] Simplify configuration of reports when generating readme files

### LINKED ISSUE #3321: [reporter] Simplify the report configuration for specific use cases.
A user can configure their own reports using the `-exportreport` and `-reportconfig` instructions and export them with the Bnd CLI `exportreport` command or the maven `export` goal. 

While this is fine when the user wants a custom report, it is not when Bnd can provide a end-to-end report (data -> template -> outputFile). The idea is to create additional Bnd CLI `exportreport` sub-commands and maven goals for these end-to-end reports.

The first use case we have is to generate readme files. So we can add two sub-commands  `exportreport readme` and `exportreport jarreadme` and one maven goal `reporter:readme`. Then, the user will not have to configure the reports in their `*.bnd` and `pom.xml` files and could provide optional template parameters directly from the command line.

cdelg:
Closes #3321 

* Add Bnd CLI sub-commands to generate readme files.
* Add a new Mojo in the reporter maven plugin to generate readme files.

With this PR users do not need to configure their reports if they need readme files. They just call `bnd exportreport readme` or `mvn bnd-reporter:readme`.

User can provide template parameters with system properties starting with `bnd.reporter.*` (eg; `-Dbnd.reporter.show.confsnippet.configadmin=true` )


cdelg:
> There were failing tests. See the report at: file:///home/travis/build/bndtools/bnd/biz.aQute.launchpad/generated/reports/tests/test/index.html

Maybe this PR's build should be restarted.

stbischof:
@cdelg I'm not sure that system propetries are the best way to control the templates. I think we should add some configs or parameter list to the readmeMojo.

cdelg:
> I'm not sure that system propetries are the best way to control the templates. I think we should add some configs or parameter list to the readmeMojo.

* So the user will have to configure the report in their pom.xml, is it possible then to override the configuration from the command line?
* Your comment are only for the Mojo or also for the Bnd sub-command?

cdelg:
@stbischof I have added parameters to the ReadmeMojo but I have also kept the system properties, is it fine or it can lead to issues?

maven/bnd-reporter-maven-plugin/src/main/java/aQute/bnd/maven/reporter/plugin/ReadmeMojo.java
```diff
@@ -0,0 +1,153 @@
...
+		}
+	}
+
+	private String getParameters() {
+		Map<String, String> params = new HashMap<>(parameters);
+
+		System.getProperties()
+			.stringPropertyNames()
+			.stream()
+			.filter(k -> k.startsWith("reporter."))
```
stbischof: May use "bnd.reporter." like at the "bnd.reporter.skip" parameter

stbischof:
Im fine with the parameter solution in the mojo.

maven/bnd-reporter-maven-plugin/src/main/java/aQute/bnd/maven/reporter/plugin/ReadmeMojo.java
cdelg: Done

cdelg:
Done, this PR can be reviewed.

biz.aQute.bnd.reporter/src/biz/aQute/bnd/reporter/exporter/ReportExporter.java
```diff
@@ -68,7 +68,7 @@
 		final Map<String, ExportReportInstruction> instructions = getScopedExportInstruction();
 		instructions.entrySet()
 			.forEach(config -> {
-				final File destination = IO.getFile(_processor.getBase(), config.getKey());
+				final File destination = IO.getFile(_processor.getBase(), cleanKey(config.getKey()));
```
bjhargrave: I think you are looking for `Processor.removeDuplicateMarker(config.getKey())`. Which means you do not need the `cleanKey` method.

cdelg: Thanks!

PR #3336, opened 2019-07-16
WIP: [report] gogo plugin produces gogo-command info of the bundle

stbischof:
wip
A reporter-plugin that collects gogo-command information from components an Produces a DTO wich can be used by the ganerator to document all gogo-commands of the bundle.

Problems:
@bjhargrave 
I have som problems to read the Method-Parameters from a MethodDef.
There are 2 nearly the same test, that tries to get the MethodParameters from a MethodDef.
* biz.aQute.bndlib.tests/test/aQute/bnd/osgi/ClazzTest.java -> **works**
* biz.aQute.bnd.reporter/test/biz/aQute/bnd/reporter/plugins/entries/bundle/GoGoClazzTest.java -> **fails**

*Inside the biz.aQute.bnd.reporter project* I did not any get MethodParameters


bjhargrave:
> I have som problems to read the Method-Parameters from a MethodDef.
> There are 2 nearly the same test, that tries to get the MethodParameters from a MethodDef.
> 
>     * biz.aQute.bndlib.tests/test/aQute/bnd/osgi/ClazzTest.java -> **works**
> 
>     * biz.aQute.bnd.reporter/test/biz/aQute/bnd/reporter/plugins/entries/bundle/GoGoClazzTest.java -> **fails**
> 
> 
> _Inside the biz.aQute.bnd.reporter project_ I did not any get MethodParameters

This is because we do not compile all the projects with the java compiler argument to generate the MethodParameters attribute. It is an optional attribute which is only generated when requested. See https://docs.oracle.com/javase/tutorial/reflect/member/methodparameterreflection.html.

In the Bnd build, we only generate the MethodParameters attribute for the test classes in biz.aQute.bndlib.tests.

https://github.com/bndtools/bnd/blob/8edde856f651540b0adb1aacffab2a44efc70300/biz.aQute.bndlib.tests/build.gradle#L1-L3

https://github.com/bndtools/bnd/blob/8edde856f651540b0adb1aacffab2a44efc70300/biz.aQute.bndlib.tests/.settings/org.eclipse.jdt.core.prefs#L32

So it is fine to turn it on for test code since we don't ship the test code. But we do not want to turn it on for non-test code since we ship the class files.

So, in general, you cannot expect class files to have MethodParameter attributes, since, by default, the compilers do not generate MethodParameter attributes in class files. So you code must not rely upon MethodParameter attributes to be present in order to function properly. For example, the DS annotation processing code has a fallback when the classes being processed https://github.com/bndtools/bnd/blob/8edde856f651540b0adb1aacffab2a44efc70300/biz.aQute.bndlib/src/aQute/bnd/component/DSAnnotationReader.java#L1081-L1089

stbischof:
@bjhargrave 
Thank you for the fast and good answer!
Should I remove the biz.aQute.bndlib.tests/test/aQute/bnd/osgi/ClazzTest.java?

@cdelg Could you have a look at this PR? Is that the way you expected a report-plugin?

bjhargrave:
> Should I remove the biz.aQute.bndlib.tests/test/aQute/bnd/osgi/ClazzTest.java?

No! It is testing that the Clazz parser can parse out the MethodParamter attribute properly. This is why the build turns on MethodParameter attribute generation for that project.

stbischof:
I close this and merge it together with #3338

PR #3337, opened 2019-07-16
[launchpad] Added timeout of 1 minute

pkriens:

### DIFF
--- a/biz.aQute.launchpad/test/aQute/launchpad/LaunchpadTest.java
+++ b/biz.aQute.launchpad/test/aQute/launchpad/LaunchpadTest.java
@@ -464,7 +464,10 @@ public class LaunchpadTest {
 					}
 				});
 		}
-		semaphore.acquire(n);
+		System.out.println("Waiting for the threads to finish");
+		boolean success = semaphore.tryAcquire(n, 60, TimeUnit.SECONDS);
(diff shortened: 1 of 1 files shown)


biz.aQute.launchpad/test/aQute/launchpad/LaunchpadTest.java
```diff
@@ -464,7 +464,9 @@ public void testLaunchpadStressByCreatingLotsOfFrameworksInDifferentThreads() th
 					}
 				});
 		}
-		semaphore.acquire(n);
+		System.out.println("Waiting for the threads to finish");
+		boolean success = semaphore.tryAcquire(n, 60, TimeUnit.SECONDS);
+		assertTrue(success);
```
bjhargrave: I think you want to softly assert this so that you still close the frameworks.

pkriens: Well, the test is theoretically not supposed to be able to timeout, if the time out happens you're in deeper shit than a missed resource?

But the softly asserts sounds nice

bjhargrave: Theoretically, hopefully, none of the tests should fail... :-)

PR #3340, opened 2019-07-17
[remote] Error in API & null location

pkriens:
- API used the install method name twice, the link only supports one, last added changed to installWithData
- Added tests
- Location can now be null, will search for existing location based on BSN
  an exception is thrown when there are 2 bundles with the same bsn installed.
- Location specified via option
- Allow multiple files on the command line

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Domain.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Domain.java
@@ -39,6 +39,8 @@ import aQute.bnd.header.Parameters;
 import aQute.bnd.maven.PomParser;
 import aQute.bnd.version.Version;
 import aQute.lib.converter.Converter;
+import aQute.lib.exceptions.Exceptions;
+import aQute.lib.io.ByteBufferInputStream;
 import aQute.lib.io.IO;
 import aQute.lib.utf8properties.UTF8Properties;
 import aQute.service.reporter.Reporter;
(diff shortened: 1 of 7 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Domain.java
```diff
@@ -448,6 +450,36 @@ public Parameters getProvideCapability() {
 		return getParameters(Constants.PROVIDE_CAPABILITY, true);
 	}
 
+	public static Domain domain(byte[] data) {
+		try (JarInputStream jin = new JarInputStream(new ByteArrayInputStream(data))) {
```
bjhargrave: Some time past, we replaced all usage of ByteArrayInputStream with ByteBufferInputStream to avoid the unnecessary synchronization. Can you use ByteBufferInputStream here instead?

biz.aQute.remote/bnd.bnd
```diff
@@ -8,12 +8,16 @@ aQute.agent.server.port = 29998
 
 -buildpath: \
 	osgi.core;version=latest;maven-scope=provided,\
+    biz.aQute.bndlib;version=latest,\
```
bjhargrave: Is it wise to move bndlib ahead of aQute.libg? This would use the libg from bndlib instead of directly from the project. All the other projects put libg before bndlib.

biz.aQute.remote/src/aQute/remote/agent/AgentServer.java
```diff
@@ -54,6 +65,7 @@
  * interfaces and communicates with a Supervisor interfaces.
  */
 public class AgentServer implements Agent, Closeable, FrameworkListener {
+	Pattern													BSN_P				= Pattern.compile("\\s*([^;\\s]+).*");
```
bjhargrave: This should be `static final`, no?

biz.aQute.remote/src/aQute/remote/agent/AgentServer.java
```diff
@@ -747,4 +771,57 @@ private CapabilityDTO toDTO(int resource, Capability r) {
 		return rd;
 	}
 
+	private Entry<String, Version> getIdentity(byte[] data) throws IOException {
+		try (JarInputStream jin = new JarInputStream(new ByteArrayInputStream(data))) {
```
bjhargrave: Use ByteBufferInputStream here also.

bjhargrave:
Will this PR allow us to remove `-nojunit: true` from the `biz.aQute.remote/bnd.bnd` file?

Issue #2940, opened 2019-02-08
m2e index generator: IDE mostly busy 


maggu2810:
I migrated a PDE project that uses Maven Tycho for headless build to Maven + bnd plugins + Eclipse IDE bnd's m2e plugins.

Some of you may know the project: Eclipse SmartHome respective openHAB Core.

It is similar to (inspired by) "https://enroute.osgi.org/"

The headless command line based Maven build works as expected.
The OSGi JUnit tests using bndrun configurations works on command line and in the IDE.
Also an application is working.

But there is one thing that drives me crazy...

It's the index generation inside the IDE.

The indexes are generated fine using a Maven build.

The index generation in the IDE is triggered very often, it takes a long time and it seems to trigger itself.

I had to disable the index configuration (I modified the lifecycle mapping in the parent POM) to work with the IDE.

But it complicates the whole setup.
If the indexes are generated separate by triggering the respective goal using Maven without building the whole reactor, the artifacts are taken from the local repository...

Could you perhaps have a look at, what I am doing wrong?

I am using three different index projects:
* runtime-index
  The runtime index contains the external dependencies that are necessary to run stuff inside an OSGi runtime.
* test-index
  The test index is about external dependencies that are necessary to run OSGi based tests (named integration tests in that project)
* openhab-core-index
  An index that contains all bundles of that reactor itself (not the test ones, but the ones that will be used in some runtime)

Using a separate "openhab-core-index" allows me to use one for every test (+ the one that is generated for the test bundle itself) and for every possible application subset.

Perhaps you can checkout the repo and have a look at how it works with and without the explicit lifecycle mapping. This would be great!

There is a Oomph setup file that will provide you the IDE setup that is used for normal development.
See: https://raw.githubusercontent.com/maggu2810/openhab-core.bnd-playground/master/tools/oomph/openHAB-Core.setup

rotty3000:
Hey @maggu2810 I am interested in taking a closer look at this but it will take me some time to deeply investigate. Significant work has been done both on the maven plugins and the bndtools m2e integration in the 4.2.0 development phase (we're ramping down for a release soon).

I suspect that it will be about 2 weeks before I can dive in but I think that we can probably address some of these issues simply be using the new features.

In the meantime perhaps you would like to investigate the features of 4.2.0-SNAPSHOT which include:
- the maven plugins and the m2e integration have better cross-support for multiple executions
[...]

PR #3343, opened 2019-07-18
Slow indexing

timothyjward:
I encountered a workspace that appears to trigger slower indexing

This may help with issue #2940

### DIFF
--- a/bndtools.core/bndtools.shared.bndrun
+++ b/bndtools.core/bndtools.shared.bndrun
@@ -88,16 +88,22 @@ m2e.bundles:\
 
 m2e.debug.bundles:\
 	com.google.guava;version='[21,22)',\
-	org.apache.xerces,\
+	javax.xml,\
+	org.apache.xerces;version='[2.9.0,3.0.0)',\
 	org.apache.xml.resolver,\
 	org.apache.xml.serializer,\
 	org.eclipse.core.externaltools,\
 	org.eclipse.emf.ecore.edit,\
 	org.eclipse.emf.edit,\
+	org.eclipse.equinox.p2.discovery,\
+	org.eclipse.equinox.p2.discovery.compatibility,\
+	org.eclipse.equinox.p2.transport.ecf,\
+	org.eclipse.equinox.p2.ui.discovery,\
 	org.eclipse.jem.util,\
 	org.eclipse.m2e.archetype.common,\
 	org.eclipse.m2e.core,\
 	org.eclipse.m2e.core.ui,\
+	org.eclipse.m2e.discovery,\
 	org.eclipse.m2e.editor,\
 	org.eclipse.m2e.editor.xml,\
 	org.eclipse.m2e.jdt,\
@@ -115,6 +121,7 @@ m2e.debug.bundles:\
 	org.eclipse.wst.common.core,\
 	org.eclipse.wst.common.ui,\
 	org.eclipse.wst.common.emf,\
+	org.eclipse.wst.common.emfworkbench.integration,\
 	org.eclipse.wst.common.environment,\
 	org.eclipse.wst.common.frameworks,\
 	org.eclipse.wst.common.frameworks.ui,\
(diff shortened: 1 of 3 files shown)


timothyjward:
Overall performance increase was from ~55 seconds to ~ 40 seconds for a full clean/build cycle. It would be good to know if a similar (or better) effect is seen by users who have reported even slower building.

bndtools.m2e/src/bndtools/m2e/IndexConfigurator.java
```diff
@@ -176,12 +178,31 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
...
-					facade.getProject()
-						.build(FULL_BUILD, buildMonitor);
+							list.size());
+						list.forEach(me -> {
+							try {
+								MavenPlugin.getMaven()
+									.execute(getMavenProject(facade, indexMonitor), me, indexMonitor);
+							} catch (CoreException e) {
+								// TODO Auto-generated catch block
+								e.printStackTrace();
```
bjhargrave: This should be logged to the eclipse logger.

cnf/ext/repositories.bnd
```diff
@@ -48,15 +48,15 @@ mavencentral:           https://repo.maven.apache.org/maven2
...
-#        url="https://download.eclipse.org/modeling/emf/emf/builds/release/2.13";\
-#        location="${workspace}/cnf/cache/stable/EclipseEMF",\
-#    aQute.bnd.repository.p2.provider.P2Repository; \
-#        name="Eclipse Webtools 3.6.3"; \
-#        url="https://download.eclipse.org/webtools/downloads/drops/R3.6.3/R-3.6.3-20150216091848/repository/";\
-#        location="${workspace}/cnf/cache/stable/EclipseWTP",\
+-plugin.m2e.debug:\
+    aQute.bnd.repository.p2.provider.P2Repository; \
+        name="Eclipse EMF 2.13"; \
+        url="https://download.eclipse.org/modeling/emf/emf/builds/release/2.13";\
```
bjhargrave: This is a no-no. Relying on Eclipse download servers kills the reliability of our builds which is why we no longer do. We now use https://dl.bintray.com/bndtools/eclipse-repo/4.7.3a/. 

This section is commented out for normal builds and only to be uncommented for debugging purposed. If you want to add the desired jars to the normal build, they need to be added to https://dl.bintray.com/bndtools/eclipse-repo/4.7.3a/.

bndtools.core/bndtools.shared.bndrun
```diff
@@ -175,7 +182,7 @@ crossplatform.bundles: \
 	${help.bundles},\
 	${p2.bundles},\
 	${pde.bundles},\
-	${m2e.bundles},\
+	${m2e.debug.bundles},\
```
bjhargrave: I don't think this is something to commit to the normal build. It is only for m2e debugging purposes. See my comment on the p2 repo with the m2e debug bundles.

cnf/ext/repositories.bnd
timothyjward: > Relying on Eclipse download servers kills the reliability of our builds which is why we no longer do.
I have no problem with avoiding Eclipse download servers.

> This section is commented out for normal builds and only to be uncommented for debugging purposes.

For normal Bndtools development it's really necessary to be able to run up what you've just been working on without a full build cycle. This whole section was broken (i.e. it didn't work at all) because the previous download site had been archived. Even switching to the new archive location didn't work because the versions were too old for Bndtools new Eclipse baseline. My view is therefore that we need to maintain this piece of the build because it's important.

bndtools.m2e/src/bndtools/m2e/IndexConfigurator.java
timothyjward: Oops - fixed, amended and force pushed

cnf/ext/repositories.bnd
bjhargrave: OK, then someone will need to identify the necessary jars to add to the bintray repo so we can up date it.

timothyjward:
As discussed on the Bndtools call this PR will be split to separate the M2E debug fix from the indexing speedup. The main purpose of this PR is therefore superseded by #3353

PR #3356, opened 2019-07-23
[bndtools] Use sensible defaults when launching using context menu

kriegfrj:
* When finding an existing launch configuration, it will match on the Eclipse project only unless a `bnd.bnd` file or a `bndrun` file has been selected.
* Clears the `tester.names` property on JUnit launches unless a testable Java element has been selected.

### DIFF
--- a/bndtools.core/src/bndtools/launch/JUnitShortcut.java
+++ b/bndtools.core/src/bndtools/launch/JUnitShortcut.java
@@ -78,23 +78,39 @@ public class JUnitShortcut extends AbstractLaunchShortcut {
 		IProject targetProject = elements.get(0)
 			.getJavaProject()
 			.getProject();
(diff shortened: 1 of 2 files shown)


bjhargrave:
@kriegfrj does this PR need any updating based upon the recent merges of your other PRs? Or is ti good to merge?

kriegfrj:
I think it is good to merge, but I was hoping @rotty3000 ore someone else might try it out with the m2e connector to make site it still works there. As I noted, I think it still should but I don't have enough experience with bndtools m2e to know of I'd broken something.

kriegfrj:
I have managed to learn a bit of build Bndtools/m2e now and I tried it out with this patch. Seems to work ok.

bjhargrave:
Thanks!

PR #3365, opened 2019-08-01
Extra system packages, ordering snapshot

pkriens:


biz.aQute.launchpad.tests/bnd.bnd
```diff
@@ -8,4 +8,5 @@
     slf4j.api, \
     slf4j.simple, \
 	${junit},\
-	biz.aQute.launchpad;version=latest
+	biz.aQute.launchpad;version=latest, \
+	aQute.libg
```
bjhargrave: This should be `aQute.libg;version=project` to be consistent with other projects.

Issue #3366, opened 2019-08-02
Launcher should not start fragments after update


pkriens:
! Unexpected error in the run body: Fragment bundles can not be stopped: com.qivicon.binding.bitronvideo.cgi.camera [16]
org.osgi.framework.BundleException: Fragment bundles can not be stopped: com.qivicon.binding.bitronvideo.cgi.camera [16]
        at org.apache.felix.framework.Felix.stopBundle(Felix.java:2753)
        at org.apache.felix.framework.BundleImpl.stop(BundleImpl.java:1038)
        at org.apache.felix.framework.BundleImpl.stop(BundleImpl.java:1024)
        at aQute.launcher.Launcher.installEmbedded(Launcher.java:800)
        at aQute.launcher.Launcher.update(Launcher.java:505)
        at aQute.launcher.Launcher.activate(Launcher.java:422)
        at aQute.launcher.Launcher.launch(Launcher.java:333)
        at aQute.launcher.Launcher.run(Launcher.java:143)
        at aQute.launcher.Launcher.main(Launcher.java:121)
        at aQute.launcher.pre.EmbeddedLauncher.executeWithRunPath(EmbeddedLauncher.java:154)
        at aQute.launcher.pre.EmbeddedLauncher.findAndExecute(EmbeddedLauncher.java:106)
        at aQute.launcher.pre.EmbeddedLauncher.main(EmbeddedLauncher.java:51)
org.osgi.framework.BundleException: Fragment bundles can not be stopped: com.qivicon.binding.bitronvideo.cgi.camera [16]
        at org.apache.felix.framework.Felix.stopBundle(Felix.java:2753)
        at org.apache.felix.framework.BundleImpl.stop(BundleImpl.java:1038)
        at org.apache.felix.framework.BundleImpl.stop(BundleImpl.java:1024)
        at aQute.launcher.Launcher.installEmbedded(Launcher.java:800)
        at aQute.launcher.Launcher.update(Launcher.java:505)
        at aQute.launcher.Launcher.activate(Launcher.java:422)
        at aQute.launcher.Launcher.launch(Launcher.java:333)
        at aQute.launcher.Launcher.run(Launcher.java:143)
        at aQute.launcher.Launcher.main(Launcher.java:121)
        at aQute.launcher.pre.EmbeddedLauncher.executeWithRunPath(EmbeddedLauncher.java:154)
[...]

PR #3367, opened 2019-08-02
[launcher] Fragments and start/stop

pkriens:
Put all start/stop handling in one place so that we
correctly check for fragments and handle eager act.


pkriens:
Fixes #3366

PR #3374, opened 2019-08-12
builder: Avoid rejarring for -includeresource jars from classpath

bjhargrave:
We try to use FileResource instead of JarResource when including
jars from the project's classpath. A FileResource will keep the jar file
unchanged. A JarResource will write a new jar into the bundle which can
result in SHA changes due to potential zip directory reordering.


bjhargrave:
@azzazzel This is the fix for the rejarring issue you identified.

PR #3407, opened 2019-09-06
[contributions] Add description about WIP PRs

kriegfrj:


seanbright:
Would it make more sense to create a "wip" or "work-in-progress" label?

kriegfrj:
Possibly - I just documented the process as it currently is. :smile:

bjhargrave:
> Would it make more sense to create a "wip" or "work-in-progress" label?

Maybe, but WIP in the PR title is pretty common and more in-your-face than some label I wont notice :-) Frankly, we do not use labels very consistently or at all in this project.

kriegfrj:
Also, labels can only be added/removed by the project maintainers, which means the author of the PR may not be able to change it.

Issue #3360, opened 2019-07-29
[launcher] Launcher ignoring changes to properties file


kriegfrj:
The resource listener in Bndtools listens for changes to any of the bundles in `-runbundles` and restarts them if necessary.

It also watches the launch file, in case any of the launch configuration has changed (eg, change in the composition of `-runbundles` itself), and installs/removes bundles as appropriate.

However, the resource watcher will always watch your project's `bnd.bnd`. This may not be the same file that was used to launch the project, which means it will pick up irrelevant changes and ignore the relevant ones if you're launching with anything other than `bnd.bnd`.

It will not be too hard to fix this so that it watches the bnd(run) file that was actually used to start the launch instead of always watching `bnd.bnd`. However, given that it touches the same part of the code as #3300 I'd like to wait for that to get merged so as to avoid conflicts.

rotty3000:
But the `bnd.bnd` file plays a role regardless of which run was used because it defines the bundle. It's not the possibility that it contains run options that is the relevant part, it's the fact that it plays a role in defining the bundle.

kriegfrj:
> But the `bnd.bnd` file plays a role regardless of which run was used because it defines the bundle. It's not the possibility that it contains run options that is the relevant part, it's the fact that it plays a role in defining the bundle.

That's true, but that role is indirect. The running application is only directly affected by the bundles and the run configuration, not the build configuration. If the build configuration in a bnd.bnd file changes, its corresponding output jar will change, and this change will get picked up by the resource listener and the bundle reinstalled. Conversely, if somehow bnd.bnd changes but it's output bundle does not (maybe build automatically is off), then there is nothing (yet) to update in the running application.

rotty3000:
I think we're talking about the logic around here: https://github.com/bndtools/bnd/blob/master/bndtools.core/src/bndtools/launch/OSGiRunLaunchDelegate.java#L244

kriegfrj:
Yes. On the surface of it it looks like it is trying to do what I wanted. Either I broke it somehow in #3145 (which I'm currently using) or else is not quite working properly.

kriegfrj:
[...]

PR #3413, opened 2019-09-09
launcher: Fix handling of system properties and launcher parms on update

bjhargrave:
The watcher, now updated to use a scheduled executor, failed to properly
handle updating the system properties and the launcher parms when the
properties file was updated.

Also remove use of Vector and regex.


bjhargrave:
@kriegfrj Can you verify this PR fixes your issue with the launcher properties update (#3360)? Thanks.

kriegfrj:
Verified - this does indeed fix #3360 when I tested locally.

I think we could do with a regression test though, especially that that this seems to have been a regression?

kriegfrj:
:+1: Thanks for the fix, BJ!

PR #3414, opened 2019-09-10
Made ProjectLauncher more flexible

swimmesberger:
It would be nice to define a different launcher type when exporting a "fatjar" currently I would have to copy much code of the ProjectLauncherImpl - with this change it's fairly easy to extend the ProjectLauncherImpl class to define a different Launcher class on the runpath classpath.

Signed-off-by: Simon Wimmesberger <wimmesberger@gmail.com>

### DIFF
--- a/biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
+++ b/biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
@@ -54,13 +54,14 @@ import aQute.libg.cryptography.SHA1;
 import aQute.libg.glob.Glob;
 
 public class ProjectLauncherImpl extends ProjectLauncher {
-	private final static Logger		logger				= LoggerFactory.getLogger(ProjectLauncherImpl.class);
-	private static final String		EMBEDDED_RUNPATH	= "Embedded-Runpath";
-	private static final String		LAUNCHER_PATH		= "launcher.runpath";
-	private static final String		EMBEDDED_LAUNCHER	= "aQute.launcher.pre.EmbeddedLauncher";
-	static final String				PRE_JAR				= "biz.aQute.launcher.pre.jar";
+	private final static Logger		logger					= LoggerFactory.getLogger(ProjectLauncherImpl.class);
+	private static final String		EMBEDDED_RUNPATH		= "Embedded-Runpath";
+	private static final String 	EMBEDDED_LAUNCHER 		= "Embedded-Launcher";
+	private static final String		LAUNCHER_PATH			= "launcher.runpath";
+	private static final String		EMBEDDED_LAUNCHER_TYPE	= "aQute.launcher.pre.EmbeddedLauncher";
+	static final String				PRE_JAR					= "biz.aQute.launcher.pre.jar";
 	private final Container			container;
-	private final List<String>		launcherpath		= new ArrayList<>();
+	private final List<String>		launcherpath			= new ArrayList<>();
 
 	private File					preTemp;
 
@@ -152,7 +153,16 @@ public class ProjectLauncherImpl extends ProjectLauncher {
 
 	@Override
 	public String getMainTypeName() {
-		return EMBEDDED_LAUNCHER;
+		return EMBEDDED_LAUNCHER_TYPE;
+	}
+
+	/**
+	 * The class name of the main launcher which should be started by the EmbeddedLauncher.
+	 *
+	 * @return if null the default Launcher will be used
+	 */
+	public String getLauncherTypeName() {
+		return null;
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


swimmesberger:
Yeah I expected that ProjectLauncherImpl is simply the default implementation as "impl" suggets but I don't see any reason why a custom launcher should implement all stuff for creating a executable itself when most of the stuff in ProjectLauncherImpl and EmbeddedLauncher is actually pretty generic. Would be fine for me to refactor the ideas into the ProjectLauncher class 😊

bjhargrave:
> Would be fine for me to refactor the ideas into the ProjectLauncher class

Some may make sense as long as they don't require changes to other ProjectLauncher subclasses (like adding new abstract methods). Perhaps it may may sense to refactory ProjectLauncherImpl into a base class which extends ProjectLauncher and then ProjectLauncherImpl (and your impl class) would extend it.

pkriens:
I disagree that the ProjectLauncherImpl is pretty generic. If you look at the remote launcher you see how different a launcher can be. We recently refactored the code because a lot of the ProjectLauncherImpl had unfortunately crept into the ProjectLauncher.

What is it that your launcher does that this launcher doesn't? With the Embedded Activators it might be better to try to use delegation instead of inheritance. I am a tad reluctant to let ProjectLauncherImpl be inherited because it will restrict us in the future. And I am not sure where we would locate a base class for this launcher.

swimmesberger:
We need some additional static methods (start, stop) in the EmbeddedLauncher class because we have a external application which requires certain methods to exist on the main classpath of a jar. The external application doesn't know anything about the bundles in the jar directory of the jar it only knows about the "default" jar classes which when exporting a "fatjar" is simply the EmbeddedLauncher. We don't really need to modify the way the export works or what it does we simply need to supply our own class in the main classpath which delegates (in certain cases) to the default EmbeddedLauncher.

swimmesberger:
With the change in this pull request we would had the ability to do that without copying the code of the ProjectLauncherImpl and doing only these simple changes.

pkriens:
You can just place these JARs on the -runpath, they are automatically included

swimmesberger:
They are unfortunately included at runtime not at build time - for a external application only the EmbeddedLauncher is on the classpath. The jars on the runpath are included by the EmbeddedLauncher on startup. This does not work for our external application.
Unfortunately there are a couple of (legacy) reasons someone would need to replace or extend the EmbeddedLauncher/Launcher Mechanism so the argument still stands that there should be some mechanism to do some customizations to the launching of a bnd application without repeating code.

pkriens:
Did you consider the Classpath manifest header? Or just rewrap the uber jar?

As said, I am not comfortable with making the launcher extendable because it restricts what we can do with it in the future. I regard the launcher as an implementation, the interface is the ProjectLauncher class. It is usually bad practice to depend on implementations.

swimmesberger:
I have tried this approaches but they seem uncomfortable. The Classpath way just doesn't feel like "the bnd way" to do it. Rewrapping would need some custom build code.
The launcher approach seems the most "clean" because I just have to create a bundle in the bnd worspace define "Launcher-Plugin" header in this bundle und I would be done (when the bundle is on the runpath). In the bundle I would simply extend from a "base" Launcher which provides me with all the basic launcher functionality.
For now I must stick with my own ProjectLauncher implementation which basically does mostly the same as the ProjectLauncherImpl but instead provides a custom class which delegates to the EmbeddedLauncher 90% of the time. It just doesn't feel right that there is no default implementation for local execution launchers but I also understand that you wouldn't be to be tied to the API because that would restrict future changes to the ProjectLauncherImpl class.
Maybe keep it as internal class/api and simply provide some extension points (via protected methods) which can be used but are subject to change?

pkriens:
Due to lack of information what your actual problem I cannot say for sure but I find it highly unlikely that your problem cannot be solved with the launcher as is, there are so many extension points already. And I am even willing to adapt the code to help you but I really do not want to make it extendable. Have been bitten too often lately.

For example, we could allow you to set the Main-Class in the bndrun file, you can then add a JAR with your code on the -runpath.

swimmesberger:
I think the main issue here is that the is no way I can tell bnd to use a jar/bundle as the main class files of the final executable uber jar. After the build the result is always the same:

1. EmbeddedLauncher class file is on the main classpath
2. EmbeddedLauncher loads the classes on the runpath at runtime
3. Application launches

1. I would like to have my class file on the main classpath
2. My class file to delegate to the default "boot" mechanism on certain conditions
3. Main Application launches

Currently putting stuff on the runpath doesn't include the classes in the main class path as .class files in the jar. Which is needed by our external application. The application needs to find a .class in the jar without looking into subdirectories like "jar" where all the jars on the runpath reside.

pkriens:
You could use `-includeresource: @myjar.jar` in the bndrun file, or just include a single class this way. This will expand the `myjar.jar` in the main launcher jar and you can whatever class you want. Then just set the Main-Class to your code. You can then call the EmbeddedLauncher or do something else.

swimmesberger:
Does overriding the "Main-Class" work? Because as far a I can tell is the "Main-Class" property always overridden here:
https://github.com/bndtools/bnd/blob/057e7849b2c3c9f3da8fda264b6ad2c050d2c017/biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java#L386

So the properties set in the bndrun are always superceded by the hardcoded properties in the ProjectLauncherImpl.

swimmesberger:
I have tried to change the idea from inheritance as a class to configure options via bnd properties in the following pull request:
https://github.com/bndtools/bnd/pull/3417

PR #3416, opened 2019-09-11
clazz: Add support for -noclassforname instruction

bjhargrave:


kriegfrj:
I was thinking it might be useful to make this an enum rather than a boolean. Thus:
`-classforname: [ none | required | optional | dynamic ]`

For backward compatibility, `required` could be the default. I'd argue, however, that at some point we should consider making "optional" the default, as people don't typically use `Class.forName()` to load a class for required dependencies - mostly for optional dependencies.

If it is too difficult to implement `optional` and `dynamic` in the short term, then you could start with `none` and `required` and add the other two later. Eg, perhaps we could implement `none` and `required` and cherry-pick it back into 4.3, and add `optional` and `dynamic` into 4.4 at some later point.

bjhargrave:
> I was thinking it might be useful to make this an enum rather than a boolean. Thus:
> `-classforname: [ none | required | optional | dynamic ]`
> 

This would be very complex to implement. Now the Clazz only collects reference to packages which are returned by `Clazz.getReferred()`. It does not have anyway to distinguish why/how the class referred to the package. So we would have to modify the class to also collect information on why/how the package references occurs. We would then also need to teach the Analyzer that this new information exists so it could make decisions based up on the information. So this would be a major change.

Also, I do not see the value of doing it. The `-noclassforname` switch merely tells `Clazz` to ignore Class.forName references. The user can then add specific packages manually to the Import-Package statement. The user can also make certain package dynamic ('resolution:=dynamic`) as well. The user has full control of this. What they don't have control over is whether we decide that Class.forName references are included in the result of `Clazz.getReferred()` which is what this PR provides control over.

The only enhancement to this PR I considered was to add fine grained control over which classes are processed for Class.forName references such as how `-dsannotations` provides fine-grained control over which classes are scanned for DS component annotations. But I am not sure this is worth it either.

kriegfrj:
Fair enough, it would be a big change. But if you at least adopt the syntax, there's a clean migration path if/when the extra options ever become feasible? Don't have to implement all the options right away. 

Just a thought. Happy for you to go worth whatever you decide.

PR #3417, opened 2019-09-11
WIP: [launcher]: Made ProjectLauncher more flexible via bnd property

swimmesberger:
With this change it's possible to define custom jars to include in the final uberjar distribution. It's also possible to define a custom class via the "Main-Class" MANIFEST property.

By using the:
-prejar property in a "bndrun" file it's possible to define one or more jars which will be directly included in the main classpath.

example:
Main-Class: org.example.custom.launcher.MainLauncher
-prejar: \
	biz.aQute.launcher.pre.jar;inline=true,\
	org.example.custom.launcher.jar

This will result in the Embedded Launcher and the custom launcher in the main classpath. The MainLauncher class will be executed when the jar is started. The MainLauncher can also delegate the start process to the EmbeddedLauncher because it's on the classpath.

* WIP: currently it's only an idea and not tested just wanted to hear your opinions

Signed-off-by: Simon Wimmesberger wimmesberger@gmail.com


swimmesberger:
I know that it's unfortunate that this introduces a new property "prejar" instead of reusing "includeresource" but with prejar we more flexibility for launcher specific attributes in the future. Also "includeresource" doesn't necessarily mean "put it on the classpath" like we have to do in the  "updateFromProject()" method.

pkriens:
I really do not want more options :-( this beast is already way too complicated. 

However, you are right that we would now have different handling of the executable and the launch, which is not good.

I think now that the easiest solution then is that you create a new launcher while replacing the `pre.jar` with your own version? Then we do not need to change the launcher at all? A wrapping project with the following bnd file would do the trick I think:

    -includeresource: @${repo;biz.aQute.launcher},pre.jar=mypre.jar

If you use the same class name for the Main-Class attribute things should work out of the box. We could of course set the Main-Class attribute in pre.jar and make the `updateFromProject()` read the `Main-Class` attribute, that would be a bit cleaner.

swimmesberger:
I have to try that with the current snapshot version and might actually work with the changes for the Main-Class attribute - because for our usecase we want to delegate to the default EmbeddedLauncher I we provide in out pre.jar a custom EmbeddedLauncher we can't delegate to the default launcher. But when we do the Main-Class change we could provide our own Main-Class AND the default EmbeddedLauncher class files in the pre.jar and delegate to the default launcher when certain conditions are met.
It just seems a bit tedious to wrap the default launcher to do all this stuff but it might work 😅
I will change this pull request accordingly to only do the required changes.

biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
```diff
@@ -369,6 +369,10 @@ public Jar executable() throws Exception {
 		Manifest m = new Manifest();
 		Attributes main = m.getMainAttributes();
 
+		logger.debug("Use Embedded launcher");
+		main.putValue(MAIN_CLASS, EMBEDDED_LAUNCHER);
+		main.putValue(EMBEDDED_RUNPATH, join(classpath));
+
```
bjhargrave: Moving this up make it subject to damage by the REMOVEHEADERS work below. If for some reason REMOVEHEADERS contains MAIN_CLASS or EMBEDDED_RUNPATH, we are screwed now. I think we will need to make sure to remove MAIN_CLASS and EMBEDDED_RUNPATH from `result` before the call to `main.keySet().removeAll(result)` below.

swimmesberger: Should be fixed

biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
```diff
@@ -369,6 +370,12 @@ public Jar executable() throws Exception {
 		Manifest m = new Manifest();
 		Attributes main = m.getMainAttributes();
 
+		logger.debug("Use Embedded launcher");
+		// contains all whitelisted properties which should not be removed in any way
+		Set<String> whitelistProperties = new HashSet<>(Arrays.asList(MAIN_CLASS, EMBEDDED_RUNPATH));
```
bjhargrave: Since this is a static list value, why create a Set and why create it each invocation. 

```java
private static final List<String> whitelistProperties = Arrays.asList(MAIN_CLASS, EMBEDDED_RUNPATH);
```

should be fine as a static variable.

swimmesberger: You are right - I just wanted to have it in the same line as setting the manifest to ensure if a additional property is added changing the whitelist is not forgotten - but i changed it 🙂.

biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
```diff
@@ -152,7 +155,7 @@ public void cleanup() {
 
 	@Override
 	public String getMainTypeName() {
-		return EMBEDDED_LAUNCHER;
+		return getProject().get(MAIN_CLASS, EMBEDDED_LAUNCHER);
```
bjhargrave: Looks good. Can you squash all the commits in to a single commit including changing the above line to

```java
getProject().getProperty(MAIN_CLASS, EMBEDDED_LAUNCHER);
```

`get` is the Domain method which just delegates to the Processor `getProperty` method anyway.

swimmesberger: Done

PR #3426, opened 2019-09-14
maven plugins doing implicit indexing result in many logged `[INFO] Undefined property in POM[<gav>] : key project.version`

rotty3000:


rotty3000:
if this fix gets approved, can we backport to 4.3.0? The current state is confusing to users and erroneous.

rotty3000:
get is only used in the constructor and it's private.

rotty3000:
the values are not really missing, they are supposed to be inferred from the defaults.

rotty3000:
updated

PR #3428, opened 2019-09-14
bnd-resolver-maven-plugin failure output has regressed

rotty3000:


rotty3000:
I know that ideally this should be fixed in Bndrun, and we discussed this on at least one call and it was going to be fixed but never was. This late stage I just want a fix for 4.3.0 that will make maven usage user friendly again.

bjhargrave:
> This late stage I just want a fix for 4.3.0 that will make maven usage user friendly again.

I think the gradle resolve task will have the same problem. So it would be better to make a new `resolveCLI` method in Bndrun for the CLI (maven/gradle) plugins to both use when resolving.

rotty3000:
@bjhargrave see updated.

bjhargrave:
replaced by #3429

Issue #3431, opened 2019-09-17
Support for Eclipse/Equinox Splash Screen in BND Launcher


juergen-albert:
In order to get all the Eclipse RCP Developers to use BND instead of PDE, we need a support for the SplashScreen. This was at least one of the requirements I've heard all the time. 

I'm currently finalizing a first Implementation for a ProjectLauncher, that can start a bndrun with a splashscreen and also provides an export using the equinox executables. The splashscreen is pulled up, before the actual Launcher is running and a Runnable, that simply is called to end the splashscreen is handed to the Launcher. This registers it in the framework as a service and finally removes the splash, if nobody else does it. 

I had a copy of the Launcher and realized, that the changes are relatively minor and do not bring any additional dependencies. Thus I would love to integrate this in the BND Launcher, instead of maintaining a copy with myself. A PR will follow soon.

juergen-albert:
https://github.com/bndtools/bnd/pull/3432#issuecomment-534028158

bjhargrave:
I think we can close this since it is replaced by #3445, right?

Issue #3410, opened 2019-09-09
[launcher] Support per-bundle reference directive


kriegfrj:
The launcher will by default try to use "reference:" url schema to install bundles into the running OSGi instance, which is much more efficient because the framework doesn't have to copy the bundles but references them in their existing locations. The `-runnoreference` directive allows you to override this if necessary.

Unless you're on Windows, in which case it will never use the "reference:" schema. The documentation mistakenly says that the default on Windows is not to use `reference:`, but the definition of `Launcher.useReferences()` will always return ~~true~~`false` on Windows. (*Edited: originally I wrote `true` when it should have been `false`.*)

There are good reasons for this. Unfortunately, the frameworks typically hold a lock on the bundles that they install, which means that you can't build them while the framework is running, which means you can't get hot updates. However, when you are testing a framework with lots of bundles there is a high startup cost associated with it.

I would like to propose a compromise implementation that allows you to specify "noreference" semantics on a per-bundle basis. So this would become an attribute of the `runbundles` property (similar to the startlevel attribute).

This would be useful because in any given test deployment, a large percentage of the bundles are quasi-static - it's really only the bundles in the Workspace repository that might change. An optimum setup on Windows would be to configure all of the Workspace bundles as "noreference" and the rest of the bundles as "reference". This way you could get 95% of the performance benefit of reference bundles but still allow your hot updates to work.

kriegfrj:
> > but the definition of `Launcher.useReferences()` will always return true on Windows.
> 
> I am not sure I see this. Can you explain why you think this is the case? If `isWindows()` returns `true` then `!isWindows()` is `false` and `false && anything` is `false`.

Sorry, you're right, I mistyped. I should have written "always returns false on Windows". The rest of the post assumes that useReferences I is always false on Windows.

pkriens:
I am hesitant. On modern systems the copying is very fast and this adds a lot of complexity.

Did you measure the performance difference? With fast disks I would expect it to be negligent.

kriegfrj:
> I am hesitant. On modern systems the copying is very fast and this adds a lot of complexity.
> 
> Did you measure the performance difference? With fast disks I would expect it to be negligent.

I'll try and get some solid benchmarks for you, but anecdotally I feel it would be a useful gain with large AUTs. Eg, all the Eclipse bundles when you launch the Bndtools launch in the bnd workspace run to a couple hundred MB, from memory. Even with an SSD it takes a while to copy that much.

[...]

PR #3432, opened 2019-09-17
#3431 Support for SplashHandler in BND Launcher

juergen-albert:
Commit for #3431 

### DIFF
--- a/biz.aQute.launcher/src/aQute/launcher/Launcher.java
+++ b/biz.aQute.launcher/src/aQute/launcher/Launcher.java
@@ -211,13 +211,19 @@ public class Launcher implements ServiceListener {
 
 	/**
 	 * Regsiters the given Runnable as endSplashHandler
-	 * 
+	 *
 	 * @param endSplashHandler
 	 * @param systemContext
 	 */
 	private static void publishSplashScreen(final Runnable endSplashHandler, BundleContext systemContext) {
 		if (endSplashHandler == null)
 			return;
+
+		// register the splashHandler
+		Dictionary<String, Object> splashProperties = new Hashtable<>();
+		splashProperties.put("name", "endSplashHandler"); //$NON-NLS-1$//$NON-NLS-2$
+		systemContext.registerService(Runnable.class.getName(), endSplashHandler, splashProperties);
+
 		// register the output stream to the launcher if it exists
 		try {
 			Method method = endSplashHandler.getClass().getMethod("getOutputStream", new Class[0]); //$NON-NLS-1$
--- a/biz.aQute.launcher/src/aQute/launcher/Launcher.java
+++ b/biz.aQute.launcher/src/aQute/launcher/Launcher.java
@@ -13,6 +13,7 @@ import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.OutputStream;
 import java.io.PrintStream;
 import java.io.PrintWriter;
 import java.io.Reader;
@@ -37,6 +38,7 @@ import java.security.Policy;
 import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.Collections;
+import java.util.Dictionary;
 import java.util.Enumeration;
 import java.util.Formatter;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


pkriens:
I'd rather not have a specific splash handler. The intention of the launcher is that the embedded activators do this. Embedded activators are on the run path.

A valid concern is that they are not run until after all bundles are installed (start happens after). In a recent discussion, a large setup took 2.5 seconds so I do not think that is a real problem. However, a change that also allows embedded activators to run before install is ok with me. Currently the before/after start flag (static boolean IMMEDIATE) could be turned into a String or so to support multiple possibilities.

Another alternative is to add code to the launcher and change the main class.  See #3410

bjhargrave:
> I'd rather not have a specific splash handler.

I agree. I am not excited by this PR as it currently stands. It is too specific and should be handled by a more generic means such as @pkriens suggests.

juergen-albert:
I see your points. The whole eclipse start process is kind of special and it took me a while to solve a lot of their idiosyncrasies. The splashscreen and JNI part that handles this, however is something, I can't really change without touching the binary launcher. I still need to speak to Tom to find out, why this is entangled as it is. 

I hope that by offering a Eclipse capable export and Launcher, to get more people moving from PDE to BND. I hope that this helps us, to get more Eclipse Projects to produce better OSGi Bundles. Unfortunately, we/I have to make a few concessions in order to make it work, without changing things in to many eclipse core projects.

As an alternative suggestion, I'd like to make the create Framework Method protected. With this, my Launcher can inherit from the bnd Launcher and can do what I need to do without maintaining a copy of the Launcher.

@pkriens I've create a new Ticket for the Embedded Activator, because this would solve a few other issues for the eclipse start process.

pkriens:
I do not want (as should be clear from the other discussion) make Launcher API. As said, I am ok to make it possible to start the Embedded Activators a bit earlier (although I would find a 2.5-sec delay acceptable.) 

If there are other functions missing let us discuss them here as well. Making launcher API will let everybody go in its own direction and forces the launcher to stay compatible. I think the launcher is powerful enough, or can be extended in a generic way, to solve most problems.

juergen-albert:
Okay, to boil my problem down a bit:

Ehe equinox launcher executables that eclipse uses basically do bit of magic. They get a couple of properties (found e.g. in the eclipse.ini). At first it locates the jvm native library (e.g. the jvm.dll).  They take the jar mentioned as -startup, look for a main class declaration, instantiating it reflectively by calling the jvms native methods and call a run method with the arguments from the ini. 
In Eclipse pendant to the EmbeddedLauncher is in a separate jar. Before starting anything, it identifies the another native library and where to find the splashscreen. It then initializes its JNIBridge with the native library and displays the splash screen. this is then wrapped in a runnable (run removes the splash) and hands it over to the EclipseStarter (This is basically our Launcher), which then does the registering of the runnable in the framework and a lot more. 
For some reason, the native launcher calls in the JVM after the start of the first MainClass and expects to find the org.eclipse.equinox.launcher.JNIBridge class and wants to know about the native splash library location. If this class is not in the initial launch jar, a NoClassDef Exception is printed.
All 3 of their start components are together like 5k LOC. 

I want to get rid of as much non OSGi behaviour as possible and only cater to the few things that are really necessary to get eclipse running and exported with BND. Thus my solution looks at the moment as follows:

I have a pre launcher jar, which basically has something like the Embedded Launcher inside it. It additionally has the org.eclipse.equinox.launcher.JNIBridge the executable requires. It handles the splash and calls my Launcher similar to the bnd embedded launcher, but hands over the runnable. My Launcher itself looks exactly as the one in the PR. 

In addition to that I have a Launcher Plugin, that uses my pre launcher, so a splash can be shown, when started via bnd(tools). Additionally it overrides the executable method and can build something  that looks like an eclipse product export and does not use the f***ed up start mechanism of eclipse.

pkriens:
What cannot be handled with Embedded Activators? (If necessary with before install activation)

juergen-albert:
It might work with before install activation. I will implemente this and try it afterwards.

bjhargrave:
I think we can close this since it is replaced by #3449, right?

juergen-albert:
Yes, we can close this. The splash even works with the embedded activator

PR #3438, opened 2019-09-20
[cdi] CDI processing finds beans on the Bundle-ClassPath when the main discover mode is none

rotty3000:


rotty3000:
not quite right yet

biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
```diff
@@ -83,6 +84,19 @@ public boolean analyzeJar(Analyzer analyzer) throws Exception {
...
 
+		// If, after removing every instruction where 'discover=none', there are
+		// no items left we should just return.
+		if (header.entrySet()
+			.stream()
+			.map(e -> e.getValue()
+				.get("discover"))
+			.filter(Objects::nonNull)
+			.filter(discover -> Discover.valueOf(discover) == Discover.none)
+			.count() == 0) {
```
bjhargrave: Instead of `count() == 0`, just replace the final `filter` call with:

```java
.allMatch(discover -> Discover.valueOf(discover) == Discover.none);
```
But handling the IllegalArgumentException issue.

`allMatch` will early complete with `false` if a non-`none` value is found.

biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
```diff
@@ -83,6 +84,19 @@ public boolean analyzeJar(Analyzer analyzer) throws Exception {
...
 		}
 
+		// If, after removing every instruction where 'discover=none', there are
+		// no items left we should just return.
+		if (header.entrySet()
+			.stream()
+			.map(e -> e.getValue()
+				.get("discover"))
+			.filter(Objects::nonNull)
+			.filter(discover -> Discover.valueOf(discover) == Discover.none)
```
bjhargrave: If the string is not a valid enum name, you will need to catch IllegalArgumentException from the `Discover.valueOf(String)` call.

For example:

https://github.com/bndtools/bnd/blob/abdf289ce7599d1595a29ec546774d56eb824f99/biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java#L136-L139

rotty3000:
I'm adding a test case right now.

PR #3441, opened 2019-09-23
[cdi] ignore anonymous classes

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java
```diff
@@ -472,8 +472,9 @@ String getSourceFile() {
 		boolean isInnerClass() {
 			String binary = type.getBinary();
 			return attributes(InnerClassesAttribute.class).flatMap(a -> Arrays.stream(a.classes))
-				.anyMatch(inner -> (inner.inner_name != null) && !Modifier.isStatic(inner.inner_access)
-					&& inner.inner_class.equals(binary));
+				.anyMatch(inner -> inner.inner_class.equals(binary)
+					&& (((inner.inner_name != null) && !Modifier.isStatic(inner.inner_access))
+						|| ((inner.inner_name == null) && (inner.inner_access == 0))));
```
rotty3000: I wonder if all of this should really just be:
```
.anyMatch(inner -> inner.inner_class.equals(binary) && !Modifier.isStatic(inner.inner_access));
```

bjhargrave:
See https://github.com/bndtools/bnd/pull/2938 and  https://github.com/bndtools/bnd/commit/b8df8562bb9e6835717fca4f79f34e2ab05f0c5c where you added the check for `inner_name != null` that you are now proposing to remove.

So is removing this check going to break something your fixed with #2938?

rotty3000:
Hey BJ, yeah, I guess it was wrong.

Now that we have a test for anon classes along with tests for nested and non-static inner classes I feel more confident that I'm not wrong this time (and that you were right in the beginning).

biz.aQute.bndlib.tests/test/test/ClazzTest.java
```diff
@@ -855,4 +855,15 @@ public void testInnerClass() throws Exception {
 			assertThat(clazz.isInnerClass()).isTrue();
 		}
 	}
+
+	Object anon = new Object() {};
+
+	public void testInnerAnonymousClass() throws Exception {
+		File file = IO.getFile("bin_test/test/ClazzTest$1.class");
```
bjhargrave: This will be a fragile test if anyone adds another anonymous class to this class. To be safe, I think you need to create a static inner class to hold the anonymous inner class for this test.

```java
public static class AnonymousClassHolder {
   Object anon = new Object() {};
}
...
File file = IO.getFile("bin_test/test/ClazzTest$AnonymousClassHolder$1.class");
```

rotty3000: I can do that!

rotty3000:
Fixed!

bjhargrave:
> Fixed!

This will make the test actually test that anonymous inner class. I looked and the ClazzTest class already had 14 anonymous inner classes. `$1` was not even the anonymous inner class you added for the test :-)

rotty3000:
Right, of course.

Issue #3445, opened 2019-09-23
Enable Embedded Activators to be started right after Framework Init and before any bundle is installed


juergen-albert:
As Peter suggested in his PR Comment:

I'd rather not have a specific splash handler. The intention of the launcher is that the embedded activators do this. Embedded activators are on the run path.

A valid concern is that they are not run until after all bundles are installed (start happens after). In a recent discussion, a large setup took 2.5 seconds so I do not think that is a real problem. However, a change that also allows embedded activators to run before install is ok with me. Currently the before/after start flag (static boolean IMMEDIATE) could be turned into a String or so to support multiple possibilities.

Another alternative is to add code to the launcher and change the main class.  See #3410

_Originally posted by @pkriens in https://github.com/bndtools/bnd/pull/3432#issuecomment-533458038_

juergen-albert:
I would suggest the following:
We leave the IMMEDIATE field as it is, in order top avoid breaking existing behaviour or make it awkward and just print out a warning that marks this as deprecated..
We look for a second field of type String (I would name it START_LEVEL and look for the values: AFTER_FRAMEWORK_START, BEFORE_BUNDLES_START and DEFAULT. if this is set IMMEDIATE is ignored.

pkriens:
Just keep IMMEDIATE but allow these strings as well as being boolean. We're inspecting this reflectively so no need to deprecate anything. And anyway, a deprecation is useless since you cannot mark anything deprecated since there is no type binding. Also, don't add `DEFAULT`, call it `AFTER_BUNDLES_START`. Providing an invalid string means log error and start `AFTER_BUNDLES_START`.

PR #3449, opened 2019-09-24
Implement after-framework-init support for embedded activators

juergen-albert:
fixes https://github.com/bndtools/bnd/issues/3445

### DIFF
--- a/biz.aQute.launcher/src/aQute/launcher/Launcher.java
+++ b/biz.aQute.launcher/src/aQute/launcher/Launcher.java
@@ -43,6 +43,7 @@ import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.IllegalFormatException;
 import java.util.LinkedHashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -99,7 +100,6 @@ public class Launcher implements ServiceListener {
 	private boolean							security;
 	private SimplePermissionPolicy			policy;
 	private Callable<Integer>				mainThread;
-	private final List<BundleActivator>		embedded			= new ArrayList<>();
 	private final Map<Bundle, Throwable>	errors				= new HashMap<>();
 	private final Map<File, Bundle>			installedBundles	= new LinkedHashMap<>();
 	private File							home				= new File(System.getProperty("user.home"));
@@ -425,7 +425,6 @@ public class Launcher implements ServiceListener {
 
 		doSecurity();
 
-		List<Bundle> tobestarted = update(System.currentTimeMillis() + 100);
 
 		int result = LauncherConstants.OK;
 
@@ -434,6 +433,8 @@ public class Launcher implements ServiceListener {
 		systemContext.addServiceListener(this, "(&(|(objectclass=" + Runnable.class.getName() + ")(objectclass="
 			+ Callable.class.getName() + "))(main.thread=true))");
 
+		List<BundleActivator> startBeforeBundleStart = new LinkedList<>();
+		List<BundleActivator> startAfterBundleStart = new LinkedList<>();
 
 		// Start embedded activators
 		trace("start embedded activators");
@@ -443,22 +444,30 @@ public class Launcher implements ServiceListener {
 				try {
 					Class<?> clazz = loader.loadClass((String) token);
 					BundleActivator activator = (BundleActivator) newInstance(clazz);
-					if (isImmediate(activator)) {
-						start(systemContext, result, activator);
+					if (isAfterFrameworkInit(activator)) {
+						result = start(systemContext, result, activator);
+					} else if (isBeforeBundlesStart(activator)) {
+						startBeforeBundleStart.add(activator);
+					} else {
+						startAfterBundleStart.add(activator);
 					}
-					embedded.add(activator);
 					trace("adding activator %s", activator);
 				} catch (Exception e) {
 					throw new IllegalArgumentException("Embedded Bundle Activator incorrect: " + token, e);
 				}
 			}
 		}
+		List<Bundle> tobestarted = update(System.currentTimeMillis() + 100);
+
+		for (BundleActivator activator : startBeforeBundleStart) {
+			result = start(systemContext, result, activator);
+		}
 
 		startBundles(tobestarted);
 
-		for (BundleActivator activator : embedded)
-			if (!isImmediate(activator))
-				result = start(systemContext, result, activator);
+		for (BundleActivator activator : startAfterBundleStart) {
+			result = start(systemContext, result, activator);
+		}
 
 		systemBundle.start();
 		trace("system bundle started ok");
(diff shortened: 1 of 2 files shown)


biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -443,22 +444,30 @@ private int activate() throws Exception {
 				try {
 					Class<?> clazz = loader.loadClass((String) token);
 					BundleActivator activator = (BundleActivator) newInstance(clazz);
-					if (isImmediate(activator)) {
-						start(systemContext, result, activator);
+					if (isAfterBundleStart(activator)) {
```
bjhargrave: This PR needs include the associated doc changes to https://github.com/bndtools/bnd/blob/master/docs/_chapters/300-launching.md#embedded-activators.

So it seem to me that previously: immediate=true means start now, otherwise start after. That names of the new is methods do not make any sense when reading this code because it reads like if isAfter, then start immediately!

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -443,22 +444,30 @@ private int activate() throws Exception {
...
+						result = start(systemContext, result, activator);
+					} else if (isBeforeBundleStart(activator)) {
+						startBeforeBundleStart.add(activator);
+					} else {
+						startAfterBundleStart.add(activator);
```
bjhargrave: So if it is not `isAfter` and not `isBefore` then you `startAfter`? That does not seem to read correctly.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
juergen-albert: My mistake. I named the method wrong and didn't realize. If fixed it and hope the new name makes it clearer now. I added the docu as well.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -471,15 +480,39 @@ private int activate() throws Exception {
...
 				.getField("IMMEDIATE");
+			if (f.getType()
+				.equals(Boolean.class)
+				|| f.getType()
+					.equals(boolean.class)) {
+				return false;
+			} else if (f.getType()
+				.equals(String.class)) {
+				return "AFTER_FRAMEWORK_START".equals(f.get(activator));
+			}
```
pkriens: Could you extract the commonality with the next method? It is almost identical. Just make a method that returns the value as an Object

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -471,15 +480,39 @@ private int activate() throws Exception {
 		return result;
 	}
 
-	private boolean isImmediate(BundleActivator activator) {
+	private boolean isAfterFrameworkStart(BundleActivator activator) {
```
bjhargrave: This is still a confusing name. The FrameworkEvent.STARTED event is defined to mean:

```
The Framework has started.

This event is fired when the Framework has started after all installed bundles that are marked to be started have been started and the Framework has reached the initial start level. 
```

but you really mean `AFTER_FRAMEWORK_INIT`. That is, after the framework is initialized by Framework.init and before any bundles are installed and started.

docs/_chapters/300-launching.md
```diff
@@ -84,10 +84,12 @@ The launcher registers a service with object class `Object` that provides some r
...
+The launcher supports _embedded activators_. These are like normal Bundle Actviator classes but are instead found on the `-runpath`. Any bundle that has the header `Embedded-Activator` will be started. The start can happen at 3 points: Before any bundle is installed, after all bundles are installed and after all bundles are started. The default is after all bundles are started. To configure the starting point add a static or instance field called `IMMEDIATE` of type String and set one of the following values: `AFTER_FRAMEWORK_START` or `BEFORE_BUNDLES_START`. Any other value will result in the default start after all bundles are started. 
+
+Until recently only before and after bundle start was support. This was determined by looking for a static boolean field `IMMEDIATE`. This is still supported and `true` will result in a start before any bundles is started. `false` trigger a start after all bundles are started.
```
bjhargrave: You should state what new value IMMEDIATE=true and IMMEDIATE=false map to. It is not clear.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
juergen-albert: That is true. I had this thought at some point as well. I will change the value and method name to correspond to  ```AFTER_FRAMEWORK_INIT```

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -480,37 +480,29 @@ private int activate() throws Exception {
...
-				return f.getBoolean(activator);
-			} else if (f.getType()
-				.equals(String.class)) {
-				return "BEFORE_BUNDLES_START".equals(f.get(activator));
+
+		Object result = getImmediateFieldValue(activator);
+		if (result != null) {
+			if (result instanceof String) {
+				return "BEFORE_BUNDLES_START".equals(result);
+			} else if (result instanceof Boolean || Boolean.TYPE.isInstance(result)) {
```
bjhargrave: You do not need the second check since the reflective call will never return a `boolean` since that cannot be represented as `Object`. It will return a `Boolean` object for a `boolean` field.

So you can implement `isBeforeBundlesStart` simply as:

```java
private boolean isBeforeBundlesStart(BundleActivator activator) {
  Object result = getImmediateFieldValue(activator);
  return "BEFORE_BUNDLES_START".equals(result) || Boolean.TRUE.equals(result);
}
```

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -480,37 +480,29 @@ private int activate() throws Exception {
...
-					.equals(boolean.class)) {
-				return false;
-			} else if (f.getType()
-				.equals(String.class)) {
-				return "AFTER_FRAMEWORK_START".equals(f.get(activator));
-			}
+			return f.get(activator);
 		} catch (Exception e) {}
-		return false;
+		return null;
```
bjhargrave: I would return `Boolean.FALSE` here.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -471,15 +480,31 @@ private int activate() throws Exception {
...
 
-			return f.getBoolean(activator);
-		} catch (Exception e) {
-			return false;
+	private boolean isAfterFrameworkInit(BundleActivator activator) {
+		Object result = getImmediateFieldValue(activator);
+		return "AFTER_FRAMEWORK_INIT".equals(result);
+	}
+
+	private boolean isBeforeBundleStart(BundleActivator activator) {
```
bjhargrave: There is a plurality mismatch between the method name (Bundle singular) and the constant value name (BUNDLES plural). They should match.

bjhargrave:
Also, please squash your work into a single commit for the next update. Thanks.

PR #3452, opened 2019-09-25
[launcher] Fix reporting output

### LINKED ISSUE #3442: [launcher] file: URLs get mangled on Windows during reporting
My experience in #3410 has taught me two things: 1. that I ought not to be so patient in waiting for bnd to do things (including launch), and 2. that I shouldn't be so quick to accept "it's Windows => it sucks" as the final answer to every performance issue faced on Windows. :smile:

Having learned these valuable lessons, I turned my attention to the other part of launching that was taking a long time (though not as long as installing the bundles) but which seemed like it shouldn't: the call to `report()` once all the bundles have started. I always observed the following when launching:

* Once installed, the bundles were started fairly quickly.
* The launcher would quickly print the report header and the line containing the information about the system bundle.
* The output would then pause for several seconds.
* Finally, the rest of the report's output would appear more-or-less instantaneously.
* Though I didn't realise it at the time, there was another symptom: the "Modified" field was printed as "<>" for every bundle (no timestamps, even for file-based bundles).

Closer inspection revealed a problem in this reporting code, in particular these lines:

https://github.com/bndtools/bnd/blob/9cc7029010dc3c4637cc18eb5cf49a989cb1553f/biz.aQute.launcher/src/aQute/launcher/Launcher.java#L1202-L1204

The intent here seems to be to see if the specified location was a real file, by parsing out the scheme and seeing if the rest is a resolvable file. The problem is that `file:` URLs on Windows typically contain a segment with a drive letter (which also includes a colon). Here is an example cut-and-paste directly from my system:

`file:/C:/workspace/bnd/cnf/cache/stable/EclipseOxygen/javax.el-2.2.0.v201303151357.jar`

The call to `String.replaceAll()` on L1203 would convert the above to the following:

`//workspace/bnd/cnf/cache/stable/EclipseOxygen/javax.el-2.2.0.v201303151357.jar`

This looks to Windows like a fully-qualified UNC path name. So the pause in the output was in the call to `f.exists()` (L1208), as the OS hit the network looking for the machine called `workspace` with a shared directory `bnd`. Mercifully, the result of this lookup failure must have been cached as the subsequent calls to `f.exists()` were quick (though if it hadn't been, possibly the performance impact would have been so high and unacceptable as to trigger an investigation sooner).

This also explains the missing timestamps in the report output - it didn't find the bundle files (as it was looking in the wrong place) and thus couldn't report their timestamps. (Having never used Bnd on non-Windows, and as this problem seems to have existed on Windows for some time, I didn't realise this was abnormal - had I known, I might have discovered this problem quicker.)

PR to follow shortly. Arguably, this should be squeezed into the upcoming release of 4.3 if possible as the impact is pretty significant.

### LINKED ISSUE #3451: [launcher] Report mangling the timestamp
While writing tests for my fix for #3442 I discovered that the report was not generating the correct timestamps.

https://github.com/bndtools/bnd/blob/9cc7029010dc3c4637cc18eb5cf49a989cb1553f/biz.aQute.launcher/src/aQute/launcher/Launcher.java#L1241-L1246

Two problems with this code:

* `Calendar.get(Calendar.MONTH)` returns a zero-based value for month (ie, January = 0, not 1);
* `fill()` left-aligns the month value.

So 01:01 the 1st of October 2018 would be output as `201890101010` instead of `2018100101`.

Fix to follow.

kriegfrj:
Closes #3442, closes #3451.

### DIFF
--- a/aQute.libg/src/aQute/libg/uri/URIUtil.java
+++ b/aQute.libg/src/aQute/libg/uri/URIUtil.java
@@ -3,6 +3,11 @@ package aQute.libg.uri;
 import java.io.File;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.nio.file.InvalidPathException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Locale;
+import java.util.Optional;
 import java.util.regex.Pattern;
 
 public final class URIUtil {
@@ -41,4 +46,74 @@ public final class URIUtil {
 		return baseURI.resolve(refURI);
 	}
 
+	private static final URI EMPTYURI = URI.create("");
+
+	/**
+	 * Attempts to fetch a path on the file system for the given URI string.
+	 * Tries a few more tricks than the standard method of
+	 * <tt>Paths.get(new URI(uriString))</tt> - it can handle plain paths, and
+	 * also nested URI pseudo-schemes like <tt>reference:</tt> and
+	 * <tt>jar:</tt>. Examples:
+	 * <ul>
+	 * <li>/path/to/file => /path/to/file</li>
+	 * <li>reference:file:/path/to/file => /path/to/file</li>
+	 * <li>jar:file:/path/to/file.jar!/some/contained/element =>
+	 * /path/to/file.jar</li>
+	 * <li>http://server/path/to/file => <tt>null</tt></li>
+	 * </ul>
+	 *
+	 * @param uriString The string containing the URI for which we are
+	 *            attempting to construct a path.
+	 * @return The method's best guess as to which file on the local filesystem
+	 *         this URI refers, or <tt>null</tt> if it's invalid or not a local
+	 *         filesystem URI.
+	 */
+	public static Optional<Path> pathFromURI(String uriString) {
+		if (uriString == null) {
+			return Optional.empty();
+		}
+
+		try {
+			URI uri = resolve(EMPTYURI, uriString);
+			while (!EMPTYURI.equals(uri)) {
+
+				final String scheme = uri.getScheme();
+				if (scheme == null) {
+					try {
+						return Optional.of(Paths.get(uriString));
+					} catch (InvalidPathException e) {
+						return Optional.empty();
+					}
+				}
+
+				switch (scheme.toLowerCase(Locale.ROOT)) {
+					case "file" :
+						return Optional.of(Paths.get(uri));
+					case "jar" :
+					case "bundle" :
+					case "zip" :
+						uriString = uri.getSchemeSpecificPart();
+						int index = uriString.indexOf("!/");
+						if (index >= 0) {
+							uriString = uriString.substring(0, index);
+						}
+						uri = resolve(EMPTYURI, uriString);
+						break;
+					case "reference" :
+						uriString = uri.getSchemeSpecificPart();
+						uri = resolve(EMPTYURI, uriString);
+						break;
+					default :
+						uriString = uri.getSchemeSpecificPart();
+						uri = new URI(uriString);
+						if (uri.getScheme() == null) {
+							return Optional.empty();
+						}
+						break;
+				}
+			}
+		} catch (URISyntaxException e1) {
+		}
+		return Optional.empty();
+	}
 }
--- a/aQute.libg/test/aQute/libg/uri/URIUtilsTest.java
+++ b/aQute.libg/test/aQute/libg/uri/URIUtilsTest.java
@@ -4,10 +4,17 @@ import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.condition.OS.WINDOWS;
 
 import java.net.URI;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.Optional;
 
+import org.assertj.core.api.SoftAssertions;
+import org.assertj.core.api.junit.jupiter.SoftAssertionsExtension;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledOnOs;
+import org.junit.jupiter.api.extension.ExtendWith;
 
+@ExtendWith(SoftAssertionsExtension.class)
 public class URIUtilsTest {
 
 	@Test
@@ -58,5 +65,61 @@ public class URIUtilsTest {
 		assertThat(result).hasToString("file:////server/share/foo.txt");
 		result = URIUtil.resolve(URI.create("file:/C:/Users/jim/base.txt"), "//server/share/foo.txt");
 		assertThat(result).hasToString("file:////server/share/foo.txt");
+	}
+
+	@Test
+	public void fromURI_returnsPath_platformIndependent(SoftAssertions softly) {
+		fromURI_returnsPath("invalid url:", null, softly);
+		fromURI_returnsPath("", null, softly);
+		fromURI_returnsPath("http://some/URL", null, softly);
+		fromURI_returnsPath("unknownsheme:invalid scheme:somepath", null, softly);
+		fromURI_returnsPath("/some/path", "/some/path", softly);
+		fromURI_returnsPath("some/path", "some/path", softly);
+		fromURI_returnsPath("file:/some/path", "/some/path", softly);
+		fromURI_returnsPath("reference:file:/some/path", "/some/path", softly);
+		fromURI_returnsPath("reference:/some/path", "/some/path", softly);
+		fromURI_returnsPath("file:/some/path", "/some/path", softly);
+		fromURI_returnsPath("jar:file:/workspace/bnd/cache/mybundle.jar!/some/contained/element",
+			"/workspace/bnd/cache/mybundle.jar", softly);
+		fromURI_returnsPath("jar:file:/workspace/bnd/cache/mybundle.jar!/", "/workspace/bnd/cache/mybundle.jar",
+			softly);
+		fromURI_returnsPath("jar:file:/workspace/bnd/cache/mybundle.jar", "/workspace/bnd/cache/mybundle.jar", softly);
+		fromURI_returnsPath("bundle:file:/workspace/bnd/cache/mybundle.jar!/some/contained/element",
+			"/workspace/bnd/cache/mybundle.jar", softly);
+		fromURI_returnsPath("bundle:/workspace/bnd/cache/mybundle.jar!/some/contained/element",
+			"/workspace/bnd/cache/mybundle.jar", softly);
+		fromURI_returnsPath("zip:file:/workspace/bnd/cache/mybundle.jar", "/workspace/bnd/cache/mybundle.jar", softly);
+	}
+
+	private static void fromURI_returnsPath(String uri, String expected, SoftAssertions softly) {
+		Path ePath = expected == null ? null : Paths.get(expected);
+		softly.assertThat(URIUtil.pathFromURI(uri))
+			.as("uri '%s' maps to path '%s'", uri, ePath)
+			.isEqualTo(Optional.ofNullable(ePath));
+	}
+
+	@Test
+	@EnabledOnOs(WINDOWS)
+	public void fromURI_returnsPath_onWindows(SoftAssertions softly) {
+		fromURI_returnsPath("C:\\some\\path", "C:\\some\\path", softly);
+		fromURI_returnsPath("D:\\some\\other\\path", "D:\\some\\other\\path", softly);
+		fromURI_returnsPath("C:/some/path", "C:\\some\\path", softly);
+		fromURI_returnsPath("file:/C:/some/path", "C:\\some\\path", softly);
+		fromURI_returnsPath("reference:file:/X:/some/path", "X:\\some\\path", softly);
+		fromURI_returnsPath("reference:X:/some/path", "X:\\some\\path", softly);
+		fromURI_returnsPath("reference:X:\\some\\path", "X:\\some\\path", softly);
+		fromURI_returnsPath("file://myserver/myshare/some/path", "\\\\myserver\\myshare\\some\\path", softly);
+		fromURI_returnsPath("unknownsheme:file://someserver/somepath", "\\\\someserver\\somepath", softly);
+		fromURI_returnsPath("jar:file:/C:/workspace/bnd/cache/mybundle.jar!/some/contained/element",
+			"C:\\workspace\\bnd\\cache\\mybundle.jar", softly);
+		fromURI_returnsPath("jar:file:/C:/workspace/bnd/cache/mybundle.jar!/",
+			"C:\\workspace\\bnd\\cache\\mybundle.jar",
+			softly);
+		fromURI_returnsPath("jar:file:/C:/workspace/bnd/cache/mybundle.jar", "C:\\workspace\\bnd\\cache\\mybundle.jar",
+			softly);
+		fromURI_returnsPath("bundle:file:/C:/workspace/bnd/cache/mybundle.jar!/some/contained/element",
+			"C:\\workspace\\bnd\\cache\\mybundle.jar", softly);
+		fromURI_returnsPath("zip:file:/C:/workspace/bnd/cache/mybundle.jar", "C:\\workspace\\bnd\\cache\\mybundle.jar",
+			softly);
 	}
 }
--- a/biz.aQute.launcher/src/aQute/launcher/Launcher.java
+++ b/biz.aQute.launcher/src/aQute/launcher/Launcher.java
@@ -29,13 +29,16 @@ import java.net.InetSocketAddress;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.nio.ByteBuffer;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.security.AllPermission;
 import java.security.CodeSource;
 import java.security.Permission;
 import java.security.PermissionCollection;
 import java.security.Policy;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
-import java.util.Calendar;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.Formatter;
@@ -46,6 +49,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.StringTokenizer;
 import java.util.concurrent.Callable;
@@ -81,6 +85,7 @@ import aQute.lib.io.ByteBufferDataOutput;
 import aQute.lib.io.IO;
 import aQute.lib.startlevel.StartLevelRuntimeHandler;
 import aQute.lib.strings.Strings;
+import aQute.libg.uri.URIUtil;
 
 /**
  * This is the primary bnd launcher. It implements a launcher that runs on Java
@@ -1199,16 +1204,19 @@ public class Launcher implements ServiceListener {
 					out.println("Id    Levl State Modified      Location");
 
 					for (int i = 0; i < bundles.length; i++) {
+						String lastModified = "<>";
 						String loc = bundles[i].getLocation();
-						loc = loc.replaceAll("\\w+:", "");
-						File f = new File(loc);
+						Optional<Path> p = URIUtil.pathFromURI(loc);
+						if (p.isPresent() && Files.exists(p.get())) {
+							lastModified = FORMAT.format(Files.getLastModifiedTime(p.get())
+								.toInstant());
+							loc = p.get()
+								.toString();
+						}
 						out.print(fill(Long.toString(bundles[i].getBundleId()), 6));
 						out.print(fill(startLevelhandler.getBundleStartLevel(bundles[i]) + "", 4));
 						out.print(fill(toState(bundles[i].getState()), 6));
-						if (f.exists())
-							out.print(fill(toDate(f.lastModified()), 14));
-						else
-							out.print(fill("<>", 14));
+						out.print(fill(lastModified, 14));
 
 						if (errors.containsKey(bundles[i])) {
 							out.print(fill(loc, 50));
@@ -1238,16 +1246,8 @@ public class Launcher implements ServiceListener {
 		out.println();
 	}
 
-	String toDate(long t) {
-		Calendar c = Calendar.getInstance();
-		c.setTimeInMillis(t);
-		return fill(c.get(Calendar.YEAR), 4) + fill(c.get(Calendar.MONTH), 2) + fill(c.get(Calendar.DAY_OF_MONTH), 2)
-			+ fill(c.get(Calendar.HOUR_OF_DAY), 2) + fill(c.get(Calendar.MINUTE), 2);
-	}
-
-	private String fill(int n, int width) {
-		return fill(Integer.toString(n), width, '0', -1);
-	}
+	static final DateTimeFormatter FORMAT = DateTimeFormatter.ofPattern("YYYYMMddHHmm")
+		.withZone(ZoneOffset.UTC);
 
 	private String fill(String s, int width) {
 		return fill(s, width, ' ', -1);
--- a/aQute.libg/src/aQute/libg/uri/packageinfo
+++ b/aQute.libg/src/aQute/libg/uri/packageinfo
@@ -1,1 +1,1 @@
-version 3.2.0
+version 3.3.0
--- a/biz.aQute.bndall.tests/test/biz/aQute/launcher/AlsoLauncherTest.java
+++ b/biz.aQute.bndall.tests/test/biz/aQute/launcher/AlsoLauncherTest.java
@@ -16,13 +16,20 @@ import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Properties;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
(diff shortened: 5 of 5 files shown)


biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -1199,16 +1205,39 @@ public void report(PrintStream out) {
 					out.println("Id    Levl State Modified      Location");
 
 					for (int i = 0; i < bundles.length; i++) {
+						String lastModified = "<>";
 						String loc = bundles[i].getLocation();
-						loc = loc.replaceAll("\\w+:", "");
-						File f = new File(loc);
+						try {
```
pkriens: I think this warrants a method, maybe in `aQute.libg.uri.URIUtil`?

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -1199,16 +1205,39 @@ public void report(PrintStream out) {
...
+									.toLowerCase()) {
+									case "file" :
+										Path p = Paths.get(locURI);
+										long time = Files.getLastModifiedTime(p)
+											.toMillis();
+										lastModified = FORMAT.format(Files.getLastModifiedTime(p)
+											.toInstant());
+										loc = p.toString();
+										break WHILE;
+									case "reference" :
```
pkriens: I think this can be the default case? This will allow you to handle other schemes that do the same trick like I think `jar:`?

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -1199,16 +1205,39 @@ public void report(PrintStream out) {
...
+										locURI = new URI(locURI.getRawSchemeSpecificPart());
+										break;
+									default :
+										break WHILE;
+								}
+								count++;
```
pkriens: I do not see how recursion could be possible so no use to check some limit. If you insist, make the length of the location. Dont think this is necessary though

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -1238,16 +1267,8 @@ private void row(PrintStream out, Object... parms) {
...
-		Calendar c = Calendar.getInstance();
-		c.setTimeInMillis(t);
-		return fill(c.get(Calendar.YEAR), 4) + fill(c.get(Calendar.MONTH), 2) + fill(c.get(Calendar.DAY_OF_MONTH), 2)
-			+ fill(c.get(Calendar.HOUR_OF_DAY), 2) + fill(c.get(Calendar.MINUTE), 2);
-	}
-
-	private String fill(int n, int width) {
-		return fill(Integer.toString(n), width, '0', -1);
-	}
+	static final DateTimeFormatter FORMAT = DateTimeFormatter.ofPattern("YYYYMMddHHmm")
```
pkriens: I see the zone was already in there but I would MUCH prefer to use UTC always in bnd.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
kriegfrj: I think I was just being overly paranoid. I'll remove it.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
kriegfrj: Ok.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
kriegfrj: For storing values or for computation, I agree. But for displaying values, presumably the intent is for it to be human intelligible? In which case, wouldn't the system default time done be more useful? If the user really wanted to see values in UTC presumably they would have UTC set as their system default time zone?

If you still disagree, this is an easy change.

kriegfrj: By way of comparison, file timestamps in Linux are stored in UTC but when you do an ls they are displayed in system time zone (from memory - been a while since I used Linux).

biz.aQute.launcher/src/aQute/launcher/Launcher.java
kriegfrj: :+1: Will do.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
bjhargrave: I think UTC is better as the output is consistent and does not vary based upon the location of the system. This is nice when your CI system is in the cloud and you don't care what time zone it is in. You only care that your output is consistent.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
kriegfrj: Done.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
kriegfrj: Done.

kriegfrj:
I think I've done everything that was asked by @pkriens and @bjhargrave, and pushed the result.

Latest push depends on #3454 for the testing of `URIUtil`. If necessary, I can re-write the tests to use JUnit 4 but they won't look as clean nor will their output be as clean.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
kriegfrj: I gave this a solid try, but I don't think it's feasible. When I make it the default, I have to special-case all of the schemes that we know don't refer to files on the local filesystem. Otherwise (eg) it resolves http://server/some/path to the local filesystem UNC path //server/some/path, and we end up with the same bug that triggered all of this.

The only easy way around this is to special-case all of the exceptions that ought not to be recursively resolved, and don't try and recursively resolve them. But if I have to list a bunch of special cases anyway, I think it's safer to special-case all the schemes that do the same trick rather than try and special-case all those that don't (opt-in rather than opt-out).

kriegfrj:
This is proving a bit hard for me to develop and test `URIUtil.fromURI()` because I don't have easy access to a Unix machine to test behaviour on Unix. The turnaround time between change -> push -> wait for Travis makes it a bit impractical. I won't be upset if someone else takes it and runs with it from here...

bjhargrave:
> This is proving a bit hard for me to develop and test `URIUtil.fromURI()` because I don't have easy access to a Unix machine to test behaviour on Unix.

LOL. Welcome to our world. I have the same issue with Windows behavior testing! Thank goodness for Azure Pipelines' Windows builds.

kriegfrj:
>I gave this a solid try, but I don't think it's feasible. When I make it the default, I have to special-case all of the schemes that we know don't refer to files on the local filesystem. Otherwise (eg) it resolves http://server/some/path to the local filesystem UNC path //server/some/path, and we end up with the same bug that triggered all of this.
>
>The only easy way around this is to special-case all of the exceptions that ought not to be recursively resolved, and don't try and recursively resolve them. But if I have to list a bunch of special cases anyway, I think it's safer to special-case all the schemes that do the same trick rather than try and special-case all those that don't (opt-in rather than opt-out).

Ok, the heuristic I used to try and get around this problem:

1. If the URI is of a scheme known to refer to files, then it will try and dereference the file. This will work both with raw paths (reference:/to/some/path) and file URLs (reference:file:/to/some/path).
2 If the URI is not of a known scheme type, it will only try and dereference the file if the scheme specific part contains the substring "file:". So http://server/some/path won't match, but http:file:/server/some/path will map to file:/server/some/path.

Hopefully this ticks all the boxes. :smile:

aQute.libg/src/aQute/libg/uri/URIUtil.java
```diff
@@ -41,4 +44,84 @@ public static URI resolve(URI baseURI, String reference) throws URISyntaxExcepti
...
+		try {
+			URI uri = resolve(EMPTYURI, uriString);
+			while (uriString != null && !uriString.isEmpty()) {
+
+				final String scheme = uri.getScheme();
+				if (scheme == null) {
+					try {
+						return Paths.get(uriString);
+					} catch (InvalidPathException e) {
+						e.printStackTrace();
```
bjhargrave: This should not be in production code. If it is an unrecoverable error that we wish to communicate, then we need to throw a meaningful exception. Printing to System.out can be lost, in Bndtools for example.

aQute.libg/src/aQute/libg/uri/URIUtil.java
```diff
@@ -41,4 +44,84 @@ public static URI resolve(URI baseURI, String reference) throws URISyntaxExcepti
...
+						return null;
+					}
+				}
+
+				switch (uri.getScheme()
```
bjhargrave: See

https://github.com/bndtools/bnd/blob/672b366c691ebed6973cbb40a22f7abfe0c15e52/biz.aQute.bndlib/src/aQute/bnd/osgi/Resource.java#L36-L61

where we handle peeling off schemes (in a URL but basically similar). But the key thing in that code is to never use a `jar:file:` URL since the code in Java will lock the file in Windows for the duration of the VM.

I don't see that we need a loop since we are not going to see `jar:jar:jar:` for example. The worst case would seem to be `reference:jar:file`

aQute.libg/src/aQute/libg/uri/URIUtil.java
```diff
@@ -41,4 +44,84 @@ public static URI resolve(URI baseURI, String reference) throws URISyntaxExcepti
...
+					.toLowerCase()) {
+					case "file" :
+						return Paths.get(uri);
+					case "jar" :
+					case "bundle" :
+					case "zip" :
+						uriString = uri.getRawSchemeSpecificPart();
```
bjhargrave: Not sure why you are using `raw` methods in this method. I did not use them in `Resource.fromURL`.

aQute.libg/test/aQute/libg/uri/URIUtilsTest.java
```diff
@@ -1,56 +1,125 @@
...
+import static org.junit.jupiter.api.condition.OS.WINDOWS;
+
 import java.net.URI;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.stream.Stream;
 
-import aQute.lib.io.IO;
-import junit.framework.TestCase;
+import org.junit.jupiter.api.Test;
```
bjhargrave: I guess it would be nice if this PR did not depend on the Junit 5 PR. Then it is held out waiting for that to be integrated.

aQute.libg/test/aQute/libg/uri/URIUtilsTest.java
```diff
@@ -1,56 +1,125 @@
...
-			result = URIUtil.resolve(URI.create("file:/C:/Users/jim/base.txt"), "//server/share/foo.txt");
-			assertEquals("file:////server/share/foo.txt", result.toString());
-		}
+		URI result = URIUtil.resolve(URI.create("file:/C:/Users/jim/base.txt"), "\\\\server\\share\\foo.txt");
+		assertThat(result).hasToString("file:////server/share/foo.txt");
+		result = URIUtil.resolve(URI.create("file:/C:/Users/jim/base.txt"), "//server/share/foo.txt");
+		assertThat(result).hasToString("file:////server/share/foo.txt");
+	}
+
+	private static Stream<Arguments> fromURI_returnsPath_platformIndependent() {
```
bjhargrave: I must say this is some seriously unreadable stuff. I can see that it is powerful but it makes it impossible to set a break point on the one assert data that you want to debug. So I would rather not do this and have the test method just do `assertThat(URIUtil.fromURI("uri")).isEqualTo("expected");` repeatedly. Then I can put a break point on whatever line I want to debug that test data.

And then we don't need Junit 5 for this PR.

biz.aQute.bnd.gradle/bnd.bnd
```diff
@@ -22,6 +22,9 @@ Bundle-Description: The Bnd Gradle plugin.
 	biz.aQute.resolve;version=latest, \
 	biz.aQute.repository;version=latest
 
+-testpath: \
```
bjhargrave: The Junit 5 commit must not be part of this PR.

bjhargrave: The JUnit 5 PR has been merged now. So this commit should be removed from this PR.

aQute.libg/src/aQute/libg/uri/URIUtil.java
kriegfrj: Sorry, a relic from my debugging. It should be ignored.

biz.aQute.bnd.gradle/bnd.bnd
kriegfrj: Yes. The plan was to rebase onto matter once the JUnit Platform change was merged.

aQute.libg/test/aQute/libg/uri/URIUtilsTest.java
kriegfrj: I anticipated that the JUnit 5 PR would get merged before this one (as indeed it has).

aQute.libg/test/aQute/libg/uri/URIUtilsTest.java
kriegfrj: This was the primary reason for me upgrading to JUnit 5, and now you want me to abandon it! 😆

> I must say this is some seriously unreadable stuff.

I'd like a bit more clarity on what you mean by "unreadable". Originally, I had the Arguments stream laid out quite a bit neater, in a more readable, tabular way, but the Eclipse formatting rules turned it into the mess that you currently see. Would fixing the formatting fix this objection? Or is there more to it?

> I can see that it is powerful but it makes it impossible to set a break point on the one assert data that you want to debug. 

I think this is worth a bit of a discussion, because as it currently stands a practical consequence of this objection is a blanket ban on the use of parameterized tests. But I get what you are saying about debugging a single result. Does the following alternative workflow work for you:

1. Set your breakpoint on the parameterized test method.
2. Run the test in the IDE.
3. Look at the results in the JUnit tab, find the parameter combination that you'd like to debug.
4. Right-click on the result for that parameter combination, and click "Debug".

The Eclipse JUnit runner will now re-run the test for that specific parameter combination, and it will stop at your breakpoint.

I've gone ahead and made the change you requested anyway, but still interested in your thoughts on the above for future reference.

aQute.libg/src/aQute/libg/uri/URIUtil.java
kriegfrj: I vaguely thought if there was a path that had a : in it that had been URI-encoded, you wouldn't want that to be decoded as it might then be interpreted as a scheme delimiter when it shouldn't. But that's a pretty esoteric case. None of my test cases covered that case - as : is not a legal path character on Windows it wasn't easy for me to test. Maybe you could try some tests for files with ':' in the path on Mac and see whether it is appropriate to use `raw`.

aQute.libg/src/aQute/libg/uri/URIUtil.java
kriegfrj: > But the key thing in that code is to never use a jar:file: URL since the code in Java will lock the file in Windows for the duration of the VM.

I understand this reasoning and I agree, but this method is not generating URIs - it is interpreting URIs that have been generated elsewhere. So I don't quite see the relevance of that comment to the current PR?

> I don't see that we need a loop since we are not going to see jar:jar:jar: for example. The worst case would seem to be reference:jar:file

I was trying to be as generic as possible without trying to pre-empt what all the possible combinations might be, so that it won't suddenly stop working in the future unexpectedly on some unforeseen combination. I can easily change it to handle only known valid combinations.

aQute.libg/src/aQute/libg/uri/URIUtil.java
kriegfrj: In the meantime, I've change it to not use the `raw`, for the sake of getting this merged.

aQute.libg/src/aQute/libg/uri/URIUtil.java
kriegfrj: `printStackTrace()` is gone now.

aQute.libg/src/aQute/libg/uri/URIUtil.java
bjhargrave: `:` is not generally use in *ix file names either since `:` is used as the path separator.

aQute.libg/test/aQute/libg/uri/URIUtilsTest.java
bjhargrave: In general, I like test case methods to be simple, A then B then C. Easy to read, easy to debug, easy to change. It may mean some more code repetition but it is a test case not production code. Sometime extracting common code to a function does make sense to be shared among test methods. But in this case the common code was a single assert. So there it little saved by sharing.

I have seen test cases which are so complicated that no one left on the project understands them or how to modify, extends or fix them. So I am very much of the opinion that test cases should be plain dead simple.

This example took advantage of a JUnit 5 feature to derive multiple test methods from a single test method and a stream of data input. For this case, it seems unnecessarily complicated for what we need to do. You could also have softly asserted all the asserts to get the maximum results if you wanted to avoid an early failure stopping the remaining asserts.

aQute.libg/src/aQute/libg/uri/URIUtil.java
```diff
@@ -41,4 +44,83 @@ public static URI resolve(URI baseURI, String reference) throws URISyntaxExcepti
...
+				if (scheme == null) {
+					try {
+						return Paths.get(uriString);
+					} catch (InvalidPathException e) {
+						return null;
+					}
+				}
+
+				switch (uri.getScheme()
+					.toLowerCase()) {
```
bjhargrave: This should use `.toLowerCase(Locale.ROOT)` to provide consistent results in all locales.

aQute.libg/src/aQute/libg/uri/URIUtil.java
```diff
@@ -41,4 +44,83 @@ public static URI resolve(URI baseURI, String reference) throws URISyntaxExcepti
...
+					case "file" :
+						return Paths.get(uri);
+					case "jar" :
+					case "bundle" :
+					case "zip" :
+						uriString = uri.getSchemeSpecificPart();
+						int index = uriString.indexOf('!');
```
bjhargrave: Perhaps this should more specifically look for `"!/"`? I think `'!'` is not specific enough here.

aQute.libg/src/aQute/libg/uri/URIUtil.java
```diff
@@ -41,4 +44,83 @@ public static URI resolve(URI baseURI, String reference) throws URISyntaxExcepti
...
+						if (index >= 0) {
+							uriString = uriString.substring(0, index);
+						}
+						if (uriString == null) {
```
bjhargrave: How can uriString ever be null here? It was just dereferenced above.

aQute.libg/src/aQute/libg/uri/URIUtil.java
```diff
@@ -41,4 +44,83 @@ public static URI resolve(URI baseURI, String reference) throws URISyntaxExcepti
...
+						uri = resolve(EMPTYURI, uriString);
+						break;
+					default :
+						uriString = uri.getSchemeSpecificPart();
+						try {
+							uri = new URI(uriString);
+							if (uri.getScheme() == null) {
+								return null;
+							}
+						} catch (URISyntaxException e) {
```
bjhargrave: Is this additional try/catch for URISytaxException needed since the code is already in a try/catch for URISytaxException which will return null?

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -1199,16 +1203,20 @@ public void report(PrintStream out) {
 					out.println("Id    Levl State Modified      Location");
 
 					for (int i = 0; i < bundles.length; i++) {
+						String lastModified = "<>";
 						String loc = bundles[i].getLocation();
-						loc = loc.replaceAll("\\w+:", "");
-						File f = new File(loc);
+						Path p = URIUtil.fromURI(loc);
+						if (p != null && Files.exists(p)) {
+							long time = Files.getLastModifiedTime(p)
```
bjhargrave: What is this `time` variable used for? It seems to be dead code.

aQute.libg/src/aQute/libg/uri/URIUtil.java
```diff
@@ -41,4 +44,83 @@ public static URI resolve(URI baseURI, String reference) throws URISyntaxExcepti
...
+	 *         filesystem URI.
+	 */
+	public static Path fromURI(String uriString) {
+		if (uriString == null || uriString.isEmpty()) {
+			return null;
+		}
+
+		try {
+			URI uri = resolve(EMPTYURI, uriString);
+			while (uriString != null && !uriString.isEmpty()) {
```
bjhargrave: How can uriString ever be `null` here? I do see that this is possible.

aQute.libg/src/aQute/libg/uri/URIUtil.java
```diff
@@ -41,4 +44,83 @@ public static URI resolve(URI baseURI, String reference) throws URISyntaxExcepti
...
+	 * <li>http://server/path/to/file => <tt>null</tt></li>
+	 * </ul>
+	 *
+	 * @param uriString The string containing the URI for which we are
+	 *            attempting to construct a path.
+	 * @return The method's best guess as to which file on the local filesystem
+	 *         this URI refers, or <tt>null</tt> if it's invalid or not a local
+	 *         filesystem URI.
+	 */
+	public static Path fromURI(String uriString) {
```
bjhargrave: I think the method name `fromURI` is not descriptive enough. Perhaps `pathFromURI` is more appropriate. Then its use reads better: `Path p = URIUtil.pathFromURI(loc);` or `Path p = pathFromURI(loc);` if you statically import the method.

aQute.libg/test/aQute/libg/uri/URIUtilsTest.java
kriegfrj: I like test methods to be as simple as possible too. But was the parameterized test really that complicated? I'd agree that it was a little unreadable (primarily due to the bad formatting), but conceptually it's quite simple - a single test method (with a single assert), and a parameter list. However, I agree that in this case with a single assert there wasn't a lot of boilerplate saved - in fact, probably when you look at the clutter introduced by `Arguments.of` it was even a net loss! I could have used more compact ways of doing the parameter list, but I've found them to be less type safe (I'm still learning here myself).

You're right that soft assertions can be used to prevent crashing out the test on the first assertion (this is what the latest push looks like). However, I think there are still some reasons to prefer parameterized approach:
1. The output in Eclipse is IMHO more easily navigable. The results are hierarchically grouped, one line per test. You can see which one(s) failed and click on it/them for the stack trace. The output of multiple soft assertion failures can be a bit difficult to follow if there are many failures.
2. Soft assertions prevent an assertion short-circuiting the test, but there's nothing to stop an uncaught exception from short-circuiting the test. Each run of a parameterized test is completely isolated from the others.

aQute.libg/src/aQute/libg/uri/URIUtil.java
kriegfrj: Not generally used, yes; but it is not illegal - you can use it in a path if it's escaped. On Windows you simply can't use it at all.

aQute.libg/src/aQute/libg/uri/URIUtil.java
kriegfrj: I think we're flying blind either way because there is no official spec for jar: syntax (at least, none that I've found).

Done.

aQute.libg/src/aQute/libg/uri/URIUtil.java
kriegfrj: Done.

aQute.libg/src/aQute/libg/uri/URIUtil.java
kriegfrj: A relic. I've fiddled with this method so much that there's probably a few of them. Gone.

aQute.libg/src/aQute/libg/uri/URIUtil.java
kriegfrj: No, it is not.

aQute.libg/src/aQute/libg/uri/URIUtil.java
kriegfrj: You're right. Originally I didn't have the opening null check in the if statement, so this actually did something.

aQute.libg/src/aQute/libg/uri/URIUtil.java
kriegfrj: Originally I thought that the path prefix would be redundant, as in an expression like `Path p = fromURI(loc)` it is clear from the LHS that it's a path. However, this would not be clear in all contexts so I'll put the path prefix in.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
kriegfrj: Debugging relic. Removed.

kriegfrj:
Another thing I thought about: how do we feel about `pathToURI()` returning an `Optional<Path>` rather than a nullable `Path`? I haven't done that because the Bnd codebase doesn't seem to make use of that pattern, but I thought to double check.

kriegfrj:
Just the local rebuild failed - looks like a transient network error.

bjhargrave:
> returning an `Optional<Path>` rather than a nullable `Path`

I don't object to this. It seems to have a fit here since not all URIs can be mapped to a local file system path.

We have started to use Optional in some places (e.g. Jar class). I am not sure it would make the code in the Launcher using the API smaller. But give it try if you wish.

kriegfrj:
> > returning an `Optional<Path>` rather than a nullable `Path`
> 
> I don't object to this. It seems to have a fit here since not all URIs can be mapped to a local file system path.
> 
> We have started to use Optional in some places (e.g. Jar class). I am not sure it would make the code in the Launcher using the API smaller. But give it try if you wish.

I think you're right - Optional doesn't make the client code smaller - I think it's main strength is yo make client code safer & less prone to NPEs. I'll give it a try.

kriegfrj:
Method signature now changed to return `Optional<Path>`.

bjhargrave:
Thanks!

PR #3454, opened 2019-09-26
Use JUnit 5 for JUnit test cases

kriegfrj:
Example of how JUnit 5 support might be added to the Bnd build.

This PR does the following:

* Adds library packages for JUnit Platform, and the Jupiter and Vintage engines to ext/central.mvn
* Adds macros to group them logically to ext/junit.bnd
* Converts `aQute.libg` to use the JUnit Platform.
  * `build.gradle` modified to cause Gradle to use the JUnit Platform rather than the JUnit 4 runner.
  * Converted `aQute.lib.io.IOTest` and `aQute.libg.uri.URIUtils` to use the Jupiter engine.
  * All other test classes remain as they were (JUnit 3 or 4 tests) and are executed by JUnit Platform using the Vintage engine.
* All other projects continue to use the JUnit 4 runner. To migrate them:
   1. Replace the ${junit} dependency in `-testpath` with ${junit-platform}
   2. Add the engines that you need to `-testpath` too (eg, ${junit-vintage} to run the old JUnit 3/4 tests, and ${junit-jupiter} if you add any Jupiter tests).

As a bonus, using the `@DisabledOnOs` annotation, I have managed to get rid of the test failures for the symlink tests on my machine.

Suggested future enhancements:
1. Get the bnd-gradle-plugin to add the necessary configuration when it detects junit-platform on the class path, so that a `build.gradle` file is not necessary.
2. Similar customisation (if appropriate) for the bnd-maven-m2e integration.

Comments welcomed/appreciated.

### DIFF
--- a/aQute.libg/bnd.bnd
+++ b/aQute.libg/bnd.bnd
@@ -10,7 +10,6 @@
 
 -testpath: \
 	${junit},\
-	osgi.core;version=latest,\
 	slf4j.simple;version=latest
 
 -exportcontents: *
--- a/aQute.libg/test/aQute/lib/io/IOTest.java
+++ b/aQute.libg/test/aQute/lib/io/IOTest.java
@@ -1,7 +1,13 @@
 package aQute.lib.io;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
+
 import java.io.File;
 import java.io.InputStream;
+import java.lang.reflect.Method;
 import java.nio.ByteBuffer;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -9,11 +15,24 @@ import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInfo;
+
 import aQute.lib.io.IO.EnvironmentCalculator;
-import junit.framework.TestCase;
 
-public class IOTest extends TestCase {
+public class IOTest {
 
+	private String testName;
+
+	@BeforeEach
+	public void before(TestInfo info) {
+		testName = info.getTestMethod()
+			.map(Method::getName)
+			.get();
+	}
+
+	@Test
 	public void testEnvVarsForHome() throws Exception {
 		Map<String, String> map = new HashMap<>();
 
@@ -46,6 +65,7 @@ public class IOTest extends TestCase {
 		assertEquals("C:\\Documents and Settings\\foobar", ec2.getSystemEnv("HOME"));
 	}
 
+	@Test
 	public void testSafeFileName() {
 		if (IO.isWindows()) {
 			assertEquals("abc%def", IO.toSafeFileName("abc:def"));
@@ -58,6 +78,7 @@ public class IOTest extends TestCase {
 		}
 	}
 
+	@Test
 	public void testFilesetCopy() throws Exception {
 		File destDir = new File("generated/fileset-copy-test");
 
@@ -78,6 +99,7 @@ public class IOTest extends TestCase {
 		assertTrue(new File(destDir, "root").exists());
 	}
 
+	@Test
 	public void testCopyURLToByteArray() throws Exception {
 		File src = new File("testresources/unzipped.dat");
 		byte[] file = IO.read(src);
@@ -91,6 +113,7 @@ public class IOTest extends TestCase {
 		}
 	}
 
+	@Test
 	public void testCopyToExactHeapByteBuffer() throws Exception {
 		File src = new File("testresources/unzipped.dat");
 		byte[] file = IO.read(src);
@@ -105,6 +128,7 @@ public class IOTest extends TestCase {
 		}
 	}
 
+	@Test
 	public void testCopyToSmallerHeapByteBuffer() throws Exception {
 		File src = new File("testresources/unzipped.dat");
 		byte[] file = IO.read(src);
@@ -118,6 +142,7 @@ public class IOTest extends TestCase {
 		}
 	}
 
+	@Test
 	public void testCopyToLargerHeapByteBuffer() throws Exception {
 		File src = new File("testresources/unzipped.dat");
 		byte[] file = IO.read(src);
@@ -131,6 +156,7 @@ public class IOTest extends TestCase {
 		}
 	}
 
+	@Test
 	public void testCopyToExactDirectByteBuffer() throws Exception {
 		File src = new File("testresources/unzipped.dat");
 		byte[] file = IO.read(src);
@@ -145,6 +171,7 @@ public class IOTest extends TestCase {
 		}
 	}
 
+	@Test
 	public void testCopyToSmallerDirectByteBuffer() throws Exception {
 		File src = new File("testresources/unzipped.dat");
 		byte[] file = IO.read(src);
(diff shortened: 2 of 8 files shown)


aQute.libg/build.gradle
```diff
@@ -0,0 +1,3 @@
+test {
+    useJUnitPlatform()
```
bjhargrave: I don't like this. We do NOT want to put build.gradle files in every project. Anything we need to do for gradle for Junit 5 needs to be handled centrally in the root build.gradle file.

aQute.libg/bnd.bnd
```diff
@@ -9,8 +9,9 @@
 	osgi.core;version=latest
 
 -testpath: \
-	${junit},\
-	osgi.core;version=latest,\
+	${junit-platform},\
```
bjhargrave: We want to have the `${junit}` macro represent the same jars for all projects to use for their `-testpath`. I don't want to have each project pick its own adventure here. This is why `${junit}` currently includes AssertJ. So I think we need a value for `${junit}` which all bundles will use.

Change this will require the `-buildpath` for `biz.aQute.junit` be changed to replace its use of `${junit}` with the current value of `${junit}`.

cnf/ext/junit.bnd
```diff
@@ -1,6 +1,20 @@
 junit: org.apache.servicemix.bundles.junit;version="[4.11,5)",\
- org.assertj.core;version="[3,4)"
+	org.assertj.core;version="[3,4)"
+junit-platform-version: 1.5.2
```
bjhargrave: Not sure I like these version numbers in here. I think they should just be directly specified in central.mvn.

aQute.libg/test/aQute/lib/io/IOTest.java
```diff
@@ -297,8 +322,10 @@ public void testCreateSymlinkOrCopyWillDeleteOriginalLink() throws Exception {
 		}
 	}
 
-	public void testCreateDirectory_Symlink() throws Exception {
-		Path rootDirectory = Paths.get("generated/tmp/test/" + getName());
+	@Test
+	@DisabledOnOs(WINDOWS)
```
bjhargrave: Why did you disable this on Windows? This test currently runs on Windows.

aQute.libg/test/aQute/lib/io/IOTest.java
```diff
@@ -317,8 +344,10 @@ public void testCreateDirectory_Symlink() throws Exception {
 			.exists());
 	}
 
-	public void testCreateDirectory_SymlinkMissingTarget() throws Exception {
-		Path rootDirectory = Paths.get("generated/tmp/test/" + getName());
+	@Test
+	@DisabledOnOs(WINDOWS)
```
bjhargrave: Why did you disable this on Windows? This test currently runs on Windows.

aQute.libg/test/aQute/lib/io/IOTest.java
```diff
@@ -297,8 +322,10 @@ public void testCreateSymlinkOrCopyWillDeleteOriginalLink() throws Exception {
 		}
 	}
 
-	public void testCreateDirectory_Symlink() throws Exception {
-		Path rootDirectory = Paths.get("generated/tmp/test/" + getName());
+	@Test
+	@DisabledOnOs(WINDOWS)
+	public void testCreateDirectory_Symlink(TestInfo info) throws Exception {
+		Path rootDirectory = Paths.get("generated/tmp/test/" + info.getTestMethod());
```
bjhargrave: Also, this is wrong since `getTestMethod()` returns `Optional<Method>` and not a string. I think you need `info.getTestMethod().map(Method::getName).get()`

bjhargrave:
```java
	private String	testName;
	@BeforeEach
	public void before(TestInfo info) {
		testName = info.getTestMethod()
			.map(Method::getName)
			.get();
	}
```

aQute.libg/build.gradle
kriegfrj: I didn't want to go big bang on all subprojects. I suppose it can still be added selectively from the root build.gradleI. The ideal solution, IMO, as I mentioned, is for the Gradle plugin to add it when it notices bnd.bnd has specified Junit Platform on the classpath.

aQute.libg/test/aQute/lib/io/IOTest.java
kriegfrj: My bad. It didn't work on my box because I hadn't configured siblings to work properly. Easy to renable

aQute.libg/test/aQute/lib/io/IOTest.java
kriegfrj: Ditto.

cnf/ext/junit.bnd
kriegfrj: Your call. There are several modules in each group of dependencies that need to be kept at the same version, I thought using a macro for DRY was appropriate.

aQute.libg/test/aQute/lib/io/IOTest.java
kriegfrj: I had falsely assumed that Method.toString() would be a friendly String of the method name but it's not that friendly at all. I've used getDisplayName() which prints as `testCreateDirectory_Symlink(TestInfo)` which should be good enough for a temp directory.

kriegfrj: Changed it again - looked more closely at the second post there and saw that was better still & I've implemented that.

aQute.libg/build.gradle
kriegfrj: I've changed it so that all projects are now using junit-platform.

kriegfrj:
All changes implemented, ready for round 2.

kriegfrj:
Once I converted all subprojects to use junit-platform, the tests for `biz.aQute.bnd.gradle` started failing because it (along among all projects). I fixed it by adding `-testpath: ${junit}` to `bnd.bnd` (as I thought that would be the team's preference) but you could also possibly fix it by adding into `build.gradle`.

bjhargrave:
I will note, that I had to delete all my Junit launch configurations in Eclipse to properly launch Junit tests again. Since all my existing Junit launch configurations were using the Junit4 Runner, doing Run As > Junit Test for a project silently ignored the Junit 5 tests and only ran the Junit 4/3 tests!

Once I deleted all the Junit launch configurations, doing Run As > Junit Test for a project properly runs all tests. We will need to advise the other Bnd devs of this issue.

bjhargrave:
> Once I converted all subprojects to use junit-platform, the tests for `biz.aQute.bnd.gradle` started failing because it (alone among all projects) didn't have ${junit} on `-testpath` (in fact, it didn't have a `-testpath` at all). I fixed it by adding `-testpath: ${junit}` to `bnd.bnd` as I thought that would be the team's preference, but I also thought maybe the empty `-testpath` was a sign that we prefer to manage test dependencies through `build.gradle` instead for this project. This is also possible (in theory - haven't tried it in practice).

I assume that `test { useJUnitPlatform() }` in build.gradle requires Junit 5 jars on the test path which weren't otherwise there. The gradle project uses Spock which runs fine on the vintage platform. So your solution to set `-testpath` in the bnd file was correct.

kriegfrj:
> I will note, that I had to delete all my Junit launch configurations in Eclipse to properly launch Junit tests again. Since all my existing Junit launch configurations were using the Junit4 Runner, doing Run As > Junit Test for a project silently ignored the Junit 5 tests and only ran the Junit 4/3 tests!
> 
> Once I deleted all the Junit launch configurations, doing Run As > Junit Test for a project properly runs all tests. We will need to advise the other Bnd devs of this issue.

Yes, that's a trap - Junit 4 runner will silently ignore Jupiter tests. I don't know of an easy way to make that less silent.

Also might be worth pointing out you can set the default runner for the workspace. It's JUnit 5 for most workspaces already, but some may have set it to JUnit 4.

PR #3455, opened 2019-09-26
Use 'this' as lock in prepare

### LINKED ISSUE #3453: Nice example of a dreadful deadlock
image

pkriens:
Using two different locks for the same file creates deadlocks.

Fixes #3453

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
@@ -270,7 +270,7 @@ public class Project extends Processor {
 			return;
 		}
 
-		synchronized (preparedPaths) {
+		synchronized (this) {
 			if (preparedPaths.get()) {
 				// ensure output folders exist
 				getSrcOutput0();


pkriens:
I do not disagree here but locking on the atomic is also wrong. I think this change is better than it was.

However, I could get the global bbd lock from Central in OSGiRunDelegate?

bjhargrave:
> locking on the atomic is also wrong

I don't think it is wrong. I think it is the right granularity. The issue is that code using bndlib should not be synchronizing on bndlib objects as they do not know bndlib's synchronization rules.

> However, I could get the global bnd lock from Central in OSGiRunDelegate?

You could, but I am not sure it is necessary here anyway. Remember, the deadlock you saw was artificially constructed by stopping threads in a debugger. There is no normal case where this happens or I am sure we would have seen the reported deadlock by now.

The risk of changing the locking in prepare is not justified by the artificial deadlock you created. 

Better to fix bndtools code to not synchronize on bndlib objects and use the provided Central.bndLock method.

pkriens:
There is no artificial deadlock, this was **real**. All threads were humming along but the system was hanging. Clicking on the pause gave me this deadlock analysis. And if you look at the flow, it can happen and I am quite confident I've seen this case in the past. It can happen when you click launch a number of times.

It is a mess but just removing the lock in bndtools feels not correct.

bjhargrave:
How it is that 3 separate threads were all in the process of launching the same project? That does not sound like something that will happen to a human using Bndtools.

The issue is the non-bndlib code improperly synchronizing on bndlib objects. There is no valid reason someone (other than bndlib itself) should synchronize on a bndlib object. We don't specify anywhere that this is an OK thing to do. I am sure anyone can write some code which synchronizes on random bndlib objects and creates a deadlock.

pkriens:
It was terribly slow for some reason and I clicked launch in the run tab multiple times.

pkriens:
I am not sure why you continue to doubt this, this clearly can happen when you look at the code?

bjhargrave:
> I am not sure why you continue to doubt this, this clearly can happen when you look at the code?

I don't doubt that deadlocks can happen. I just think the proper solution is to fix the places that synchronize on Project objects in Bndtools.

bjhargrave:
BTW, if you plan to put this fix in anyway, you need to amend your commit to include a `Signed-off-by`. The DCO check has failed.

pkriens:
Well, I am first trying get consensus. I think we agree on that this is bad practice. 

I tried to ask what your idea is about using the global bnd lock, remove the locks and hope for the best, or have some other lock.

bjhargrave:
> I tried to ask what your idea is about using the global bnd lock, remove the locks and hope for the best, or have some other lock.

In this instance, I think you can simply remove the `synchronized (model)` from the `OSGiRunLaunchDelegate.initialiseBndLauncher` method since I don't think it does anything useful. No other callers to `Project.getProjectLauncher()` synchronize on the project object before calling that method. So that synchronization does nothing except surface this deadlock potential. 

Separately, we should probably review all the other places in the Bndtools code that synchronizes on a bndlib object to decide what action should be taken there. Actions may include removing the uninteresting synchronization (as I suggest for the example above) or replacing with Central.bndLock synchronization if warranted.

PR #3457, opened 2019-09-27
Issue 3443 - Incorrect behaviour for HttpsVerification.hostnameVerify

coheigea:
The hostname verification logic in HttpsVerification is incorrect. If "verify" is true then it actually disables hostname verification. If "verify" is false, then it always fails hostname verification.

I added two tests using a cert with an incorrect CN for the hostname. Both tests fail with the old code and pass with the new.


pkriens:
Thanks! You did take a deep dive and did not kill any other test cases. I see you figured out the magic Httpbin class :-)

Could you sign your commit? https://github.com/bndtools/bnd/pull/3457/checks?check_run_id=238573528

coheigea:
@pkriens done

PR #3464, opened 2019-10-03
A failing test case when using the 4.3.0 resolver plugin

timothyjward:


rotty3000:
@timothyjward I just ran this PR rebased on 4.3.0.REL tag (it was one or two commits earlier than REL) and I did not get any failures.

rotty3000:
.. wait! :( I did something wrong. Re-verifying.

rotty3000:
Ok, so what I've discovered is that resolving each `bndrun` file individually works as expected. It's only when the resolve task has both bndrun files together that it misbehaves. I'm still investigating.

At least, if you were to separate the files into separate resolve executions it should work.

rotty3000:
This test was included in the fix PR. Thx @timothyjward

PR #3484, opened 2019-10-16
fixed error printout on launch without any error having appeared

juergen-albert:
If the launcher ran in this peace of code, it always communicated an error "could not resolve the bundles: []" which is apparently no error.


bjhargrave:
Thanks!

PR #3486, opened 2019-10-17
[maven] do not store a processor which is predicated on a changing workspace

rotty3000:


biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
```diff
@@ -258,10 +258,6 @@ private String getNamePart(File runFile) {
 	}
 
 	private Processor getProcessor(Workspace workspace) {
-		if (processor != null) {
-			return processor;
-		}
-
```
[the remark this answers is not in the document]
bjhargrave: I have this same comment. The private processor field is now unused and should show a warning in Eclipse. Please remove the field.

biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java
```diff
@@ -511,13 +512,20 @@ private void loadPreferences() {
...
+	}
+
 	@Override
 	protected void postProcessProviders(Requirement requirement, Set<Capability> wired, List<Capability> candidates) {
 		if (candidates.isEmpty())
 			return;
 
 		// Call resolver hooks
-		for (ResolverHook resolverHook : registry.getPlugins(ResolverHook.class)) {
+		for (ResolverHook resolverHook : getResolverHooks()) {
```
bjhargrave: I am not sure I understand the value of this change. Is this necessary for correctness? Or is is some optimization? It seems the safe thing is to always get the current hooks rather than only get them once.

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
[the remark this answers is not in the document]
rotty3000: yup, was just an oversight.

biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java
rotty3000: This was an optimization. I noticed while debugging that the code was passing through the plugin calculation logic (which is not trivial) to find resolve hooks about a billion times.
Ask yourself if it makes sense if, once a resolve has already started, a new resolve hook should suddenly appear. Wouldn't that wreak havoc over the result having had only a subset be subject to any resolve hooks?

rotty3000: Of minor consequence, it also sped up the build a bit, but enough to notice.

bjhargrave: OK, It just wasn't clear to me whether this was part of the bug fix or just some optimization you tucked into this PR. It is the latter.

PR #3492, opened 2019-10-18
[m2e] Bndtools does not show bundles in dependencyManagement

rotty3000:


rotty3000:
thanks to @gamerson for finding the solution.

rotty3000:
@bjhargrave I'd like to include this fix in any maintenance release plans.

PR #3503, opened 2019-10-26
tests: Some tests depend upon jars in `-dependson` projects

bjhargrave:
So we configure gradle to make sure those project's `jar` tasks are
up-to-date before starting the `test` task.

Fixes https://github.com/bndtools/bnd/issues/3500


pkriens:
thanks

PR #3505, opened 2019-10-27
fix more cases of `Bndtools does not show bundles in dependencyManagement`

rotty3000:


rotty3000:
This was verified on top of 4.3.1.RC if we're inclined to add this to the release.

PR #3510, opened 2019-10-30
[maven] bnd-*-maven-plugin cannot load additional repository definitions using`-plugin` from bndrun

rotty3000:


rotty3000:
yeah, you're right. The issue I guess is that I don't typically use `-standalone true` since the behavior if the maven plugins implies it. However, it seems that having an explicit value of `-standalone true` behaves differently. I think we should fix that.

rotty3000:
I'll create another issue for the fact that omitting `-standalone` is not the same as `-standalone: true` even in scenarios where standalone is the only option (maven, non-workspace gradle).

PR #3531, opened 2019-11-05
[tester] Avoid adding tester bundle to runbundles if already present

### LINKED ISSUE #3527: [tester] Tester adds itself to runbundles even if already there
When you launch tests using the `-tester` directive, the tester bundle will add itself to the end of the `-runbundles` directive at launch time.

However, sometimes the tester bundle has already been explicitly included in `-runbundles` manually, in which case you end up with it installed twice. It would be good if the tester bundle first checked to see if it was already included before adding itself.

This behaviour is found in both `biz.aQute.tester` and `biz.aQute.tester.junit-platform`.

kriegfrj:
Fixes #3527.


bjhargrave:
It would be better to check for the duplication in the launcher to avoid duplicating this check.

This is what I do in https://github.com/bndtools/bnd/pull/3533/commits/0d3a18c627c9057c27222446f334425db3cfabc7

bjhargrave:
#3533 fixes this.

kriegfrj:
> It would be better to check for the duplication in the launcher to avoid duplicating this check.
> 

Agreed.

PR #3537, opened 2019-11-06
Win loop

rotty3000:


rotty3000:
@kriegfrj if you have a moment could you try this?

bndtools.m2e/src/bndtools/m2e/BndConfigurator.java
```diff
@@ -145,6 +145,9 @@ public AbstractBuildParticipant getBuildParticipant(final IMavenProjectFacade pr
 					return build;
 				}
 
+				final File classesDir = maven.getMojoParameterValue(projectFacade.getMavenProject(), execution,
+					"classesDir", File.class, monitor);
```
bjhargrave: What is `classesDir`? Are you reaching into BndMavenPlugin to read a field?

https://github.com/bndtools/bnd/blob/e5d1d948b5d323c5fefecf445c00feb3d0ba2e71/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/BndMavenPlugin.java#L95-L96

That seems highly fragile. And at the least deserves comments on both sides of the connection to document the connection. 

What if `includeClassesDir` is false?

rotty3000: Ok, I've added a comment w.r.t. `classesDir`.

`includeClassesDir` won't have any effect on this change since it's not about the same thing. Even when `includeClassesDir` is false, `classesDir` is still the output dir we're getting bytecode and resources from.

kriegfrj:
> @kriegfrj if you have a moment could you try this?

This seems to have fixed the build loop - I've now got the osgi-test workspace building on my local machine. Thanks Ray!

PR #3577, opened 2019-11-16
[bndtools.m2e] junit launcher doesn't understand maven profiles

### LINKED ISSUE #3576: [bndtools.m2e] junit launcher doesn't understand maven profiles
Eclipse M2E supports maven profiles. You can set a profile in the project configuration

Screenshot from 2019-11-15 21-18-50

But the bndtools launchers don't understand this and so do not show the launcher shortcuts.

rotty3000:
fixes #3576

### DIFF
--- a/bndtools.core/_plugin.xml
+++ b/bndtools.core/_plugin.xml
@@ -605,9 +605,13 @@
 
                            <!--  Java element (src-folder, package or JUnit class) -->
                            <adapt type="org.eclipse.jdt.core.IJavaElement">
-                              <test property="org.eclipse.jdt.core.hasTypeOnClasspath" value="junit.framework.Test" />
-                              <test property="org.eclipse.jdt.junit.canLaunchAsJUnit" forcePluginActivation="true" />
                               <test property="org.bndtools.core.isInBndJavaProject" forcePluginActivation="true" />
+                              <test property="org.eclipse.jdt.junit.canLaunchAsJUnit" forcePluginActivation="true" />
+                              <or>
+                                 <test property="org.eclipse.jdt.core.hasTypeOnClasspath" value="org.junit.Test" />
+                                 <test property="org.eclipse.jdt.core.hasTypeOnClasspath" value="junit.framework.Test" />
+                                 <test property="org.eclipse.jdt.core.hasTypeOnClasspath" value="org.junit.jupiter.api.Test" />
+                              </or>
                            </adapt>
 
 						   <!-- bnd.bnd or *.bndrun file -->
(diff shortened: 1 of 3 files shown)


bndtools.core/_plugin.xml
```diff
@@ -605,9 +605,13 @@
...
                            <!--  Java element (src-folder, package or JUnit class) -->
                            <adapt type="org.eclipse.jdt.core.IJavaElement">
-                              <test property="org.eclipse.jdt.core.hasTypeOnClasspath" value="junit.framework.Test" />
-                              <test property="org.eclipse.jdt.junit.canLaunchAsJUnit" forcePluginActivation="true" />
                               <test property="org.bndtools.core.isInBndJavaProject" forcePluginActivation="true" />
+                              <test property="org.eclipse.jdt.junit.canLaunchAsJUnit" forcePluginActivation="true" />
+                              <or>
+                                 <test property="org.eclipse.jdt.core.hasTypeOnClasspath" value="org.junit.Test" />
+                                 <test property="org.eclipse.jdt.core.hasTypeOnClasspath" value="junit.framework.Test" />
+                                 <test property="org.eclipse.jdt.core.hasTypeOnClasspath" value="org.junit.jupiter.api.Test" />
```
bjhargrave: There are more Junit 5 types to consider. You may need to map this `<test>` to a set of `<tests>`:

https://github.com/bndtools/bnd/blob/9c99bbd913df0c2c78114868b61da4d3d50e0fea/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/TestCases.java#L7-L11

bndtools.m2e/_plugin.xml
```diff
@@ -52,6 +52,7 @@
 							<or>
 								<test property="org.eclipse.jdt.core.hasTypeOnClasspath" value="org.junit.Test" />
 								<test property="org.eclipse.jdt.core.hasTypeOnClasspath" value="junit.framework.Test" />
+								<test property="org.eclipse.jdt.core.hasTypeOnClasspath" value="org.junit.jupiter.api.Test" />
```
bjhargrave: Need more junit 5 tests in `<or>` as in the above comment.

bndtools.core/_plugin.xml
rotty3000: I don't think so. Please not this is not a check of the classes and how they are annotated, it's just checking if a class is on the classpath, Which means you only need a single one from each junit framework version.

bndtools.m2e/_plugin.xml
rotty3000: No I don't think so.

bndtools.core/_plugin.xml
bjhargrave: OK

Issue #3584, opened 2019-11-18
Resolve handles snapshot bundles differently in bndtools 4.3 than in 4.2


tobigun:
The resolve behavior of snapshot bundles is different between bndtools 4.3 and 4.2.

To reproduce:
1. add a dependency to a snapshot bundle to the ```-runrequires``` section of a bndrun file:
e.g. ```osgi.identity;filter:='(osgi.identity=<bundlename>)'```
2. press the resolve button in the Eclipse bndtools plugin view of the bndrun file
3. the bundle dependency is now written as ```<bundlename>;version=[0.1.0,0.1.1)``` into the ```-runbundles``` section of the bndrun file

With bndtools 4.2 ```<bundlename>;version=snapshot``` was written to the ```-runbundles``` section.

A side effect of this change is that the change from "snapshot" to "[0.1.0,0.1.1)" is  recognized as a version change and all snapshot bundles are reordered in the resolve process.

It is unclear whether this is a bug or the desired behavior.

pkriens:
Are you sure this was not done with Gradle? I am looking into this because also have bug #3387.

I am working on always making it snapshot but I am not sure I understand why this was different in 4.2?

tobigun:
> Are you sure this was not done with Gradle?

Just tested it again with bndtools 4.2. and a simple workspace that only has one bundle.
runrequires contains "bnd.identity;id='application'" which becomes "application;version=snapshot" in the runbundles list after resolution.

So it is "snapshot" in bndtools 4.2 and it does not seem to be related to Gradle.

pkriens:
Thanks Tobias, that helped. I just fixed it.

pkriens:
Fixed in #3585

tobigun:
Hello Peter,
Thanks for fixing this.

In which release version was the fix included? Seems as if bndtools 5.0.1 still has the changed behavior.

bjhargrave:
There were a number of fixes required to get this right. It should all be properly fixed in 5.1.0 which was just released.

PR #3585, opened 2019-11-18
[snapshot] Difference between Bndtools and other drivers in resolve o…

pkriens:
…utput

The output of a resolve operations is a list of -runbundles. This list
is calculated nowadays in one place (RunResolution) but unfortunately
there is a different workspace used in Bndtools and in other drivers. This
meant that Bndtools workspace resources were marked with the micro range
while these same workspaces in Gradle were marked with 'snapshot'.

A previous change added a WorkspaceRepositoryMarker to the repos that 
model the workspace projects. The new one used for Gradle added a 
WORKSPACE_NAMESPACE capability but the old one in Bndtools did not. 

This fix adds the same capability the old WorkspaceR5Repository so that
the versioning to identify the resource is now independent of the driver.

Long term solution should be to replace the WorkspaceR5Repository with
something more modern so that we have a single unified workspace 
resource repo.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java
@@ -935,4 +935,22 @@ public class ResourceBuilder {
 
 	}
 
+	/**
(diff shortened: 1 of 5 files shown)


pkriens:
This seems to be the behavior of 4.2

Fixes #3584

Issue #3580, opened 2019-11-18
when running JUnit5 tests from the maven cli specifying -Dtest= with a single test method, no tests are run


rotty3000:
Here's an example from cli with trace enabled:
```
mvn verify -Pbnd-next -Dtest=org.osgi.test.junit5.context.BundleContextExtensionExampleTest:test1

[snip]

# framework=org.eclipse.osgi.launch.Equinox@399c4be1
# registered launcher with arguments for syncing
# will call main
# running in main thread
# test cases org.osgi.test.junit5.context.BundleContextExtensionExampleTest:test1
# run unresolved true
# using /home/rotty/projects/osgi-test/org.osgi.test.junit5/target/test-reports/test, path: /home/rotty/projects/osgi-test/org.osgi.test.junit5/target/test-reports/test
# received names of classes to test org.osgi.test.junit5.context.BundleContextExtensionExampleTest:test1
# testing request org.junit.platform.launcher.core.DefaultDiscoveryRequest@6a55299e
# TestPlan Execution Started: org.junit.platform.launcher.TestPlan@5553d0f5
# Execution Started: Bnd JUnit Platform Bundle Engine - [engine:bnd-bundle-engine]
# Execution Finished: Bnd JUnit Platform Bundle Engine - [engine:bnd-bundle-engine] - TestExecutionResult [status = SUCCESSFUL, throwable = null]
# TestPlan Execution Finished: org.junit.platform.launcher.TestPlan@5553d0f5

Test run finished after 9 ms
[         1 containers found      ]
[         0 containers skipped    ]
[         1 containers started    ]
[         0 containers aborted    ]
[         1 containers successful ]
[         0 containers failed     ]
[         0 tests found           ]
[         0 tests skipped         ]
[         0 tests started         ]
[         0 tests aborted         ]
[         0 tests successful      ]
[         0 tests failed          ]

[INFO] No Errors
```

rotty3000:
See branch https://github.com/rotty3000/osgi-test-1/tree/junit5.initial

rotty3000:
Do note that specifying just the class alone works as expected.

kriegfrj:
Is there magic in m2e to turn the Maven "test=" into a "tester.names" OSGi property setting?

kriegfrj:
> Is there magic in m2e to turn the Maven "test=" into a "tester.names" OSGi property setting?

Scratch that, I can see from the tester's trace output that the tester.names property has made it as far as the tester's Activator.

[...]

PR #3588, opened 2019-11-19
[junit-platform tester] Allow quoted test names

kriegfrj:
This PR is a partial fix for #3580.

### DIFF
--- a/biz.aQute.junit/src/aQute/junit/plugin/ProjectTesterImpl.java
+++ b/biz.aQute.junit/src/aQute/junit/plugin/ProjectTesterImpl.java
@@ -9,6 +9,7 @@ import static aQute.junit.constants.TesterConstants.TESTER_TRACE;
 import static aQute.junit.constants.TesterConstants.TESTER_UNRESOLVED;
 
 import java.util.Collection;
+import java.util.stream.Collectors;
 
 import org.osgi.annotation.bundle.Header;
 
@@ -16,7 +17,6 @@ import aQute.bnd.build.Project;
 import aQute.bnd.build.ProjectLauncher;
 import aQute.bnd.build.ProjectTester;
 import aQute.bnd.osgi.Constants;
-import aQute.bnd.osgi.Processor;
 import aQute.bnd.service.EclipseJUnitTester;
 
 @Header(name = Constants.TESTER_PLUGIN, value = "${@class}")
@@ -27,6 +27,10 @@ public class ProjectTesterImpl extends ProjectTester implements EclipseJUnitTest
 
 	public ProjectTesterImpl(Project project) throws Exception {
 		super(project);
+	}
+
+	static String maybeQuote(String s) {
+		return (s.indexOf(',') == -1) ? s : '"' + s + '"';
 	}
 
 	@Override
@@ -57,7 +61,9 @@ public class ProjectTesterImpl extends ProjectTester implements EclipseJUnitTest
 			Collection<String> testnames = getTests();
 			if (testnames.size() > 0) {
 				launcher.getRunProperties()
-					.put(TESTER_NAMES, Processor.join(testnames));
+					.put(TESTER_NAMES, testnames.stream()
+						.map(ProjectTesterImpl::maybeQuote)
+						.collect(Collectors.joining(",")));
 			}
 			// This is only necessary because we might be picked
 			// as default and that implies we're not on the -testpath
--- a/biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/plugin/ProjectTesterImpl.java
+++ b/biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/plugin/ProjectTesterImpl.java
@@ -10,6 +10,7 @@ import static aQute.junit.constants.TesterConstants.TESTER_TRACE;
 import static aQute.junit.constants.TesterConstants.TESTER_UNRESOLVED;
 
 import java.util.Collection;
+import java.util.stream.Collectors;
 
 import org.osgi.annotation.bundle.Header;
 
@@ -18,7 +19,6 @@ import aQute.bnd.build.Project;
 import aQute.bnd.build.ProjectLauncher;
 import aQute.bnd.build.ProjectTester;
 import aQute.bnd.osgi.Constants;
-import aQute.bnd.osgi.Processor;
 import aQute.bnd.service.EclipseJUnitTester;
 
 @Header(name = Constants.TESTER_PLUGIN, value = "${@class}")
@@ -32,6 +32,10 @@ public class ProjectTesterImpl extends ProjectTester implements EclipseJUnitTest
 	public ProjectTesterImpl(Project project, Container me) throws Exception {
 		super(project);
 		this.me = me;
+	}
+
+	static String maybeQuote(String s) {
+		return (s.indexOf(',') == -1) ? s : '"' + s + '"';
 	}
 
 	@Override
@@ -68,7 +72,9 @@ public class ProjectTesterImpl extends ProjectTester implements EclipseJUnitTest
 			Collection<String> testnames = getTests();
 			if (testnames.size() > 0) {
 				launcher.getRunProperties()
-					.put(TESTER_NAMES, Processor.join(testnames));
+					.put(TESTER_NAMES, testnames.stream()
+						.map(ProjectTesterImpl::maybeQuote)
+						.collect(Collectors.joining(",")));
 			}
 
 			//
--- a/biz.aQute.tester/src/aQute/tester/plugin/ProjectTesterImpl.java
+++ b/biz.aQute.tester/src/aQute/tester/plugin/ProjectTesterImpl.java
@@ -9,6 +9,7 @@ import static aQute.junit.constants.TesterConstants.TESTER_TRACE;
 import static aQute.junit.constants.TesterConstants.TESTER_UNRESOLVED;
 
 import java.util.Collection;
+import java.util.stream.Collectors;
 
 import org.osgi.annotation.bundle.Header;
 
@@ -17,7 +18,6 @@ import aQute.bnd.build.Project;
 import aQute.bnd.build.ProjectLauncher;
 import aQute.bnd.build.ProjectTester;
 import aQute.bnd.osgi.Constants;
-import aQute.bnd.osgi.Processor;
 import aQute.bnd.service.EclipseJUnitTester;
 
 @Header(name = Constants.TESTER_PLUGIN, value = "${@class}")
@@ -30,6 +30,10 @@ public class ProjectTesterImpl extends ProjectTester implements EclipseJUnitTest
 	public ProjectTesterImpl(Project project, Container me) throws Exception {
 		super(project);
 		this.me = me;
+	}
+
+	static String maybeQuote(String s) {
+		return (s.indexOf(',') == -1) ? s : '"' + s + '"';
 	}
 
 	@Override
@@ -60,7 +64,9 @@ public class ProjectTesterImpl extends ProjectTester implements EclipseJUnitTest
 			Collection<String> testnames = getTests();
 			if (testnames.size() > 0) {
 				launcher.getRunProperties()
-					.put(TESTER_NAMES, Processor.join(testnames));
+					.put(TESTER_NAMES, testnames.stream()
+						.map(ProjectTesterImpl::maybeQuote)
+						.collect(Collectors.joining(",")));
 			}
 
 			//
--- a/aQute.libg/src/aQute/lib/strings/Strings.java
+++ b/aQute.libg/src/aQute/lib/strings/Strings.java
@@ -164,10 +164,14 @@ public class Strings {
 	}
 
 	public static Stream<String> splitQuotedAsStream(String s) {
+		return splitQuotedAsStream(s, true);
+	}
+
+	public static Stream<String> splitQuotedAsStream(String s, boolean retainQuotes) {
 		if ((s == null) || (s = s.trim()).isEmpty()) {
 			return Stream.empty();
 		}
-		return new QuotedTokenizer(s, COMMA, false, true).stream()
+		return new QuotedTokenizer(s, COMMA, false, retainQuotes).stream()
 			.filter(Strings::notEmpty);
 	}
 
--- a/biz.aQute.bndall.tests/test/biz/aQute/launcher/AlsoLauncherTest.java
+++ b/biz.aQute.bndall.tests/test/biz/aQute/launcher/AlsoLauncherTest.java
@@ -772,6 +772,16 @@ public class AlsoLauncherTest {
 	}
 
 	@Test
+	public void testTesterQuoted() throws Exception {
+		ProjectTester pt = project.getProjectTester();
+		pt.addTest("'test.Junit4TestCase'");
+		pt.addTest("\"test.TestCase2:m1\"");
+		pt.addTest("\"test.TestCase2:m2\"");
+		assertThat(pt.getTests()).containsExactly("test.Junit4TestCase", "test.TestCase2:m1", "test.TestCase2:m2");
+		assertThat(pt.test()).isEqualTo(1);
+	}
+
+	@Test
 	public void testTimeoutActivator() throws Exception {
 		try (ProjectLauncher l = project.getProjectLauncher()) {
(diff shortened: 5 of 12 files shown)


bjhargrave:
>     3\. Do we need to do anything to the gradle or maven plugins for quoting test names?

I think the answer here is that the ProjectTesterImpls need to quote, if necessary, each element in the generated `tester.names` property.

kriegfrj:
> > 
> > 3\. Do we need to do anything to the gradle or maven plugins for quoting test names?
> > 
> 
> I think the answer here is that the ProjectTesterImpls need to quote, if necessary, each element in the generated `tester.names` property.

Hmm, I think that's part of the answer. However I've started looking at this and I think it's worth pausing to take stock and quickly looking at all of the consumers to figure out an implementation that is consistent. There are a few different players here:

1. JUnit 3, 4 & 5
2. maven-surefire-plugin and maven-failsafe-plugin
3. Gradle's built-in test task
4. The old Bnd tester
5. Bndtools
6. Bnd gradle plugin
7. Bnd maven tester plugin

===JUnit===

* JUnit Platform syntax is: `<fully-qualified classname>#<method name>(parameter1Type, parameter2Type)`
* The bit in parentheses is required for parameterized methods, but optional for methods that don't take parameters.
* There is no easy way to filter based on a certain test iteration for a parameterized test.

===Maven plugins===

Documentation is sparse (https://maven.apache.org/surefire/maven-surefire-plugin/examples/single-test.html and https://maven.apache.org/surefire/maven-failsafe-plugin/examples/single-test.html), but:

* They use syntax -Dtest=<filter> (-Dit.test for failsafe plugin).
* `#` is used as the method name delimiter. Thus, syntax is loosely based on JUnit.
* There doesn't appear to be a way to specify a particular method if it is overloaded with different parameter counts.
* Individual parameter iterations can be filtered using [] syntax. Filter is based on the display name of the test.
* What documentation exists is for JUnit 4 parameterized tests. I"m not sure to what extent the above applies to JUnit 5. However, I have verified empirically that JUnit 5 parameterized tests work, but the [] filter syntax does not.

===Gradle===

Appears to be two syntacies - one similar to maven-surefire (-D), and the `--tests` command line. The latter seems to be more recent.

The syntax seems to be similar to the maven plugins, except that the method delimiter is "." rather than "#". Given the inherent ambiguity arising from this (as . also delimits packages and classes), they also seem to use the convention that the test classes starts with capital letters.

https://docs.gradle.org/current/userguide/java_testing.html#test_filtering

===Old Bnd tester===

* Exact matches only, using FQN of class
* Uses ":" as the method delimiter for `tester.names`.
* No support for parameterized tests.

===Bndtools===

* Does whatever the tester does.
* Not sure how well it supports parameterized test methods out-of-the-box.
* We have the most flexibility here because it's all handled under-the-hood.

===Bnd maven and gradle testing plugins===

* Seem to be caught between emulating the syntax of their respective environment, or simply using Bnd tester syntax.
* Eg, @rotty3000 was using the `-Dtest=` of maven, but the `:` delimiter of the Bnd tester syntax.

So we need to make a decision here about whether these plugins try to emulate the maven/gradle syntax, or else whether they simply use the Bnd tester syntax.

Thoughts?

bjhargrave:
For OSGi testings (vs normal JUnit testing), all test execution flows through 

https://github.com/bndtools/bnd/blob/2171d1a6462402843e290284919d3c8ab11c2b7b/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L2356

which takes a `List<String>` of test names. 

Gradle:

https://github.com/bndtools/bnd/blob/2171d1a6462402843e290284919d3c8ab11c2b7b/biz.aQute.bnd.gradle/src/aQute/bnd/gradle/TestOSGi.groovy#L94

Maven:

https://github.com/bndtools/bnd/blob/2171d1a6462402843e290284919d3c8ab11c2b7b/maven/bnd-testing-maven-plugin/src/main/java/aQute/bnd/maven/testing/plugin/TestingMojo.java#L188

Bndtools uses ProjectTester.addTest directly:

https://github.com/bndtools/bnd/blob/2171d1a6462402843e290284919d3c8ab11c2b7b/bndtools.core/src/bndtools/launch/OSGiJUnitLaunchDelegate.java#L305-L321

This `List<String>` of test names is converted into the `tester.names` property by the ProjectTester impls

https://github.com/bndtools/bnd/blob/2171d1a6462402843e290284919d3c8ab11c2b7b/biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/plugin/ProjectTesterImpl.java#L68-L72

https://github.com/bndtools/bnd/blob/2171d1a6462402843e290284919d3c8ab11c2b7b/biz.aQute.junit/src/aQute/junit/plugin/ProjectTesterImpl.java#L57-L61

So this is where any quoting should be done if necessary since this is where the list is flattened to a property value.

Each test activator (`aQute.junit` or `aQute.tester.junit.platform`) should then be responsible for  parsing the property back into a list and mapping the value into the form expected by the respective testing engine.

So the form in the `tester.names` property should not need any changing other than to support quoting elements if the element uses a `,` in parameters.

We can change Bndtools to provide more precise element values since it should know the full parameters of the selection in the UI.

Gradle and Maven users can supply more precise forms on the command line invocation which will be passed into the ProjectTester.addTest method.

So, in summary, the ProjectTester impls need to quote elements, if necessary, when assembling the `tester.names` property value. The test activators need to handle quoted elements when parsing the `tester.names` property value back into a list. The test activators need to handle any mapping to the forms expected by there respective test engines.

rotty3000:
Speaking mostly to the maven scenario, in an ideal world, bnd would be able to run it's tests within `maven-failsafe-plugin` using its cli interface. However, we also don't want to spend an inordinate amount of time nor be burdened with baggage just to make that work. So, barring the direct use of `maven-failsafe-plugin` we'd try to emulate it as closely as possible in order to make developers comfortable.

So, I think the overarching goal here is that while we empower developers with the best possible OSGi testing experience we also make them comfortable. Whatever _that_ is is what we should do. Oh, and let's also be practical.

kriegfrj:
Sorry, I was in a bit of a rush when I posted that comparison of syntacies yesterday, so perhaps the main point I was trying to make got lost. 

@bjhargrave , I definitely agree that the `ProjectTesterImpl` is the best place to quote the elements of `tester.names`. I'm more concerned about divergence other parts of the `tester.names` syntax. Like what @rotty3000 pointed out here:

> Speaking mostly to the maven scenario, in an ideal world, bnd would be able to run it's tests within `maven-failsafe-plugin` using its cli interface. However, we also don't want to spend an inordinate amount of time nor be burdened with baggage just to make that work. So, barring the direct use of `maven-failsafe-plugin` we'd try to emulate it as closely as possible in order to make developers comfortable.
> 

At the moment, as @bjhargrave pointed out here https://github.com/bndtools/bnd/issues/3580#issuecomment-557142977, the gradle and maven plugins simply pass in the `-Dtest=` / `--tests` parameters directly to the tester unmodified.

I think this is what I was getting at. The current syntax used by the Bnd testers for `tester.names` is incompatible with every other syntax - maven-failsafe/surefire-plugin, Gradle, and JUnit 5. `BundleEngine` uses JUnit 5 internally (method delimiter = `#`) and the junit.platform.Activator translates from the Bnd tester syntax (method delimiter = `:`) to JUnit 5 syntax. I'm questioning the value of maintaining a fourth syntax that is incompatible with everything else. I think it would be better if we simply adopted JUnit 5 syntax for the elements of `tester.names`. If necessary/convenient, we should also get the Maven/Gradle plugin wrappers to translate from maven/Gradle syntax to the tester syntax.

Perhaps this part of the discussion should be split into a different ticket and leave this one to deal with the quoting.

bjhargrave:
> fourth syntax

What are all these syntax's?

For Bnd's Gradle TestOSGi task, `--tests` uses the normal Bnd `tester.names` format which is `class:method`. The format Gradle uses for plain JUnit testing to complicated with wildcard matching and we don't attempt to support it for OSGi testing. The reason we cannot support doing anything with wildcards is that we cannot map them to concrete test names in the gradle plugin as the plugin does not know what the execution runtime looks like. The TestOSGi task is just a simple driver of Bnd and we should not teach it to duplicate Bnd function.

For Bnd's Maven bnd-testing-maven-plugin, it also uses the normal Bnd `tester.names` format which is `class:method`. The maven-failsafe-plugin uses a complicated format supporting wildcard which I also don't think we can support. I am not even sure why we are discussing maven-failsafe-plugin here since we have bnd-testing-maven-plugin and no plan or understanding of how maven-failsafe-plugin could replace it for OSGi testing. Like as in the Bnd gradle support, the maven plugin is also a simple driver of Bnd. It also does not know about the execution environment to turn wildcards into concrete test names.

So for both our gradle and maven support we use the original Bnd `tester.names` format of `class:method` which we should continue to use for compatibility reasons. We should, however, expand `tester.names` format to allow parameters so that Bndtools can pass more precise information. `class:method(type1,type2)`. This format extension requires we support quoting elements in the `tester.names` property value so we can parse it correctly in the junit activator side.

So I see one format, extended to support parameters, here.

Now, at the engine level (aQute.junit, aQute.tester.junit.platform), we need to map this one syntax into a form the engine understands. The `replace(':',"#')` in junit-platform is basically an internal implementation choice since ultimately, we must supply DiscoverySelectors to the engine. How we map the Bnd `tester.names` format into DiscoverySelectors is our internal implementation problem.

But externally, the Bnd `tester.names` format has one syntax we plan to extend for method parameter specification. And that syntax is what Gradle users will put in `--tests` and Maven users will put in `-Dtest`. So nothing needs to be done in the maven or gradle plugins. They just pass the collected test values into the Project.test method in a `List<String>` (using the Bnd `tester.names` format) and we take it from there.

If we think we want to support some form of wildcard in the maven and gradle plugins, then we need to add that to the ONE Bnd `tester.names` format so that there is consistency and also so that the activators who are IN the runtime can map the wildcards to concrete tests.

kriegfrj:
To answer your question: why are we discussing maven-failsafe/surefire-plugin, I'll quote @rotty3000 :  "barring the direct use of maven-failsafe-plugin we'd try to emulate it as closely as possible in order to make developers comfortable." In other words, to reduce the learning curve, the bnd-tester-maven-plugin should try and implement the same syntax if possible, or at least as similar as possible. Ditto for Gradle --tests support.

What I'm getting from your post is that this is impractical, partly because you need the OSGi environment to properly scan for the tests, and the maven/gradle wrappers are responsible for launching that environment. Makes sense.

I thought perhaps we could do it in a limited form - eg, support "#" for Maven, "." for Gradle. This could be done either by supporting these alternatives natively in the testers' format, or else having the maven/gradle plugins perform the translation to the native testers' format. This should be much easier that supporting all of the wildcard syntax (though I think that wildcards and tag support is something we should consider at some point too).

biz.aQute.junit/src/aQute/junit/plugin/ProjectTesterImpl.java
```diff
@@ -29,6 +30,10 @@ public ProjectTesterImpl(Project project) throws Exception {
 		super(project);
 	}
 
+	static String maybeQuote(String s) {
+		return (s.indexOf(',') == -1) ? s : '"' + s + '"';
```
bjhargrave: It is a bit more complicated than this. The user may have already quoted the string! `--tests="class:method(int,int)"`.

See https://github.com/bndtools/bnd/blob/593fc85bbb09dd6b4c1e65ae4637142987e2019a/aQute.libg/src/aQute/libg/qtokens/QuotedTokenizer.java#L222-L250 for a robust quoting example.

biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/plugin/ProjectTesterImpl.java
```diff
@@ -34,6 +35,10 @@ public ProjectTesterImpl(Project project, Container me) throws Exception {
 		this.me = me;
 	}
 
+	static String maybeQuote(String s) {
+		return (s.indexOf(',') == -1) ? s : '"' + s + '"';
```
bjhargrave: Need to handle better as above.

biz.aQute.tester/src/aQute/tester/plugin/ProjectTesterImpl.java
```diff
@@ -32,6 +33,10 @@ public ProjectTesterImpl(Project project, Container me) throws Exception {
 		this.me = me;
 	}
 
+	static String maybeQuote(String s) {
+		return (s.indexOf(',') == -1) ? s : '"' + s + '"';
```
bjhargrave: Need to handle better as above.

biz.aQute.junit/src/aQute/junit/plugin/ProjectTesterImpl.java
```diff
@@ -57,7 +62,9 @@ public boolean prepare() throws Exception {
 			Collection<String> testnames = getTests();
 			if (testnames.size() > 0) {
 				launcher.getRunProperties()
-					.put(TESTER_NAMES, Processor.join(testnames));
+					.put(TESTER_NAMES, Processor.join(testnames.stream()
```
bjhargrave: It seems busy to stream/map the list into a list and then join into a string. We can go straight there

```java
.put(TESTER_NAMES, testnames.stream()
  .map(ProjectTesterImpl::maybeQuote)
  .collect(Collectors.joining(",")));
```

biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/plugin/ProjectTesterImpl.java
```diff
@@ -68,7 +73,9 @@ public boolean prepare() throws Exception {
 			Collection<String> testnames = getTests();
 			if (testnames.size() > 0) {
 				launcher.getRunProperties()
-					.put(TESTER_NAMES, Processor.join(testnames));
+					.put(TESTER_NAMES, Processor.join(testnames.stream()
```
bjhargrave: It seems busy to stream/map the list into a list and then join into a string. We can go straight there

```java
.put(TESTER_NAMES, testnames.stream()
  .map(ProjectTesterImpl::maybeQuote)
  .collect(Collectors.joining(",")));
```

biz.aQute.tester/src/aQute/tester/plugin/ProjectTesterImpl.java
```diff
@@ -60,7 +65,9 @@ public boolean prepare() throws Exception {
 			Collection<String> testnames = getTests();
 			if (testnames.size() > 0) {
 				launcher.getRunProperties()
-					.put(TESTER_NAMES, Processor.join(testnames));
+					.put(TESTER_NAMES, Processor.join(testnames.stream()
```
bjhargrave: It seems busy to stream/map the list into a list and then join into a string. We can go straight there

```java
.put(TESTER_NAMES, testnames.stream()
  .map(ProjectTesterImpl::maybeQuote)
  .collect(Collectors.joining(",")));
```

biz.aQute.junit/src/aQute/junit/plugin/ProjectTesterImpl.java
kriegfrj: :+1: wasn't sure if `Processor.join()` was doing some other magic.

biz.aQute.junit/src/aQute/junit/plugin/ProjectTesterImpl.java
kriegfrj: Wouldn't it be better if the process responsible for populating testernames ensure that the strings that go into the list are normalized to begin with?

bjhargrave: This means that every driver must do this instead of in fewer places. Better to do this at the choke points.

kriegfrj: Better to do this at the choke points, yes, but I think that the choke point is where the model is populated (`ProjectTester.addTest()`). That way, the drivers don't have to worry about it either (as they all use `addTest()`). If I do it in `ProjectTesterImpl` I have to do it three times, plus any other third-party consumers of `getTests()` will also have to handle it.

kriegfrj: I have added code to strip quotes (if present) in `ProjectTester.addTest()`.

PR #3590, opened 2019-11-19
[embedded activators] Framework start point

### LINKED ISSUE #3582: Embedded Activator: (still) different behavior with in aQute launcher
With the launcher included in bndtools 4.2 the system bundle was started before any Embedded Activator was started. This was also the case for Immediate Activators (which are now the startBeforeBundles Activators):
https://github.com/bndtools/bnd/blob/4.2.0.REL/biz.aQute.launcher/src/aQute/launcher/Launcher.java#L399

With bndtools 4.3 the system bundle was originally started after all Embedded Activators.
A few days ago a patch was applied that at least the system bundles is started for the startAfterBundles:
https://github.com/bndtools/bnd/commit/a121ba2448513ec82519dc1bf0492b5732eab58d#diff-44c67ee8afec40b1760cc3196bca58a3

But this still does not address the startBeforeBundles as the system bundle is still started after these bundles.

Isn't it possible to start the system bundle before all Embedded Activators to have the same behavior as before?

Code example (systemBundle.start() moved above start of startBeforeBundleStart):
>List<Bundle> tobestarted = update(System.currentTimeMillis() + 100);
>
>systemBundle.start();
>trace("system bundle started ok");
>
>for (BundleActivator activator : startBeforeBundleStart) {
>       result = start(systemContext, result, activator);
>}
>
>startBundles(tobestarted);
> 
>~~systemBundle.start();~~
>~~trace("system bundle started ok");~~
> 
>for (BundleActivator activator : startAfterBundleStart) {
>       result = start(systemContext, result, activator);
>}
> 
>startLevelhandler.afterStart();

pkriens:
The original model had 2 phases:

* framework inited
* framework started
* IMMEDIATE=true -> before bundles were started
* IMMEDIATE=false (or not set)-> after all bundles were started

After the start level and the work done for launching Eclipse
we had the following situation:

* framework inited
* IMMEDIATE = "AFTER_FRAMEWORK_INIT"
* IMMEDIATE=true || BEFORE_BUNDLES_STARTED -> before bundles were started
* IMMEDIATE=false (or not set) || AFTER_BUNDLES_STARTED -> after all bundles were started
* framework started

This meant that there was no different between BEFORE_BUNDLES_STARTED and
AFTER_BUNDLES_STARTED. Therefore, for 4.3.1 we moved to during the meeting:

* framework inited
* IMMEDIATE = "AFTER_FRAMEWORK_INIT"
* IMMEDIATE=true || BEFORE_BUNDLES_STARTED -> before bundles were started
* framework started
* IMMEDIATE=false (or not set) || AFTER_BUNDLES_STARTED -> after all bundles were started

However, this still made it impossible to run the Embedded Activator
as before since some embedded activators assumed that the framework was
started. This was easy to fix by starting the framework from the embedded
activator. However, it is not backward compatible and embedded activators
should not touch the framework this way out of cohesion. There might be
other embedded activators.

This patch therefore makes it again compatible with 4.3 butt now with the new
option AFTER_FRAMEWORK_INIT and a proper extensible model.

* framework inited
* IMMEDIATE = "AFTER_FRAMEWORK_INIT"
* framework started
* IMMEDIATE=true || BEFORE_BUNDLES_STARTED -> before bundles were started
* IMMEDIATE=false (or not set) || AFTER_BUNDLES_STARTED -> after all bundles were started

I've attempted to describe (or define) it now better in the documentation and
put some caveats around it. I've also refactored the code a bit to have the 
decision points closer to each other.

Fixes #3582

### DIFF
--- a/biz.aQute.launcher/src/aQute/launcher/Launcher.java
+++ b/biz.aQute.launcher/src/aQute/launcher/Launcher.java
@@ -114,6 +114,28 @@ public class Launcher implements ServiceListener {
 	private AtomicReference<DatagramSocket>	commsSocket			= new AtomicReference<>();
 	private StartLevelRuntimeHandler		startLevelhandler;
 
+	enum EmbeddedActivatorPhase {
+
+		AFTER_FRAMEWORK_INIT,
+		BEFORE_BUNDLES_START,
+		AFTER_BUNDLES_START,
+		UNKNOWN;
+
+		static EmbeddedActivatorPhase valueOf( Object o) {
+			if ( AFTER_FRAMEWORK_INIT.name().equals(o))
+				return AFTER_FRAMEWORK_INIT;
+
+			if (Boolean.TRUE.equals(o) || BEFORE_BUNDLES_START.name()
+				.equals(o))
+				return BEFORE_BUNDLES_START;
+
+			if (Boolean.FALSE.equals(o) || AFTER_BUNDLES_START.name()
+				.equals(o))
+				return AFTER_BUNDLES_START;
+
+			return UNKNOWN;
+		}
+	}
 	/**
 	 * Called from the commandline (and via EmbeddedLauncher in the embedded
 	 * mode.)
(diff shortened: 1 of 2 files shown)


bjhargrave:
And kudos on the nice detailed commit comment!

Issue #3339, opened 2019-07-17
bndrun configurations cannot be exported


amitjoy:
I have a project comprising multiple bndrun configurations and some sub bnd files. 

The project structure is as follows:

```
project.application
   bnd.bnd
   abc.bnd
   def.bnd
   a.bndrun
   b.bndrun
   c.bndrun
```

The contents of the bnd.bnd -

```
-sub        : *.bnd
-export     : \
	a.bndrun,\
	b.bndrun,\
	c.bndrun
```

Since `a.bndrun` includes the aforementioned sub bundles, the build fails in Eclipse with the following error message:

```
/Users/amit/QIVICON/repos/sample/personal/project.application/target/project.application.abc.jar" does not exist, given error is null

```

PR #3591, opened 2019-11-19
[export] bndrun configurations cannot be exported #3339

pkriens:
The -export instruction can export bndrun files. However,
these exports were done in the Project Builder. At that
time any bundles are build but not yet present on the file
system. If the bndrun file included the bundles (Jars) 
then it got an error on a clean build or a previous
version.

This patch moves the -export execution into Project
and executes it after the bundles/jars are stored
on the file system.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
@@ -2217,6 +2217,11 @@ public class Project extends Processor {
 		return null;
 	}
 
+	/**
+	 * The keep flag is really awkward since it overrides the -runkeep flag in
+	 * the file. Use doExport instead
+	 */
+	@Deprecated
 	public void export(String runFilePath, boolean keep, File output) throws Exception {
 		Map<String, String> options = Collections.singletonMap("keep", Boolean.toString(keep));
 		Entry<String, Resource> export;
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
@@ -43,8 +43,6 @@ import java.util.TreeMap;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.function.Function;
-import java.util.jar.Attributes;
-import java.util.jar.Attributes.Name;
 import java.util.jar.Manifest;
 import java.util.regex.Pattern;
 
@@ -100,7 +98,6 @@ import aQute.bnd.service.release.ReleaseBracketingPlugin;
 import aQute.bnd.service.specifications.RunSpecification;
 import aQute.bnd.version.Version;
 import aQute.bnd.version.VersionRange;
-import aQute.lib.collections.ExtList;
 import aQute.lib.collections.Iterables;
 import aQute.lib.converter.Converter;
 import aQute.lib.exceptions.ConsumerWithException;
@@ -1931,16 +1928,16 @@ public class Project extends Processor {
 
 			for (Map.Entry<File, Resource> ee : exports.entrySet()) {
 				File outputFile = ee.getKey();
-					File actual = write(f -> {
-						IO.copy(ee.getValue()
-							.openInputStream(), f);
-					}, outputFile);
+				File actual = write(f -> {
+					IO.copy(ee.getValue()
+						.openInputStream(), f);
+				}, outputFile);
 
-					if (actual != null) {
-						buildFilesSet.add(actual);
-					} else {
-						error("Could not save %s", ee.getKey());
-					}
+				if (actual != null) {
+					buildFilesSet.add(actual);
+				} else {
+					error("Could not save %s", ee.getKey());
+				}
 			}
 
 			if (!isOk()) {
@@ -2249,7 +2246,7 @@ public class Project extends Processor {
 	 * since the extension is defined by the exporter)
 	 *
 	 * @param type an exporter type like `bnd.executablejar` or null as default
-	 *            `bnd.pack`, the original export function
+	 *            `bnd.executablejar.pack`, the original export function
 	 * @param options parameters to the exporters
 	 * @return and entry with the name and resource for the export
 	 * @throws Exception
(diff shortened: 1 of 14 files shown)


biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -1887,16 +1888,25 @@ boolean isStale(Set<Project> visited) throws Exception {
...
+					}
 				}
-				builtFiles.add(file);
-				if (lastModified < jar.lastModified()) {
-					lastModified = jar.lastModified();
+			}
+
+			if (isOk()) {
+				for (Run export : builder.getExportedRuns()) {
+					lastModified = export(builder, builtFiles, lastModified, export);
```
bjhargrave: Perhaps this loop and the private export method the loop calls should be moved into ProjectBuilder as a method which is called from this site? This way we keep all the building logic in ProjectBuilder but we call it at the proper time in the Project build process?

pkriens: Well, there is a tension because the ProjectBuilder should not touch files. I am also thinking of using the Exporter services instead of hardcoding the model. Just need to check if there is a difference in how they build. This -export supports -profile, but that is useful for the ExecutableJar exporter as well.

pkriens:
I made a lot of changes. Afraid you have to look again.

[-export] Reconciliation of -export and export functions

Exporting of projects happened all over the place. This
is an attempt to align the -export with the existing
exporters that had ignored this instruction. It also
fixes 2 issues

## Fixes #3339 bndrun configurations cannot be exported

The executable file was made before the bundle jars
were stored on the file system. Did made it impossible to
include the just build JARs.

The processing is moved after the saving of the bundle
jars in the buildLocal method. The biz.aQute.launcher.AlsoLauncherTest.testExportWithIncludedBundlesInBndrun()
method tests for this case.

## Fixes 3178 "-runkeep:true" will not be considered during export as jar file

There was a method in Project that took a boolean keep
flag. Using this route, the keep flag was basically always
reset even if the -runkeep flag was set in the Project.
I could only find a reference in a PackageTask for ant
but I assume this is also used form Gradle?

I've deprecated this method because we should use the
existing properties, just the keep as a configuration
is pretty minimal. Worst case we could put the options
in the Run file but making a single setting a boolean
in the method prototype does not scale.

The deprecated method is adjusted. It now sets -runkeep
if either the keep flag is true or the property is set
to true. 

## More properties on the -export

Added a few properties on the -export instruction and
documented them.

Since the old way had a rater awkward output name it
was necessary to override the name with a `name` attribute.

The idea was to reconcile the exporters with this
instruction so there is a type attribute now. If neither
the name nor the type is set then the code takes on
backward compatibility mode and calculates the old name,
basically the name of the bndrun file (including the bndrun)
suffixed with .jar. However, it ran the name through
getOutputFile which did macro stuff. This is now
only used in backward compatible mode since some of
the exporters have different suffixes. So in non-backward
compatible mode the file name is not run through getOutputFile.

It is now also possible to set the bsn of the output
if the exporter supports it.

These options are tested in biz.aQute.launcher.AlsoLauncherTest.testExportOptions()

## Two export modes

The Project supported the pack() method and the
export(). The export method uses the plugin exporters and
should be the way forward. However, the pack method did a
bit more and was used by -export. To handle this in a unified
way, the ExecutableJarExporter was extended to support a new
type that matched the old -export pack() method. This became
the default type for -export.

biz.aQute.bndlib/src/aQute/bnd/exporter/executable/ExecutableJarExporter.java
```diff
@@ -28,26 +41,48 @@ public ExecutableJarExporter() {}
...
+
+				jar = launcher.executable();
+				launcher.removeClose(jar);
+				project.getInfo(launcher);
 			}
-			name = name + Constants.DEFAULT_JAR_EXTENSION;
-			return new SimpleEntry<>(name, new JarResource(jar, true));
+		} else if (EXECUTABLE_PACK.equals(type)) {
+
+			String profile = options.getOrDefault(Constants.PROFILE, project.getProperty(Constants.PROFILE));
```
bjhargrave: There is a `Processor.getProfile()` method which may make sense for the default value.

Although this method has an odd implementation and no one seems to use it.

docs/_instructions/export.md
```diff
@@ -1,6 +1,51 @@
...
+* `version` – Set the version of the resulting JAR file if possible
+* other – Any remaining properties are added to the properties of the run file. For example, `-profile` can be used to select a specific property profile.
+
+The `-exporttype` can be used to set some default attributes for a specific type.
+
+## Exporters
+
+The Exporters use a plugin mechanism and therefore the list is not closed. The following exporters are supported by bnd out of the box:
+
+* `bnd.executablejar.pack` – Exports a JAR file using the launcher from the `-runpath` or the default if no launcher is specified.
```
bjhargrave: Perhaps this name should be more distinct from `bnd.executablejar`? I do not understand the difference between pack and executablejar and this description does not help :-)

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -2186,23 +2243,47 @@ public void release() throws Exception {
 		release(false);
 	}
 
+	/**
+	 * Export this project via the exporters. The return is the calculated name
+	 * (can be overridden with the option `name`, which must be the basename
+	 * since the extension is defined by the exporter)
+	 *
+	 * @param type an exporter type like `bnd.executablejar` or null as default
+	 *            `bnd.pack`, the original export function
```
bjhargrave: This comment does not match the name.

biz.aQute.bndlib/src/aQute/bnd/exporter/executable/ExecutableJarExporter.java
```diff
@@ -28,26 +41,48 @@ public ExecutableJarExporter() {}
...
+			jar = project.pack(profile);
```
bjhargrave: Wouldn't it make more sense to have the logic in the pack method pulled down here and then have the pack method (like the export and exportRunBundles methods) call the exporter? It is odd to have doExports call the exporter and then the exporter call back to the project.pack method.

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -2925,55 +3007,56 @@ public Jar pack(String profile) throws Exception {
...
-							continue;
-						Name name = new Name(key);
-						String trimmed = value.trim();
-						if (trimmed.isEmpty())
-							main.remove(name);
-						else if (EMPTY_HEADER.equals(trimmed))
-							main.put(name, "");
-						else
-							main.put(name, value);
+			try (Builder b = subBuilders.iterator()
```
bjhargrave: It seems that calling pack on a project can be a bit non-deterministic since it always uses the first subbuilder but the first subbuilder is not well defined since the order of subbuilderscan be based upon File.listFiles() which is defined to not have a prescribed order of its results.

But oddly, this method only seems to use the subbuilder to obtain a BSN if there is no BSN in the launcher's executable jar...

biz.aQute.bndlib/src/aQute/bnd/exporter/executable/ExecutableJarExporter.java
```diff
@@ -28,26 +41,48 @@ public ExecutableJarExporter() {}
...
-			launcher.setKeep(configuration.keep());
-			Jar jar = launcher.executable();
-			project.getInfo(launcher);
-
-			String name = jar.getName();
-			String[] baseext = Strings.extension(name);
-			if (baseext != null && ("bnd".equals(baseext[1]) || "bndrun".equals(baseext[1]))) {
-				name = baseext[0];
+		Jar jar;
+		if (EXECUTABLE_JAR.equals(type)) {
```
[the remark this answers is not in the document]
pkriens: nope, then I need to add additional code for null

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -3524,4 +3607,9 @@ public Container getBundle(org.osgi.resource.Resource r) throws Exception {
 	public boolean isStandalone() {
 		return getWorkspace().getLayout() == WorkspaceLayout.STANDALONE;
 	}
+
+	public boolean isRun() {
```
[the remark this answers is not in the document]
pkriens: k

biz.aQute.bndlib/src/aQute/bnd/exporter/executable/ExecutableJarExporter.java
```diff
@@ -28,26 +41,48 @@ public ExecutableJarExporter() {}
...
+			try (ProjectLauncher launcher = project.getProjectLauncher()) {
+				//
+				// Ensure that we do not override the -runkeep property
+				// if keep is not explicitly set. We therefore take the `or`
+				// of the keep option and the -runkeep
+				//
+
+				Configuration configuration = Converter.cnv(Configuration.class, options);
+				if (configuration.keep() || Processor.isTrue(project.getProperty(Constants.RUNKEEP)))
```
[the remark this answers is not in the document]
pkriens: k

biz.aQute.bndlib/src/aQute/bnd/exporter/executable/ExecutableJarExporter.java
pkriens: k

docs/_instructions/export.md
pkriens: made more clear

biz.aQute.bndlib/src/aQute/bnd/exporter/executable/ExecutableJarExporter.java
pkriens: good idea!

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
pkriens: they were actually not needed, cleaned up

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
pkriens: k

Issue #3589, opened 2019-11-19
[junit-platform tester] BundleSelectorResolver to resolve MethodSelectors


kriegfrj:
`BundleSelectorResolver` currently constructs `MethodSelector` objects using the `(Class<?>, String, String)` constructor:
https://github.com/bndtools/bnd/blob/c79dcbfa1b11e1f22c8a8790580de60778c95ba9/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java#L334-L335
This leaves the `org.junit.platform.engine.discovery` package responsible for resolving the method name and parameters into a `Method` object. However, because the built-in JUnit-Platform method parameter resolver is not class-loader aware, it always tries to load the parameter classes using the default class loader. This can result in it not finding the right method or not finding it at all, if the parameter classes are not shared between the test bundle and the launcher framework.

I have raised an issue for this on the JUnit 5 Github page (ref: junit-team/junit5#2104). In the meantime, if junit-platform tester is to work with older/existing versions of junit-platform, we'll need to work around this by resolving the `Method` reference in `BundleSelectorResolver` (similar to what we already do for resolving the `Class<?>` reference).

bjhargrave:
> we'll need to work around this by resolving the `Method` reference in `BundleSelectorResolver` (similar to what we already do for resolving the `Class<?>` reference).

I agree. If the user does not supply parameter, we should find them and supply them in the MethodSelector. If the method name is overloaded, I suggest we just use the first encountered method.

kriegfrj:
> I agree. If the user does not supply parameter, we should find them and supply them in the MethodSelector. If the method name is overloaded, I suggest we just use the first encountered method.

Having just gone through the process of implementing the original purpose of this issue, I circled back to this point and realised that it is not that simple.

The built-in JUnit support for method selectors and finding methods looks for an exact match on the number of parameters accepted by the test method. In other words, if you don't supply any parameter types, it will look specifically for a method that doesn't take any parameters - it won't simply use the first encountered.

To get the support that you're after, we'd need to re-implement a large chunk of the method-finding code in JUnit from scratch. This is probably doable, but not trivial.

If we decide that we still wish to proceed with this feature, I suggest we open a new issue for it.

PR #3602, opened 2019-11-24
[junit-platform tester] Add missing tests for method selectors

kriegfrj:
There were no tests to check that method selectors with unresolved classes would be handled properly. There was a TODO in the code to add these, so it is now TODONE.

This was done partially in preparation for #3589 to make sure that we don't regress during that implementation.


bjhargrave:
Thanks!

PR #3604, opened 2019-11-25
[junit-platform] Use Launcher builder framework

kriegfrj:
Now that we're on a more recent version of junit-platform-launcher, there is a builder API that can be used to construct launcher configs. This is a neater way of constructing a launcher that uses only the `BundleEngine` as a test engine, rather than using the ThreadContextClassLoader hack that we were forced to use in the earlier release.


bjhargrave:
Thanks some more!

PR #3607, opened 2019-11-25
MapStream

bjhargrave:
This class is a stream class for Maps. It extends BaseStream like other
stream types and has most of the operations like Stream for Map objects.


biz.aQute.bndlib/src/aQute/bnd/stream/EntryPipeline.java
```diff
@@ -0,0 +1,312 @@
...
+		entries().forEachOrdered(e -> consumer.accept(e.getKey(), e.getValue()));
+	}
+
+	@Override
+	public long count() {
+		return entries().count();
+	}
+
+	@Override
+	public boolean anyMatch(BiPredicate<? super K, ? super V> predicate) {
```
rotty3000: Not sure it's really necessary but for completeness should we have `anyMatchByKey|Value` and so forth for `allMatch`, `noneMatch`?

bjhargrave: I don’t think `anyMatchByKey`, etc. is useful since you can just ignore the key or value argument. `mapKey` is helpful since you avoid having to make an `Entry` for the result. These `any`/`all`/`none` method just return boolean. 

A reason to add them would be to use an existing `Predicate` instead of having to make a `BiPredicate`. Do you think this makes adding them worthwhile?

bjhargrave: The more I think about it, perhaps it is a good idea to add them.

bjhargrave: Actually, I now recall why these methods are not necessary. Since the match methods are terminals which return a boolean, there is not need to preserve the entries in the stream like we must do for filterKey/filterValue. To match by key call the match method on `keys()` and to match by value call the match method on `values()`. For example `mapStream.keys().anyMatch(predicate)`. This is basically what an `anyMatchKey` method would do anyway. Similarly if you wanted to find the max key, you would do `mapStream.keys().max(comparator)`.

rotty3000: Ok makes sense!

PR #3613, opened 2019-11-27
comm.tests: Use different port range for windows

bjhargrave:
We were having intermittent BindException because the chosen port was in
use.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ env:
     - GRADLE_OPTS=-Dorg.gradle.parallel=false
     - JAVA_JDK=https://cdn.azul.com/zulu/bin/zulu8.42.0.23-ca-jdk8.0.232-linux_x64.tar.gz
 
-matrix:
+jobs:
   include:
     - name: "OpenJDK8"
       env:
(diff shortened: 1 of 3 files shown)


pkriens:
Should we move this to aQute.lib? I did something similar last week and noticed there are some ranges that are free. Would be nice to have this knowledge central

bjhargrave:
Well this is a hack. If we do anything in IO, we should make sure to return an available ephemeral port not just the next number in some range. And the free range for windows depends on the version of windows. https://support.microsoft.com/en-us/help/832017/service-overview-and-network-port-requirements-for-windows. The change I  made here works because Azure uses Windows Server 2016 to run the CI build for WIndows.

https://github.com/bndtools/bnd/blob/b4b07aff36c858b859a761e4ec0ea1222ad63e02/azure-pipelines.yml#L76

pkriens:
It is better to have a hack in one place than in many places? 

This was exactly the same place that I fixed in the pkriens/beforelaunch branch.

bjhargrave:
I guess it would have been nice if you had released this little fix by itself earlier :-) Would have saved everyone a lot of time.

PR #3626, opened 2019-12-03
Proposal for a 4.3.2 as discussed last Friday

pkriens:
I'd like to release a 4.3.2 with at least these changes.

397ae60 -> Launching is incompatible with 4.2, this fixes it
e9f3502 -> Start level could not handle non-bundles, blowing up some tests
f058ccf -> Restore use of snapshot in -rundbundles for workspace bundles

### DIFF
--- a/aQute.libg/src/aQute/lib/startlevel/StartLevelRuntimeHandler.java
+++ b/aQute.libg/src/aQute/lib/startlevel/StartLevelRuntimeHandler.java
@@ -198,6 +198,11 @@ public class StartLevelRuntimeHandler implements Closeable {
 							if (bundle.getBundleId() == 0)
 								return;
 
+							if (bundle.getSymbolicName() == null) {
+								logger.trace("Found bundle without a bsn %s, ignoring", bundle);
+								return;
+							}
+
 							BundleIdentity id = installed.computeIfAbsent(bundle, BundleIdentity::new);
 							if (event.getType() == BundleEvent.INSTALLED || event.getType() == BundleEvent.UPDATED) {
 								setStartlevel(bundle, id);
--- a/biz.aQute.launcher/src/aQute/launcher/Launcher.java
+++ b/biz.aQute.launcher/src/aQute/launcher/Launcher.java
@@ -43,6 +43,7 @@ import java.util.HashMap;
 import java.util.Hashtable;
 import java.util.IllegalFormatException;
 import java.util.LinkedHashMap;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -99,7 +100,6 @@ public class Launcher implements ServiceListener {
 	private boolean							security;
 	private SimplePermissionPolicy			policy;
 	private Callable<Integer>				mainThread;
-	private final List<BundleActivator>		embedded			= new ArrayList<>();
 	private final Map<Bundle, Throwable>	errors				= new HashMap<>();
 	private final Map<File, Bundle>			installedBundles	= new LinkedHashMap<>();
 	private File							home				= new File(System.getProperty("user.home"));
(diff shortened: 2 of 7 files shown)


bjhargrave:
I don't make PRs on next to do a maint release. I just cherry pick the agreed upon commits.

bjhargrave:
What are the issue numbers for the issues you want to fix in a maint release? (Sorry I was not able to make the discussion due to US Thanksgiving holiday.)

bjhargrave:
I see at least 2 packages with version bumps in this PR which is pretty risky for a maint release.

pkriens:
I can only see one package? `aQute.bnd.osgi.resource`

This is a small method. I could work around this but that does seem more risk than using it as is because that will be new code. The surrounding fix is fixing a rather annoying bug when you don't have and use the resolver. It selects the actual version of workspace bundles.

kriegfrj:
If we're looking at a 4.3.2, I'd like to submit #3452 for consideration. Mainly because of the performance benefit in the launcher for Windows users.

It does bump some dependencies, but only on the test side (we might need to include #3454 as well for the CI build to pass).

pkriens:
I'll be (trying) to handle the release this time. I am in Germany this week but can do an RC1 probably on Friday.

However, what's the performance benefit? It also seems a rather large commit?

It also introduces a new package which would bump aQute.lib's minor?

I changed the proposal to not have any version changes on BJ's request. I am not as strong on this as BJ but it does make a nice simple rule that a micro release should really reflect a micro release. I.e. no new API anywhere.

However, in principle I do not have a problem with this.

kriegfrj:
> I'll be (trying) to handle the release this time. I am in Germany this week but can do an RC1 probably on Friday.

All the best!

> However, what's the performance benefit?

On Windows without this fix, there is a ~4-6s delay when running the launcher with tracing enabled as it prints the report (this it the time it takes for the faux network lookup to fail). This is annoying, bur really only if you're stopping/restarting frequently.

> It also seems a rather large commit?

I'm not sure what your metric is here. #3452 by itself doesn't seem much larger than any of the others that you've suggested for inclusion in this minor release. A lot of it is on the test side.

#3454 is a bit bigger, but it only affects the test code rather than the production code. 

I could re-write the tests for #3452 to downgrade them to JUnit 4, as I ended up not using the `@ParameterizedTest` feature of Jupiter anyway. Then it wouldn't depend on #3454. As long as the production code doesn't change, this is still low risk?

> It also introduces a new package which would bump aQute.lib's minor?

A new method, but I don't think a new package? But yes, it does bump the minor version of `aQute.libg.uri` from 3.2.0 to 3.3.0.

> I changed the proposal to not have any version changes on BJ's request. I am not as strong on this as BJ but it does make a nice simple rule that a micro release should really reflect a micro release. I.e. no new API anywhere. However, in principle I do not have a problem with this.

That makes sense. I'll leave you & BJ to debate the merits of this versioning policy vs the merits of including the patch.

pkriens:
The aQute.lib change then also bumps the bundle's version. It is a nasty constrain but there is something to say for a simple rule.

I would be ok to not include the test cases since they are not that relevant for this micro release. This makes the change a lot smaller and clearer what the scope is.

Could you prepare a PR on the next branch? Then I can include it.

bjhargrave:
> Could you prepare a PR on the next branch? Then I can include it.

Yes, I think you need to restructure the change against next to avoid the JUnit 5 tests since JUnit 5 test support is not in 4.3.x. And in general, we should try to avoid bumping package versions in maintenance releases. So you should be able to pull the URIUtil change into the launcher as a private method to address the issue in this maintenance release.

bjhargrave:
Perhaps we should instead focus our collecting energies on getting a 5.0 release done soon to get new function out. If we are concerned that the new JUnit 5 support is not fully baked, we can characterize it as a "preview".

This would seems a much better use of our collective time.

kriegfrj:
Yes, I think you're right. To much mucking around with the PR has is own risks, especially as it falls with the launcher. Better to invest that time polishing 5.0.0.

bjhargrave:
We agreed to do a 5.0 release in early January. So closing this PR.

PR #3630, opened 2019-12-03
[view] Bndtools Explorer

pkriens:
A replacement for the package explorer in the
bndtools perspective. The bndtools explorer
has a filter field for projects. Filtering
is only done on project name (glob), all its
resources are then visible.


pkriens:
Check it out!! Finally a filter for the package explorer!

I've made it default for the perspective but we could discuss this. It inherits the Package Explorer so it should be identical

PR #3634, opened 2019-12-03
[cdi] allow the CDI container to process interfaces

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
```diff
@@ -115,7 +115,7 @@ public boolean analyzeJar(Analyzer analyzer) throws Exception {
 		for (Clazz c : analyzer.getClassspace()
 			.values()) {
 
-			if (c.isModule() || c.isEnum() || c.isInterface() || c.isInnerClass() || c.isSynthetic()
+			if (c.isModule() || c.isEnum() || c.isInnerClass() || c.isSynthetic()
```
bjhargrave: This also means you will process annotations which are interfaces. Is that intended?

rotty3000: hmm, good question... I suspect no.

biz.aQute.bndlib/src/aQute/bnd/cdi/CDIAnnotations.java
```diff
@@ -115,7 +115,7 @@ public boolean analyzeJar(Analyzer analyzer) throws Exception {
 		for (Clazz c : analyzer.getClassspace()
 			.values()) {
 
-			if (c.isModule() || c.isEnum() || c.isInterface() || c.isInnerClass() || c.isSynthetic()
+			if (c.isModule() || c.isEnum() || c.isAnnotation() || c.isInnerClass() || c.isSynthetic()
 				|| !c.is(QUERY.CONCRETE, null, analyzer)) {
```
bjhargrave: FYI, `!c.is(QUERY.CONCRETE, null, analyzer)` can be simplified to `c.isAbstract()` and this reads better.

https://github.com/bndtools/bnd/blob/0355bd26c0a91d570d69dfa607182a2f75ba040a/biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java#L1771-L1772

bjhargrave: Speaking of abstract classes, if you want to look at interfaces for CDI annotations, why not abstract classes?

PR #3642, opened 2019-12-07
gradle: Use dependency's jars for compilation

bjhargrave:
If the `java-library` plugin is applied, Gradle will compile against
the dependency's classes dir instead of building the jar and compiling
against the jar. This is not good for bundles since Bnd can pull in code
from other places and also set the package metadata.

So we configure the sourceSet's compileClasspath configuration to use
jars for compilation.

This support requires Gradle 5.6 or later.

See https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_classes_usage

### DIFF
--- a/biz.aQute.bnd.gradle/publish.gradle
+++ b/biz.aQute.bnd.gradle/publish.gradle
@@ -18,7 +18,7 @@ apply plugin: 'java'
 
 ext.groupId = 'biz.aQute.bnd'
 ext.artifactId = 'biz.aQute.bnd.gradle'
-ext.artifactVersion = '4.3.0'
+ext.artifactVersion = '4.3.1'
 
 repositories {
   jcenter()
@@ -29,7 +29,9 @@ configurations {
   plugin
   plugin.dependencies.all { Dependency dep ->
     if (dep instanceof ExternalDependency) {
-      dep.force = true
+      dep.version {
+        strictly dep.getVersion()
+      }
     }
     if (dep instanceof ModuleDependency) {
       dep.transitive = false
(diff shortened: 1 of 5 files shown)


bjhargrave:
@rotty3000 Can check if this fix removes the need for 

https://github.com/junit-team/junit5/blob/6928c029875593f657c2952790f252ff75b03c6f/gradle.properties#L20-L24

bjhargrave:
For reference, this is how Gradle processes the `org.gradle.java.compile-classpath-packaging` system property.

https://github.com/gradle/gradle/blob/3ff62230dd9fb408da2e7766be28772f0ad2984b/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaBasePlugin.java#L82-L90

https://github.com/gradle/gradle/blob/3ff62230dd9fb408da2e7766be28772f0ad2984b/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaBasePlugin.java#L105-L110

https://github.com/gradle/gradle/blob/3ff62230dd9fb408da2e7766be28772f0ad2984b/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaBasePlugin.java#L298

rotty3000:
@bjhargrave sorry I didn't get a chance to look at this yet, but I will try sometime later this week with the junit5 project. Thanks for tackling this.

rotty3000:
@bjhargrave is this too heavy a change for inclusion into the discussed 4.3.2 maintenance release?

bjhargrave:
> is this too heavy a change for inclusion into the discussed 4.3.2 maintenance release?

It is fairly safe, but the commit may not merge easily due to lots of changes in the gradle plugin code for Bnd 5.0.

rotty3000:
got it!

Issue #3644, opened 2019-12-11
[docs] Windows tips page


kriegfrj:
I'd like for other people not to have to go through this [performance issues on Windows] again. Do you guys think that it is worth having a page in the Bnd/Bndtools manual with tips for Windows users that summarises my findings? It would be fairly simple and I'm putting my hand up to write it. Just let me know where is the most appropriate place for it and I'll submit it as a fresh PR.

_Originally posted by @kriegfrj in https://github.com/bndtools/bnd/issues/3410#issuecomment-533951698_

PR #3645, opened 2019-12-11
Windows tips page

kriegfrj:
Manual page with tips for Windows users.

Implements #3644.

### DIFF
--- a/docs/_chapters/650-windows.md
+++ b/docs/_chapters/650-windows.md
@@ -1,0 +1,70 @@
+---
+title: Tips for Windows users
+layout: default
+author: Fr Jeremy Krieg
+---
+
+Bnd and Bndtools runs on any system that can run Eclipse. However, every operating system is different and has its own quirks. Windows in particular has a few that impact developers using Bnd. While the Bnd experience may never be as smooth on Windows as it is on macOS or Linux, there are a few tweaks that make it a lot smoother than what comes out of the box.
+
+It is possible that some of the tips listed here will also help for Eclipse development work in general, or perhaps any development work on Windows. 
+
+## Key problems on Windows 
+
+Windows has two key architectural differences to *nix-based OSes that really impact development work in Java:
+
+### File locking
+
+Most \*nix systems use advisory file locking, rather than mandatory locking. Windows uses mandatory locking rather than advisory. Most Java programs are developed on \*nix and tend to assume \*nix locking semantics. This means that they are not as careful at cleaning up after themselves and ensuring that file locks are released.
+
+In particular, there are problems when an application exits using `System.exit()` - the normal way for Java applications to clean up is to release a lock is in a `finally` clause. However, when an application calls `System.exit()`, these normal channels are bypassed, which means the JVM never gets a chance to release the locks and hence can't delete the locked files.
+
+### Real-time protection
+
+From version 10, Windows' built-in anti-malware program Windows Defender has become very tightly integrated with the OS - seemingly, it is written into the Windows kernel API IO functions themselves, which means it's nearly impossible to bypass. This makes it extremely well suited to catching malware in the act. Unfortunately, it means it slows down every file write. When you are writing lots of small files, this delay is significant.
+
+## Workarounds and solutions
+
+### Use Bnd 5+
+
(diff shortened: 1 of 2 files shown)


docs/_chapters/650-windows.md
```diff
@@ -0,0 +1,70 @@
+---
+title: Tips for Windows users
+layout: default
+author: Fr Jeremy Krieg
+---
+
+Bnd and Bndtools runs on any system that can run Eclipse. However, every operating system is different and has its own quirks. Windows in particular has a few that impact developers using Bnd. While the Bnd experience may never be as smooth on Windows as it is on iOS or Linux, there are a few tweaks that make it a lot smoother than what comes out of the box.
```
bjhargrave: I think you mean `macOS` not `iOS` as `iOS` is the iPhone operating system.

docs/_chapters/650-windows.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Real-time protection
+
+From version 10, Windows' built-in anti-malware program Windows Defender has become very tightly integrated with the OS - seemingly, it is written into the Windows kernel API IO functions themselves, which means it's nearly impossible to bypass. This makes it extremely well suited to catching malware in the act. Unfortunately, it means it slows down every file write. When you are writing lots of small files, this delay is significant.
+
+## Workarounds and solutions
+
+### Use Bnd 5+
+
+Prior to Bnd 5.0, there was a bug that caused the launcher to pause for several seconds if you had `[-runtrace](/instructions/runtrace.html)` enabled. This bug has been fixed as of 5.0 (currently in pre-release at the time of writing), so it is worthwhile upgrading to use 5.0 if you are able.
```
bjhargrave: You can't put the back tiks around the the whole link. Just put them inside the square braces around `-runtrace`. Otherwise the link is not rendered as a link.

docs/_chapters/650-windows.md
```diff
@@ -0,0 +1,70 @@
...
+It is possible that excluding other directories may help performance - eg, your `.m2` directory. You need to weigh the merits of excluding each folder vs the incremental security risk.
+
+### Create a custom temp directory
+
+As noted above, for best performance you should exclude the temp directory, however for best safety you may not wish to. A good compromise is:
+
+* Create a temp directory somewhere else on your machine that is specifically for your development work (eg, `C:\temp`).
+* Exclude this directory from antivirus scanning.
+* Configure Eclipse to use use this directory by adding the `-Djava.io.tmpdir=<path to tmp>` vm arg to the `eclipse.ini` file.
+* Configure your launches' `.bndrun` files to use this temp directory by using the `-runvm` instruction (eg: `java.io.tmpdir=C:\\temp` - noting the double-backslash.
```
bjhargrave: I think the `-runvm` instruction value will also need the `-D` prefix too. `-runvm: -Djava.io.tmpdir=...`

kriegfrj: From the docs for `-runvm`:

> Additional arguments for the VM invocation. Keys that start with a - are added as options, otherwise they are treated as -D properties for the VM.

To be honest, it's been a while since I actually tried it, but if it doesn't work that way we need to update the docs.

docs/_chapters/650-windows.md
kriegfrj: Fixed.

docs/_chapters/650-windows.md
kriegfrj: You mean you're not running Bndtools on your iPhone? 😄

Fixed.

docs/_chapters/650-windows.md
bjhargrave: I do not see anywhere in the code where that "magic" happens. Also, https://bnd.bndtools.org/instructions/runvm.html, does not contain any such text regarding that "magic" (that whole page is just a place holder). So I don't know what doc for `-runvm` to which you are referring.

kriegfrj: I'm quoting the instruction index page rather than the specific page for runvm itself.

https://bnd.bndtools.org/chapters/825-instructions-ref.html

bjhargrave: That doc is still wrong :-)

kriegfrj: Thanks for confirming - I'll fix both!

bjhargrave:
Thanks!

PR #3653, opened 2019-12-13
[docs] Fixed misleading doc for -runvm

### LINKED ISSUE #3652: [docs] runvm doc incorrectly states that -D is not necessary
The doc on the instruction index page currently states that -D is implied if there is no - on a parameter.

As @bjhargrave noted in https://github.com/_render_node/MDIzOlB1bGxSZXF1ZXN0UmV2aWV3VGhyZWFkMjE5OTA4MDYxOnYy/pull_request_review_threads/discussion this is incorrect.

kriegfrj:
Doc incorrectly stated that -D would be inferred if no leading - was found on the property. This is not the case - arguments are added as-is.

Fixes #3652.

### DIFF
--- a/docs/_instructions/runvm.md
+++ b/docs/_instructions/runvm.md
@@ -2,7 +2,7 @@
 layout: default
 class: Project
 title: -runvm KEYS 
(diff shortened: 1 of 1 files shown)


bjhargrave:
Thanks!

PR #3656, opened 2019-12-14
[bndtools] launch m2e

rotty3000:


rotty3000:
@bjhargrave I left the full index in there just so you could play with the PR if you needed. Did we really need EMF btw?

FYI, the first commit is pure sort. I did this so that I could use diff on the results as I iterated to get this working. It's much easier to do this with sorted values. Can we keep it sorted?

rotty3000:
FYI, if you try loading an existing workspace from a newer version of eclipse, it won't show any projects. You need to re-import the projects and then everything seemed to work fine.

bndtools.core/bndtools.gtk.linux.x86_64.bndrun
```diff
@@ -1,15 +1,24 @@
 -include: bndtools.shared.bndrun
 
 -runbundles.linux: \
+	org.eclipse.swt.gtk.linux.x86_64,\
+	org.eclipse.e4.ui.workbench.renderers.swt,\
```
bjhargrave: this is already in the shared bndrun, so don't duplicate it here.

bndtools.core/bndtools.gtk.linux.x86_64.bndrun
```diff
@@ -1,15 +1,24 @@
...
+	${crossplatform.vmargs}
 
 -runproperties: \
 	${crossplatform.runprops},\
 	osgi.ws=gtk,\
 	osgi.arch=x86_64,\
 	osgi.os=linux
+
+-runblacklist.linux: \
+	osgi.identity;filter:='(osgi.identity=*macosx*)'
```
bjhargrave: Shouldn't `*win32*` also be in here?

bndtools.core/bndtools.shared.bndrun
```diff
@@ -13,46 +13,9 @@ crossplatform.runprops: \
...
-	bnd.identity;id='org.eclipse.emf.codegen.ecore.ui',\
-	bnd.identity;id='org.eclipse.emf.codegen.ui',\
-	bnd.identity;id='org.eclipse.emf.common.ui',\
-	bnd.identity;id='org.eclipse.equinox.event',\
-	bnd.identity;id='org.eclipse.ant.core',\
-	bnd.identity;id='org.eclipse.ant.ui',\
-	bnd.identity;id='org.eclipse.ui.ide.application',\
-	bnd.identity;id='org.eclipse.ui.ide',\
-	bnd.identity;id='org.eclipse.sdk',\
-	bnd.identity;id='org.eclipse.pde.ua.ui',\
```
bjhargrave: why did you pull pde? I think it should remain for the bndtools.pde project, no?

cnf/ext/repositories.bnd
```diff
@@ -37,6 +37,11 @@ mavencentral:           https://repo.maven.apache.org/maven2
         locations="https://dl.bintray.com/bndtools/eclipse-repo/4.10/index.xml.gz";\
         poll.time=-1;\
         cache="${workspace}/cnf/cache/stable/Eclipse-2018-12",\
+    aQute.bnd.repository.osgi.OSGiRepository;\
```
bjhargrave: Obviously we don't want to commit this. Are there any artifacts missing from the bintray repo I need to add?

bndtools.core/bndtools.gtk.linux.x86_64.bndrun
rotty3000: right.

bndtools.core/bndtools.gtk.linux.x86_64.bndrun
rotty3000: added for completeness though it was never getting resolved.

bndtools.core/bndtools.shared.bndrun
rotty3000: re-added!

cnf/ext/repositories.bnd
rotty3000: Yes, there are arfiacts missing from the slim repo, see the diff. I removed the repo reference.

PR #3658, opened 2019-12-15
[pom] Eliminate logging to System.err

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/maven/support/ProjectPom.java
```diff
@@ -153,9 +161,9 @@ public boolean excludes(String name) {
 
 	@Override
 	protected String replace(String in) {
-		System.err.println("Replce: " + in);
+		reporter.trace("Replace: %s", in);
 		if (in == null) {
-			System.err.println("null??");
+			reporter.error("null??");
```
bjhargrave: I guess a more helpful message would be useful. This message seems like a debug "how did I get here?" message rather than something useful.

rotty3000: done.

Issue #3611, opened 2019-11-27
LaunchPad in MavenWorkspace


stbischof:
It would be great to have LaunchPad together with the MavenWorkspaces.
I tryed to find some ways but the static call of RemoteWorkspaceClientFactory in a Field makes it impossible for me to handle it.
```

final static RemoteWorkspace	workspace = RemoteWorkspaceClientFactory.create(projectDir,
		new RemoteWorkspaceClient() {});
```

https://github.com/bndtools/bnd/blob/28fe908a2b2026b09c084838db2c89fd7054f77d/biz.aQute.launchpad/src/aQute/launchpad/LaunchpadBuilder.java#L51

```
LauchPadBuilder.withWorkspace(repoIndexFile),

```

would be nice.

PR #3660, opened 2019-12-16
EXPERIMENT [launchpad] standalone launchpad (do not merge)

rotty3000:


rotty3000:
Hey gents, here's a standalone launchpad. Naming is up for grabs, I just went with what seemed natural.

rotty3000:
FYI, while this works standalone and all tests pass, a glaring omission is still how we would pump Maven (or gradle) details into it.

I'm thinking something of the service loader nature, or some form of adaptor that lives somewhere on the test path to bridge across.

kriegfrj:
> I'm thinking something of the service loader nature, or some form of adaptor that lives somewhere on the test path to bridge across.

I don't pretend to understand 100% what this PR is about. :smile: However, in response to the above, I'm in favour of not re-inventing the wheel unless there is a compelling reason. Ie, don't invent a new "form of adaptor that lives somewhere on the test path" when the `ServiceLoader` form already exists (provided that the `ServiceLoader` will do what you need).

stbischof:
That looks really nice!!!

It would be nice to have the StandaloneLaunchpad also without an .bndrun File.
It should take the Auto-Generates index File.
Refs:#3611

biz.aQute.launchpad/src/aQute/launchpad/StandaloneLaunchpadBuilder.java
```diff
@@ -0,0 +1,290 @@
...
+import aQute.bnd.service.Strategy;
+import aQute.bnd.service.remoteworkspace.RemoteWorkspace;
+import aQute.bnd.service.specifications.BuilderSpecification;
+import aQute.bnd.service.specifications.RunSpecification;
+import aQute.bnd.version.Version;
+import aQute.bnd.version.VersionRange;
+import aQute.lib.utf8properties.UTF8Properties;
+import aQute.libg.glob.Glob;
+
+public class StandaloneLaunchpadBuilder extends AbstractLaunchpadBuilder<StandaloneLaunchpadBuilder> {
```
pkriens: We should not have a separate builder as far as I can see. That makes the source code incompatible.

The only problem with the current approach is that the client searches for the workspace using the working directory to find the workspace cache directory. This all is only needed to find the port the 'current' workspace is registered on.

In the maven case, we create the workspace ourselves. We could this easily find out the port of the workspace and register this as a system property when launching the test process.

biz.aQute.launchpad/src/aQute/launchpad/StandaloneLaunchpadBuilder.java
```diff
@@ -0,0 +1,290 @@
...
+			return About.CURRENT.toString();
+		}
+
+		@Override
+		public RunSpecification getRun(String pathToBndOrBndrun) {
+			return unchecked(() -> {
+				File file = new File(pathToBndOrBndrun);
+
+				if (file.isDirectory()) {
+					Project project = run.getWorkspace()
```
pkriens: I went out of my way to not require bnd on the classpath for the test code because it creates to many dependencies. Notice that this contains a lot of code: SLF4j, all the resolve and repo code, etc. The whole idea is that the code runs on the process that launches the test process, not inside the test process. This is a big no-no for me. Launchpad should keep the Workspace and bnd Project code all on the launcher process and never in the launched process.

biz.aQute.launchpad/src/aQute/launchpad/StandaloneLaunchpadBuilder.java
rotty3000: Each execution instance in maven or gradle.builder has a unique workspace so there is nothing to link to. Also, I am remiss to handle this as a remote process.

biz.aQute.launchpad/src/aQute/launchpad/StandaloneLaunchpadBuilder.java
rotty3000: (note that when I say gradle I mean non-workspace `biz.aQute.bnd.builder`)

@pkriens it's all good. This is an experiment. I didn't expect it to be so easy. I just wanted to get my hands dirty to see what could be done.

However, a couple of points:
- there is no stable workspace in maven/gradle to link to
- there are no _projects_ in bnd project terms
- everything in maven/gradle is modeled around Run/Bndrun. The workspace is merely an internal detail of the model rather the main concern.

pkriens: I understand there is no _stable_ workspace, but there **must** be a workspace object. If you look deeper, then you see that the Run _isa_ Project and therefore the Workspace object is absolutely mandatory. A standalone bndrun fakes a workspace as far as I know.

So you should link to this temporary Workspace and let it do the heavy lifting.

rotty3000: this would mean making N server backend come up per build!

rotty3000: Why remote to something that is instantiated on demand, in process?

bjhargrave:
Sounds like a compelling experiment. I never like the idea that Launchpad discovers a port number and reaches back over a socket to a remote workspace. This already created issues with parallel builds in our own build.

biz.aQute.launchpad/src/aQute/launchpad/StandaloneLaunchpadBuilder.java
pkriens: a) They can share the same backend
b) You cannot do it in the same process because you do not want the whole bnd code on your test class path. 
c) Multiple runs can be faster since you only need to do it once per test process

rotty3000: I fail to see who will orchestrate starting this remote process! There is zero bnd code executed until the builder instance is created in the test class at which point starting a remote process is really fruitless!

Furthermore, if what we need is classloader isolation, then let's just do that!

pkriens: Well, on the maven side you 'somewhere' have the bndrun file? That implies you have a workspace?

rotty3000: but that workspace is created only when the bndrun file is interpreted which doesn't happen until the test case class is loaded.

kriegfrj: > Furthermore, if what we need is classloader isolation, then let's just do that!

Would this work, @pkriens? Are there any use cases that you've encountered where (eg) implicitly adding a few extra "excludeExport()" directives to exclude Bnd-related packages wouldn't solve the problem of having the whole bnd code on your test class path?

pkriens: I am pretty sure you're gone get in lots of conflicts. 

I guess the detection code could be instructed with a system property to start a maven workspace for maven in another class loader that has all the bnd code. They could communicate over the remote workspace interface. I guess the analysis code will be  quite different. Maybe we should have a discussion about this.

stbischof:
Any news?

PR #3661, opened 2019-12-16
[maven] bndrun implicit repo, repos of maven dependencies not considered

rotty3000:


biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/DependencyResolver.java
```diff
@@ -176,10 +176,14 @@ private void discoverArtifacts(Map<File, ArtifactResult> files, List<DependencyN
 				.getScope())) {
 				continue;
 			}
+
+			List<RemoteRepository> copyOfRepositories = new ArrayList<>(remoteRepositories);
```
bjhargrave: Not crazy about the var name. `combinedRepositories` would be more meaningful.

rotty3000: done

Issue #3664, opened 2019-12-17
[junit-platform tester] BundleEngine to report reason for resolution failure


kriegfrj:
We have introduced a regression into the `BundleEngine` (more specifically, `BundleSelectorResolver`). Here is the relevant diff (between revisions 301dab8e83c24b53bba316508f5fb300fc3d4597 and d0bcd4d5f3):

````
            .forEach(bundle -> {
-                   final String bundleDesc = uniqueIdOf(bundle);
-                   try {
-                         bundle.start();
-                   } catch (BundleException e) {
+                   if (!frameworkWiring.resolveBundles(Collections.singleton(bundle))) {
                            if (testUnresolved) {
-                                   unresolvedBundlesDescriptor
-                                           .addChild(new StaticFailureDescriptor(unresolvedBundlesDescriptor.getUniqueId()
-                                                   .append("bundle", bundleDesc), displayNameOf(bundle), e));
+                                   unresolvedBundlesDescriptor.addChild(new StaticFailureDescriptor(
+                                           unresolvedBundlesDescriptor.getUniqueId()
+                                                   .append("bundle", uniqueIdOf(bundle)),
+                                           displayNameOf(bundle),
+                                           new BundleException("Unable to resolve", BundleException.RESOLVE_ERROR)));
                            }
                    }
            });
````

Undoubtedly, my original code was less elegant. However, in the process of this refactoring, we have lost a useful feature.

In my original implementation, if the attempt to start the bundle threw a `BundleException`, that `BundleException` was reported as-is through the JUnit framework. If the exception was due to a resolution failure, that reason was normally reported in the exception message. This meant the reason for the failure was immediately visible in the JUnit output and could often be fixed without any need for further debugging.

In the current implementation, if the resolution fails we get "Unable to resolve" and that's it. To debug the reason why resolution failed requires further digging (eg, re-run the test with launcher tracing on and look at the console output).

It would be nice if this feature could be re-introduced. It is probably my fault that this has regressed because I didn't put a regression test in to test for this feature. :smile:

@bjhargrave, do you think you could have a quick look at this one?

PR #3668, opened 2019-12-17
junit-platform: Capture real BundleException with resolve failure

bjhargrave:
We don't want to start a bundle to try and resolve it. But if attempting
to resolve a bundle does not succeed, we then call Bundle.start to
capture a BundleException with resolution failure information for the
test execution.

Fixes https://github.com/bndtools/bnd/issues/3664

### DIFF
--- a/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java
+++ b/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java
@@ -180,10 +180,15 @@ public class BundleSelectorResolver {
 			.filter(BundleUtils::isNotFragment)
 			.filter(BundleUtils::isNotResolved)
 			.forEach(bundle -> {
-				if (!frameworkWiring.resolveBundles(Collections.singleton(bundle))) {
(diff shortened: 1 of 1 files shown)


kriegfrj:
Innovative solution, well done!

Is there a gap in the OSGi standard somewhere that this information cannot be obtained any way other than by starting the bundle?

bjhargrave:
> Is there a gap in the OSGi standard somewhere that this information cannot be obtained any way other than by starting the bundle?

Perhaps, but in the OSGi spec, resolving a bundle is something a framework can do at any time. So Bundle.start is a convenient time to report resolution failure since a bundle must be resolved then.

kriegfrj:
> > Is there a gap in the OSGi standard somewhere that this information cannot be obtained any way other than by starting the bundle?
> 
> Perhaps, but in the OSGi spec, resolving a bundle is something a framework can do at any time. So Bundle.start is a convenient time to report resolution failure since a bundle must be resolved then.

I get that resolution is something that the framework can do at any time. But in this instance you're using the `FrameworkWiring` interface that is supposed to resolve on demand and synchronously. It would be nice if `resolveBundles()` was spec'd to throw a `BundleException` on failure (which could contain the information that we're after) rather than simply returning false.

PR #3670, opened 2019-12-17
[processor] add _version_cleanup macro farther down the stack so it can be used in bndrun files

rotty3000:


docs/_macros/maven_version.md
```diff
@@ -21,3 +21,5 @@ summary: Cleanup a potential maven version to make it match an OSGi Version synt
 		return null;
 	}
```
bjhargrave: Remove the place holder code above.

biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
```diff
@@ -2981,4 +2981,19 @@ public String getChecksum() {
 			throw Exceptions.duck(e);
 		}
 	}
+
+	/**
+	 * A macro to cleanup a version string (like Maven versions) to an OSGi
+	 * version
+	 */
+	public String _version_cleanup(String args[]) {
```
bjhargrave: This should use the more modern style:

```java
static final String _version_cleanupHelp = "${version_cleanup;<version>}, Cleanup a potential maven version to make it match the OSGi Version syntax.";

public String _version_cleanup(String args[]) throws Exception {
  Macro.verifyCommand(args, _version_cleanupHelp, null, 2, 2);
  return Analyzer.cleanupVersion(args[1]);
}
```

You should also push this macro down even further into the `Macro` class since this macro actually needs nothing from `Processor`.

docs/_macros/version_cleanup.md
```diff
@@ -0,0 +1,26 @@
+---
+layout: default
+class: Builder
+title: version_cleanup ';' VERSION-STRING
+summary: Cleanup a potential maven version to make it match an OSGi Version syntax.
+---
+
+
+```java
```
bjhargrave: This place holder code should be removed and some words put here to describe what the macro does to the version argument.

docs/_macros/version_cleanup.md
```diff
@@ -0,0 +1,26 @@
...
+	else if (args.length < 2)
+		error("${version_cleanup} macro has no arguments, use ${version_cleanup;1.2.3-SNAPSHOT}");
+	else {
+		return Analyzer.cleanupVersion(args[1]);
+	}
+	return null;
+}
+```
+
+**Note:** This macro should be used in place of the `${maven_version}` macro since it can be used from more places.
```
bjhargrave: I don't think you need this note here. Just maven_version should point here.

docs/_macros/version_cleanup.md
```diff
@@ -0,0 +1,26 @@
+---
+layout: default
+class: Builder
```
bjhargrave: This should be `class: Macro` if you move the method to `Macro` or `class: Processor` if you leave it in `Processor`.

docs/_macros/version_cleanup.md
```diff
@@ -0,0 +1,26 @@
...
+title: version_cleanup ';' VERSION-STRING
+summary: Cleanup a potential maven version to make it match an OSGi Version syntax.
```
bjhargrave: `the OSGi Version syntax`

PR #3677, opened 2019-12-20
[junit-platform tester] Check for mixed JUnit 3 and 4 tests

### LINKED ISSUE #3671: [junit-platform tester] Check for mixed JUnit 3/4 tests
If a test class contains JUnit 4-style `@Test` annotations *and* extends from `junit.framework.TestCase` (JUnit 3-style), then the engine will silently ignore the `@Test`-annotated tests and only run those that start with `testXXX()` (JUnit-3 style).

Although JUnit 3 is old, there are still quite a few tests out there using it and getting migrated to 4 or 5. The above is a trap that can make you think that your tests are all passing when really there are a bunch that aren't running at all.

The old tester had code to check for and warn of this situation:

https://github.com/bndtools/bnd/blob/a116b9446ff64fa3d90e7c7880b5832c70bb96d2/biz.aQute.junit/src/aQute/junit/Activator.java#L513-L518

I was hoping that the JUnit Team might add similar support to the Vintage engine (as this is surely not an issue that is unique to Bnd testing), but they have recently declined this request (junit-team/junit5#1964) as they see it as a legacy issue.

So we should perhaps consider adding this check to `BundleEngine` too, if we think this is still a case that is likely to catch people out.

kriegfrj:
Will check classes for mixed JUnit 3 and 4 tests and generate a test error for each one it finds under the bundle in which the class was found.

Fixes #3671.

### DIFF
--- a/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java
+++ b/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java
@@ -7,6 +7,8 @@ import static java.util.stream.Collectors.toSet;
 import static org.junit.platform.engine.discovery.DiscoverySelectors.selectClass;
 import static org.junit.platform.engine.discovery.DiscoverySelectors.selectMethod;
 
+import java.lang.annotation.Annotation;
+import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Method;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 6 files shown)


biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java
```diff
@@ -398,6 +401,51 @@ private void markClassesResolved(Bundle bundle) {
...
+		return false;
+	}
+
+	private boolean elementHasJUnit4Annotations(AnnotatedElement element) {
+		Annotation[] annotations = element.getAnnotations();
+		if (annotations != null) {
+			for (Annotation a : annotations) {
+				final String annotationName = a.annotationType()
+					.getName();
+				if (annotationName.startsWith("org.junit.") && !annotationName.contains(".jupiter.")) {
```
bjhargrave: I wonder if this is focused enough in JUnit 4? Perhaps some future JUnit 6 will have annotations which start with `org.junit.` but do not contain `.jupiter.`. Then we will be mistaken. Perhaps we need a list of JUnit 4 annotation names in a collection and can ask that collection if it contains the observed annotation name. The doesn't-contain test is not future proof enough.

kriegfrj: Fair point - the latest push has gone with a fairly exhaustive check of specifically JUnit 4 annotations rather than a generic pattern.

PR #3684, opened 2019-12-21
[maven] bnd-process-test mojo has wrong resolution scope

rotty3000:


rotty3000:
Yes, I would like to tag this for 5.0.0. Also bare with me on this since I identified some cleanup I should do.

rotty3000:
@bjhargrave ok, I'm ready on this now.

PR #3688, opened 2019-12-21
[cli] flag to print resolution results from plugins

rotty3000:


rotty3000:
@bjhargrave the gradle plugin already does this so this brings parity.

rotty3000:
Please mark for inclusion in 5.0.0!

PR #3691, opened 2019-12-22
[maven] EE inference needs to be more robust

rotty3000:


rotty3000:
@bjhargrave please mark this for 5.0.0.

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
```diff
@@ -1,12 +1,13 @@
 package aQute.bnd.maven.lib.resolve;
 
+import static java.util.Optional.ofNullable;
```
bjhargrave: I am not a fan of this static import. `ofNullable` in code is too ambiguous. In Java 9, Stream has the same static method name for example.

biz.aQute.bndlib/src/aQute/bnd/build/model/EE.java
```diff
@@ -1,5 +1,7 @@
 package aQute.bnd.build.model;
 
+import static java.util.Optional.of;
```
bjhargrave: I am not a fan of this static import. `of` in code is way too ambiguous. Stream has `of`. In Java 9, Map has `of`.

PR #3702, opened 2020-01-03
[BndTools] added project refresh after add nature

### LINKED ISSUE #3629: "Add Bndtools Project Nature" not completely working
Bndtools 4.3.1.REL-201911131708-g281e71d
Possibly related to #2961

How to reproduce:
1. Use a working bnd workspace 
2. Create a new java project in eclipse 
    (make sure it's actual location is next to the cnf folder)
3. Right click project and choose "Configure" > "Add Bndtools Project Nature"
    (.classpath and .project files are correctly modified, empty bnd.bnd file is created)
4. Add any bundle from "Repositories" Part to the Build Path of the new bnd.bnd file.
    Bnd Bundle Path entry is not showing up, required bundles can not be used by own classes, nothing working.
5. Hit all clean, rebuild and reload buttons available, nothing changes. However after restarting eclipse it works.

Since it works when you use "File" > "New" > "Bnd OSGi Project" I checked the source code and found that the following thing is executed on "new project" but not on "add nature":

**bndtools.central.Central.getWorkspace().refreshProjects();**

After a test where i only executed this line after the nature adding everything works perfectly. So my suggestion would be to add this line to "Add Bndtools Project Nature" code.

juergen-albert:
* fixes #3629 "Add Bndtools Project Nature" not completely working

Signed-off-by: Signed-off-by: Juergen Albert <j.albert@data-in-motion.biz>

### DIFF
--- a/bndtools.core/bndtools.shared.bndrun
+++ b/bndtools.core/bndtools.shared.bndrun
@@ -89,7 +89,8 @@ crossplatform.runprops: \
 	bnd.identity;id='org.eclipse.ui.ide.application',\
 	bnd.identity;id='org.junit.platform.engine',\
 	bnd.identity;id='org.junit.platform.launcher',\
-	bnd.identity;id='org.slf4j.impl.log4j12'
+	bnd.identity;id='org.slf4j.impl.log4j12',\
+	bnd.identity;id='org.eclipse.osgi.compatibility.state'
 
 -runee: JavaSE-1.8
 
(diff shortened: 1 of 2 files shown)


juergen-albert:
> Why do you need to refresh the entire Bnd and Eclipse workspaces when only one project's nature changes?

I have to refresh the Bnd Workspace, because if I don't Bnd does not see the new project when the Eclipse Project was refreshing. This was the only way I found to make Bnd aware of the new Project, when it needs to.

I need to check what happens if I don't refresh the whole Eclipse Workspace. I had to fiddle around a bit to get it working and it is possible, that this is just a leftover of this process.

PR #3703, opened 2020-01-03
Pkriens/jareditor

pkriens:
Rewrite of the JAREditor


biz.aQute.bnd/src/aQute/bnd/main/RepoCommand.java
```diff
@@ -541,7 +541,7 @@ public void _diff(diffOptions options) throws UnsupportedEncodingException, IOEx
 					.put(map)
 					.flush();
 			else if (!options.diff())
-				bnd.printMultiMap(map);
+				MultiMap.format(map);
```
bjhargrave: This just makes a String and then discards it. Is that what you meant? I assume you want to print the String here?

biz.aQute.bndlib/src/aQute/bnd/service/result/Result.java
```diff
@@ -225,4 +234,12 @@
 	 */
 	Result<V, E> recover(FunctionWithException<? super E, ? extends V> recover);
 
+	/**
+	 * Terminal function that processes the result or the error
+	 *
+	 * @param ok the consumer called when ok
+	 * @param err the consumer called when not ok
+	 */
+	void then(ConsumerWithException<? super V> ok, ConsumerWithException<? super E> err);
```
bjhargrave: I would call this something like `accept` rather than `then`. `then` implies some chaining in a monadic style. This is really about consuming the Result and `accept` is the Java method name for consumers.

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -294,7 +294,10 @@ private Jar buildFromInputStream(InputStream in, long lastModified) throws IOExc
 				int size = (int) entry.getSize();
 				try (ByteBufferOutputStream bbos = new ByteBufferOutputStream((size == -1) ? BUFFER_SIZE : size + 1)) {
 					bbos.write(jin);
-					putResource(entry.getName(), new EmbeddedResource(bbos.toByteBuffer(), lastModified), true);
+					long time = entry.getTime();
```
bjhargrave: Should this be `long time = ZipUtil.getModifiedTime(entry)`? We seem to prefer that in other places to `ZipEntry.getTime()`.

aQute.libg/src/aQute/lib/strings/Strings.java
```diff
@@ -345,7 +345,7 @@ public static String ensureSuffix(String s, String suffix) {
 		if (s.endsWith(suffix))
 			return s;
 
-		return s + suffix;
+		return s;
```
bjhargrave: This seems to defeat the purpose of the method since it now always returns `s`.

aQute.libg/src/aQute/lib/io/IO.java
```diff
@@ -186,6 +187,19 @@ public static OutputStream copy(InputStream in, OutputStream out) throws IOExcep
 		}
 	}
 
+	public static OutputStream copy(InputStream in, OutputStream out, int limit) throws IOException {
```
bjhargrave: We already have `LimitedInputStream` for this purpose. So we don't need this method which duplicates the behavior. Just do

```java
IO.copy(new LimitedInputStream(in, limit), out);
```

aQute.libg/src/aQute/lib/io/IO.java
```diff
@@ -1157,6 +1171,40 @@ public static PrintWriter writer(OutputStream out, Charset encoding) {
...
+		public void flush() throws IOException {
+			if (appendable instanceof Flushable) {
+				((Flushable) appendable).flush();
+			}
+		}
+
+		@Override
+		public void close() throws IOException {
+			flush();
+			if (appendable instanceof Closeable) {
```
bjhargrave: Shouldn't this be `AutoCloseable`?

aQute.libg/src/aQute/lib/io/IO.java
```diff
@@ -1157,6 +1171,40 @@ public static PrintWriter writer(OutputStream out, Charset encoding) {
...
+
+		private Appendable appendable;
+
+		public AppendableWriterAdapter(Appendable appendable) {
+			this.appendable = appendable;
+		}
+
+		@Override
+		public void write(char[] cbuf, int off, int len) throws IOException {
+			appendable.append(String.valueOf(cbuf), off, len);
```
bjhargrave: Maybe `appendable.append(String.valueOf(cbuf, off, len));`? That is, pass `off` and `len` to `String.valueOf`.

aQute.libg/src/aQute/lib/io/IO.java
```diff
@@ -1157,6 +1171,40 @@ public static PrintWriter writer(OutputStream out, Charset encoding) {
 		return new PrintWriter(new OutputStreamWriter(out, encoding));
 	}
 
+	static class AppendableWriterAdapter extends Writer {
+
+		private Appendable appendable;
```
bjhargrave: Make final.

biz.aQute.bndlib/src/aQute/bnd/service/result/Err.java
```diff
@@ -165,8 +165,18 @@ public V orElseGet(Supplier<? extends V> orElseSupplier) {
 		}
 	}
 
+	@Override
+	public void accept(ConsumerWithException<? super V> ok, ConsumerWithException<? super E> err) {
```
bjhargrave: Need to add `requireNonNull(ok);` for consistent behavior regardless of whether the Result is Ok or Err.

biz.aQute.bndlib/src/aQute/bnd/service/result/Ok.java
```diff
@@ -170,4 +170,13 @@ public V orElseGet(Supplier<? extends V> orElseSupplier) {
 	public String toString() {
 		return String.format("Ok(%s)", value);
 	}
+
+	@Override
+	public void accept(ConsumerWithException<? super V> ok, ConsumerWithException<? super E> err) {
```
bjhargrave: Need to add `requireNonNull(err);` for consistent behavior regardless of whether the Result is Ok or Err.

PR #3711, opened 2020-01-13
[jareditor] JarEditor Temp project support

gamerson:


gamerson:
thanks @pkriens  for the JarEditor update, it's really a pleasure to use.  This PR adds a couple of things.

1. Changes the label provider to workbench label provider so that we get pretty icons :)
Screen Shot 2020-01-13 at 8 55 15 AM

2. Adds a "Temp" project where all of the extracted jar files go instead of using "linked" folders.  Here are the reasons why I'm proposing the new temp project 

- Doesn't modify an existing project's .project file and potentially dirty-ing the git working tree
Screen Shot 2020-01-13 at 8 51 36 AM
- Doesn’t require an existing project to be available, e.g. you can use jar editor with empty workspace (in case of opening an external file)
Screen Shot 2020-01-13 at 9 13 08 AM

However, others committers may decide to go with the current mechanism "linked folders" approach which is certainly a valid approach, just comes down to preference of committers.

Here is a screenshot of how the "filter" shows up in the workbench preferences:
Screen Shot 2020-01-13 at 8 47 22 AM

bndtools.jareditor/src/bndtools/jareditor/internal/TemporaryProjectFilter.java
```diff
@@ -0,0 +1,27 @@
...
+	@Override
+	public boolean select(Viewer viewer, Object parentElement, Object element) {
+		if (element instanceof IJavaProject) {
+			element = ((IJavaProject) element).getProject();
+		}
+
+		return !Optional.ofNullable(element)
+			.filter(IProject.class::isInstance)
+			.map(IProject.class::cast)
+			.filter(Objects::nonNull)
```
bjhargrave: Why do you need the non-null filter? The `Class.isInstance` will already filter out null values.

bndtools.jareditor/src/bndtools/jareditor/internal/TemporaryProject.java
```diff
@@ -0,0 +1,133 @@
...
+import org.eclipse.core.runtime.SubMonitor;
+import org.eclipse.jdt.core.IClasspathEntry;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.launching.JavaRuntime;
+
+public class TemporaryProject {
+
+	public static final String PROJECT_NAME = "JAREditorTempFiles";
```
bjhargrave: Perhaps `Bndtools` should be in the name so people can associate seeing this with Bndtools?

bndtools.jareditor/src/bndtools/jareditor/internal/TemporaryFile.java
```diff
@@ -63,7 +64,14 @@
...
 
-	private static Result<IFolder, String> select() throws CoreException, IOException {
+	private static Result<IFolder, String> selectTempProject() throws CoreException, IOException {
+		IProject project = tempProject.getJavaProject()
+			.getProject();
+
+		return Result.ok(project.getFolder("temp"));
+	}
+
+	private static Result<IFolder, String> selectProject() throws CoreException, IOException {
```
bjhargrave: I don't see where this private method `selectProject` is ever called.

bndtools.jareditor/resources/plugin.xml
```diff
@@ -14,4 +14,55 @@
 			<run class="bndtools.jareditor.internal.JarFileSystem" />
 		</filesystem>
 	</extension>
+	<extension point="org.eclipse.ui.navigator.navigatorContent">
+		<commonFilter
```
bjhargrave: Does this hiding of the temp project also work with the new `Bndtools Explorer` view?

bndtools.jareditor/src/bndtools/jareditor/internal/TemporaryProjectFilter.java
gamerson: Good point, I can remove it.

bndtools.jareditor/src/bndtools/jareditor/internal/TemporaryProject.java
gamerson: Agreed, that would let people know where it is coming from when they encounter it.

bndtools.jareditor/src/bndtools/jareditor/internal/TemporaryFile.java
gamerson: It can be removed, I just left it in case we needed both mechanisms for some reason.

bndtools.jareditor/resources/plugin.xml
gamerson: Yes: That was added here: https://github.com/bndtools/bnd/pull/3711/files#diff-05adbc15428a40a3be593eb756feb105R51-R58

pkriens:
Thanks!

I'd not realized this changed the .project, agree it is a bad idea. However, I need to be in a Java project to make the .class files work. Does this work in your case?

gamerson:
> Thanks!
> 
>  However, I need to be in a Java project to make the .class files work. Does this work in your case?

Yes, the `BndtoolsJAREditorTempFiles` project we create is a standard java project with nominal classpath.  So .class files will load with class editor.

Issue #3528, opened 2019-11-04
[tester] Tester bundle exits if stopped while running in continuous mode


kriegfrj:
If you run the tester in continuous mode, and then stop the bundle (eg, using the gogo shell), then it will call `System.exit()` and terminate the entire process.

Terminating makes sense in one-shot and automatic mode; not so much in continuous mode.

This behaviour applies to both `biz.aQute.tester` and `biz.aQute.tester.junit-platform`.

bjhargrave:
I would like to see the whole launcher/tester process stop calling System.exit. See #2308.

kriegfrj:
Yes, I had noticed that the tmp files don't get cleaned up properly on Windows, and it is quite annoying especially when you're running on a small SSD.

Actually, the problem in the OP is slightly different to what I thought: it's not the tester that calls exit(), but it its run() method returns and control returns back to the Launcher (`mainThread.call()`), which then exits. To achieve what I was hoping for in the OP, we'd need to change the tester not to run as a runnable. Perhaps removing this line would fix it:
https://github.com/bndtools/bnd/blob/33b36ba0fbbceae2922bc09f5e5661b9868b2f44/biz.aQute.junit/src/aQute/junit/Activator.java#L77
... or else, only set `main.thread` if running in one-shot or automatic mode.

I agree that we should move towards getting the tester to not call `exit()`, but I think that this is a separate issue.

kriegfrj:
Ok, I've had a look at this one.

Simply removing the line that adds "main.thread=true" to the Runnable registration doesn't run. The registered runnables get run by the Launcher, and they are only run if the `main.thread` property is set to `true`.

> Yes, I had noticed that the tmp files don't get cleaned up properly on Windows, and it is quite annoying especially when you're running on a small SSD.
> 
> Perhaps removing this line would fix it:
> https://github.com/bndtools/bnd/blob/33b36ba0fbbceae2922bc09f5e5661b9868b2f44/biz.aQute.junit/src/aQute/junit/Activator.java#L77
> 
> 
> ... or else, only set `main.thread` if running in one-shot or automatic mode.

 I've had a quick look at the above, and this simple suggestion will not work.

The magic that automatically runs a registered `Runnable` service lives in the Launcher. The Launcher will only find & run runnables that have their `main.thread` property set to `true`. Removing this property means that the `run()` method is never called.

[...]

PR #3717, opened 2020-01-16
[tester] OSGiJUnitLaunchDelegate to handle restarting tester bundle

kriegfrj:
Loop around the call to `accept()` to accept reconnection attempts (such as what happens when the tester bundle is restarted).

Partial fix for #3528.

### DIFF
--- a/bndtools.core/src/bndtools/launch/OSGiJUnitLaunchDelegate.java
+++ b/bndtools.core/src/bndtools/launch/OSGiJUnitLaunchDelegate.java
@@ -183,20 +183,17 @@ public class OSGiJUnitLaunchDelegate extends AbstractOSGiLaunchDelegate {
 		ILaunch					launch;
 		IJavaProject			project;
 		TestRunSessionAndPort	testRunSession;
-		int						port;
 		DataOutputStream		outStr;
 
-		public ControlThread(ILaunch launch, IJavaProject project, int port) {
+		public ControlThread(ILaunch launch, IJavaProject project, int port) throws IOException {
 			this.launch = launch;
 			this.project = project;
-			this.port = port;
+			listener = new ServerSocket(port);
 		}
 
 		@Override
 		public void run() {
 			try {
-				listener = new ServerSocket(port);
-
 				// Note that this loop will terminate when accept() throws a
 				// SocketException. This ordinarily happens when the
 				// ServerSocket is closed in ControlThread.close(), which is in
--- a/bndtools.core/src/bndtools/launch/OSGiJUnitLaunchDelegate.java
+++ b/bndtools.core/src/bndtools/launch/OSGiJUnitLaunchDelegate.java
@@ -178,12 +178,12 @@ public class OSGiJUnitLaunchDelegate extends AbstractOSGiLaunchDelegate {
 
 	private static class ControlThread implements Runnable, AutoCloseable {
 
-		ServerSocket			listener;
-		Socket					socket;
-		ILaunch					launch;
-		IJavaProject			project;
-		TestRunSessionAndPort	testRunSession;
-		DataOutputStream		outStr;
+		final ServerSocket				listener;
+		final ILaunch					launch;
+		final IJavaProject				project;
+		volatile Socket					socket;
+		volatile TestRunSessionAndPort	testRunSession;
+		volatile DataOutputStream		outStr;
 
 		public ControlThread(ILaunch launch, IJavaProject project, int port) throws IOException {
 			this.launch = launch;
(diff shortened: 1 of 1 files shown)


bndtools.core/src/bndtools/launch/OSGiJUnitLaunchDelegate.java
```diff
@@ -197,12 +197,14 @@ public void run() {
 			try {
 				listener = new ServerSocket(port);
 
-				socket = listener.accept();
-				outStr = new DataOutputStream(socket.getOutputStream());
-				while (socket.getInputStream()
-					.read() != -1) {
-					testRunSession = new TestRunSessionAndPort(launch, project);
-					outStr.writeInt(testRunSession.getPort());
+				while (true) {
```
bjhargrave: So the only way out of this loop is throwing an exception? That seems like poor flow control.

kriegfrj: The loop spends the idle time in the call to `accept()`, which blocks until it receives an incoming connection attempt. The only way (that I have found) to interrupt it is by closing the socket from another thread, which causes `accept()` to throw a `SocketException`. If you know a better way, I'm all ears!

bjhargrave: In the current code, it waits at the accept, then enters the read loop until the socket is closed and then exits the thread. In you change, it never, exits the thread. It just loop back up again. There is no control mechanism to tell this thread it is done and should exit. This makes this class generate thread which are stuck waiting on a server socket.

kriegfrj: Sorry, I may not have been clear. When I said "the only way to interrupt [`accept()`] is by closing the [server] socket from another thread", I should have perhaps noted that this mechanism was already in place. This code should not leave any threads behind stuck waiting on a server socket.

Zombie threads were still a potential problem that needed to be handled even in the old code, because it would still wait at the `accept()` until the first connection attempt was made - a zombie thread could still arise if that first connection attempt never came.

The fix was to call `close()` on the server socket at termination time:

https://github.com/bndtools/bnd/blob/5905abe02ef9c1687ca01192aa0c0d59b407dbd7/bndtools.core/src/bndtools/launch/OSGiJUnitLaunchDelegate.java#L219

https://github.com/bndtools/bnd/blob/5905abe02ef9c1687ca01192aa0c0d59b407dbd7/bndtools.core/src/bndtools/launch/OSGiJUnitLaunchDelegate.java#L245-L246

This in turn causes the `accept()` call in the background thread to throw a `SocketException`, which exits the loop and allows the thread to die naturally.

The same mechanism still works in the new implementation. You can verify that this is the case by running a test - when the app-under-test exits, you should see this log message in your log:

https://github.com/bndtools/bnd/blob/5905abe02ef9c1687ca01192aa0c0d59b407dbd7/bndtools.core/src/bndtools/launch/OSGiJUnitLaunchDelegate.java#L207-L208

bjhargrave: So this only applies to rerunIDE mode for tester.junit-platform?

kriegfrj: Yes.

It also occurred to me if that if the loop's control flow is not immediately obvious to someone with your experience then a little commentary might be in order. I'll add some comments.

bndtools.core/src/bndtools/launch/OSGiJUnitLaunchDelegate.java
```diff
@@ -186,29 +186,36 @@ public void stopTestRun() {
 		int						port;
 		DataOutputStream		outStr;
 
-		public ControlThread(ILaunch launch, IJavaProject project, int port) {
+		public ControlThread(ILaunch launch, IJavaProject project, int port) throws IOException {
 			this.launch = launch;
 			this.project = project;
-			this.port = port;
+			listener = new ServerSocket(port);
```
bjhargrave: You also need to delete the `port` variable since it is no longer used. Also make `listener`, `launch`, and `project` variables final. The other variables should be volatile since they are written by one thread (run method) and read by another thread (close method).

kriegfrj: Done, done and done. I made the second two changes in a separate commit as they fix a separate issue. I also put a lock around the attempt to close `testRunSession` to avoid a race condition that could cause an NPE.

bjhargrave:
Thanks for all the work!

PR #3727, opened 2020-01-21
[contract] bnd finds 2 contracts in bundle that provides only 1 if `-contract` instruction is parameterized

rotty3000:


bjhargrave:
> is this too late for the release?

Well we still can. But we should discuss if we want another RC.

rotty3000:
@bjhargrave after further thought it's got a work around and I think it's too small to hold up the release.

PR #3732, opened 2020-01-23
More fixes of LGTM warnings

bjhargrave:


pkriens:
I am not too happy with the //lgtm... comments that disable a warning. I've seen at customers how cluttered the code becomes when you move  in that direction. Aren't there other ways to ignore those errors?

bjhargrave:
You can use SuppressWarnings but I found that to be more cumbersome and also required changing Eclipse setting to stop Eclipse from complaining about SuppressWarnings values Eclipse did not understand.

bjhargrave:
I made PR #3734 to replace the comments except for the one in AugmentRepository since I have no idea why the list exists. Things are put in it but nothing is ever read from it. @pkriens can you help with AugmentsRepository?

PR #3736, opened 2020-01-23
jareditor: Plugin must be started for the Bundle-Activator to be called

bjhargrave:
The code NPE'd trying to get the state location because no one started
the bundle. So we mark it lazy activation to get Eclipse to start it.


gamerson:
Nice catch.

PR #3741, opened 2020-01-24
Baseline 5.0

bjhargrave:
@rotty3000, @pkriens This is another step that needs to be taken after the new release is published in maven central.


rotty3000:
baseline, got it!

Issue #3745, opened 2020-01-28
[junit-platform tester] Sub-request to handle plain DiscoverySelector


kriegfrj:
The latest version of junit-platform-engine has changed its internal discovery algorithm, and in doing so exposed a bug in the JUnit Platform tester.

The bug relates to the following code:

https://github.com/bndtools/bnd/blob/362f0ae0eb6f26f4fceb0f36c0dd501fca7bcd4d/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java#L520-L533

The aim of this method is to resolve any raw `ClassSelector` or `MethodSelector` selectors (ie, those specified using Strings for the class names) into the versions that have fully loaded Class instances. This allows it to take care of the OSGi aspects of the classloading (which plain JUnit Platform is not adequately aware of), and it also turns a `BundleSelector` (which only `BundleEngine` understands) into `ClassSelector`s that can be understood by the sub-engines. 

Under the older version of JUnit Platform, calls to `EngineDiscoveryRequest.getSelectorsByType()` were always made using the individual concrete subclasses `ClassSelector` and `MethodSelector`. So the code path would go via L524 and return our fully-resolved version of the selectors rather than the original selector instance (which may include a String version).

Under the new version, it first calls `getSelectorsByType()` with the interface `DiscoverySelector.class`:

https://github.com/junit-team/junit5/blob/master/junit-platform-engine/src/main/java/org/junit/platform/engine/support/discovery/EngineDiscoveryRequestResolution.java#L80

This causes it to bypass L524 of `BundleSelectorResolver` and instead delegate to the original request.

In the common use case (running all tests in one or more bundles), the original request will typically comprise one or more `BundleSelector`s and nothing else. So these `BundleSelectors` will be returned. These are in turn passed on to the sub-engine for discovery, but the sub-engine, not understanding what a `BundleSelector` is, will simply ignore them. Thus, it doesn't discover any tests (hence the behaviour @bjhargrave discovered in osgi/osgi-test#70).

The fix:
* Instead of `selectorType.equals()`,  in L523, we should use something like `selectorType.isAssignableFrom()`.
* Always delegate to the original request, filter the result to remove `BundleSelector`s, `ClassSelector`s and `MethodSelector`, and add the pre-resolved versions of `ClassSelector`/`MethodSelector` instances to the return value if appropriate.

[...]

PR #3746, opened 2020-01-28
junit-platform: Fix Sub-request to handle plain DiscoverySelector

bjhargrave:
JUnit 5.5 changed to call getSelectorsByType with the base
DiscoverySelector type which caused us to fail to substitute the OSGi
ready selectors.

We change to call the original request and then filter out the selector
we will supply. We then add the selectors we will supply to the stream.

Fixes https://github.com/bndtools/bnd/issues/3745

### DIFF
--- a/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java
+++ b/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java
@@ -5,12 +5,10 @@ import static aQute.tester.bundle.engine.BundleEngine.CHECK_UNRESOLVED;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toSet;
 import static org.junit.platform.engine.discovery.DiscoverySelectors.selectClass;
-import static org.junit.platform.engine.discovery.DiscoverySelectors.selectMethod;
 
 import java.lang.annotation.Annotation;
 import java.lang.reflect.AnnotatedElement;
 import java.lang.reflect.Method;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


bjhargrave:
This fix solves the problem identified in https://github.com/osgi/osgi-test/pull/70#issuecomment-578339474.

biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java
```diff
@@ -520,16 +520,14 @@ private boolean elementHasJUnit4Annotations(AnnotatedElement element) {
...
 			info(() -> "Getting selectors from sub-request for: " + selectorType);
-			if (selectorType.equals(ClassSelector.class) || selectorType.equals(MethodSelector.class)) {
-				return selectors.stream()
-					.filter(selectorType::isInstance)
-					.map(selectorType::cast)
-					.collect(toList());
-			}
-			if (selectorType.equals(BundleSelector.class)) {
-				return new ArrayList<>();
-			}
```
kriegfrj: I'm wondering if it might be a useful optimisation to retain this as a short-circuit just before the main return statement?

kriegfrj:
I'd also like to see a regression test or two - I can put these together and test against your patch.

biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java
bjhargrave: I don't think this is worth it. It might sometimes be useful with pre 5.5 versions of JUnit 5 but will be counterproductive on later versions.

bjhargrave: I also found that no one (in our tests) ever calls this method with a `BundleSelector.class` argument anyway.

PR #3750, opened 2020-01-30
tester: Update JUnit 4 to 4.13 and Assertj to 3.15

bjhargrave:
We provide JUnit 4 packages in biz.aQute.junit but don't want to import
the latest version of these packages in biz.aQute.tester and
biz.aQute.tester.junit-platform since we want to work with slightly
older versions of JUnit 4.

We change to use the JUnit 4 (non-OSGi) artifacts instead of servicemix
since there does not appear to be any progress in updating servicemix.

We relocate the aQute.junit packages from the biz.aQute.junit project
to the biz.aQute.tester project so can compile them against the older
JUnit 4 packages.

The latest Assertj artifact changes the Bundle-Symbolic name to
assertj-core.


rotty3000:
Servicemix is presently releasing junior 4.13.

kriegfrj:
I notice that you've parameterized the JUnit versions and used those parameters in `central.mvn`. Should we perhaps do the same with JUnit Platform and Jupiter versions?

bjhargrave:
> Should we perhaps do the same with JUnit Platform and Jupiter versions?

Yes. See https://github.com/bndtools/bnd/pull/3753.

PR #3757, opened 2020-02-03
Pkriens/diverse

pkriens:


org.bndtools.templating.gitrepo/resources/processed/org/bndtools/templating/jgit/initialrepos.txt
```diff
@@ -1 +1 @@
-bndtools/workspace;branch=origin/${version;==;${base.version}}
```
bjhargrave: This is wrong. We must have version specific branches so older versions of Bndtools don't stop working when we update master for newer branches. We just forgot to update the repo for 5.0.0. (Need to add this task to the release process checklist.)

org.bndtools.templating.gitrepo/src/org/bndtools/templating/jgit/Cache.java
```diff
@@ -10,14 +10,21 @@
...
-    public byte[] download(URI uri) throws IOException {
+	public byte[] download(URI uri) throws Exception {
         byte[] data;
-        try (HttpClient client = new HttpClient()) {
+		try {
+			HttpClient client = Central.getWorkspace()
+				.getPlugin(HttpClient.class);
+
+			if (client == null)
+				client = new HttpClient();
```
bjhargrave: You should close this HttpClient. The old code does this.

pkriens: I've removed it, it must be there always

PR #3764, opened 2020-02-07
Indexing improvements

bjhargrave:


biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
```diff
@@ -312,12 +316,12 @@ public String put(String key, String value) {
 						}
 						break;
 				}
-				return map.put(attribute, value);
+				return attribute;
 			}
 		}
 		// default String type
 		types.remove(key);
-		return map.put(key, value);
```
pkriens: Why did you change the return type?

biz.aQute.bndlib/src/aQute/bnd/print/JarPrinter.java
```diff
@@ -286,25 +286,27 @@ public JarPrinter doManifest(Manifest manifest) {
...
-				sb.append(nibble(c));
+				sb.append('\\')
+					.append('u')
+					.append(nibble(c >> 12))
+					.append(nibble(c >> 8))
+					.append(nibble(c >> 4))
+					.append(nibble(c));
 			}
 		}
-		return sb.toString();
```
pkriens: If you're going to uglify the code all over the place, should you not at least move this into a central method? Repeating this low level checks all over the place is error prone.

biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
bjhargrave: I split the method in two. One part handles computing the type and modifying the `type` map and the other part modifies the `map` map.

PR #3767, opened 2020-02-08
Add parameter to skip bundle version check

### LINKED ISSUE #3446: Baseline - allow skipping exported package versions check and bundle version checks individually
We are in the process of migrating a large numble of bundles to using the `bnd-maven-plugin` and the `bnd-baseline-maven-plugin` instead of the `maven-bundle-plugin`. One part which is a bit of a bump in the process is that the baselining functionality now fails in case the expected bundle version was not changed.

This is a problem for us since we use a different versioning conventions:

- development releases are always odd
- release versions are always even

We recently had a case where a change was made, and the baseline correctly flagged the bundle version as needing a version bump from 1.2.6 to 1.3.0. However, we do not do this at development time, so we would need to change to 1.3.1-SNAPSHOT and then release at 1.3.2, which is confusing for consumers - where is 1.3.0 ?

Our workflow could be best adapted by allowing to control baselining for exported package versions and bundle versions.

kwin:
This closes #3446

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/baseline/BaselineTest.java
+++ b/biz.aQute.bndlib.tests/test/test/baseline/BaselineTest.java
@@ -491,6 +491,26 @@ public class BaselineTest extends TestCase {
 	}
 
 	// Adding a method to a ProviderType produces a MINOR bump (1.0.0 -> 1.1.0)
+	// in package, but bundle version should be ignored
+	public void testBundleVersionDiffignore() throws Exception {
+		Processor processor = new Processor();
+
+		DiffPluginImpl differ = new DiffPluginImpl();
+		differ.setIgnore("Bundle-Version");
+		Baseline baseline = new Baseline(processor, differ);
+
+		try (Jar older = new Jar(IO.getFile("testresources/api-orig.jar"));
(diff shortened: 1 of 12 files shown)


kwin:
Ok, I will try to come up with something...

kwin:
@bjhargrave Actually `-diffignore: Bundle-Version` works already as expected. I just added some tests for that and clarified the documentation in the bnd-baseline-maven-plugin.

bjhargrave:
Thanks for following up on this. We both learned how to disable bundle version checks. In hindsight, it seems sort of obvious :-)

saumya1singh:
Hi, how can we use this param in a maven command? @bjhargrave @kwin 
maybe something like `mvn clean install deploy -diffignore`
I know that's wrong format

bjhargrave:
> how can we use this param in a maven command?

You cannot with version 5.2.0. At your suggestion I just add properties to 6.2.0 with https://github.com/bndtools/bnd/pull/5051.

So once 6.2.0 releases (probably mid-Feb), you could say:

```
mvn clean install deploy -Dbnd.baseline.diffignores=Bundle-Version
```

saumya1singh:
sounds great. thanks @bjhargrave 👍🏻

PR #3777, opened 2020-02-14
[doc] Manifest Annotations

pkriens:
- Added shortcuts for MavenBndRepository GAVs
- Explained filter generation with the @Requirement

### DIFF
--- a/docs/_chapters/230-manifest-annotations.md
+++ b/docs/_chapters/230-manifest-annotations.md
@@ -8,9 +8,7 @@ Manifest headers are challenging to keep in sync with the code in the bundle. It
 
 One of the goals of bnd is to eliminate such issues by relying on Java's type system to express the semantics of OSGi metadata.
 
-## Bundle Annotations
-
-To address this bnd developers pioneered _manifest annotations_ which evolved into OSGi's [_bundle annotations_](https://osgi.org/specification/osgi.core/7.0.0/framework.api.html#org.osgi.annotation.bundle). A **bundle annotation** is used to express metadata that cannot otherwise be derived from code.
+To address this bnd pioneered _manifest annotations_ which evolved into OSGi's [_bundle annotations_](https://osgi.org/specification/osgi.core/7.0.0/framework.api.html#org.osgi.annotation.bundle). A **bundle annotation** is used to express metadata that cannot otherwise be derived from code.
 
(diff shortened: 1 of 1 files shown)


docs/_chapters/230-manifest-annotations.md
```diff
@@ -72,6 +71,26 @@ Require-Capability: \
 
 The invisible link created between user code and the indirect requirement is a powerful mechanism that enables automatic validation of a bundle closure.
 
+The actual requirement `filter:` directive is constructed from an `AND` of the `filter()`, `name()`, and `version()` annotation methods. All fields are optional. The name field will create an assertion that the given namespace equals the value of the `name()` annotation method. For example, if the namespace is `com.example.foo` and the `name()` method has the value `bar` then the filter is `(com.example.foo=bar)`. If a version is specified, it will be expanded to a filtered version-range expression.  The convention of using the namespace name as the property key is commonly used in OSGi specification. For example, the filter `(osgi.wiring.package=com.example.foo)` is the filter for an Import-Package `com.example.foo` while `osgi.wiring.package` is the namespace for the packages.
+
+The filter will **not** be processed for macros like the `@Header` annotation is. The filter can therefore not refer to any annotation methods in meta-annotated classes.
```
bjhargrave: This is not entirely true. See https://bnd.bndtools.org/chapters/235-accessor-properties.html.

pkriens:
I spent half an hour going over the `getFilter()` method and I fail to see where -any- macro processing takes place on the filter. Am I missing something?

docs/_chapters/230-manifest-annotations.md
pkriens: I spent half an hour going over the getFilter() method and I fail to see where -any- macro processing takes place on the filter. Am I missing something?

rotty3000: All results of AnnotationsHeaders are parsed by macros. It's not that the _filter_ itself is processed, but the entire end result is processed here https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java#L979
In this way, any string contents of any attribute or value can contain macros.

rotty3000: As @bjhargrave referenced, we have an entire feature built around this behaviour (which _is_ tested) :) Furthermore, additional bnd features beyond bundle annotations use this such as SPI Annotations.

pkriens: You are right, they are processed properly. But looking at Tim I was not the only one confused.

Issue #3773, opened 2020-02-13
[bndtools] Editor support for bnd files outside of current workspace


kriegfrj:
Occasionally, when I'm working in Eclipse on a Bnd workspace, I want to open a Bnd file from another workspace for reference. However, when you do this the editor component fails with an NPE:

![BndEditorNPE](https://user-images.githubusercontent.com/7224087/74406386-6f19db00-4e7f-11ea-9b1f-9d31ca570660.png)

Easy enough to work around by opening the file in another text editor.

pkriens:
The NPE is caused by the code getting an IEditorInput it cannot transform to a Resource. Imho this IEditorInput is a mess (you need to look at the type to see how to convert it to a resource and there are basically an undefined set of types.) 
Could you track where the file comes from you try to open?

kriegfrj:
Hi @pkriens, thanks for your response. I'm not sure I 100% understand your question.

Here are the steps to reproduce:

1. Open a bnd workspace in Eclipse.
2. Go **File|Open File...**
3. Open a `bnd.bnd` file somewhere on your file system that is *not* part of the the bnd workspace that you opened in 1.

My use case is that I often want to be able to open a `bnd.bnd` file from the Bndtools workspace for reference, while I'm working on my own workspace.

Your description of the cause of the NPE makes sense - my (admittedly limited) understanding of the Eclipse workspace model is that `IResources` can only represent resources that are part of the current Eclipse workspace, so when it tries to resolve the "foreign" `bnd.bnd` file it probably resolves to `null`.

I think that the Bndtools editor should be able to handle this case more gracefully - ideally to be able to show the file with appropriate syntax highlight, but if that's not feasible then at the very least get out of the way and let Eclipse's default text editor display the result if it can't.

pkriens:
It is not only the IResource itself. In several places we need to know the workspace and project. The IResource is used to find the project the bnd.bnd belongs to.

I've made an attempt to handle this: #3779

PR #3779, opened 2020-02-14
[bndeditor] Fix for NPE while opening a file outside the workspace

pkriens:
When the IEditorInput is not mappable to a resource
but we can map it to a Java File, we allow inputResource
to be null.

To load the editor model, we then create a standalone
Run for the project.

This class could use some major cleanup. One wonders
if by now there is not better editors support in 
Eclipse?


bjhargrave:
Fixes #3773

PR #3788, opened 2020-02-18
Add Example for using -javaagent parameter

chrisrueger:
* example about JaCoCo as a javaagent
* based on discussion in https://groups.google.com/forum/#!topic/bndtools-users/fjavKvFERBg

Signed-off-by: Christoph Rueger <chrisrueger@gmail.com>


bjhargrave:
Thanks for the PR!

PR #3792, opened 2020-02-20
Allow ProjectTester to be called on ProjectLauncher.update

bjhargrave:


biz.aQute.bndlib/src/aQute/bnd/build/ProjectLauncher.java
```diff
@@ -61,6 +61,7 @@
 
 	final static Logger					logger				= LoggerFactory.getLogger(ProjectLauncher.class);
 	private final Project				project;
+	private final List<Runnable>		onUpdate			= new ArrayList<>();
```
kriegfrj: Does this need to be made thread-safe? Or are we assuming/mandating that all calls to `update()`/`onUpdate()` are made from the same thread?

bjhargrave: There are many other things in the launcher which are not fully thread safe. The calls to onUpdate and update don't need to be on the same thread, there just needs to be some memory barriers between them and there are generally plenty of these. I don't see this as a real problem here now.

PR #3793, opened 2020-02-20
Adds a log warning if no sources available

juergen-albert:
If the maven-release instruction sets a source path, it is not obvious
why and if no sources are packaged. Thus a warning will be logged.

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
@@ -249,6 +249,11 @@ public class MavenBndRepository extends BaseRepository implements RepositoryPlug
 					.orElse(null);
 			} else {
 				pom = createPomFromFile(options.context.getFile(instructions.pom.path));
+				if (pom == null) {
+					logger.warn(
+						"A pom path was set in the -maven-release instuction, but no file could be found in {} (base path: {}) ",
+						instructions.pom.path, options.context.getBase());
+				}
 			}
 		} else {
 			if (!configuration.ignore_metainf_maven()) {
@@ -291,6 +296,11 @@ public class MavenBndRepository extends BaseRepository implements RepositoryPlug
 
 	private Jar getSources(Tool tool, Processor context, String path) throws Exception {
 		Jar jar = toJar(context, path);
+		if (path != null && jar == null) {
+			logger.warn(
+				"A sources path was set in the -maven-release instuction, but no sources could be found in {} (base path: {}) ",
+				path, context.getBase());
+		}
 		if (jar != null) {
 			tool.setSources(jar, "");
 		} else {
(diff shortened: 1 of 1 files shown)


juergen-albert:
True, the javadoc needs something like this too and while we are at it, the pom as well. 

I would not add it to the toJar method, because we don't know where we came from. Meaning we could not give a meaningful message, that states the he has trouble find the sources/javadoc/pom

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
```diff
@@ -249,6 +249,12 @@ private Resource getPom(PutOptions options, ReleaseDTO instructions, Jar binary)
 					.orElse(null);
 			} else {
 				pom = createPomFromFile(options.context.getFile(instructions.pom.path));
+				if (pom == null) {
+					logger.warn(
+						"A pom path was set in the -maven-release instuction, but no file could be found in {} (base path: {}) ",
+						instructions.pom.path, options.context.getBaseURI()
```
bjhargrave: I think the 2nd arg should just be `options.context.getBase()`. I don't see why we would want to use a URI here.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
```diff
@@ -291,6 +297,12 @@ private boolean isLocal(ReleaseDTO instructions) {
 
 	private Jar getSources(Tool tool, Processor context, String path) throws Exception {
 		Jar jar = toJar(context, path);
+		if (path != null && jar == null) {
+			logger.warn(
+				"A sources path was set in the -maven-release instuction, but no sources could be found in {} (base path: {}) ",
+				path, context.getBaseURI()
```
bjhargrave: I think the 2nd arg should just be `options.context.getBase()`. I don't see why we would want to use a URI here. We also should not `toString()` it in case logging does not happen. Leave the `toString()` call to the logging code.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
```diff
@@ -305,6 +317,12 @@ private Jar getSources(Tool tool, Processor context, String path) throws Excepti
 	private Jar getJavadoc(Tool tool, Processor context, String path, Map<String, String> options, boolean exports)
 		throws Exception {
 		Jar jar = toJar(context, path);
+		if (path != null && jar == null) {
+			logger.warn(
+				"A javadoc path was set in the -maven-release instuction, but no javadoc could be found in {} (base path: {}) ",
+				path, context.getBaseURI()
```
bjhargrave: I think the 2nd arg should just be `options.context.getBase()`. I don't see why we would want to use a URI here. We also should not `toString()` it in case logging does not happen. Leave the `toString()` call to the logging code.

bjhargrave:
Thanks!

PR #3804, opened 2020-02-26
[docs] Update/correct tester docs with latest version information

kriegfrj:
* Old docs referred to 4.4 release which became the 5.0 release.
* Old docs referred to Eclipse Oxygen, which is now Eclipse 2018-12.
* Added note about 5.6 release of JUnit being bundled.
* Added note about incompatibility with JUnit 5.5.

Fix includes HEAD and 5.0.0 release notes.

### DIFF
--- a/docs/_chapters/310-testing.md
+++ b/docs/_chapters/310-testing.md
@@ -72,7 +72,7 @@ You can find a bndtools project that shows this at [Github](https://github.com/b
 
 ## Testing With JUnit 5 - `biz.aQute.tester.junit-platform`
 
-As of 4.4, bnd includes a new tester bundle `biz.aQute.tester.junit-platform` that supports JUnit 5.
+As of Bnd 5.0, bnd includes a new tester bundle `biz.aQute.tester.junit-platform` that supports JUnit 5.
 
 As per the [JUnit 5 documentation](https://junit.org/junit5/docs/current/user-guide/#overview-what-is-junit-5), JUnit 5 is comprised of three modules:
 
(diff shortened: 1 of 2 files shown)


docs/_chapters/310-testing.md
```diff
@@ -147,7 +147,14 @@ As noted above, `biz.aQute.tester.junit-platform` requires JUnit Platform (and i
...
 ```
-Alternatively, it is not difficult to download the required (non-OSGi) modules from Maven Central and include them as-is on `-runpath`, or else (preferably) wrap them into bundles and include them in `-runrequires`/`-runbundles`.
+
+Alternatively, it is not difficult to download the required (non-OSGi) modules from Maven Central and include them as-is on `-runpath`, or else (preferably) wrap them into bundles and include them in `-runrequires`/`-runbundles`. 
+As of JUnit 5.6, the JUnit jars already have the OSGi metadata and so can be used as bundles 
+direct from Maven Central.
+
+Also note that unfortunately, due to a bug in `biz.aQute.tester.junit-platform`, Bndtools 5.0 
+*does not work with JUnit 5.5+*. This has already been fixed in the main development branch and we expect the 
+fix to be included in Bndtools 5.0.1 (to be released soon).
```
bjhargrave: 5.0.1 is possible. I have made a test build. But we still need to collectively discuss and agree on a 5.0.1 release. So is it safer to simply say the release after 5.0.

docs/releases/5.0.0/chapters/310-testing.html
```diff
@@ -497,9 +497,11 @@ <h4 id="finding-the-junit-platform-bundles">Finding the JUnit Platform bundles</
...
 </code></pre></div>    </div>
-    <p>Alternatively, it is not difficult to download the required (non-OSGi) modules from Maven Central and include them as-is on <code class="highlighter-rouge">-runpath</code>, or else (preferably) wrap them into bundles and include them in <code class="highlighter-rouge">-runrequires</code>/<code class="highlighter-rouge">-runbundles</code>.</p>
   </li>
 </ul>
+<p>Alternatively, it is not difficult to download the required (non-OSGi) modules from Maven Central and include them as-is on <code class="highlighter-rouge">-runpath</code>, or else (preferably) wrap them into bundles and include them in <code class="highlighter-rouge">-runrequires</code>/<code class="highlighter-rouge">-runbundles</code>. As of JUnit 5.6, the JUnit jars already have the OSGi metadata and so can be used as bundles direct from Maven Central.</p>
+
+<p>Also note that unfortunately, due to a bug in <code class="highlighter-rouge">biz.aQute.tester.junit-platform</code>, Bndtools 5.0 <em>does not work with JUnit 5.5+</em>. This has already been fixed in the main development branch and we expect the fix to be included in Bndtools 5.0.1 (to be released soon).</p>
```
bjhargrave: 5.0.1 is possible. I have made a test build. But we still need to collectively discuss and agree on a 5.0.1 release. So is it safer to simply say the release after 5.0.

bjhargrave:
Thanks!

PR #3806, opened 2020-02-26
Adds org.osgi.dto Package to fixe ClassNotFoundException in gradle build when using P2Repository

juergen-albert:
adds osgi.dto pckg to fix CNF Exception in gradle for p2 repo

* addionally counts the package version up, because baselineing
complained about it

### DIFF
--- a/biz.aQute.bndlib/bnd.bnd
+++ b/biz.aQute.bndlib/bnd.bnd
@@ -5,7 +5,8 @@ Bundle-Description: bndlib: A Swiss Army Knife for OSGi
 
 -includepackage: \
 	aQute.bnd.*,\
-	aQute.lib.spring
+	aQute.lib.spring,\
+	org.osgi.dto
 
 Export-Package: \
     aQute.bnd.build;-noimport:=true,\
@@ -55,6 +56,7 @@ Export-Package: \
 Import-Package: \
 	org.osgi.service.repository,\
 	org.osgi.service.log,\
+	org.osgi.dto,\
 	${replacelist;${retainall;${packages;CONDITIONAL};${packages;NAMED;org.osgi.*}};$;\\;provide:=false},\
 	*
 
--- a/biz.aQute.bnd.reporter/bnd.bnd
+++ b/biz.aQute.bnd.reporter/bnd.bnd
@@ -14,18 +14,10 @@ jtwig.version: 5.86.1.RELEASE
 	artifactId=jtwig-reflection;\
 	version=${jtwig.version};\
 	scope=compile
-
-dto.version: 1.0.0
--maven-dependencies.dto:\
-	dto;\
-	groupId=org.osgi;\
-	artifactId=org.osgi.dto;\
-	version=${dto.version};\
-	scope=compile
  
 -buildpath: \
 	osgi.annotation;version=latest;maven-scope=provided,\
-	org.osgi.dto;version=${dto.version},\
+	org.osgi.dto;version=latest,\
 	aQute.libg;version=project,\
 	biz.aQute.bndlib;version=latest,\
 	slf4j.api;version=latest,\


biz.aQute.repository/bnd.bnd
```diff
@@ -41,7 +41,8 @@ Export-Package: \
 	aQute.maven.*, \
 
 -includepackage: \
-	aQute.p2.provider
+	aQute.p2.provider,\
+	org.osgi.dto;-split-package:=first
```
bjhargrave: I think it is wrong to put a copy of a core OSGi package in this jar. It if needs to go somewhere, it should centrally be in biz.aQute.bndlib.

juergen-albert: @pkriens wrote to me on slack, that we should put it in the gradle plugin. My gut feeling tells me, that bndlib is the better place. Outside of an OSGi FW the package is missing anyway and inside the FW, it does not matter if we export it too. I furthermore take an educated guess by predicting, that the P2Repo might not be the only thing that will use DTOs in the future.
BTW: Is there any significant difference between the ```aQute.bnd.util.dto.DTO ```and the ```org.osgi.dto.DTO```?

pkriens: This means that bndlib potentially conflicts in Eclipse since there is already an OSGi framework around. For this reason, I think it makes more sense to get the OSGi classes from the driver? If not, you might run into a class space separation with the uses constraints if you run on a framework that has a different version? This is a fuzzy area so I am not dogmatic.

@juergen-albert The aQute classes were the inspiration for the OSGi DTO. So they existed long before the OSGi variant.

bjhargrave: bndlin already contains many OSGi Core packages that it imports but does not export. So when used in an OSGi runtime, bndlib uses the imported package. When used in a flat classpath, bndlib can use its own copy of the classes. So adding org.osgi.dto in this manner, like org.osgi.framework and org.osgi.resource, make sense.

If we want to make the driver provide these classes, we will need to make larger changes. I am fine with this also. It will affect maven, gradle and bnd (standalone). It will also affect anyone else who is using bndlib as they will need to change to provide these OSGi Core classes.

```
org/osgi/framework/AdaptPermission$1.class
org/osgi/framework/AdaptPermission.class
org/osgi/framework/AdaptPermissionCollection.class
org/osgi/framework/AdminPermission$1.class
org/osgi/framework/AdminPermission.class
org/osgi/framework/AdminPermissionCollection.class
org/osgi/framework/AllServiceListener.class
org/osgi/framework/Bundle.class
org/osgi/framework/BundleActivator.class
org/osgi/framework/BundleContext.class
org/osgi/framework/BundleEvent.class
org/osgi/framework/BundleException.class
org/osgi/framework/BundleListener.class
org/osgi/framework/BundlePermission.class
org/osgi/framework/BundlePermissionCollection.class
org/osgi/framework/BundleReference.class
org/osgi/framework/CapabilityPermission$1.class
org/osgi/framework/CapabilityPermission$Properties.class
org/osgi/framework/CapabilityPermission.class
org/osgi/framework/CapabilityPermissionCollection.class
org/osgi/framework/Configurable.class
org/osgi/framework/Constants.class
org/osgi/framework/Filter.class
org/osgi/framework/FrameworkEvent.class
org/osgi/framework/FrameworkListener.class
org/osgi/framework/FrameworkUtil$1.class
org/osgi/framework/FrameworkUtil$CaseInsensitiveMap.class
org/osgi/framework/FrameworkUtil$DNChainMatching.class
org/osgi/framework/FrameworkUtil$FilterImpl$Parser.class
org/osgi/framework/FrameworkUtil$FilterImpl.class
org/osgi/framework/FrameworkUtil$ServiceReferenceMap.class
org/osgi/framework/FrameworkUtil$SetAccessibleAction.class
org/osgi/framework/FrameworkUtil.class
org/osgi/framework/InvalidSyntaxException.class
org/osgi/framework/PackagePermission$1.class
org/osgi/framework/PackagePermission.class
org/osgi/framework/PackagePermissionCollection.class
org/osgi/framework/PrototypeServiceFactory.class
org/osgi/framework/ServiceEvent.class
org/osgi/framework/ServiceException.class
org/osgi/framework/ServiceFactory.class
org/osgi/framework/ServiceListener.class
org/osgi/framework/ServiceObjects.class
org/osgi/framework/ServicePermission$1.class
org/osgi/framework/ServicePermission$Properties$Entry.class
org/osgi/framework/ServicePermission$Properties.class
org/osgi/framework/ServicePermission.class
org/osgi/framework/ServicePermissionCollection.class
org/osgi/framework/ServiceReference.class
org/osgi/framework/ServiceRegistration.class
org/osgi/framework/SignerProperty.class
org/osgi/framework/SynchronousBundleListener.class
org/osgi/framework/UnfilteredServiceListener.class
org/osgi/framework/Version.class
org/osgi/framework/VersionRange.class
org/osgi/framework/launch/Framework.class
org/osgi/framework/launch/FrameworkFactory.class
org/osgi/framework/launch/package-info.class
org/osgi/framework/launch/packageinfo
org/osgi/framework/package-info.class
org/osgi/framework/packageinfo
org/osgi/resource/Capability.class
org/osgi/resource/Namespace.class
org/osgi/resource/Requirement.class
org/osgi/resource/Resource.class
org/osgi/resource/Wire.class
org/osgi/resource/Wiring.class
org/osgi/resource/package-info.class
org/osgi/resource/packageinfo
org/osgi/util/tracker/AbstractTracked.class
org/osgi/util/tracker/BundleTracker$Tracked.class
org/osgi/util/tracker/BundleTracker.class
org/osgi/util/tracker/BundleTrackerCustomizer.class
org/osgi/util/tracker/ServiceTracker$AllTracked.class
org/osgi/util/tracker/ServiceTracker$Tracked.class
org/osgi/util/tracker/ServiceTracker.class
org/osgi/util/tracker/ServiceTrackerCustomizer.class
org/osgi/util/tracker/package-info.class
org/osgi/util/tracker/packageinfo
```

pkriens: a) if they are private I guess it should be ok since we do not rely on OSGi classes in bndlib for internal functionality. I think they are all about OSGi knowledge?
b) The driver approach seems a lot cleaner

I leave it up to you ...

bjhargrave: So you are ok with removing the org.osgi.framework, org.osgi.resource packages from bndlib and requiring all users of bndlib, such as the maven plugins, gradle plugins, bnd.jar and all others to provide any needed OSGi packages?

For maven and gradle, it is easy to make the required OSGi package available by using maven dependencies in the POM. For bnd.jar, we just need to -includepackage the necessary packages. External users of bndlib will need to know which packages they need if they are not using maven dependencies in the POM to build their classpath.

pkriens: hmm, that is true. Shall we discuss this Friday?

biz.aQute.bndlib/bnd.bnd
```diff
@@ -73,6 +75,7 @@ Import-Package: \
 	org.osgi.service.repository;version=latest,\
 	org.osgi.util.function;version=latest,\
 	org.osgi.util.promise;version=latest,\
+	org.osgi.dto;version=latest,\
```
bjhargrave: osgi.core (above) already contains the dto package. So you do not need to add org.osgi.dto to the buildpath.

bjhargrave: I think you deleted the wrong line. You need org.osgi.dto in the Import-Package statement. You do not need org.osgi.dto in the -buildpath instruction since osgi.core is already in the -buildpath instruction and it contains the org.osgi.dto package.

juergen-albert:
This should be it now. We only require this Package if we are not running in a OSGi Framework like gradle. Thus the Package simply needs to be included without any Import or Export Statement. The Import is not necessary, because the Bundles itself is not using the DTO class itself. An Export is not necessary because in a Framework it should be provided by the framework, as we are building against R6.

I initially had the dto bundle as build dependency to get the dependency entry in the pom.xml. I've overlooked that the core was missing. Therefore I added the maven dependency part for the core bundle.

biz.aQute.bndlib/bnd.bnd
```diff
@@ -62,6 +63,12 @@ Import-Package: \
 
 -maven-scope: provided
 
+-maven-dependencies.core: \
```
bjhargrave: This is still wring. You don't need this since bndlib itself contains the package for non-OSGi use cases. Doing thing just adds a second copy of the package to the classpath for maven/gradle.

Also, you need to readd the Import-Package statement for org.osgi.dto. Even though bndlib itself does not currently use the org.osgi.dto package, it does no harm to import it and it prevents us from making silly mistakes in the future when we actually do start to use it in bndlib and then forget to add the Import-Package for it.

bjhargrave:
Thanks!

PR #3809, opened 2020-02-28
Update version and fixed appearance of links

lefou:


lefou:
Thanks for merging.

There is one think I can't understand. Why do you drop authorship when merging. All my previous contributions lost my name in the process of merging. There is no technical explanation to me, as git differentiates between author and committer. It even supports co-authors as does GitHub, to not loose author attribution. Please review your handling of contributions! It's not the proper open source way.

lefou:
Thanks for merging.

(Edit: I'm sorry, I miss-read the git history and blamed you for dropping authorship in a previous version of the comment. That's not the case. I'm sorry for that.)

PR #3814, opened 2020-03-02
[launcher] Added a connect mode to the launcher.

pkriens:
The OSGi Alliance is working on a Connect specification.
This specification will allow an SPI to take over some
of the module handling in an OSGi Framework.

This change will first look if there is a Module
Connector loadable via the Service Loader. If so,
it will try to use OSGi Connect with the found
Module Connector. If anything fails, it will fall
back to the classic mode. I.e, the types of OSGi
Connect are not required to be part of the class path.

Also, a static method is added to call the launcher
externally. This method can take an optional 
Properties and/or an optional Bundle Activator. The
Bundle Activator is called before the bundles are started.

An implicit change is that the code now uses the Service Loader
for the framework as well. This code was originally 
artisan but we're now solidly on Java 8.

### DIFF
--- a/biz.aQute.launcher/src/aQute/launcher/Launcher.java
+++ b/biz.aQute.launcher/src/aQute/launcher/Launcher.java
@@ -182,23 +182,23 @@ public class Launcher implements ServiceListener {
 	 * `launcher.properties`.
 	 *
 	 * @param args
-	 * @param propertiesOrNull the properties or null
-	 * @param activatorOrNull null, or a Bundle Activator that will be called
+	 * @param properties the properties or null
+	 * @param activator null, or a Bundle Activator that will be called
 	 *            back before the bundles are started, see
 	 *            {@link EmbeddedActivatorPhase#BEFORE_BUNDLES_START}.
 	 * @return the exit code
 	 * @throws Throwable
 	 */
-	public static int launch(String[] args, Properties propertiesOrNull, BundleActivator activatorOrNull)
+	public static int launch(String[] args, Properties properties, BundleActivator activator)
 		throws Throwable {
 		Launcher target;
-		if (propertiesOrNull == null) {
+		if (properties == null) {
 			target = new Launcher();
 			target.loadPropertiesFromFileOrEmbedded();
 		} else {
-			target = new Launcher(propertiesOrNull);
+			target = new Launcher(properties);
 		}
-		target.externalActivator = activatorOrNull;
+		target.externalActivator = activator;
 		return target.launch(args);
 	}
 
--- a/biz.aQute.launcher/src/aQute/launcher/Launcher.java
+++ b/biz.aQute.launcher/src/aQute/launcher/Launcher.java
@@ -9,7 +9,6 @@ import static java.lang.invoke.MethodHandles.publicLookup;
 import static java.lang.invoke.MethodType.methodType;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
-import java.io.BufferedReader;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -45,6 +44,7 @@ import java.util.Enumeration;
 import java.util.Formatter;
 import java.util.Hashtable;
 import java.util.IllegalFormatException;
+import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.LinkedList;
 import java.util.List;
@@ -52,6 +52,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.ServiceLoader;
 import java.util.StringTokenizer;
 import java.util.concurrent.Callable;
 import java.util.concurrent.Executors;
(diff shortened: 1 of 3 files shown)


biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -1122,37 +1161,30 @@ public void run() {
...
 				"unchecked", "rawtypes"
 			})
 			Map<String, String> configuration = (Map) p;
-			systemBundle = factory.newFramework(configuration);
-			trace("framework instance %s", systemBundle);
+
+			systemBundle = createConnect(loader, configuration);
+			if (systemBundle == null) {
+				systemBundle = createClassic(systemBundle, loader, configuration);
+			}
```
bjhargrave: It seems wrong to always try connect first. Equinox and Felix will eventually support connect and thus will have the service for connect. So this code will then always assume connect mode rather than normal OSGi framework mode. I think using connect must be some option on `-runfw` or some other bndrun file instruction.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -1175,38 +1207,58 @@ public void run() {
...
-
-			try (BufferedReader rdr = IO.reader(url.openStream(), UTF_8)) {
-				String line;
-				while ((line = rdr.readLine()) != null) {
-					trace("%s", line);
-					line = line.trim();
-					if (!line.startsWith("#") && line.length() > 0) {
-						factories.add(line);
-					}
+	private Framework createConnect(ClassLoader loader, Map<String, String> configuration) {
```
bjhargrave: How do we test this actually works? We have tests now for normal OSGi framework launching (both explicit and implicit tests).

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -169,6 +176,32 @@ public static int run(String args[]) throws Throwable {
...
+	 *
+	 * @param args
+	 * @param propertiesOrNull the properties or null
+	 * @param activator null, or a Bundle Activator that will be called back
+	 *            before the bundles are started, see
+	 *            {@link EmbeddedActivatorPhase#BEFORE_BUNDLES_START}.
+	 * @return the exit code
+	 * @throws Throwable
+	 */
+	public static int launch(String[] args, Properties propertiesOrNull, BundleActivator activator)
```
bjhargrave: We have `propertiesOrNull` but not `activatorOrNull`? I am not sure I like the `OrNull` variable name suffix, but we should at least be consistent.

biz.aQute.launcher/src/org/osgi/framework/connect/ConnectContent.java
```diff
@@ -0,0 +1,71 @@
...
+ * limitations under the License.
+ */
+package org.osgi.framework.connect;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Map;
+import java.util.Optional;
+public interface ConnectContent {
```
bjhargrave: I think it is wrong for us to have a copy of the package's source code in our repo. We can easily miss important changes as the OSGi spec changes in development. We should instead reference the OSGi snapshot jar to access the package. https://oss.sonatype.org/content/repositories/snapshots/org/osgi/org.osgi.framework/1.10.0-SNAPSHOT/

biz.aQute.launcher/src/aQute/launcher/Launcher.java
pkriens: Nope, it only uses connect when it can find a Module Connector. Why would you put a Module Connector on the -runpath when you do not intend to use it?

biz.aQute.launcher/src/aQute/launcher/Launcher.java
pkriens: I want it in there so we can experiment. Once the frameworks are available we can have tests. You can only activate the code by putting a Module Connector services on your -runpath, which seems a good safeguard this is not activated accidentally..

biz.aQute.launcher/src/org/osgi/framework/connect/ConnectContent.java
pkriens: I wanted to placate you :-) I talked to Thomas and Karl and they seemed rather anxious in this area. But if we can import this snapshot, that would be great.

pkriens: Hmm, but then we take all Framework packages from the snapshot?

bjhargrave: If we have the snapshot jar following the released jar on the `-buildpath`, and we use `-split-package:=first` for the package instruction, we will get the released packages except for connect which is not in the released jar, no?

biz.aQute.launcher/src/aQute/launcher/Launcher.java
bjhargrave: We should then doc this to make sure no one is surprised that a thing on runpath changes the launcher behavior.

biz.aQute.launcher/src/org/osgi/framework/connect/ConnectContent.java
pkriens: It is only used to build, the API is not ending up in any binary so split package is not needed. I made it the second entry so this is perfect! Much better than embedded these sources.

Issue #3810, opened 2020-02-28
[Feature Request] Introduction of launcher flag to use symbolic name as location


amitjoy:
**_Current State_**: Launcher uses the bundle path as its location which makes sense to support multiple versiond of bundles having same bundle symbolic name.

**_Requirement_**: We might need a flag for launcher to allow using bsn instead of bundle path.

**_Scenario_**: Let's say, we have the following bundle in the launcher exported uber JAR.
`com.example.bundle.a Version 1.0`
Now, once we use the uber JAR, the OSGi storage will have all the bundles expanded in their proper locations - in this scenario, the bundle location of Bundle A will be - `jar/com.example.bundle.a-1.0.0.jar`
(considering `com.example.bundle.a`  will be consumed from remote maven repo, such as, maven central). Now, if we update the bundle in uber JAR to use 2.0.0, the storage will now contain Bundle A in two different locations - `jar/com.example.bundle.a-1.0.0.jar` and `jar/com.example.bundle.a-2.0.0.jar`.

Hence, we might need a flag for launcher that enables us to use `bsn` instead of `path`.

PR #3818, opened 2020-03-03
[launcher] Location format

pkriens:
The current launcher calculates a unique name for
each embedded JAR when creating an executable JAR.
This unique name is based on the name of the file
in the repo. Some of these files are just the bsn
but sometimes the full version is in there. 

Unfortunately, the launcher uses this path in the
executable JAR to create the location of the 
corresponding bundle. 

If the version is in that name, then this creates
a problem if the storage area is kept. If the version
of a bundle gets changed, it will create a new location
creating unwanted duplicates.

To generate a proper location, you need a strategy.
You could use a location that is unique for a
bsn, implying that all bundles are singletons. A sensible
strategy seems to include the major version in the 
location, allowing unique locations for incompatible
bundles.

Anyway, this patch allows the exporter to set a
strategy by specifying a format that uses macros. Available
are ${@version} and ${@bsn}.

### DIFF
--- a/biz.aQute.bndall.tests/test/biz/aQute/launcher/AlsoLauncherTest.java
+++ b/biz.aQute.bndall.tests/test/biz/aQute/launcher/AlsoLauncherTest.java
@@ -226,6 +226,63 @@ public class AlsoLauncherTest {
 	}
 
 	@Test
+	public void testLocationFormat() throws Exception {
+		project.setProperty(Constants.RUNPROPERTIES, "test.cmd=exit");
+		project.setProperty(Constants.RUNTRACE, "true");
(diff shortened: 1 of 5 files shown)


rotty3000:
@pkriens just so I understand, is the issue that updating a bundle (i.e. to a new version) should result in an overwrite in the storage?

pkriens:
Yes. If you keep the storage, the location will change when the version changes, resulting in duplicates

pkriens:
and if you're unlucky the launcher can hang

rotty3000:
got it!

bjhargrave:
Fixes #3810

PR #3823, opened 2020-03-04
LGTM fix for (allegedly) possible NPE

kriegfrj:
This was not really a possible NPE as there was a catch() clause that would have handled any NPE appropriately. However, it's nice not having any LGTM alerts.

### DIFF
--- a/biz.aQute.remote/src/aQute/remote/agent/SocketRedirector.java
+++ b/biz.aQute.remote/src/aQute/remote/agent/SocketRedirector.java
@@ -62,6 +62,9 @@ public class SocketRedirector implements Redirector {
 						Thread.sleep(1000);
 					}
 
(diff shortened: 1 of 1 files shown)


kriegfrj:
Looks like @pkriens beat me to it. His solution was neater than mine anyway. :smile:

pkriens:
I am not going to fix the other one because it is not an error. The path is concatenated with other characters for output, so if there is a .. in there you **want** to see it. 
Not sure how to ignore that one?

kriegfrj:
Perhaps we can suppress it: https://lgtm.com/help/lgtm/alert-suppression

pkriens:
I've reported a false positive

https://github.com/Semmle/ql/issues/2996

PR #3825, opened 2020-03-06
[repository] Error reporting on Repositories in RepositoriesView

pkriens:
Currently, the error handling of repositories is quite bad. Some repositories throw an exception in the setProperties, others throw one at the first usage. These errors go into the Workspace but there are no mechanisms to see those errors. 

This patch provides a mechanism to allow a repository to provide its status. Additionally, the RepositoriesView was adapted to show an icon (defined by the repository itself). This icon is decorated with the error status if it's a read-only repo and a remote repo. 

The extra information we need from the repository is obtained via the RepositoryPlugin interface. The new methods all have defaults.

The MavenBndRepository was adjusted to take advantage of these new capabilities. It was changed to not throw exceptions on errors. It keeps a status field and reports this status. If it is still called, it will silently return a valid result like not found or empty collection. 

We need a discussion to define this strategy. Likely, before a build, we should verify that all repositories -workspace have a null status. This should be cheap and aborts builds that will be erroneous anyway. After we have this discussion, I'll adapt all the other repositories.


pkriens:
image

aQute.libg/src/aQute/lib/exceptions/Exceptions.java
```diff
@@ -63,4 +65,24 @@ public static String causes(Throwable t) {
 		return sj.toString();
 	}
 
+	static Pattern DISPLAY_P = Pattern.compile("(.*)(Exception|Error)");
```
bjhargrave: This should be `final`, no?

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -881,4 +882,24 @@ private boolean isUpdateInfo(int code, URLConnection con) {
...
 
 	}
+
+	/**
+	 * Validate a URI to see if it is supported by this client
+	 *
+	 * @param u the uri
+	 * @return null if ok, otherwise a reason why it is invalid
+	 */
+	public String validateURI(URI u) {
```
bjhargrave: Could this be a `static` method? It does not seem to use any instance members.

pkriens: No, that would make it impossible to add schemes via settings if necessary

PR #3832, opened 2020-03-13
Use Maven Toolchain JAVA_HOME to launch test runtime

njbartlett:
Some Maven projects need to support different versions of Java across
different modules, e.g. a shared library that builds with Java 1.8 for
maximum compatibility plus an application module that takes advantage of
new language features in Java 11. The canonical way to handle this is
with Maven's Toolchain feature, which is supported by core plugins such
as the maven-compiler-plugin and maven-surefire-plugin.

bnd-testing-maven-plugin does not currently support Toolchains.
Instead, it launches java using whatever JDK was used to run mvn itself.
This make it impossible to support different Java versions in different
modules.

This change very simply adds the JAVA_HOME from the Maven Toolchain
manager if and only if one is defined by the POM. Projects that do not
specify a Toolchain are unaffected and will run in exactly the same way
as they always have.

### DIFF
--- a/.github/scripts/build.sh
+++ b/.github/scripts/build.sh
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 ./gradlew --no-daemon --version
 ./mvnw --version
-./gradlew --no-daemon -Dmaven.repo.local=maven/target/m2 --continue :build :maven:deploy
+./gradlew --no-daemon -Dmaven.repo.local=maven/target/m2 --continue :build :maven:deploy "$@"
--- a/.github/scripts/docs.sh
+++ b/.github/scripts/docs.sh
@@ -4,4 +4,4 @@ gem --version
 gem install bundler -v '~> 2.0'
 bundle --version
 cd docs
-./build.sh
+./build.sh "$@"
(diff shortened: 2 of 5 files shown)


njbartlett:
Anybody know how to fix these CI Build checks? All failed on git checkout...

bjhargrave:
> Anybody know how to fix these CI Build checks? All failed on git checkout...

I restarted them and they all failed again. It may be a problem at GitHub. So lets wait a few hours and I will restart the jobs again.

njbartlett:
Thanks BJ.

bjhargrave:
Perhaps there is something off with your repo or this PR. It still fails on the git checkout. Peter just submitted a PR which is git checkouts just fine. Perhaps you can rebase you branch on the latest origin/master and force push to try again.

njbartlett:
@bjhargrave Thanks for your advice. A rebase and force push indeed fixed the problem with git checkouts.

Now there is a test failure that seems to be completely unrelated to the change in my commit. Do you know if this test is known to be broken?

bjhargrave:
> Do you know if this test is known to be broken?

It is not a broken test. Its is a broken sonatype snaphot repo. The request to get the org.osgi.framework snapshot jar failed. So the dependencies on that jar then failed. Too many communications failure modes in the build...

bjhargrave:
Something is borked with github actions. Another PR build is hung for 3+ hrs  I think we try to rebuild again tomorrow.

bjhargrave:
I tried to push an updated commit to the PR, but since you used your master branch for this PR, something weird happened. It allowed me to update your master branch to match the tip of bndtools/bnd master and then closed this PR! So #3834 is now a replacement for this PR.

Issue #3808, opened 2020-02-27
[bndtools] Add Content Assist for imports


kriegfrj:
bndtools/bndtools#1850 added quick-fix suggestions for missing bundles for imports.

Another related feature that Eclipse has is suggestions as you type. Eg, you start typing `import org.osgi...` and it gives you a drop-down list of known packages to import.

The built-in Eclipse suggester will only suggest packages that are already on the build path. Bndtools could potentially add suggestions of exported packages from all bundles known in the repositories -even those that are not yet in the build path. It could then (like the Quick Fix processor) automatically add the relevant bundle to `-buildpath`.

pkriens:
This would be heaven ... when do we get a PR? :-)

pkriens:
I just looked at the code. I saw here https://github.com/bndtools/bndtools/pull/1850/files#diff-d864d16e149e1242dd5554a5620a8229R354-R366 that you get the RepositoryPlugins. Why not use `Workspace.getRepositories()` directly? This will give you all the Repository objects that end uses for resolving. You current might miss repositories because there is an assumption that RepositoryPlugin implements Repository which is false. Or am I missing something?

You can also search for all packages you're looking for in one go using the multi requirement facility on the current OSGi. Or do the search in parallel since it might involve downloading.

If you only have the class name it, of course, becomes harder. We could first ask JDT then. If it is anywhere on the classpath then it should be able to have reference to the JAR. Otherwise, we need to index all repositories ...

Since we're already indexing the classes for each repository we could extend the ResourceBuilder to include a Bloom filter or so for class names. We could even add this to the XML as a proprietary namespace so that remote indexes also contain this info. The Bloom filter would allow us to detect the presence so we can then do a proper check.

I would strongly prefer that any logic goes into bndlib. I.e. the package/class search function should go to Workspace and not be done in bndtools. The push for Intellij is increasing and any function inside Eclipse cannot be reused. It would even be an interesting function in bed to search repositories.

kriegfrj:
> I just looked at the code. I saw here https://github.com/bndtools/bndtools/pull/1850/files#diff-d864d16e149e1242dd5554a5620a8229R354-R366 that you get the RepositoryPlugins. Why not use `Workspace.getRepositories()` directly? This will give you all the Repository objects that end uses for resolving. You current might miss repositories because there is an assumption that RepositoryPlugin implements Repository which is false. Or am I missing something?
>
[...]

PR #3833, opened 2020-03-13
Pkriens/importclassname

pkriens:
Supports importing by class name


biz.aQute.bndlib/src/aQute/bnd/osgi/Descriptors.java
```diff
@@ -654,4 +675,112 @@ public TypeRef getTypeRefFromPath(String path) {
...
+	/**
+	 * Return a 2 element array based on the fqn. The first element is the
+	 * package name, the second is the class name. Each can be absent, but not
+	 * both. The class name can be a nested class (will contain a '.' then)
+	 *
+	 * @param fqn a Java identifier name, either a simple class name, a
+	 *            qualified class name, or a package name
+	 * @return a Result with 2 element array with [package, class]
+	 */
+	public static Result<String[], String> determine(String fqn) {
```
bjhargrave: Would the return type be better as `Result<Map.Entry<String,String>,String>` or `Result<Pair<String,String>,String>`? Using `String[]` is pretty open and provides no clear evidence that the array contains exactly 2 strings.

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java
```diff
@@ -638,15 +646,15 @@ public void addRequirements(List<Requirement> requires) {
 		return capabilities;
 	}
 
-	public Map<Capability, Capability> from(Resource bundle) {
-		Map<Capability, Capability> mapping = new HashMap<>();
+	public Map<Capability, CapReqBuilder> from(Resource bundle) {
```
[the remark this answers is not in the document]
bjhargrave: Actually, upon further thinking, this change is, from a practical point of view, a binary incompatible change. Any call site getting a value from a map entry will have a compiler generated cast to `Capability`. So existing binaries doing something like:

```java
Capability cap = from(bundle).get(someCap);
```

will result in a ClassCastException at runtime since the compiler expects the value to be of type `Capability` and generates a cast to verify this.

So this sort of change is, in effect, binary incompatible.

biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java
```diff
@@ -134,13 +138,17 @@ public Resource build() {
...
+			.map(cr -> cr.buildCapability())
+			.collect(Collectors.toSet()));
 		resource.setRequirements(requirements.values());
 		return resource;
 	}
 
+	@Deprecated
 	public List<Capability> getCapabilities() {
-		return new ArrayList<>(capabilities.values());
+		throw new UnsupportedOperationException("Must first be build");
```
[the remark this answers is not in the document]
pkriens: There is only one place where we use it. I'll take a look

pkriens:
Could you wait with reviewing until I ask you? This is annoying

bjhargrave:
> Could you wait with reviewing until I ask you? This is annoying

Perhaps you should put `WIP` in the PR title to communicate to others the PR is not ready for review? Otherwise, I assume the PR is reviewable. 

Also, instead of opening a PR, you could only push to your fork and let your fork's github actions build and test for you. I do that all the time before making a PR.

pkriens:
> On 16 Mar 2020, at 17:43, BJ Hargrave <notifications@github.com> wrote:
> Could you wait with reviewing until I ask you? This is annoying
> 
> Perhaps you should put WIP in the PR title to communicate to others the PR is not ready for review? Otherwise, I assume the PR is reviewable.
> 

Well, then don't make such assumptions. I call you when I need you. There is button for asking for a review.

> Also, instead of opening a PR, you could only push to your fork and let your fork's github actions build and test for you. I do that all the time before making a PR.
> 
Yes people differ.

aQute.libg/src/aQute/lib/zip/ZipUtil.java
```diff
@@ -54,6 +54,7 @@ public static String cleanPath(String path) {
 				int lastSlash = clean.lastIndexOf("/");
 				if (lastSlash == -1) {
 					if (clean.length() == 0) {
+						// bad design, this is a common outcome
```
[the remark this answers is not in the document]
pkriens: We should not throw exceptions for input validation. Wrong input is part of the contract of this method. Exceptions, at least in bnd, are a failure of the code because the inputs are outside the contract.

pkriens:
Fr. Krieg, could you take a look at the Eclipse code? And maybe try it out? I want to make sure I do not miss functionality.

bndtools.core/src/org/bndtools/core/editors/quickfix/BuildpathQuickfixProcessor.java
```diff
@@ -0,0 +1,165 @@
...
+							if (test && !testpath.contains(id) && !buildpath.contains(id))
+								proposals.add(propose(e.getKey(), id, context, location, project, "-testpath"));
+
+							if (!buildpath.contains(id))
+								proposals.add(propose(e.getKey(), id, context, location, project, "-buildpath"));
+
+						}
+					});
+			}
+			return proposals.toArray(new IJavaCompletionProposal[0]);
```
kriegfrj: For the common case where we haven't found any proposals, perhaps we should return `null` rather than an empty array. Saves the overhead of the extra object allocations. Relatively small optimisation though.

bndtools.core/src/org/bndtools/core/editors/quickfix/BuildpathQuickfixProcessor.java
```diff
@@ -0,0 +1,165 @@
...
+		} catch (Exception e) {
+			throw Exceptions.duck(e);
+		}
```
kriegfrj: Because this is the entry point from Eclipse into the quick fix processor, maybe it would be better to wrap this in a `CoreException` rather than allowing it to propagate up? I'm not 100% sure on Eclipse best practice here, but that seems to be what we've done elsewhere in Bndtools.

bndtools.core/test/org/bndtools/core/editors/BuildpathQuickFixProcessorTest.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.junit.Test;
+
+public class BuildpathQuickFixProcessorTest {
+
+	@Test
+	public void testBasic() {
+		// not sure how to test this
+		// the majority of the code is tested in bnd
+	}
```
kriegfrj: I don't think it's accurate to say that "the majority of the code is tested in bnd". I'm sure that all of the *searching* code is tested in Bnd - but searching is only half of the quick fix processor's responsibility. The other thing that the quick fix processor has to do (equally important) is understand Eclipse's error message formats so that it knows what to search for and can translate that into a form that Bnd understands, and then conversely take Bnd's response and translate that into a form that Eclipse can use. The first part in particular is a non-trivial operation and accounted for a significant chunk of the original quick-fix code.

The original `ImportPackageQuickFixProcessorTest` was close to a full end-to-end test - it used actual source code with errors which could be quick-fixed, then it used Eclipse's compiler to compile these into an actual AST, and finally it created actual `IProblemLocation` objects (which were built based on observed values determined by experimentation on a live Eclipse instance). It passed these in to `getCompletions()` to ensure that the correct set of `ICompletionProposal`s was created.

In order to ensure that we haven't had any regressions, we should do something similar. The old test should form the basis, as I know it had pretty good coverage of lots of weird corner cases. Your code is admittedly a lot simpler, but I'm not sure if this simplification has come at the expense of us not handling some of these weird corner cases.

The more I think about this as I type, the more I am of the opinion that the best way forward for me to review your changes will be to port `ImportPackageQuickFixProcessor` to `BuildpathQuickFixProcessorTest`. For starters, I can't remember all of the weird corner cases, and even if I could testing them all manually would be onerous. If I instead go to the effort of porting the test, then we have the added benefit of a solid regression test moving forward so that others can make changes without fear of breaking existing functionality (that is, after all, why I went to the effort of writing the original regression test in the first place!)

kriegfrj:
Also wanted to address this in your commit comment:

> In the bug, it is indicated that there was also a completion list. This did not work for me, and I could also not really see why I never got triggered.

Sorry, the original issue (#3808) was not meant to indicate that there was a completion list - in fact, it was meant to indicate the opposite. I originally implemented a quick fix processor only, and not an auto-suggestion tool. The former only triggers when the builder reports an error, whereas the latter triggers as you are typing and possibly when there is no build error. Auto-suggestion is a different interface in Eclipse and has a different plugin entry point to register, and I never got around to implementing this. Issue #3808 was basically to report this fact, and to suggest that we add auto-suggestions alongside the quick fixes.

bndtools.core/test/org/bndtools/core/editors/BuildpathQuickFixProcessorTest.java
pkriens: That was what I was secretly hoping for :-) 

It is always very nice if you have a different person doing the test.

bndtools.core/src/org/bndtools/core/editors/quickfix/BuildpathQuickfixProcessor.java
pkriens: less is more ...

bndtools.core/src/org/bndtools/core/editors/quickfix/BuildpathQuickfixProcessor.java
pkriens: If you look at the code you see that CoreException is thrown when we have something useful to tell, see line 85. All other exceptions are basically _out of contract_. No way to do anything useful there. If there are common exceptions that we want to handle we should change the API to use Result. 

In bnd, exceptions are treated as a boolean, they should just signal an out of contract failure. The new Result is really helpful here. Since not everybody follows that model, sometimes you need to catch an exception that contains information you cannot get in any other way. That is then necessary but we should try to treat exceptions as booleans, indicating you're in lala land, no way to known where you are anymore.

bndtools.core/src/org/bndtools/core/editors/quickfix/AddBundleCompletionProposal.java
```diff
@@ -0,0 +1,142 @@
...
+	public int getRelevance() {
+		return relevance;
+	}
+
+	@Override
+	public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
+		try {
+			IStatus status = Central.bndCall(() -> {
+				BndEditModel model = new BndEditModel(project);
+				model.load();
```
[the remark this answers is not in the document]
pkriens: Yes, it is been on my mind for some time to bring the monitor as concept in bnd. I'll add something, but not in this branch.

bndtools.core/src/org/bndtools/core/editors/quickfix/AddBundleCompletionProposal.java
```diff
@@ -0,0 +1,142 @@
...
+				context.getCompilationUnit()
+					.createImport(fqn + ".*", null, monitor);
+			} else {
+				context.getCompilationUnit()
+					.createImport(fqn, null, monitor);
+			}
+			return status;
+		} catch (Exception e) {
+			throw new CoreException(
+				new Status(IStatus.ERROR, Plugin.PLUGIN_ID, "Failed to add bundle to -buildpath", e));
```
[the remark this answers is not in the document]
pkriens: fixed

bndtools.core/src/org/bndtools/core/editors/quickfix/AddBundleCompletionProposal.java
```diff
@@ -0,0 +1,142 @@
...
+			assert determine[0] != null : "We must have found a package";
+
+			if (determine[1] == null) {
+				context.getCompilationUnit()
+					.createImport(fqn + ".*", null, monitor);
+			} else {
+				context.getCompilationUnit()
+					.createImport(fqn, null, monitor);
+			}
+			return status;
```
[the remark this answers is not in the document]
pkriens: - I think duplicates are automatically skipped, at least I read something in the comments
- I just cannot find a definition of the error codes. 

I tested this in the UI and it seems to work fine so far. Shall we wait until we have actually test cases. I could not find a good definition of the error codes, what arguments they supplied, and what cases they were reported. I basically could only find import errors.

Having test code would allow us to see each case we want to support. I reasoned this was the safe way and the Eclipse quick fixes support cleaning up if necessary.

I am a big fan of less is more. I'd like to start very simple that covers 80%, and then add the missing cases explicitly.

bndtools.core/src/org/bndtools/core/editors/quickfix/BuildpathQuickfixProcessor.java
```diff
@@ -0,0 +1,165 @@
...
+
+	@Override
+	public boolean hasCorrections(ICompilationUnit unit, int problemId) {
+		switch (problemId) {
+			case IProblem.IsClassPathCorrect :
+				return true;
+			case IProblem.ImportNotFound :
+				return true;
+			case IProblem.UndefinedType :
+				return true;
```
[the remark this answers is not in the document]
pkriens: I had a really hard time generating the other error ids. Just could not figure it out.

I kept them separate to visually indicate that we should try to figure out if they all need to be supported and how. I am assuming they have some semantics.

bndtools.core/test/org/bndtools/core/editors/ImportPackageQuickFixProcessorAddBundleCompletionProposalTest.java
```diff
@@ -1,57 +0,0 @@
-package org.bndtools.core.editors;
```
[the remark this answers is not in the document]
pkriens: I like tests but I agree I've got an aversion against Eclipse GUI testing. I find I spend way more time testing my own mocks & scaffolding than actual error cases. 

However, if you think, as we discuss later, that you can set up error cases for all the use cases (error codes/source code) then that would be fantastic.

bndtools.core/src/org/bndtools/core/editors/quickfix/BuildpathQuickfixProcessor.java
```diff
@@ -0,0 +1,165 @@
...
+			if (project == null)
+				return null;
+
+			boolean test = isInDir(project.getTestSrc(), compUnit.getResource());
+			Workspace workspace = project.getWorkspace();
+
+			for (IProblemLocation location : locations) {
+
+				if (!hasCorrections(context.getCompilationUnit(), location.getProblemId()))
+					return null;
```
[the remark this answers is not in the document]
pkriens: I actually got called ... I'll make it a continue. good catch.

bndtools.core/test/org/bndtools/core/editors/BuildpathQuickFixProcessorTest.java
pkriens: If you just could have a standalone test where I can give a Java source with a problem and get the quick fixes, that would be great. We could use this in other cases as well since this stuff is highly undertested.

pkriens:
I understand that the Corona virus is going to keep you away from this. 

I am going to try to get this in the main branch asap so we can work on improving it there, and other people might chip in.

bndtools.core/test/org/bndtools/core/editors/ImportPackageQuickFixProcessorAddBundleCompletionProposalTest.java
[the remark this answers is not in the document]
kriegfrj: I agree - Eclipse testing is tricky. I think writing the tests for the original quick fix processor was a bit of a turning point for me when I started to realise the limitations of mocks - yes, you spend as much time debugging the mocks as you do the actual code. Since then I've had a strong preference for testing against actual code rather than against mocks (where it makes sense to do so), effectively skipping unit testing and going straight for integration tests.

I think the good news is that I've learned a lot more about OSGi and Bnd since those days, and I think I can come up with an OSGi-type test that uses more of actual Eclipse and less mocking. This could become a model for other GUI tests.

bndtools.core/src/org/bndtools/core/editors/quickfix/BuildpathQuickfixProcessor.java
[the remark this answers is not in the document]
kriegfrj: Yes, they all have some semantics and (from memory) they are a little different.

I remember that one of the more esoteric error types I tried to handle was when there is a missing transitive dependency. Error message something like "Cannot find <xxx> - it is indirectly referenced by classes on the classpath."

bndtools.core/test/org/bndtools/core/editors/BuildpathQuickFixProcessorTest.java
kriegfrj: I'll have a look when I get a chance.

kriegfrj:
> I understand that the Corona virus is going to keep you away from this.
>
> I am going to try to get this in the main branch asap so we can work on improving it there, and other people might chip in.

Thanks for your understanding. Getting it into the main branch will also make it easier for me to submit PRs for the test if/when I get around to it.

One minor thing I forgot in the review: I would rename `BuildpathQuickfixProcessor` to `BuildpathQuickFixProcessor` for consistency.

bndtools.core/src/org/bndtools/core/editors/quickfix/BuildpathQuickfixProcessor.java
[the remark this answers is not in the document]
kriegfrj: Happens to the best of us :wink:

bndtools.core/src/org/bndtools/core/editors/quickfix/BuildpathQuickfixProcessor.java
kriegfrj: Hmm, I'm not sure I follow here. If this was a Bnd internal function, changing the API to use Result would be a sensible option. But this is part of the Eclipse API - we don't have control over it, we have to try and comply with it.

The documentation for `IQuickFixProcessor.getCorrections()` states that we may throw `CoreException` if something goes wrong. It doesn't say that we're allowed to throw any other kind of exception, and if we do Eclipse's behaviour will be undefined (especially a checked exception masked as an unchecked exception). It might prevent other quick fix processors from running. Less likely, but more seriously it might crash our editor, or our entire Eclipse instance. Surely it is better to be a good citizen and do our best to make sure that our own internal errors (even unanticipated errors caused by bugs) don't do damage to any of our "neighbours"?

pkriens: The issue is that in 'normal' code the exception would always be a runtime exception. I decided from the beginning that an Checked Exceptions are evil and should be treated as runtime exceptions. I.e. if this exception is thrown, the code is wrong. The Exceptions class was added to treat the checked exceptions as RuntimeExceptions.

You're right about the consequences but an exception there means the code is fundamentally broken. I.e. like a NPE. It just should not happen and therefore should not have a decent  handling. It is Game Over.

kriegfrj: I get that an exception there means the code is fundamentally broken, that it should not happen. But if it *is* fundamentally broken/it does happen? What if we do cause an NPE, for example? Wouldn't it be better if we *didn't* bring down Eclipse or stop other quick fix processors from running?

If your answer to this is simply "no", I guess I will shrug my shoulders and note my disagreement for the record. I would sandbox it to prevent our contagion from spreading.

bndtools.core/src/org/bndtools/core/editors/quickfix/BuildpathQuickfixProcessor.java
kriegfrj: Having re-read the spec for `getCorrections()` (per the javadoc), just to let you know that this is technically non-compliant - the javadoc calls for `null` to be returned if you don't have any proposals. It will probably work, but we might one day run into an issue if Eclipse ever relies on the assumption that it should be `null`.

bndtools.core/src/org/bndtools/core/editors/quickfix/BuildpathQuickfixProcessor.java
pkriens: But Eclipse is sandboxing it already ... Any runtime exception must be caught or Eclipse would not survive very long. It should also be logged in the error log, which Core Exceptions aren't I think. It happens to be visible here because I decided early on in bnd (1999 probably) that checked exceptions are evil and all methods are throwing Exception, basically treating all exceptions as runtime exceptions. The primary reason is to pass the original cause upwards instead of the long chains that too often miss the root cause. This strategy makes runtime exceptions visible in cases like this, cases that normal code can ignore. If this exception happens, it is like a core dump in C code or panic() in Rust. I want this to be really bad because it is a code error and must be fixed asap, not hidden in any way. The goal is to show the place where the thing failed and we should therefore not wrap it in any way. And I want it in the Eclipse log.

So far it worked fine for me. However, lambdas threw a bit of a wrench in this strategy and that makes us struggle a bit with it. I wish Java had an option to ignore checked exceptions, like every other language on earth. It is probably the worst idea since language design began. So due to lambdas, I'd probably not throw Exception on every method again. But that would increase the use of duck() because I refuse to play the game of checked exceptions unless forced by bad API design.

bndtools.core/src/org/bndtools/core/editors/quickfix/BuildpathQuickfixProcessor.java
pkriens: I disagree but it is not worth to fight about this. Fixed.

bndtools.core/src/org/bndtools/core/editors/quickfix/BuildpathQuickfixProcessor.java
kriegfrj: If Eclipse is properly sandboxing it already, then scratch what I said - I agree that yours is the best implementation in that circumstance. Unnecessary wrapping of the exception makes it harder to find the root cause in the logs.

PR #3834, opened 2020-03-14
Use Maven Toolchain JAVA_HOME to launch test runtime

bjhargrave:
Some Maven projects need to support different versions of Java across
different modules, e.g. a shared library that builds with Java 1.8 for
maximum compatibility plus an application module that takes advantage of
new language features in Java 11. The canonical way to handle this is
with Maven's Toolchain feature, which is supported by core plugins such
as the maven-compiler-plugin and maven-surefire-plugin.

bnd-testing-maven-plugin does not currently support Toolchains.
Instead, it launches java using whatever JDK was used to run mvn itself.
This make it impossible to support different Java versions in different
modules.

This change very simply adds the JAVA_HOME from the Maven Toolchain
manager if and only if one is defined by the POM. Projects that do not
specify a Toolchain are unaffected and will run in exactly the same way
as they always have.


bjhargrave:
@njbartlett If you are happy with this, go ahead and merge.

njbartlett:
Thanks BJ!

PR #3838, opened 2020-03-18
[launcher] Support -executable location format

pkriens:
This fix was added to the 5.1 master but had
API changes. This is similar without the API
changes.

basically allows a -executable: location='${@bsn}'
instruction to set the format of the path used
in the launcher executable jar. The default 
path can contain versions and since the path
is used for the OSGi location, new bundles
are created then the path changes. So you get
the same bsn installed several times. Not good

### DIFF
--- a/biz.aQute.bndall.tests/test/biz/aQute/launcher/AlsoLauncherTest.java
+++ b/biz.aQute.bndall.tests/test/biz/aQute/launcher/AlsoLauncherTest.java
@@ -96,6 +96,64 @@ public class AlsoLauncherTest {
 		System.setProperties(prior);
 	}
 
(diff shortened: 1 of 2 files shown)


bjhargrave:
We will also need to add a bullet to https://github.com/bndtools/bnd/wiki/Changes-in-5.0.1#fixes-since-500 describing this fix.

pkriens:
done

pkriens:
will you merge or do I

bjhargrave:
> will you merge or do I

I can do it since I also need to tag it RC2.

bjhargrave:
Fixes #3810 for 5.0.1 release.

PR #3852, opened 2020-03-24
unwraps the InvocationTargetException to report the real error

### LINKED ISSUE #3851: BndPomRepository only shows InvocationTargetException if an Artifact can't be downloaded for some reason
In Repository Explorer only a InvocationTargetException is shown if an Artifact can't be downloaded. We need to unwrap it to show the real one.

juergen-albert:
fixes #3851

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/Traverser.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/Traverser.java
@@ -5,6 +5,7 @@ import static org.osgi.framework.namespace.IdentityNamespace.CAPABILITY_TYPE_ATT
 import static org.osgi.framework.namespace.IdentityNamespace.IDENTITY_NAMESPACE;
 
 import java.io.File;
+import java.lang.reflect.InvocationTargetException;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -28,6 +29,7 @@ import org.slf4j.LoggerFactory;
 import aQute.bnd.http.HttpClient;
 import aQute.bnd.osgi.resource.CapabilityBuilder;
 import aQute.bnd.osgi.resource.ResourceBuilder;
+import aQute.lib.exceptions.Exceptions;
 import aQute.maven.api.Archive;
 import aQute.maven.api.IPom.Dependency;
 import aQute.maven.api.MavenScope;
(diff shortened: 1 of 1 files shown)


biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/Traverser.java
```diff
@@ -206,8 +207,14 @@ private void parseResource(Archive archive, String parent) throws Exception {
 			}
 			addInformationCapability(rb, archive.toString(), parent);
 		} catch (Exception e) {
+			Throwable theException = e;
```
pkriens: See the methods for this in Exceptions, they provide solid support to get the original cause

juergen-albert: Makes sense. I will use the ```Exceptions.unrollCause``` method.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/Traverser.java
```diff
@@ -206,8 +207,9 @@ private void parseResource(Archive archive, String parent) throws Exception {
 			}
 			addInformationCapability(rb, archive.toString(), parent);
 		} catch (Exception e) {
+			Throwable theException = Exceptions.unrollCause(e);
```
bjhargrave: This will unroll all the way back to the root cause. If you only want to unroll off any InvocactionTargetExceptions, then you want to do:

```java
Throwable theException = Exceptions.unrollCause(e, InvocationTargetException.class);
```

See for example: 

https://github.com/bndtools/bnd/blob/a6d93a4af6b9c388044f632887f682b30d5e8fe6/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java#L419-L420

juergen-albert: Maybe one day I will manage to create PR that does not need at least 3 iterations :-)

bjhargrave:
Thanks!

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/Traverser.java
pkriens: I think there is a rumour that Chuck Norris could once bypass BJ with only 2 iterations ...

PR #3861, opened 2020-03-31
[bndtools] Defensively quote vmargs

gamerson:


bndtools.core/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
```diff
@@ -301,7 +304,14 @@ public String getVMArguments(ILaunchConfiguration configuration) throws CoreExce
 		StringBuilder builder = new StringBuilder();
 		Collection<String> runVM = getProjectLauncher().getRunVM();
 		for (Iterator<String> iter = runVM.iterator(); iter.hasNext();) {
-			builder.append(iter.next());
+			String vmArg = iter.next();
+			Matcher matcher = vmArgPattern.matcher(vmArg);
+
+			if (matcher.matches())
+				vmArg = MessageFormat.format("-D{0}=\"{1}\"", matcher.group(1), matcher.group(2));
```
kriegfrj, line 311: Would it be simpler to quote the entire string? And wouldn't it be safe to do this for all args? And wouldn't it be better to use single quotes rather than double (in case there are any cases in the quoted string that will start getting interpolated)? And if we quote with a single quote, we should also escape any single quotes that appear in the value itself.

I'm not asking these questions because I know the answers of the top of my head...

pkriens:
There already is a careful quote function in Header.quote(Appendable,String) we might want to reuse. But I am waiting for @bjhargrave's comments

gamerson:
@bjhargrave  has an alternate solution in hand... closing this one.

PR #3863, opened 2020-04-02
[P2Repository] Adds support for format="packed" (pack200)

psoreide:
As the Pack200 class is deprecated in JavaSE-11 we use the unpack200 command.

### DIFF
--- a/biz.aQute.repository/bnd.bnd
+++ b/biz.aQute.repository/bnd.bnd
@@ -40,7 +40,8 @@ Export-Package: \
 	aQute.maven.*, \
 
 -includepackage: \
-	aQute.p2.provider
+	aQute.p2.provider,\
+	aQute.p2.packed
 
 -conditionalpackage:\
 	aQute.lib.*,\
--- a/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
@@ -49,6 +49,7 @@ import aQute.lib.io.IO;
 import aQute.libg.cryptography.MD5;
 import aQute.p2.api.Artifact;
 import aQute.p2.api.ArtifactProvider;
+import aQute.p2.packed.Unpack200;
 import aQute.p2.provider.P2Impl;
 import aQute.p2.provider.TargetImpl;
 import aQute.service.reporter.Reporter;
@@ -61,6 +62,7 @@ class P2Indexer implements Closeable {
 	private final static Logger			logger					= LoggerFactory.getLogger(P2Indexer.class);
 	private static final long			MAX_STALE				= TimeUnit.DAYS.toMillis(100);
 	private final Reporter				reporter;
+	private final Unpack200				processor;
 	final File							location;
 	private final HttpClient			client;
 	private final PromiseFactory		promiseFactory;
@@ -76,7 +78,10 @@ class P2Indexer implements Closeable {
 		.filter("(" + MD5_ATTRIBUTE + "=*)")
 		.buildSyntheticRequirement();
 
-	P2Indexer(Reporter reporter, File location, HttpClient client, URI url, String name) throws Exception {
+	P2Indexer(Unpack200 processor, Reporter reporter, File location, HttpClient client, URI url,
+		String name)
+		throws Exception {
+		this.processor = processor;
 		this.reporter = reporter;
 		this.location = location;
 		this.indexFile = new File(location, "index.xml.gz");
@@ -117,8 +122,9 @@ class P2Indexer implements Closeable {
 
 		Promise<File> go = client.build()
 			.useCache(MAX_STALE)
+			.asTag()
 			.async(url.toURL())
-			.map(file -> link);
+			.map(tag -> processor.unpackAndLinkIfNeeded(tag, link));
 
 		if (listeners.length == 0)
 			return go.getValue();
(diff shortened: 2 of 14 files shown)


biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -109,21 +110,38 @@ File get(String bsn, Version version, Map<String, String> properties, DownloadLi
...
 
-		if (listeners.length == 0)
-			return go.getValue();
+		if (listeners.length == 0) {
+			return handleDownloadResult(packed, link, go);
+		}
 
 		new DownloadListenerPromise(reporter, name + ": get " + bsn + ";" + version + " " + url, go, listeners);
+
+		return handleDownloadResult(packed, link, go);
```
bjhargrave, line 128: You can't do it this way. You need to do any unpacking as a map operation on the promise. Otherwise you block here in this method and the point of using listeners is to NOT block here.

biz.aQute.repository/src/aQute/p2/provider/ArtifactRepository.java
```diff
@@ -112,8 +113,8 @@ void parse() throws Exception {
 
 				Map<String, String> artifactProperties = getProperties(artifactNode, "properties/property");
 				xmlArtifact.format = artifactProperties.get("format");
-				if (xmlArtifact.format != null) {
-					continue; // we do not currently support packed format
+				if ("packed".equals(xmlArtifact.format) && !Unpack200.canUnpack()) {
```
bjhargrave, line 116: Shouldn't there be a constant for `packed`?

aQute.libg/src/aQute/libg/pack200/Unpack200.java
```diff
@@ -0,0 +1,147 @@
+package aQute.libg.pack200;
```
bjhargrave, line 1: I am not sure why this live in `aQute.libg`. We should probably just put it in `biz.aQute.repository` as a `aQute.p2.packed` package.

aQute.libg/src/aQute/libg/pack200/Unpack200.java
```diff
@@ -0,0 +1,147 @@
...
+			throw new RuntimeException(e);
+		}
+	}
+
+	/**
+	 * Get unpack200 command
+	 *
+	 * @return String [] or null
+	 */
+	public static String[] getUnpack200Commands() {
```
bjhargrave, line 116: Once this class is in biz.aQute.repository, we can use Processor.getJavaExecutable to find the unpack command. See

https://github.com/bndtools/bnd/blob/98e97befcd426ecd99bc85ed32f421bf95c98bdf/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/Tool.java#L160-L161

where we use this method to find the javadoc command.

This method lets the user set the `unpack200` property in their workspace or project to override the default command. Then we don't need `@jre`, `@path`, `@none`. We just use the value of the `unpack200` property (like we do for the `javadoc`, `java`, and `javac` properties) to find the command.

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -183,7 +201,11 @@ private Repository readRepository() throws Exception {
 				}
 				return fetch(a, 2, 1000L).map(tag -> {
 					ResourceBuilder rb = new ResourceBuilder();
-					rb.addFile(tag.getFile(), a.uri);
+					File file = tag.getFile();
```
bjhargrave, line 204: I would have assumed we would have configured the promise to have performed the unpack operation so we don't have to do it everywhere else.

bjhargrave: The unpacking should be done as a map in the fetch result:

```java
return fetch(a, 2, 1000L).map(tag -> Artifact.FORMAT_PACKED.equals(a.format) ? 
  handleDownloadResult(true, null, tag.getFile()) : tag.getFile())
.map(file -> {
  ResourceBuilder rb = new ResourceBuilder();
  rb.addFile(file, a.uri);
  ...
})
```

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -109,24 +114,44 @@ File get(String bsn, Version version, Map<String, String> properties, DownloadLi
...
 		new DownloadListenerPromise(reporter, name + ": get " + bsn + ";" + version + " " + url, go, listeners);
+
 		return link;
 	}
 
+	private File handleDownloadResult(boolean packed, File link, File source)
+		throws Exception {
+
+		File unpacked = new File(source.getParentFile(), source.getName() + ".unpacked");
+		if (packed && !unpacked.exists()) {
```
bjhargrave, line 141: I think the exists() test is a problem. It means that if the remote file is changed we never unpack the changed file it if was previously unpacked. That seems like a bad caching problem.

Also, we do we keep both files (original and original.unpacked). Perhaps we should unpack in place? Then we can know the link value before and avoid having to create it later here.

biz.aQute.repository/bnd.bnd
```diff
@@ -26,6 +26,7 @@
 Export-Package: \
 	aQute.bnd.deployer.http;bnd-plugins=true,\
 	aQute.p2.api,\
+	aQute.p2.packed,\
```
bjhargrave, line 29: I would avoid exporting this package. There is no need I can see to make this API.

Issue #3783, opened 2020-02-18
Bnd's Maven dependency scope handling not correct


maggu2810:
For me it seems that the logic bnd uses to handle resolve the Maven dependency scope is not correct.

To demonstrate the mismatch I add an example here: https://github.com/maggu2810/bnd-depscope.tmp

There are to POM file that define dependencies:
* deps/compile: dependencies used at compile time
* deps/runtime: dependencies used at runtime

There is one application that depends on both artifact:
* deps compile using scope provided
* deps runtime using scope compile

So, if the Maven dependency scope are applied correctly the dependencies of "deps compile" that are not part of "deps runtime" should use scope provided.
The Maven dependencies that are part of "deps runtime" should use scope "compile".

Maven handles this correctly.

After checking out the repo above execute `mvn clean install`.
This will build the whole reactor.

Now, inspect the dependencies and their copy of the "app.demo": `mvn dependency:list -pl apps/demo/`

It identifies that the json-b and json specification is using "compile" scope.
* org.apache.geronimo.specs:geronimo-jsonb_1.0_spec:jar:1.1:compile
* org.apache.geronimo.specs:geronimo-json_1.1_spec:jar:1.2:compile

If you open the `demo.bndrun` file in the Eclipse IDE you will realize that these bundles are not part of the implicit repository (but it should as compile and runtime scope should be added).

I assume artifacts that are added twice, the first time using scope provided and the second time using scope compile are not correctly resolved by Bnd. The "new" scope will be compile (as done by Maven itself) but Bnd stays at provided.

maggu2810:
@rotty3000 You already fixed #3361 that has been about a different dependency handling of maven compiler plugin vs. bnd's maven plugins.

Do you think you will some time to look into this one?

maggu2810:
Hi,
I started some debug session and it seems I found the cause of the problem.

Here a short example that should demonstrate the problem.

Think about a POM file that looks like this one:

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
  <modelVersion>4.0.0</modelVersion>

  <parent>
    <groupId>de.maggu2810.bnd.depscope.deps</groupId>
[...]

PR #3880, opened 2020-04-06
fix scope handling at dependency resolution

maggu2810:
At dependency resolution the scope should be used to identify if an artifact should be added or not.
It must not be used to skip the whole part of a tree as the childs of a tree node could use a different scope then one of its predecessor.

Related to: https://github.com/bndtools/bnd/issues/3783


maggu2810:
For a more detailed explanation see: https://github.com/bndtools/bnd/issues/3783#issuecomment-609567816

rotty3000:
@maggu2810 I overlooked during review that you hadn't added the github hint `fixes #3783` in the commit comments which resulted in the issue not being auto closed.
Let's both try to remember about that ;)

maggu2810:
You are right. I have chosen "related to" instead of "fixes" because this PR fixes "at least one bug" I found while inspecting the specific issue.
I did not yet check if it solved the issue in all its details.
So I decided to be more defensive. If it is already fixed by this change I could close them manually easily. Instead of realising that there is another change necessary and open the issue again...
Will try to invest more investigation next time in front of PR creation.

PR #3883, opened 2020-04-06
fix: plugin version in readme

mneundorfer:
I'm guessing this is a typo, as according to [plugins.gradle.org](https://plugins.gradle.org/search?term=bnd), only `5.0.1` is available, not `5.1.0`...?


bjhargrave:
The README in the master branch is for the version being built in the master branch which is 5.1.0. The README in the master branch may discuss new features which are in 5.1.0 and not the prior release. The README discusses this in the 3rd paragraph.

Please see the README on the 5.0.1.REL tag for the 5.0.1 level README. https://github.com/bndtools/bnd/blob/5.0.1.REL/biz.aQute.bnd.gradle/README.md

PR #3885, opened 2020-04-06
Updated documentation of findproviders macro

glimmerveen:
Updated documentation of the newly added findproviders macro.

@pkriens : I was unsure about where to place the Maven part you suggested I add. If you had another place in mind, please let me know.

### DIFF
--- a/docs/_macros/findproviders.md
+++ b/docs/_macros/findproviders.md
@@ -1,7 +1,7 @@
 ---
 layout: default
 class: Workspace
-title: findproviders  ';' namespace ( ';' FILTER )? 
+title: findproviders  ';' namespace ( ';' FILTER )?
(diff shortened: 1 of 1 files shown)


rotty3000:
This looks really great @glimmerveen but could you add your `Signed-of-by:` attribute so we can accept it? (just sign off your commit and then force push to your branch is all you need to do)

glimmerveen:
I updated my commit, is this what you had in mind @rotty3000 ?

rotty3000:
haha, I do aplogize, it's correct including the typo I made in the attribute which I tried to quicky correct (however I don't think corrections in comments get email updates).

`Signed-off-by:` (note that I missed the second `f` earlier). Please update it one last time ;) and thanks again.

glimmerveen:
Haha, I totally missed that typo :)

bjhargrave:
Thanks!

Issue #3439, opened 2019-09-21
NPE in BndPomRepository if pom file not found


kriegfrj:
I had a `BndPomRepository` configured in my `build.bnd` but the POM file was missing. Ended up with the following NPE. It would be nice if it were a little more gracious about handling it, and a little more helpful in reporting what is wrong.

eclipse.buildId=4.12.0.I20190605-1800
java.version=12.0.1
java.vendor=Oracle Corporation
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en_AU
Framework arguments:  -product org.eclipse.epp.package.java.product

bndtools.core
Error
Sun Sep 22 01:04:50 ACST 2019
Init of workspace

java.lang.NullPointerException
	at aQute.bnd.repository.maven.pom.provider.BndPomRepository.versions(BndPomRepository.java:270)
	at aQute.bnd.build.Project.getBundle(Project.java:1276)
	at aQute.bnd.build.Project.getBundles(Project.java:656)
	at aQute.bnd.build.Project.parseBuildpath(Project.java:558)
	at aQute.bnd.build.Project.prepare(Project.java:400)
	at aQute.bnd.build.Project.getDependson(Project.java:865)
	at aQute.bnd.build.Workspace.getBuildOrder(Workspace.java:571)
	at bndtools.central.Central.getWorkspace(Central.java:253)
	at bndtools.central.Central.getProject(Central.java:575)
	at bndtools.central.Central.getProject(Central.java:579)
	at org.bndtools.builder.classpath.BndContainerInitializer$Updater.<init>(BndContainerInitializer.java:230)
	at org.bndtools.builder.classpath.BndContainerInitializer.initialize(BndContainerInitializer.java:87)
	at org.eclipse.jdt.internal.core.JavaModelManager.initializeContainer(JavaModelManager.java:3195)
	at org.eclipse.jdt.internal.core.JavaModelManager$10.run(JavaModelManager.java:3079)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2295)
	at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2322)
	at org.eclipse.jdt.internal.core.JavaModelManager.initializeAllContainers(JavaModelManager.java:3141)
	at org.eclipse.jdt.internal.core.JavaModelManager.getClasspathContainer(JavaModelManager.java:2111)
[...]

PR #3915, opened 2020-04-15
created non-s versions of snapshot and release url properties

### LINKED ISSUE #3872: modify BndPomRepository properties
The BndPomRepository has the property pom and revision, but it can handle a List of each. Thus we should add "poms" and "revisions" to the internal configuration to take one or the other together with an update to the Documentation. 

Such inconsistency make a bad user experience.

juergen-albert:
* fixes #3872
* moved from Exceptions status messages in the BndPomRepository

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
@@ -10,6 +10,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -39,6 +40,7 @@ import aQute.bnd.service.RegistryPlugin;
 import aQute.bnd.service.RepositoryListenerPlugin;
 import aQute.bnd.service.RepositoryPlugin;
 import aQute.bnd.service.clipboard.Clipboard;
+import aQute.bnd.service.repository.Prepare;
 import aQute.bnd.util.repository.DownloadListenerPromise;
 import aQute.bnd.version.Version;
 import aQute.lib.converter.Converter;
@@ -57,11 +59,12 @@ import aQute.service.reporter.Reporter;
  */
 @BndPlugin(name = "BndPomRepository")
 public class BndPomRepository extends BaseRepository
-	implements Plugin, RegistryPlugin, RepositoryPlugin, Refreshable, Actionable, Closeable {
+	implements Plugin, RegistryPlugin, RepositoryPlugin, Refreshable, Actionable, Closeable, Prepare {
 	private static final String	MAVEN_REPO_LOCAL	= System.getProperty("maven.repo.local", "~/.m2/repository");
 	private static final int	DEFAULT_POLL_TIME	= 300;
 
-	private boolean				inited;
+	private Promise<Boolean>	initialized;
+
 	private PomConfiguration	configuration;
 	private Registry			registry;
 	private String				name;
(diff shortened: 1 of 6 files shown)


biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -57,11 +59,11 @@
  */
 @BndPlugin(name = "BndPomRepository")
 public class BndPomRepository extends BaseRepository
-	implements Plugin, RegistryPlugin, RepositoryPlugin, Refreshable, Actionable, Closeable {
+	implements Plugin, RegistryPlugin, RepositoryPlugin, Refreshable, Actionable, Closeable, Prepare {
 	private static final String	MAVEN_REPO_LOCAL	= System.getProperty("maven.repo.local", "~/.m2/repository");
 	private static final int	DEFAULT_POLL_TIME	= 300;
 
-	private boolean				inited;
+	private Boolean				inited				= false;
```
pkriens, line 66: Why?

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -75,43 +77,71 @@
...
+	private String				status;
 
-			startPoll();
-		} catch (Exception e) {
-			reporter.exception(e, "Init for BndPomRepository failed %s", configuration);
-			throw Exceptions.duck(e);
+	@SuppressWarnings("deprecation")
+	@Override
+	public void prepare() throws Exception {
+		synchronized (this) {
```
pkriens, line 85: prepare does not need to be synchronized, it is only called once after all plugins have been created and injected

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -151,7 +181,10 @@ private void poll() throws Exception {
 
 	@Override
 	public boolean refresh() throws Exception {
-		init();
+		prepare();
```
pkriens, line 184: don't use prepare. Refresh is about the pom file but not about the settings. If setting sin the bnd file change, a new plugin is created

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -166,49 +199,56 @@ public boolean refresh() throws Exception {
...
+				if (revisions.isEmpty()) {
+					status("Revision is neither a file nor a revision " + configuration.revision());
+				}
+			} else if (configuration.query() != null) {
+				this.query = configuration.query();
+				this.queryUrl = configuration.queryUrl("http://search.maven.org/solrsearch/select");
+			} else {
+				status("Neither pom, revision nor query property are set");
+			}
 			inited = false;
```
pkriens, line 236: don't need to do this. setProperties is only called once

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -230,7 +270,10 @@ public PutResult put(InputStream stream, PutOptions options) throws Exception {
 	@Override
 	public File get(String bsn, Version version, Map<String, String> properties, DownloadListener... listeners)
 		throws Exception {
-		init();
+		prepare();
```
pkriens, line 273: don't use prepare. You can have an init() but prepare should catch all errors & is only called once.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
juergen-albert: The old init method has been synchronized and the setProperties sets the the initialized marker in a syncronized block. I wasn't sure if this would be necessary, but I thought that their might be a reason, I could not see, so I decided to keep the sync approach. I will throw this out.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
juergen-albert: I used the init to sync on  at one point and just forgot to switch it back.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
juergen-albert: you mean the `inited = false` here?

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
juergen-albert: Hmm... If I have a prepare that simply checks the config and sets a status if something is off, I would need to check the `isOk()` in the init because it would most likely throw an Exception. The I would need to check everywhere the `isOk()` as well (as I already do), to avoid Exceptions e.g. in BndTools. Would this be right?

pkriens: The prepare method is from the Prepare interface, it has a well defined meaning, so it should not be called as the check method. In prepare() you calculate _anything_ that is related to the properties since they will not change during the life of the instance. Since you do not want to block, ideally you want to do any initialization in a background thread. An init() method could then block until that thread is finished & check isOk().

     prepare() -> process properties, registry, etc. & start init thread
     init()          -> block till init thread is ready, check status

pkriens:
Did you also look at #3439?

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -75,21 +80,90 @@
 	private String				queryUrl;
 	private ScheduledFuture<?>	pomPoller;
 
-	public synchronized void init() {
+	private String				status;
+
+	private PromiseFactory		promiseFactory		= new PromiseFactory(Executors.newSingleThreadExecutor());
```
pkriens, line 85: You should get this one from the registry (setRegistry), or `Processor.getPromiseFactory()`

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -75,21 +80,90 @@
...
+
+	@SuppressWarnings("deprecation")
+	private boolean init() {
 		try {
-			if (inited)
-				return;
-			inited = true;
+			if (initialized == null) {
```
pkriens, line 90: Generally you make sure that this is initialized before you call the method. You can create the promise in prepare() since this is _always_ called before any of the methods that can call init().

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -75,21 +80,90 @@
...
 			Workspace workspace = registry.getPlugin(Workspace.class);
 			HttpClient client = registry.getPlugin(HttpClient.class);
 			localRepo = IO.getFile(configuration.local(MAVEN_REPO_LOCAL));
 			File location = workspace.getFile(getLocation());
 
-			List<MavenBackingRepository> release = MavenBackingRepository.create(configuration.releaseUrls(), reporter,
-				localRepo, client);
-			List<MavenBackingRepository> snapshot = MavenBackingRepository.create(configuration.snapshotUrls(),
-				reporter, localRepo, client);
+			String releaseUrl = configuration.releaseUrl() != null ? configuration.releaseUrl()
```
pkriens, line 158: you should do all this work in prepare() since it is only done once.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -166,49 +244,20 @@ public boolean refresh() throws Exception {
...
 
 	@Override
 	public Map<Requirement, Collection<Capability>> findProviders(Collection<? extends Requirement> requirements) {
-		init();
+		try {
+			if (!init()) {
+				return Collections.emptyMap();
+			}
+		} catch (RuntimeException re) {
+			throw re;
```
pkriens, line 257: if you use duck, no need to do this extra Runtime check

juergen-albert:
> Did you also look at #3439?

Yes, this is taken care of as well.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
juergen-albert: good to know, I will use this factory instead

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
juergen-albert: I already do all this in the prepare method, but a lot of the existing tests didn't work this way. So I decided to do the extra check, so it would still work as it did before. As this is public API, I did not want to make it too hard on other people using it. I reasoned, that this check will not be too expensive and should be tolerable.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
juergen-albert: I had it in the prepare method in a `() -> {}` block, but the method became to long, so I thought it might be better to split it. prepare is doing the checks and kicks of the init Promise and the extra method is doing the extra initialization. I can do it in one method, but it looks ugly though.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
pkriens: yeah, for tests you sometimes have to compromise, ok

PR #3917, opened 2020-04-15
Commands-Test should also run with JAVA_HOME

stbischof:

### DIFF
--- a/aQute.libg/test/aQute/libg/command/CommandTest.java
+++ b/aQute.libg/test/aQute/libg/command/CommandTest.java
@@ -4,23 +4,28 @@ import static org.assertj.core.api.Assertions.assertThat;
 
 import org.junit.jupiter.api.Test;
 
+import aQute.lib.io.IO;
+
 public class CommandTest {
 
 	@Test
-	public void testSimple() throws Exception {
+	public void testExecution() throws Exception {
 		Command c = new Command();
-		c.add("java");
+		c.add(IO.getJavaExecutablePath("java"));
 		c.add("-Dx=\"foo bar\"");
 		c.add("-version");
-		assertThat(c.getArguments()).containsExactly("java", "-Dx=\"foo bar\"", "-version");
+		assertThat(c.getArguments()).containsExactly(IO.getJavaExecutablePath("java"), "-Dx=\"foo bar\"", "-version");
 		assertThat(c.execute(System.out, System.err)).isZero();
 	}
 
 	@Test
 	public void testFull() throws Exception {
-		Command c = new Command("java -Dx=\"foo bar\" -version");
-		assertThat(c.getArguments()).containsExactly("java", "-Dx=\"foo bar\"", "-version");
-		assertThat(c.execute(System.out, System.err)).isZero();
+
+		Command c = new Command("app -Dx=\"foo bar\" -version");
+		assertThat(c.getArguments()).containsExactly("app", "-Dx=\"foo bar\"", "-version");
+
+		c = new Command().full("app -Dx=\"foo bar\" -version");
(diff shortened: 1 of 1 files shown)


aQute.libg/test/aQute/libg/command/CommandTest.java
```diff
@@ -2,24 +2,31 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.util.Optional;
+
 import org.junit.jupiter.api.Test;
 
 public class CommandTest {
+	private static String java() {
+		return Optional.ofNullable(System.getenv("JAVA_HOME"))
```
bjhargrave, line 11: `aQute.lib.io.IO.JAVA_HOME` is a static holding the location of the JAVA_HOME folder. Please use it. Also, this should be done in a `@BeforeEach` method rather than a static method since this is JUnit 5 test case. The `@BeforeEach` method can set a field with the proper java command path which is then referenced from the test methods.

stbischof: What is the benefit to use @BeforeEach and not @BeforeAll?

bjhargrave: BeforeEach is instance based and not class based. Unless we have a real need for static initialization, we should avoid it.

aQute.libg/test/aQute/libg/command/CommandTest.java
```diff
@@ -2,24 +2,45 @@
...
+			.filter(File::isFile)
+			.map(File::toString)
+			.orElse("java");
+	}
 
 	@Test
 	public void testSimple() throws Exception {
 		Command c = new Command();
-		c.add("java");
+		c.add(java);
```
bjhargrave, line 33: I just added `IO.getJavaExecutablePath(String name)` to master. So you can rebase and then just do

```java
c.add(IO.getJavaExecutablePath("java"));
```

This also means you wont need the BeforeEach!

stbischof: done

aQute.libg/test/aQute/libg/command/CommandTest.java
```diff
@@ -4,22 +4,24 @@
...
-		assertThat(c.getArguments()).containsExactly("java", "-Dx=\"foo bar\"", "-version");
+		assertThat(c.getArguments()).containsExactly(IO.getJavaExecutablePath("java"), "-Dx=\"foo bar\"", "-version");
 		assertThat(c.execute(System.out, System.err)).isZero();
 	}
 
 	@Test
 	public void testFull() throws Exception {
-		Command c = new Command("java -Dx=\"foo bar\" -version");
-		assertThat(c.getArguments()).containsExactly("java", "-Dx=\"foo bar\"", "-version");
+		Command c = new Command(IO.getJavaExecutablePath("java") + " -Dx=\"foo bar\" -version");
```
bjhargrave, line 23: I just realized that this will fail if the absolute path string contains a space since the full string is parsed by spaces (This is why no one should really use the full command string). So for this test method, you will need to revert to the prior state (using simple `java` command) and delete the final assert on the execute method. We have confirmed the execution works in `testSimple` and this method confirms the full argument parsing, so we can avoid the assert on the execute here.

stbischof: next try. thank you for review

bjhargrave:
Maybe you can squash these 3 commits into 1? Especially since 2 of them have the same comment.

stbischof:
here it is

bjhargrave:
Thanks

PR #3926, opened 2020-04-16
Pkriens/generate

pkriens:
A number of commits to implement source code generation


pkriens:
@bjhargrave This PR contains the code generation. Let me know if you want to discuss it. The API is in `ProjectGenerate`. You can get the inputs from this & the outputs are in the ${src} set.

The switch to the CompilerParticipant worked fine, reduced the risk it seemed because the changes are now very local.

docs/_instructions/generate.md
```diff
@@ -0,0 +1,56 @@
...
+
+    src=${^src},src-gen
+
+Assuming that the input files are in the `gen` directory, the following can be used to automatically generate the output files based on the input files.
+ 
+    -generate:   \
+        gen/**.java; \ 
+            output='src-gen/' ; \ 
+            generate='javagen -o src-gen gen/'
+
```
juergen-albert, line 56: Can you add an example, that an external Plugin? It is a but unclear how such an instruction would need too look like.

docs/_chapters/875-external-plugins.md
```diff
@@ -0,0 +1,33 @@
...
+        subtype       = String.class)
+    public class CallImpl implements Callable<String> {
+        public String call() throws Exception {
+            return "hello";
+        }
+    }
+
+In Bndtools, you can declare any class as an ExternalPlugin. The automatic build features will automatically build the JAR of the plugin and this will immediately become available in the rest of the build. If you use external plugins from the local workspace, make sure to declare a  `-dependson` to the external plugin project in any project that uses it, this dependency is not automatically detected. 
+
+A JAR can contain any number of external plugins. It must ensure that it does not have any dependencies outside the bndlib it was compiled against.
```
juergen-albert, line 33: Am I seeing this right, can we use this then with `-plugin.test: my.external.plugin.MyPlugin;prop1=1`?

docs/_instructions/generate.md
pkriens: https://github.com/bndtools/bnd/pull/3926/files#diff-3ea5fc0a86dd03fb518bf256100a9e27

docs/_chapters/875-external-plugins.md
pkriens: No, external plugins are only supported by generate right now. However, they are completely independent of them. If they are appreciated, we should integrate them with the internal plugins.

juergen-albert: +1 from me!

docs/_instructions/generate.md
juergen-albert: I'm not quite sure what the diff should show me.

pkriens: Sorry, was supposed to be a link to the external-plugins

PR #3927, opened 2020-04-16
[reporter] calculate and show Checksum the bundle

stbischof:


biz.aQute.bnd.reporter/src/biz/aQute/bnd/reporter/plugins/entries/bundle/ChecksumPlugin.java
```diff
@@ -0,0 +1,95 @@
...
+		_properties.put(ReportEntryPlugin.ENTRY_NAME_PROPERTY, EntryNamesReference.CHECKSUM);
+		_properties.put(ReportEntryPlugin.SOURCE_CLASS_PROPERTY, Jar.class.getCanonicalName());
+	}
+
+	@Override
+	public ChecksumDTO extract(final Jar jar, final Locale locale) {
+		Objects.requireNonNull(jar, "jar");
+		Objects.requireNonNull(locale, "locale");
+
+		ChecksumDTO checksumDTO = new ChecksumDTO();
```
bjhargrave, line 39: See the `aQute.libg.cryptography` for classes to compute MD5, SHA1, SHA256. They provide a nice `asHex` method to return the digest value as a hex string.

stbischof: I changed it to `aQute.libg.cryptography`. I is really useful!

Had to fix SHA-512. Not sure with that. please have a look.

bjhargrave:
Thanks for the SHA512 fix!

Issue #3819, opened 2020-03-03
[bndtools] ConcurrentModificationException occurs during: "Create implicit repo"


cdelg:
eclipse.buildId=4.14.0.I20191210-0610
java.version=11.0.2
bndtools=  5.0.0.REL-202001221519-g612f80a
using maven

I get this exception each time I open a bndrun file:

    An internal error occurred during: "Create implicit repo".
    java.util.ConcurrentModificationException
        at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1660)
        at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
        at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
        at java.base/java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
        at java.base/java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
        at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
        at java.base/java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:497)
        at bndtools.m2e.MavenImplicitProjectRepository.fullRefresh(MavenImplicitProjectRepository.java:159)
        at bndtools.m2e.MavenImplicitProjectRepository.createRepo(MavenImplicitProjectRepository.java:144)
        at bndtools.m2e.MavenImplicitProjectRunListener$1.run(MavenImplicitProjectRunListener.java:43)
        at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)

Do you thing it is a bug or is there something wrong with my bndrun file:

    -standalone: \
	    ../../bom/runtime-index/target/index.xml;name="runtime-index",\
        ../../bom/core-index/target/index.xml;name="core-index",\
         target/index.xml;name="self"

    -runfw: org.eclipse.osgi

    -runee: JavaSE-11

    -resolve.effective: resolve, active

    -runrequires: bnd.identity;id='com.xxx.openhab.io.yyy.provider'

    #-distro: ${.}/openhab-distro-2.5.0.jar;version=file

    -runstartlevel: order=sortbynameversion

    -runrepos: com.xxx.openhab.io.yyy.provider (implicit)

cdelg:
[...]

PR #3929, opened 2020-04-16
m2e: Avoid possible ConcurrentModificationException

bjhargrave:
We make a copy of the Refreshable repositories before we iterate over
them.

Fixes https://github.com/bndtools/bnd/issues/3819


bjhargrave:
This is a band-aid over the real issue where the bndtools.m2e project mutates the Workspace.getRepositories() list which it should not.

bjhargrave:
Closing as this is just wallpaper.

PR #3936, opened 2020-04-18
Runbundles Version to snapshot

### LINKED ISSUE #3928: Runbundles Version to snapshot
According to the latest changes, the resolver always adds the respective bundle versions to the `-runbundles` entries.

For example,

```
-runbundles: \
	californium-osgi;version='[1.0.8,1.0.9)',\
	com.eclipsesource.jaxrs.jersey-min;version='[2.22.1,2.22.2)',\
	com.fasterxml.jackson.core.jackson-annotations;version='[2.7.0,2.7.1)',\
	com.fasterxml.jackson.core.jackson-core;version='[2.7.4,2.7.5)',\
	com.fasterxml.jackson.core.jackson-databind;version='[2.7.4,2.7.5)',\
	com.google.guava;version='[15.0.0,15.0.1)',\
	com.company.binding;version='[1.198.1,1.198.2)',\
	com.company.backend.httpclient;version='[1.198.1,1.198.2)',\
	com.company.backendclient;version='[1.198.1,1.198.2)',\
```

In the aforementioned list, the bundles having reverse domain of `com.company` (the last 3 bundles) are workspace projects.

Previously, this `-runbundles` was as follows (also prepared by the resolver):

```
-runbundles: \
	californium-osgi;version='[1.0.8,1.0.9)',\
	com.eclipsesource.jaxrs.jersey-min;version='[2.22.1,2.22.2)',\
	com.fasterxml.jackson.core.jackson-annotations;version='[2.7.0,2.7.1)',\
	com.fasterxml.jackson.core.jackson-core;version='[2.7.4,2.7.5)',\
	com.fasterxml.jackson.core.jackson-databind;version='[2.7.4,2.7.5)',\
	com.google.guava;version='[15.0.0,15.0.1)',\
	com.company.binding;version=snapshot,\
	com.company.backend.httpclient;version=snapshot,\
	com.company.backendclient;version=snapshot\
```

According to the current configuration we have, for every product release, we use the same version for all the bundles generated from the workspace projects - such as 1.x.y.TIMESTAMP. So, after every release or hotfix, we need to update all the bndrun files we have. This is really time consuming job to get them resolved and wait 40-50 seconds (we have around 300 workspace projects) for every bndrun file.

Is there any way to ensure that resolver always uses `snapshot` as versions for the bundles that exist as workspace projects?

rotty3000:
fixes #3928 even more


bjhargrave:
I looked at doing the same basic thing when I was working on my fix. However, your solution has performance issues. When even one project is rebuilt, you end up reindexing the whole workspace. This may be fine for small workspaces but imagine a workspace where there are hundreds of projects with dependencies. When you rebuild some project at the base of the dependency tree you will reindex the whole workspace each time each dependent project rebuilds. This can be a huge problem in something like Open Liberty.

So I reached the conclusion that the current setup for Eclipse is OK. It only reindexes the the files in a project as they get built. It uses the `.index` files on disk to aid in Eclipse start time performance since you don't need to actually index all the bundles in the workspace on Eclipse start, just read their `.index` files.

So I think we do need a specialized workspace repository for Eclipse/Bndtools which differs from the simple static WorkspaceResourcesRepository which works fine for Maven and non-Bnd workspace Gradle. This specialized workspace repository can and should be improved, but we cannot replace it with a simple index-the-whole-workspace repository like WorkspaceResourcesRepository.

So I don't think this PR is a good idea.

rotty3000:
Sure, I'm still working on this as you can see it's draft. ;)

What I'm thinking is that to avoid file system resources the WorkspaceResourcesRepo should actually Map the resources per file we index with it. the resource repo would then drop resources associated with a given file and readd only those as the builder updates. I'm poc'ing this now.

bjhargrave:
You still need to consider Eclipse start up time which the `.index` files assist with. So I don't think we can do away with those.

rotty3000:
aright

PR #3949, opened 2020-04-21
[parallel] Trying again to go parallel

pkriens:
There were launchpad tests failing because
the Workspace got closed.

I fixed one small bug that could cause multiple
remote workspaces but that was unlikely to 
cause problems during the testing.

There used to be tests that used the bnd workspace that opened up a test case in one
process and that might have caused problems
due to the port finding algorithm. However,
I can no longer find tests that use the bnd
workspace itself. (Maybe bndall or the gradle
tests)

This is a test to find out if the problem
is gone.


bjhargrave:
@pkriens Should we put the maven repo fix in its own PR, so we can merge it?

pkriens:
It is already in the main branch

bjhargrave:
> It is already in the main branch

Excellent! Thanks.

PR #3950, opened 2020-04-21
[workspace] Canonicalize paths in getProjectFromDir

pkriens:
In a test, the ws had '..' in the path and therefore failed to find that a project dir pointed to a directory.

This patch canonicalizes both paths.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
@@ -91,6 +91,7 @@ import aQute.bnd.version.Version;
 import aQute.bnd.version.VersionRange;
 import aQute.lib.collections.Iterables;
 import aQute.lib.deployer.FileRepo;
+import aQute.lib.exceptions.Exceptions;
 import aQute.lib.hex.Hex;
 import aQute.lib.io.IO;
 import aQute.lib.io.IOConstants;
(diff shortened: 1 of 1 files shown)


biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -320,13 +321,18 @@ public void setFileSystem(File workspaceDir, String bndDir) throws Exception {
...
 	public Project getProjectFromFile(File projectDir) {
-		projectDir = projectDir.getAbsoluteFile();
-		assert projectDir.isDirectory();
+		try {
+			projectDir = projectDir.getCanonicalFile();
+			assert projectDir.isDirectory();
 
-		if (getBase().equals(projectDir.getParentFile())) {
-			return getProject(projectDir.getName());
+			if (getBase().getCanonicalFile()
```
bjhargrave, line 328: Wouldn't be better to make sure Processor always stores it's base as a canonical file? Or does that mess up symlinks?

pkriens: That is the reason I did not have the courage ...

I agree it would be nicer but it can break existing code. I always thought getAbsoluteFile would remove the .. :-(

bjhargrave: Maybe instead of canonical path, we us IO.getPath since it already has logic to clean . and .. from paths?

https://github.com/bndtools/bnd/blob/bee81533e76d26df4445cf6099c87b18d505dd8e/aQute.libg/src/aQute/lib/io/IO.java#L679-L697

pkriens: I considered that but I am not sure it handles the Windows drive identifier (d:) properly? This is the current problem I am trying to fix

pkriens: WINDOWS SUCKS!

PR #3953, opened 2020-04-21
fixes BndPomRepo refresh error #3952

### LINKED ISSUE #3952: BndPomRepository with new revision does not work from time to time
If a BndPomRepository was initialized and an index XML was created in the cache, changes to the any property will not have any effect under some circumstances.

If a file already exists, it will only be rewritten if something mentioned in the config becomes stale. Thus if I initialize it with any revision an index is written. Now we change e.g. the Version. The repo checks, if the pom of the revision is in the local maven repo and if the pom is newer then the index file to determine if it is stale. 
This mechanism works If we work with snapshot revisions in a running system like bndtools, so it would automatically would pic up on newer Versions. If I switch between Versions it only works if the Version is really new. In every other case the old index would  persist and bnd will never know about the new config.

PR follows.

juergen-albert:
fixes #3952


bjhargrave:
@pkriens Are you ok with this PR?

PR #3955, opened 2020-04-22
Split OSGi API bundles into separate repo

kriegfrj:
As I understand it, all of the `org.osgi.*` bundles are meant for building only. At runtime, one of the OSGi framework bundles that implements these APIs should be used instead. So they should appear in `-buildpath` but not `-runbundles`. This means that you don't want the resolver to consider the `org.osgi.*` bundles when resolving.

By separating the OSGi bundles into a separate repo, it is easy to exclude that repo using `-runrepos`. Then the resolver won't consider them. This has two benefits:

1. They won't accidentally end up in `-runbundles` when they shouldn't.
2. The resolver has less alternatives to consider and so it will be able to do its job faster.


bjhargrave:
> As I understand it, all of the `org.osgi.*` bundles are meant for building only. At runtime, one of the OSGi framework bundles that implements these APIs should be used instead.

This is _NOT_ true. Most OSGi bundles are usable as runtime. 

osgi.core is for compile only. Your framework impl must provide the core API are runtime.

osgi.cmpn is discouraged from use at runtime and the latest version won't resolve. This is to encourage people to use the individual org.osgi.service.* jars at runtime.

But all other osgi jars are fine to use at runtime. See this blog post: https://blog.osgi.org/2020/01/to-embed-or-not-to-embed-your-api.html

kriegfrj:
> > As I understand it, all of the `org.osgi.*` bundles are meant for building only. At runtime, one of the OSGi framework bundles that implements these APIs should be used instead.
> 
> This is _NOT_ true. Most OSGi bundles are usable as runtime.
> 
> osgi.core is for compile only. Your framework impl must provide the core API are runtime.
> 
> osgi.cmpn is discouraged from use at runtime and the latest version won't resolve. This is to encourage people to use the individual org.osgi.service.* jars at runtime.
> 
> But all other osgi jars are fine to use at runtime.

There's also `org.osgi.service.component.annotations`, which doesn't resolve. But for the most part, I stand corrected. It's probably not worth creating the separate repo for only three jars.

PR #3958, opened 2020-04-22
Use file for jarindex

bjhargrave:


biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -121,13 +123,22 @@
...
-		Lazy<WorkspaceExternalPluginHandler>	externalPlugins	= new Lazy<>(
-			() -> new WorkspaceExternalPluginHandler(Workspace.this));
+	class WorkspaceData implements Closeable {
+		final Supplier<List<RepositoryPlugin>>			repositories;
+		volatile RemoteWorkspaceServer					remoteServer;
+		final Supplier<WorkspaceClassIndex>				classIndex;
+		final Supplier<WorkspaceExternalPluginHandler>	externalPlugins;
+
+		WorkspaceData() {
+			repositories = Memoize.supplier(Workspace.this::initRepositories);
```
pkriens, line 133: I am not happy with the replacement of Lazy with Memoize. It has no close support nor introspective facilities. Can  you explain why you made this change?

pkriens: I would like to revert until there is a compelling reason to make it less functional and more complex?

bjhargrave: Lazy has some serious issues. I am happy to discuss when you are available.

PR #3959, opened 2020-04-23
gradle: Add support for Bnd's code generation to Bnd Workspace plugin

bjhargrave:
Adds a new `generate` task upon which `compileJava` depends if the
project wants to generate code.

@pkriens I am not sure how to test this. We should add a new test project for the gradle plugin. But need help to make a test workspace like https://github.com/bndtools/bnd/tree/master/biz.aQute.bnd.gradle/testresources/workspaceplugin1 which can be used to test the `generate` task.


bjhargrave:
@pkriens I completed the test case for the Bnd gradle plugin changes. I needed to add a new method to `ProjectGenerate` to return the list of output dirs from the `-generate` instruction.

PR #3964, opened 2020-04-23
Fix workspace closeable

bjhargrave:
@pkriens I added a CloseableMemoize after our discussion today and fixed the workspace and project to be closeable and use it.

This is also based on a change to make Memoize an interface with peek and some monadic methods.


pkriens:
I had made changes as well, so I need to rebase :-(

Issue #3882, opened 2020-04-06
[junit-platform] NoClassDefFoundError in test class silently ignored


kriegfrj:
>As it happens, my initial attempts at doing this have exposed an issue with the junit-platform tester, which I will document in a new issue shortly.
>
>_Originally posted by @kriegfrj in https://github.com/bndtools/bnd/issues/3881#issuecomment-609628225_

With my first attempt at trying to port `ImportPackageQuickFixProcessorTest` to an OSGi JUnit test, I found that the tester was silently skipping over my Jupiter test case. When I added another (simpler) test in the same bundle, it ran the simpler test fine but still skipped the other.

Using the debugger, I managed to establish that `EngineDiscoveryRequestResolution.resolve()` was throwing a `NoClassDefFoundError`, which was in turn being silently swallowed by the `catch(Throwable t)` in `EngineDiscoveryRequestResolution.resolveCompletely()`.

`BundleSelectorResolver` tries to prevent this from happening by loading the class before handing off to the engines, but it seems this doesn't catch all cases - I suspect that although we're loading the class we're not triggering initialization of the class. Will look at using `Class.forName()` or something instead.

kriegfrj:
Ok, so after more wrestling with the debugger, I've figured out that the `NoClassDefFoundError` is triggered when you first call `Class.getDeclaredMethods()` (which is called by `ReflectionUtils.findMethod()`) - if one of the methods references a class that is not visible to the current class. I think the fix is to try and make sure that it is called before handing off to the engine, so that we can catch and report the error appropriately.

I will try and add some appropriate test cases to cover this and then try and submit a fix.

rotty3000:
One has to wonder how the bundle made it into runtime with missing classes? Maybe all that is needed is for `ReflectionUtils.findMethod()` ensure it always puts the classloader of the test class on the TCCL before looking further at it?

kriegfrj:
> One has to wonder how the bundle made it into runtime with missing classes? 

This is Eclipse,  complete with split packages and missing dependency information... I had to manually add the x86_64 fragment for org.eclipse.swt to runrequires and re- resolve to fix it. Gave me the idea of a quick fix processor to fix these kinds of issues

> Maybe all that is needed is for `ReflectionUtils.findMethod()` ensure it always puts the classloader of the test class on the TCCL before looking further at it?

There are some improvements to be made to the class loading in JUnit, it's true. Bit in my case the class was actually missing and Jupiter reacted by silently skipping the class.

[...]

PR #3967, opened 2020-04-24
[junit-platform] Handle singleton unresolveds

### LINKED ISSUE #3966: [junit-platform] Class resolution failure for singleton test class silently ignored
While working on #3882 I discovered another small bug: 

Normally, if you have a test class in a bundle that fails to load/resolve, `BundleEngine` will alert you to the fact and report it as an error under that bundle's node. There are tests in `BundleEngineTest` that cover this.

However, if it happens that *none* of the test cases in a bundle resolve successfully, then that bundle will be silently ignored, and no feedback given to the user.

The fix is fairly straightforward and will follow shortly.

kriegfrj:
Fixes #3966.

### DIFF
--- a/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java
+++ b/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java
@@ -68,7 +68,7 @@ public class BundleSelectorResolver {
 		};
 
 	final boolean						testUnresolved;
-	final Map<Long, BundleDescriptor>	bundleMap	= new HashMap<>();
+	final Map<Long, BundleDescriptor>	bundleMap			= new HashMap<>();
 
 	private String dump(TestDescriptor t, String indent) {
 		return indent + t + "\n" + t.getChildren()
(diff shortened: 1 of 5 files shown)


kriegfrj:
I've also included part of the fix for #3882 in this push as the two were closely linked, it was easier to do them in one push.

The fix for #3882 still doesn't handle the cases where class selectors and/or method selectors are used to select classes that don't resolve. This proved slightly more difficult so I thought to defer them to a later commit.

biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java
```diff
@@ -389,7 +395,8 @@ private void markClassesResolved(Bundle bundle) {
 						}
 						return selectMethod(testClass, DiscoverySelectors.selectMethod(testcase));
 					}
-				} catch (ClassNotFoundException cnfe) {
+				} catch (ClassNotFoundException | NoClassDefFoundError cnfe) {
```
bjhargrave, line 398: Putting `NoClassDefFoundError` here can hide our own stupidity in other parts of this code :-( Perhaps we should narrow the window for catching `NoClassDefFoundError` to only the calls to testClass.getDeclaredMethods() and	testClass.getDeclaredFields() since I think that is the only source for which we need to worry about it?

kriegfrj: Fair point. Note that `loadClass()` can throw `NoClassDefFoundError` too if the superclass doesn't resolve; see here: 
https://github.com/bndtools/bnd/blob/5693958978d52d83cf847a649af841cede89cc00/biz.aQute.tester.test/test/aQute/tester/bundle/engine/test/BundleEngineTest.java#L1227-L1244

So I have included the `loadClass()` in the scope of the catch for `NoClassDefFoundError` too.

bjhargrave:
Thanks!

PR #3974, opened 2020-04-27
[junit-platform] Check for unresolved classes when selectors used

kriegfrj:
Contains the remainder of the fix for #3882 - to handle the cases where method selectors and/or class selectors are used.


bjhargrave:
Thanks for all your work on the test support! Much appreciated.

PR #3988, opened 2020-04-29
memoize: A null consumer to accept should not trigger memoizing

bjhargrave:


aQute.libg/src/aQute/lib/memoize/CloseableMemoizingSupplier.java
```diff
@@ -107,6 +107,7 @@ public void close() throws Exception {
 
 	@Override
 	public CloseableMemoize<T> accept(Consumer<? super T> consumer) {
+		requireNonNull(consumer);
```
pkriens, line 110: Why not just return? 

We should be nice in accepting nulls but rather strict in never returning nulls. Now you might force some to have to check.

bjhargrave: I think it is a programming error to pass a null consumer. It may be ok to accept null for a collection argument as if it were an empty collection. But a consumer is code to execute.

pkriens: What is the rationale? Someone needs to check you or any of your N callers. There is a perfect case to make that null means 'do not call me'. By having such a default you just significantly reduce the error space of this function, i.e. you make it a _complete_ function, without any cost on your side and potential savings on the caller side.

bjhargrave: If is the general contract of methods taking a Consumer, that the Consumer be non-null. 

For example from `java.util.function.Consumer` itself:

`https://github.com/openjdk/jdk/blob/478773c10278d86a4b81d2ad5382f7314cee4c17/src/java.base/share/classes/java/util/function/Consumer.java#L63-L66

This code already required that. I just move the check earlier to avoid memoizing when you would NPE anyway later in the code.

By allowing null to mean no-op you avoid catching programming errors early.

I think it is a bad idea to allow null here as it hides programming errors.

pkriens: I know it is a common pattern in Java, but sadly so are checked exceptions. I just do not like the extra error space you allow by having incomplete functions. The idea that you catch programming errors this way is a bit like hitting your head hard against a wall and then when you stop saying that it feels so good. You just force the error handling on the N callers.

But it is a matter of style and we do not have a style book.

PR #3992, opened 2020-04-30
[bndtools] Removed BundleClassCompletionProposalComputer relic

kriegfrj:
Came across https://github.com/bndtools/bnd/blob/efffde045662f2ca82b208b609996592f9987689/bndtools.builder/src/org/bndtools/builder/classpath/BundleClassCompletionProposalComputer.java

This class was part of commit https://github.com/bndtools/bndtools/commit/b914890656d21130a69eeb1220641031154b241a so it looks like @njbartlett had this idea from the beginning, but it doesn't appear to have been properly finished or even seriously touched since. We now have the QuickFixProcessor since #3833 and we have #3808 to track the future addition of Content Assist, so this can go now.


bjhargrave:
Thanks!

PR #4002, opened 2020-05-02
fix an npe and only refresh when changed

rotty3000:


bndtools.m2e/src/bndtools/m2e/MavenDependenciesRepository.java
```diff
@@ -24,37 +23,39 @@
...
-				logger.debug("{}: Collected files {} for project {}", getName(), files, projectFacade.getFullPath());
-
-				return files;
-			} catch (Exception e) {
-				logger.error("{}: Failed to collect files for project {}", getName(), projectFacade.getFullPath(), e);
-				return Collections.emptySet();
-			}
-		};
+	Optional<Supplier<Set<File>>> collect(IMavenProjectFacade projectFacade, IProgressMonitor monitor) {
+		logger.debug("{}: Collecting files for project {}", getName(), projectFacade.getProject());
```
bjhargrave, line 40: This changes the code from doing work in the future (when the supplier is called) to doing work now and returning a simple supplier. I assume you mean this change to do the work now? If so, then just return a `Set<File>`. The caller can check the set for emptyness and do other things including making the simple supplier. Or a the least change to return `Optional<Set<File>>`. Having a supplier in the return implies it is actually doing something interesting in the future. But with this change, all the interesting work has already been done.

bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
```diff
@@ -358,54 +365,53 @@ private File get(final String bsn, final Version version) throws Exception {
...
 	}
 
-	Supplier<Set<File>> collect(IMavenProjectFacade projectFacade, IProgressMonitor monitor) {
-		logger.debug("{}: Adding project {}", getName(), projectFacade.getProject());
-
-		return () -> {
-			try {
-				Set<File> files = new HashSet<>();
-				MavenProject mavenProject = projectFacade.getMavenProject(monitor);
+	Optional<Supplier<Set<File>>> collect(IMavenProjectFacade projectFacade, IProgressMonitor monitor) {
```
bjhargrave, line 373: Same comment here as above.

bndtools.m2e/src/bndtools/m2e/MavenDependenciesRepository.java
rotty3000: fixed

bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
rotty3000: fixed

bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
```diff
@@ -177,25 +177,33 @@ public boolean isEmpty() {
...
-					index(iProject, collect(projectFacade, monitor));
+					Set<File> collected = collect(projectFacade, monitor);
+					if (!collected.isEmpty()) {
+						index(projectFacade.getProject(), collected);
+						changed = true;
+					}
 					break;
 				case REMOVED :
-					remove(event.getOldMavenProject()
+					changed = remove(event
```
bjhargrave, line 196: Since you are in a loop here, you probably do not want to potentially set changed from true back to false. So you want something like:

```java
if (remove(event.getOldMavenProject().getProject())) {
  changed = true;
}
```

bjhargrave: Or `changed |= remove(...`.

rotty3000: I noticed this and was already fixing it :) I actually needed to fix this exact issue in a wider scope.

bndtools.m2e/src/bndtools/m2e/MavenBndrunContainer.java
```diff
@@ -41,10 +44,19 @@
 	private static final ProjectRegistryManager			projectManager			= MavenPluginActivator.getDefault()
 		.getMavenProjectManagerImpl();
 
-	public BndrunContainer getBndrunContainer(IMavenProjectFacade projectFacade, MojoExecution mojoExecution,
+	private static final MavenRunListenerHelper			helper					= new MavenRunListenerHelper() {};
```
bjhargrave, line 47: If this class already implements `MavenRunListenerHelper`, why does it need another `helper` instance? Is it because you need to call from static methods? Perhaps this class should no longer implement `MavenRunListenerHelper`?

rotty3000: It's because the methods are static. Do note that I'm cleaning all this up in a future PR.

PR #4003, opened 2020-05-02
filter: Use improved filter impl with caching of converted values

bjhargrave:
The values from the filter strings are parsed into strings. But they
often need to be converted to typed values for the actual comparison
with the input maps. This change caches the most recent conversion since
it is highly likely that the value types are highly consistent over the
various input maps. So we can save a lot of conversion of value to type.

This can be particularly beneficial for bnd.hashes which are all
numerical and we can be searching for a specific hash value across all
the indexed resources of which can have many package capabilities each
with many hash values with which to compare. It is also very useful
for version attributes which are generally always of type Version.

This is the filter impl I wrote for OSGi R8.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/FilterImpl.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/FilterImpl.java
@@ -1,0 +1,1388 @@
+/*
+ * Copyright (c) OSGi Alliance (2005, 2020). All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


pkriens:
Did you benchmark this? A filter is matched only once during a resolve, so caching inside the filter itself should not show any performance gain for resolve? The lookup for the hashes seems to always require a new filter, so also there a cache inside the filter seems unnecessary?

bjhargrave:
The value in the filter string is used many times. For example, in the `classIndexTest` method, the filter string `(bnd.hashes=-1919773011)` is used to search for a hash value. The string `-1919773011` must be converted to a long to search the `List<Long>` values in the resource's capabilities. I instrumented the filter class to see how many times we need the long value of the string `-1919773011`. The search needs the value 9761 times. The first time it is converted to a Long and the remaining 9760 times, we can use the cached Long. So in this one test case, we avoid parsing a string into a long 9760 times! So even if you never cache the Filter object, you can save a lot of string parsing. If you also cache the Filter objects, the savings grow.

image

We will see similar savings by avoiding repeated parsing of a version string into a Version object when searching all resources for a package capability within some version range.

pkriens:
This is a very special case because we have long lists. First, the filter could easily have taken advantage of a single conversion for each collections/array comparison, that would have reduced it to low hundreds (nr of exported packages in the repos).But I guess it is nice to have the same filter as the OSGi now.

PR #4010, opened 2020-05-05
Filter matches from bnd generated jars in Java Search queries

gamerson:


pkriens:
I want to test this out so I am merging this.

Issue #3899, opened 2020-04-09
[bndtools] Automatically add source bundles when debugging


kriegfrj:
The debugger seems to the sources are embedded in the main bundle under `OSGI-OPT/src` (as is the case with most bundles built by Bnd).

However, the Eclipse bundles use the "separate bundle for the source" paradigm. I'm trying to implement #3881 and getting a little frustrated by the fact that the debugger can't find them automatically. This is compounded by the fact that Bnd doesn't even automatically download them for you into the Eclipse cache merely by being on `-runbundles`. So to get it to work, I seem to need to:

* add the bundle to `-buildpath` and open the source (force
* run in debug mode
* do "add sources" in the debugger and manually browse to bnd's Eclipse cache folder to find the source bundle that I need and add it.

I think that all of this is something that `AbstractOSGILaunchDelegate` should be able to handle on behalf of the user when it detects that it has been launched in debug mode.

pkriens:
I remember Neil was searching for <bsn>.sources and attached them as source. I also know I implemented this for the maven repos.

kriegfrj:
Yes, it seems to work for the Maven-sourced dependencies. But for BuildpathQuickFixProcessorTest I'm using Eclipse (p2)-sourced dependencies.

And even with the .source bundles, it seems to be able to find them if you have the bundle on the build path. But often you have bundles in `-runbundles` that are not on the `-buildpath`, and it seems the debugger is not finding those.

Antother possibility is that when BJ builds the OSGi Repository version of Eclipse 2018-12, he could package all of the source into the OSGI-OPT/src folder of the main bundle. I think that would work, but it would be a fix for our specific scenario. It'd be nice to have something a bit more generic.

pkriens:
+1 to the idea of adding sources of all -runbundles in the debug launching. As far as I can see, that should be pretty straightforward? That would actually be great.

kriegfrj:
I've had an initial look into this, as the lack of debug sources is really causing me problems when trying to debug the quick fix integration tests. Reporting my findings for feedback, and also for my future self in case I have to move off onto something else.

[...]

PR #4011, opened 2020-05-06
[bndtools debug] Overhaul debug source path generation

### LINKED ISSUE #3993: [bndtools debug] BndDependencySourceContainer to fully replace DefaultSourceContainer
`BndDependencySourceLookupDirector` is what populates the "source" tab in the debug launch configuration.

Currently, it adds two source containers - `DefaultSourceContainer` and `BndDependencySourceContainer`, in that order. The `DefaultSourceContainer` is the one used by the JDT core for Java projects. I am convinced that we should ideally not be using `DefaultSourceContainer` at all, as I will explain.

The JDT world assumes that the build path is identical to (or at the very least, a proper subset of) the run path. `DefaultSourceContainer` follows this assumption. Basically, it puts all of the build-time sources on the runtime source lookup path, including:

-The JRE you are building against
-The sources of any jars/libraries/projects in your dependencies
-The source of the project itself.

However, Bndtools launches are different - the run time dependencies are generally different (in fact, potentially completely different) from the build time ones. Some examples (both commonplace and more theoretical):
1. It is common (indeed, recommended by the good folk here) to build against osgi.core and then run against a particular implementation (eg, Felix or Equinox).
2. It is common to build against a lower version and then run against a higher version.
3. Using -runee it is possible to set a different JVM for runtime than what you used to build against.
4. It is even possible to create a bnd(run) launch within a project that doesn't actually include that project at all.

Having the wrong sources on the source lookup path at best uses extra computer resources and slows things down. At worst it makes debugging more difficult because it shows the wrong source. `DefaultSourceContainer` contributes to this by adding lots of the wrong sources.

At the very least, `DefaultSourceContainer` needs to be prioritized after the Bnd-specific dependencies. This would reduce the incidence of issues where the runtime versions are different (eg, the debugger will use the Felix source instead of the source from osgi.core). But ultimately, I think it would be best if `BndDependencySourceContainer` did a proper job of finding the sources for the *runtime* dependencies (including for the actual JVM that the `AbstractOSGiLaunchDelegate` chose to launch with, as opposed to the one that was built against), which would make `DefaultSourceContainer` redundant.

kriegfrj:
This commit significantly changes the layout of the source dependencies in the debug configuration window. I would be interested in hearing people's opinions and having them try it out to see if I've broken anything. See attached screenshot for an example:
BndDependencySourceContainer
Noteable features:

* `DefaultSourceContainer` is no longer used.
* `BndDependencySourceContainer` now contains a few intermediate nodes in its hierarchy to group the dependencies in a way that corresponds to the bnd(run) file - ie, separate entries for `runee`, `runfw`, `runpath` and `runbundles`
* `runpath` dependencies are now added too
* Bundle dependencies displayed as bundles using the Bndtools bundle icon
* `runee` is handled directly by `BndDependencySourceContainer` and will add
the sources for the VM that is actually used for launching (as opposed
to the one that was used for compilation).

Fixes #3993


pkriens:
Wow, that looks eye watering! I wanna have it!

Is it ready to merge?

kriegfrj:
> Wow, that looks eye watering good! I wanna have it!
> 
> Is it ready to merge?

Well, one thing I've learned in this is that source lookup in Eclipse is very complicated. Either i have not fully understood the need for this complexity, or else it is needlessly so...

Given the complexity, there's a chance i may have broken it in some obscure way (breakpint handling is one area to be wary of). So I'd feel uncomfortable about it finding its way in to 5.1 at this late stage, but I'd be fairly comfortable about inflicting it on us who can handle the bleeding edge so we can put it through its paces before the next release.

kriegfrj:
The latest commit also fixes #3899 (downloading sources automatically). It was easier to add it to this PR as it depends on the new features implemented there.

pkriens:
I think the current source code lookup is broken enough to not worry about some obscure bugs in this code. If we merge it today then we have at least 2 weeks to try it out. having source lookup of -runbundles would be an awesome feature to have in 5.1.

I am ok to merge this into 5.1.0 unless BJ explodes :-)

bjhargrave:
> I am ok to merge this into 5.1.0 unless BJ explodes :-)

This code is way out of my area of knowledge! But lets give it a go...

bjhargrave:
Thanks again for all your contributions, @kriegfrj!

kriegfrj:
> This code is way out of my area of knowledge! But lets give it a go...

And here I was as a novice hoping that the experts would review it for me... 😆 Fingers crossed then...

pkriens:
I am afraid, Fr. Krieg, that you've become the expert. And I got the distinct feeling there are not that many others in the world :-)

Issue #4012, opened 2020-05-06
[repository] MavenBndRepository source download implementation broken?


kriegfrj:
I'm trying to implement #3899, and I've bumped into this piece of code:

https://github.com/bndtools/bnd/blob/28dbfe50374b9c35369fd173ee37ec184e0e632a/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java#L916-L937

To me, this looks a bit broken. In particular, the archive referenced in lines 930 and 934 should be `sources` and not `primaryArchive` (which seems to be what `BndPomRepository` does). Otherwise, you get the binary jar returned instead of the source jar.

Either this is broken or there's some magic happening here that I don't understand. As the original author perhaps @pkriens could weigh in on this?

bjhargrave:
Fixed by #4014

PR #4014, opened 2020-05-06
[sources] Father Krieg noticed in #4012 sources were not handled corr…

pkriens:
…ectly

The following lines found the sources but then used the original archive.

https://github.com/bndtools/bnd/blob/28dbfe50374b9c35369fd173ee37ec184e0e632a/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java#L916-L937

I am sure I tested it by hand but I guess I got convinced that it worked because of the other mechanisms we use for sources. So I now added a real test.


kriegfrj:
> I am sure I tested it by hand but I guess I got convinced that it worked because of the other mechanisms we use for sources. So I now added a real test.
>

:+1: No substitute for a real test. :smile:

PR #4016, opened 2020-05-06
Better detection of maven dependencies

### LINKED ISSUE #4015: Better detection of maven dependencies
I have noticed that for some bundles which are published to our maven repository the dependencies in the pom.xml are incomplete. I debugged the cause and it seems that currently we try to determine the coordinates for an artifact based on the META-INF/maven/*/pom.properties file. This seems to work fine for most artifacts. There seems to be a couple of a maven publishes libraries which do not contain these pom.properties files - therefore we can't determine the maven coordinates and the dependency does not land in the dependencies xml tag of the pom.xml file.

Because we use the bnd workspace model with a MavenBndRepository I thought that the repository knows every maven coordinate of every dependency otherwise it would no be able to download it. Therefore my proposal to solve that problem is to let the repository optional provide some metadata information about an artifact which can then be used by the ProjectBuilder to determine the coordinates.

swimmesberger:
Feedback welcome

Fixes #4015

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Container.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Container.java
@@ -41,6 +41,7 @@ public class Container {
 	private final Project					project;
 	private volatile DownloadBlocker		db;
 	private volatile Map<String, String>	attributes;
+	private volatile Map<String, String>	metadata;
 	private long							manifestTime;
 	private Manifest						manifest;
 	private volatile File[]					bundleClasspathExpansion;
@@ -63,6 +64,7 @@ public class Container {
 			this.bundleClasspathExpansion = new File[0];
 		}
 		this.attributes = attributes;
+		this.metadata = Collections.emptyMap();
 		this.db = db;
 
 	}
@@ -205,6 +207,24 @@ public class Container {
 		if (attributes == Collections.<String, String> emptyMap())
 			attributes = new HashMap<>(1);
 		attributes.put(name, value);
+	}
+
+	public Map<String, String> getMetadata() {
+		return attributes;
+	}
+
+	public synchronized void putMetadata(String name, String value) {
+		if (metadata == Collections.<String, String> emptyMap())
+			metadata = new HashMap<>(1);
+		metadata.put(name, value);
+	}
+
+	public synchronized void putMetadata(Map<String, String> data) {
+		if (data == null)
+			return;
+		for (Map.Entry<String, String> entry : data.entrySet()) {
+			putMetadata(entry.getKey(), entry.getValue());
+		}
 	}
 
 	/**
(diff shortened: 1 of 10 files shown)


bjhargrave:
This seems like a huge new API addition and bleeds maven-ness throughout parts of the code which really shouldn't care.

If we go back to the root cause, we have code in ProjectBuilder which is trying to compute the maven coordinates for an artifact. It only tries to read pom.properties. It could also try other things like read pom.xml. It could parse the file name (if the name in inside MAVEN_REPO_LOCAL file space). While these may not be perfect, they will address most cases and don't require such a broad change across the code adding maven-ness in many places.

swimmesberger:
I tried not to make the change to specific to maven the metadata could be used for others types of things too. The repository has more information about a file than simply the path to the file on the FileSystem. With this API we have the possibility to get that information across from the repository to the project/builder.

More importantly we can finally use the maven information which the repository knows better than anyone else at hand to use.

pkriens:
Metadata in bnd should be OSGi metadata. I agree that we should be extremely careful not to shine maven metadata true.

What is the problem we're trying to solve here?

swimmesberger:
@pkriens The main issue is that the ProjectBuilder/Project does not know the correct maven coordinates for buildpath dependencies. These coordinates are needed for the pom.xml file generation. Currently the coordinates are "detected" in the ProjectBuilder by looking at the jar file content of each dependency and looking for the pom.properties file in the META-INF/maven zip entry. This works for I would say 90% of the cases because most maven publishes artifacts contains such files.

The issue is for the other 10% of cases - here we have to find a better way to get those maven coordinates.

My proposal was to get the repository to communicate the coordinates of the dependencies. The repository implementation knows best where to find the maven coordinates because it's also responsible for retrieving the file which requires the coordinates. This would in my opinion be the most solid implementation because when the repository is wrong the dependency can't be retrieved therefore a feedback loop to the user exists. When the repository can find the dependency we also know the maven coordinates.

I agree that adding additional API for that is a little bit invasive but I haven't found a better to communicate that information across those borders. There might be simpler solutions to retrieve the coordinates in the ProjectBuilder but that would again induce some knowledge about maven in the ProjectBuilder (like the knowledge when to find the pom.properties in the jar) we would add some knowledge where the local maven directory is and how the directories are structure. That is information that the repository implementation has already. Such solutions would be fine for me because it would probably work but it wouldn't be the "most solid" one.

bjhargrave:
I have an idea on how to accomplish this with less change required. I'll need some time to vet my idea and propose a PR.

bjhargrave:
See https://github.com/bndtools/bnd/pull/4020 for, what I think, is a better solution.

bjhargrave:
Closing since #4020 supersedes.

PR #4020, opened 2020-05-06
Better computation of maven dependencies

bjhargrave:
We provide a means for maven repo types to supply maven attributes about artifacts through DownloadListeners into Containers which can then be examined by the maven dependency computation logic. pom.properties processing is the fall back for when artifacts come from non-maven repos such as the workspace repo.

This is done with limited API changes.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
@@ -136,36 +136,57 @@ public class ProjectBuilder extends Builder {
 		Jar jar = new Jar(file);
 		super.addClasspath(jar);
 		project.unreferencedClasspathEntries.put(jar.getName(), c);
-		if ((dependencies != null) && !Boolean.parseBoolean(c.getAttributes()
-			.getOrDefault("maven-optional", "false"))) {
-			jar.getResources(pomPropertiesFilter)
-				.forEachOrdered(r -> {
-					UTF8Properties pomProperties = new UTF8Properties();
-					try (InputStream in = r.openInputStream()) {
-						pomProperties.load(in);
-					} catch (Exception e) {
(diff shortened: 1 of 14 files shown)


biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
```diff
@@ -136,36 +136,57 @@ private void addClasspath(Parameters dependencies, Container c) throws IOExcepti
...
-						String key = new StringBuilder().append(depGroupId)
-							.append(':')
-							.append(depArtifactId)
-							.append(':')
-							.append(depVersion)
-							.toString();
-						dependencies.add(key, attrs);
-					}
-				});
+		Map<String, String> containerAttributes = c.getAttributes();
```
pkriens, line 139: This solution is a lot better! 

However, could we somehow delegate this very mavenish code to not be in the build package? Maybe a plugin? At least make all this maven support go to one place?

Nice work!

bjhargrave: It may be possible but we need to collect the maven dependency information at the time we make the classpath here in ProjectBuilder. Any later and we have lost the necessary information (the Containers). This is why the code generates the -maven-dependencies instruction so that the information collected here, when we have the information, can be communicated to the future when PomResource is generating the pom. By the time you get to PomResource, you have lost all information on the class path.

The -maven-dependencies instruction also allows the user to fix/replace the maven dependency information in their bnd file.

For now, I think this is fine for 5.1. We can always try and make it better in the next release.

swimmesberger:
Looks good! Thanks @bjhargrave for building upon the idea about communicating the information from the repository and creating a PR that has much less impact and the APIs of bnd. I'm glad that we have a much more solid foundation for maven coordinates to build upon.

PR #4038, opened 2020-05-10
Performance fixes

bjhargrave:
I profiled the build of the org.osgi.test.cases.framework project in OSGi which is a large projects and makes a bundle with hundreds (>260) nested bundles. Building the jar for this project has always been slow so I decided to dive deeper. 

I found 2 main hot spots for which fixes are in this PR. Any times mentioned in the commits were while profiling with YourKit and aren't precise. However, I did some more measurements on the OSGi build to compare 5.1.0-SNAPSHOT (pre-fix) against the two fixes in this PR (post-fix).

A parallel, incremental build with pre-fix took 15:06 on my system. With post-fix it takes 1:55 which is a major reduction in time. Even though this was an incremental build, because the Gradle plugin changed, the build reran all the jar tasks. So the jar tasks are running must faster post-fix.

Once fully built, a no-parallel, incremental build of just the org.osgi.test.cases.framework.jar takes 9s (which is the time for gradle to check all the dependencies are up-to-date). This time is the same with both pre-fix and post-fix code.

Then I deleted the jar from the generated folder and ran a no-parallel, incremental build to remake the jar. With pre-fix, this takes 3:42. With post-fix, this now takes 0:22. Also a major reduction in time.

Overall both of these fixes make some pretty impressive gains in speeding up the OSGi build. I suspect most of the improvement for the OSGi build comes from the Builder change. Any large build also using system should see a large build speed up.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
@@ -1,5 +1,6 @@
 package aQute.bnd.osgi;
 
+import static aQute.lib.exceptions.FunctionWithException.asFunction;
 import static java.util.stream.Collectors.toList;
 
 import java.io.File;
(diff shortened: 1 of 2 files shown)


rotty3000:
I just tested this and the build is definitely much much faster. Impressive @bjhargrave!

pkriens:
Wow! Nice work, this lockdown is really good for bndtools :-)

PK

Issue #4029, opened 2020-05-07
[bndtools debug] Add proper OSGi-aware runtime source lookup


kriegfrj:
In the wake of #4011, the source lookup in Bndtools is a lot better but it is still not 100% because it is built on the basic built-in Eclipse JDT Java source lookup. This source lookup assumes a flat classpath and is not classloader aware, eg:

1. The `ISourceContainers` are asked to search for the source for a class by name only, with no hint as to which bundle it might belong to.
2. The `CompositeSourceContainer` (which is used in #4011) simply searches its children sequentially for the first match - or, if you've allowed duplicate sources, it will return all the matches and then allow the user to select one.

So if you have multiple versions of a class in `-runbundles` in different bundles, you are likely to run into issues with source selection.

The solution to this is for `BndDependencySourceLookupDirector` to override the default behaviour. As far as I can understand, the debug source lookup is in three parts in Eclipse:

1. A class that maps from a class from Eclipse's debug model (eg, IJavaStackFrame) to a String that the source lookup containers (`ISourceContainer`) will understand. This is the `ISourceLookupParticipant`
2. The container classes that take this String and find the corresponding source file.
3. The `ISourceLookupDirector` that coordinates all of the above.

For completeness, there is also `ISourcePathLocator` (which is legacy and a superinterface of `ISourceLookupDirector`), `ISourcePathComputer` (which I'm not 100% sure what it does but I think it is used by the `DefaultSourceContainer` to generate the `ISourceContainer` instances). And there is `IRuntimeClasspathEntry`, which seems to be used as an intermediary to compute the source containers. I don't think we need to worry about `ISourcePathComputer` or `ISourcePathComputer` as yet (though a deeper understanding might change that opinion) - we can do everything we need in `BndDependencySourceLookupDirector`. To see how this could be implemented, we can use `PDESourceLookupDirector` (https://github.com/eclipse/eclipse.pde.ui/blob/master/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/sourcelookup/PDESourceLookupDirector.java) as a reference. 

[...]

PR #4051, opened 2020-05-13
[classpath] Add support for dependencies with Bundle-ClassPath

### LINKED ISSUE #2415: expand-bcp does nothing, contributeFiles() never called
This is a follow-up of bndtools/bndtools#1871.

Following @bjhargrave's useful tip-off, I started looking at the `expand-bcp` option as a way of supporting bundles on the build path that have embedded jars using the `Bundle-ClassPath` feature of OSGi (implemented in d8c177a055bf051f83acb6e0355d5f280e80d1bb). However, it didn't seem to work (or in fact, to do anything).

After putting in some debug tracing, I determined that my `expand-bcp` parameter is being parsed, but that the setting is simply being ignored. Further, after digging through the code, it seems to me that there may be some code in there which has been partially orphaned.

The unit test for `expand-bcp` verifies that `Container.contributeFiles()` does the right thing when expand-bcp is set. However, it seems that `contributeFiles()` is never called when Bnd assembles the classpath in preparation for the build. Using the Eclipse hierarchy browser, I can see that `contributeFiles()` is only called:

1. recursively, by itself (for handling libraries);
2. by `SubsystemExporter.getBundles()`, which is in turn called by (and only by) `SubsystemExporter.export()`, however it is only called on containers of type `LIBRARY`;
3. by `Project.toFile()`.

`Project.toFile()` looks like it will do the right thing, but it is an orphaned method. It is package-private but is not called anywhere from within Bnd.

Also, though I don't fully understand the purpose or operation of `SubsystemExporter`, by my superficial understanding it seems that this implementation of `expand-bcp` within `contributeFiles()` will actually do the _wrong_ thing in the context of `SubsystemExporter`, as it will incorrectly list the individual embedded jars as separate bundle dependencies in the generated manifest. This bug will only trigger if `expand-bcp` is used in conjunction with a bundle using Bundle-ClassPath and that bundle is part of a library on the build path (if this configuration is even possible).

kriegfrj:
I know that we're not fans of using `Bundle-ClassPath` in our own bundles. But there are lots of bundles out there that (for better or worse) do. Sometimes it is nice to use them. The most recent example I hit was in trying to use the Eclipse JDT Debug plugin for #4029.

With this fix, if you add the `expand-bcp=true` attribute to a bundle on your `-buildpath`, then Bndtools will extract any embedded jars in its `Bundle-ClassPath` and add them to the compiler class path. The Gradle workspace plugin also supports this syntax to maintain build fidelity. Leveraging the `expand-bcp` attribute also makes it an opt-in feature, which (hopefully) makes it low risk.

I'm open to suggestions on better ways to implement this. In particular, the current implementation adds the BCP dependencies as top-level jars under Bnd Bundle Path container - I think it would be a nice bit of visual gloss if they were listed hierarchically under the parent bundle. But this seems to work so I'm not complaining for now. It also doesn't support the full gamut of `Bundle-ClassPath` usage (as the BCP entry could refer to a directory in the bundle, rather than an embedded jar).

If this functionality is needed in the Maven and Gradle non-workspace plugins, they will need their own specific implementations as bnd is not directly involved in the build dependency selection procedure in those environments.

Fixes #2415. Fixes bndtools/bndtools#1871.


pkriens:
Wouldn't it be simpler to do this on the bnd side? We already have the convention with embedded sources to create a temporary jar with the sources and prefix it with '+'. We could easily create a jar that is flattened properly in the same way.

That way all drivers will use the same model instead of having to implement support in different places?

pkriens:
BTW, I like the idea. It has always been bad that we could not compile these bundles.

kriegfrj:
> Wouldn't it be simpler to do this on the bnd side?

I thought that's how I implemented it? It's bnd that processes the `extract-bcp` parameter and extracts the jars.

> We already have the convention with embedded sources to create a temporary jar with the sources and prefix it with '+'. We could easily create a jar that is flattened properly in the same way.
>
> That way all drivers will use the same model instead of having to implement support in different places?

I thought the "+" trick was only something that `MavenBndRepository` did? If there is a common code path that all repos go through to give them the chance to normalise the bundles, that might work - but does the "+" trick work in the Gradle-non-workspace/Maven context? I had assumed not, but maybe I was wrong...

pkriens:
You're right, it are currently the maven repos doing this because it is manual. However, all bundles will become containers before they see the driver. We could implement the functionality in the container creation.

Preferably, the driver should be oblivious of this.

bjhargrave:
> Preferably, the driver should be oblivious of this.

Yes, This should all be handled in Container so that it does not bleed out everywhere.

bjhargrave:
@kriegfrj Any plans for this PR?

kriegfrj:
> @kriegfrj Any plans for this PR?

I think that, given the suggestion that was made for the alternative implementation, this depends on #4246, which I'd love to tackle but don't really have time for.

bjhargrave:
> I think that, given the suggestion that was made for the alternative implementation, this depends on #4246, which I'd love to tackle but don't really have time for.

OK. So should we close this PR for now?

kriegfrj:
Well, the PR works as-is. A better solution is no doubt possible. I guess the question is do you think that this solution is acceptable in the meantime? If not, then let's close it, if so let's merge it.

bjhargrave:
> do you think that this solution is acceptable in the meantime? If not, then let's close it, if so let's merge it.

It is an old draft PR that has not been built since May and failed the build for some reason. So I not going to review in its current state. I would suggest we close it.

PR #4053, opened 2020-05-13
add "open project" command handler

gamerson:
Use the `Quick Access` `Cmd+3` and then type in a project name (or partial name) and you should see a new "open project" command available.

This will try to open the BndtoolsExplorer first and then falls back to PackageExplorer.  If neither are open, it will open the BndtoolsExplorer.

openProject


pkriens:
LGTM, can we merge?

PR #4062, opened 2020-05-14
[BndtoolsExplorer] add build status toolbar item

gamerson:
buildStatus

@pkriens  This is just a start.  There are some edge cases and other cleanup, but I wanted to give you this to try out, as I believe you are also working on some changes to the Bndtools explorer with regards to the selection filtering.


pkriens:
This looks awesome! I only got nitpicks!

- Could we move the icon to the left, standalone? Although it has an action it is mostly a visual indicator. I really don't know why but it feels it is on the wrong side.
- There is quite a bit of space in that action bar. Could we add the Git branch that the workspace is checked out on? I realized that with our workspace model, all projects are on the same branch. So one can remove the branch per project, would make the list a lot more easy on the eyes.
- The OK icon, I think I would like it a large as possible square and exactly the same green as JUnit. Id like people to have an uneasy feeling when it is not green :-)

As said, these are nitpicks. I love it! This is will do great things for software engineering :-)

Please merge it asap so I can play with it.

BTW, love these animated GIFs. I am copying them to the bndtools.org site once they are live. E.g.:

   https://bndtools.org/manual/jareditor.html#find

bjhargrave:
>     * Could we add the Git branch

Do we really want to couple to the egit/jgit plugins? I would rather not and I don't think we should have code dorking around in the `.git/` folder :-) Egit/jgit can already decorate projects with the branch. It is also possible that a workspace could have multiple branches happening across different projects when one is using submodules or maven projects. Better to leave git decoration to jgit/egit.

rotty3000:
Also, no all workspace projects are in the same repo. I frequently have multiple projects from different repos in my Eclipse workspace and so each project can show it's git status individually.

gamerson:
> * Could we move the icon to the left, standalone? Although it has an action it is mostly a visual indicator. I really don't know why but it feels it is on the wrong side.

In my first draft I added the status icon to the "view menu" which means it will follow the other toolbar items.

If we move it to the "left" I would not prefer to put it on its own line but rather I'd want to put it to the left of the 'filter' part, like this:

Screen Shot 2020-05-18 at 11 54 22 AM


> * There is quite a bit of space in that action bar. Could we add the Git branch that the workspace is checked out on? I realized that with our workspace model, all projects are on the same branch. So one can remove the branch per project, would make the list a lot more easy on the eyes.

I think this was discussed elsewhere, so skipping this discussion for now.

> * The OK icon, I think I would like it a large as possible square and exactly the same green as JUnit. Id like people to have an uneasy feeling when it is not green :-)

The JUnit icon is only an overlay (8x8 pixels), so there isn't a full 16x16 or 128x128 (in case of MacRetinaDisplay) that JUnit uses.  However, we could use this one:
testok@2x. Maybe this is the one you were talking about the whole time :)

> * Not sure about the action for green? Maybe show some workspace statistics. I.e. total nr of projects, line of code, classes, etc. Notice that Bndtools Explorer represents a single directory workspace. We have a scope of _all_ projects. Eclipse lacks this since projects can be _anywhere_. The green icon, therefore, represents the OK state of the workspace. People tend to dislike the single directory workspace but here it can shine.

I'll remove my message dialog for now, and just disable the toolbar item for green.

> BTW, love these animated GIFs. I am copying them to the bndtools.org site once they are live. 

Sure! The program I'm using is called `LICECap` https://www.cockos.com/licecap/

PR #4064, opened 2020-05-17
bndtools: use LaunchUtils to create the bndrun used in command execut…

rotty3000:
…ions and support m2e

### DIFF
--- a/bndtools.core/src/bndtools/command/BndResolveHandler.java
+++ b/bndtools.core/src/bndtools/command/BndResolveHandler.java
@@ -1,7 +1,7 @@
 package bndtools.command;
 
-import java.io.File;
 
+import org.bndtools.api.RunMode;
 import org.eclipse.core.commands.AbstractHandler;
 import org.eclipse.core.commands.ExecutionEvent;
 import org.eclipse.core.commands.ExecutionException;
@@ -17,10 +17,10 @@ import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Status;
 
-import aQute.bnd.build.Project;
 import biz.aQute.resolve.Bndrun;
 import bndtools.Plugin;
 import bndtools.central.Central;
+import bndtools.launch.util.LaunchUtils;
 
 public class BndResolveHandler extends AbstractHandler {
 
@@ -37,13 +37,13 @@ public class BndResolveHandler extends AbstractHandler {
 			public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
 				IStatus status = Status.OK_STATUS;
 				try {
-					Bndrun bndrun = Bndrun.createBndrun(Central.getWorkspace(), new File(bndrunFile.getLocationURI()));
-					Project bndProject = Central.getProject(project);
-					bndrun.setBase(bndProject.getBase());
+					Bndrun bndrun = (Bndrun) LaunchUtils.createRun(bndrunFile, RunMode.LAUNCH);
 
 					bndrun.resolve(false, true);
 
-					bndrunFile.refreshLocal(IResource.DEPTH_ONE, monitor);
+					if (Central.isBndProject(project)) {
+						bndrunFile.refreshLocal(IResource.DEPTH_ONE, monitor);
+					}
 				} catch (Exception e) {
 					status = new Status(IStatus.ERROR, Plugin.PLUGIN_ID, "Unable to resolve jar", e);
 				}
--- a/bndtools.core/src/bndtools/command/BndrunFilesParameterValues.java
+++ b/bndtools.core/src/bndtools/command/BndrunFilesParameterValues.java
@@ -1,13 +1,14 @@
 package bndtools.command;
 
+import static java.util.stream.Collectors.toMap;
+
 import java.util.ArrayList;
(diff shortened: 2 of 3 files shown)


rotty3000:
This extends the utility of `<ctrl>+3` commands `bnd export/resolve` to m2e projects.

bndtools.core/src/bndtools/command/BndrunFilesParameterValues.java
```diff
@@ -33,25 +31,21 @@
 
 	@Override
 	public Map getParameterValues() {
-		Workspace ws = Central.getWorkspaceIfPresent();
+		if (bndrunFilesMap != null) {
```
bjhargrave, line 34: This seems like a perfect place for a memoizier to get lazy initialization and it will properly handle the synchronization for at-most-once initialization.

rotty3000: yeah we should, I'll try my hand at it.

rotty3000: check it now!

bndtools.core/src/bndtools/command/BndrunFilesParameterValues.java
```diff
@@ -21,44 +23,40 @@
...
+			Map<String, String> map = Arrays.stream(ResourcesPlugin.getWorkspace()
+				.getRoot()
+				.getProjects())
+				.filter(IProject::isOpen)
+				.map(this::findBndrunFiles)
+				.flatMap(Collection::stream)
+				.map(IResource::getFullPath)
+				.collect(toMap(IPath::toPortableString, IPath::toPortableString));
+
+			ResourcesPlugin.getWorkspace()
```
bjhargrave, line 50: Perhaps you don't want to add the listener in the supplier since you will get an additional listener every time reset is called and the new memoized is initialized. The each listener will further call reset...

I think you just need to add the listener in the constructor.

rotty3000: fixed

bjhargrave:
Is this something that should get cherry picked into next for 5.1?

bndtools.core/src/bndtools/command/BndResolveHandler.java
```diff
@@ -37,13 +37,13 @@ public Object execute(ExecutionEvent event) throws ExecutionException {
...
 				try {
-					Bndrun bndrun = Bndrun.createBndrun(Central.getWorkspace(), new File(bndrunFile.getLocationURI()));
-					Project bndProject = Central.getProject(project);
-					bndrun.setBase(bndProject.getBase());
+					Bndrun bndrun = (Bndrun) LaunchUtils.createRun(bndrunFile, RunMode.LAUNCH);
 
 					bndrun.resolve(false, true);
 
-					bndrunFile.refreshLocal(IResource.DEPTH_ONE, monitor);
+					if (Central.isBndProject(project)) {
```
gamerson, line 44: Hey @rotty3000   Thanks for wiring this up with Maven support! 

Why did you want to predicate this on being a bndProject.  Will that match maven projects?

bndtools.core/src/bndtools/command/BndrunFilesParameterValues.java
```diff
@@ -21,44 +22,37 @@
...
+import aQute.lib.memoize.Memoize;
 
 public class BndrunFilesParameterValues implements IParameterValues {
 
 	static final String BNDRUN_FILE = "bnd.command.bndrunFile";
 
-	private static Map<String, String> bndrunFilesMap;
-	private static IResourceChangeListener	listener	= new InvalidateMapListener();
+	private IResourceChangeListener					listener	= new InvalidateMapListener();
+	private volatile Supplier<Map<String, String>>	bndrunFilesMap;
```
gamerson, line 32: By changing this to non-static, every command that shared this ParameterValues map is going to recalculate.  So right now we have 2 commands but in the future we may want more.  I don't think this is a huge performance hit right now, but if we have N commands, then there will be N copies of this class and N copies of the bndrunFilesMap.  Do y'all think that will be an issue?  Maybe not since there will not be an abundance of bndfiles usually.

bndtools.core/src/bndtools/command/BndResolveHandler.java
rotty3000: Because in M2E these files are already being monitored and would result in an infinite loop if I added this other propagation.

bndtools.core/src/bndtools/command/BndrunFilesParameterValues.java
rotty3000: Got it! I can turn it back into static.

gamerson: Hey Ray, It seems the bndrunFilesMap didn't get switched back to static.  Change of plans?

gamerson:
> Is this something that should get cherry picked into next for 5.1?

I think so, seems like a safe thing to include.

bndtools.core/src/bndtools/command/BndrunFilesParameterValues.java
rotty3000: No sorry, will do momentarily!

PR #4066, opened 2020-05-17
maven repo: fix URL normalization to support glob server IDs

andrey-kuznetsov:
Rewrote `ConnectionSettings.normalize()` with plain regex: `java.net.URI` failed to parse expressions with wildcards; resulting string looked like "https://null".


bjhargrave:
I updated the commit with an improved regex and a test case. This fix needs to go into 5.1, so I wanted to make sure we had a test.

bjhargrave:
Thanks!

PR #4067, opened 2020-05-18
Copy.gavs

rotty3000:
Actions to copy GAVs from MavenBndRepository and Maven Dependencies/Workspace repositories:

copy gav-2020-05-17 21-57

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
@@ -365,7 +365,7 @@ public class BridgeRepository {
 		return null;
 	}
 
-	private static InfoCapability getInfo(Resource resource) {
+	public static InfoCapability getInfo(Resource resource) {
 		return ResourceUtils.capabilityStream(resource, BND_INFO,
 			InfoCapability.class)
 			.findFirst()
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
@@ -29,6 +29,7 @@ import org.osgi.util.promise.PromiseFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import aQute.bnd.maven.MavenCapability;
 import aQute.bnd.osgi.Constants;
 import aQute.bnd.osgi.Jar;
 import aQute.bnd.osgi.Macro;
@@ -331,6 +332,8 @@ class IndexFile {
 			BridgeRepository.addInformationCapability(rb, name, version.getOSGiVersion(), repo.toString(),
 				Constants.NOT_A_BUNDLE_S);
 		}
+		MavenCapability.addMavenCapability(rb, archive.revision.group, archive.revision.artifact,
+			archive.revision.version, archive.classifier, repo.toString());
 		Resource resource = rb.build();
 		return Collections.singletonMap(archive, resource);
 	}
@@ -458,6 +461,8 @@ class IndexFile {
 		String bsn = archive.getWithoutVersion();
 		MavenVersion version = archive.revision.version;
 		BridgeRepository.addInformationCapability(rb, bsn, version.getOSGiVersion(), repo.toString(), msg);
+		MavenCapability.addMavenCapability(rb, archive.revision.group, archive.revision.artifact,
+			archive.revision.version, archive.classifier, repo.toString());
 		return rb.build();
 	}
 
--- a/bndtools.core/src/bndtools/dnd/gav/GAVIPageListener.java
+++ b/bndtools.core/src/bndtools/dnd/gav/GAVIPageListener.java
@@ -1,0 +1,104 @@
+package bndtools.dnd.gav;
+
+import static org.eclipse.swt.dnd.DND.DROP_COPY;
+
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.swt.custom.StyledText;
+import org.eclipse.swt.dnd.TextTransfer;
+import org.eclipse.swt.dnd.Transfer;
+import org.eclipse.swt.widgets.Control;
+import org.eclipse.ui.IEditorInput;
+import org.eclipse.ui.IFileEditorInput;
+import org.eclipse.ui.IPartListener;
+import org.eclipse.ui.IStorageEditorInput;
+import org.eclipse.ui.IWorkbenchPart;
+import org.eclipse.ui.dnd.IDragAndDropService;
+import org.eclipse.ui.part.EditorPart;
+
+import aQute.lib.exceptions.Exceptions;
+
+public class GAVIPageListener implements IPartListener {
+
+	private final Transfer[] transfers = new Transfer[] {
+		TextTransfer.getInstance()
+	};
+
+	@Override
+	public void partOpened(IWorkbenchPart part) {
+		IDragAndDropService dndService = part.getSite()
+			.getService(IDragAndDropService.class);
+
+		if (dndService == null) {
+			return;
+		}
+
+		Control control = part.getAdapter(Control.class);
+
+		if ((control == null) || !(control instanceof StyledText)) {
+			return;
+		}
+
+		IPath file = null;
+
+		if (part instanceof EditorPart) {
+			EditorPart editorPart = (EditorPart) part;
+
+			IEditorInput editorInput = editorPart.getEditorInput();
+
+			if (editorInput instanceof IFileEditorInput) {
+				IFileEditorInput input = (IFileEditorInput) editorInput;
+
+				file = input.getFile()
+					.getFullPath();
+			} else if (editorInput instanceof IStorageEditorInput) {
+				IStorageEditorInput input = (IStorageEditorInput) editorInput;
+
+				try {
+					file = input.getStorage()
+						.getFullPath();
+				} catch (CoreException e) {
+					throw Exceptions.duck(e);
+				}
+			}
+		}
+
+		if (file != null) {
+			String fileName = file.lastSegment();
+			String fileExtension = file.getFileExtension();
+
+			if ("pom.xml".equals(fileName)) {
+				dndService.addMergedDropTarget(control, DROP_COPY, transfers,
+					new MavenDropTargetListener((StyledText) control));
+			} else if ("gradle".equalsIgnoreCase(fileExtension)) {
+				dndService.addMergedDropTarget(control, DROP_COPY, transfers,
+					new GradleDropTargetListener((StyledText) control));
+			} else if ("bnd".equalsIgnoreCase(fileExtension) || "bndrun".equalsIgnoreCase(fileExtension)) {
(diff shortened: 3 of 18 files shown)


pkriens:
a) Isn't drag & drop easier?
b) Shouldn't you also support the bnd native format?

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
```diff
@@ -263,6 +289,17 @@ public Resource get(String bsn, Version version) throws Exception {
 		return resourceInfo.resource;
 	}
 
+	public ResourceInfo getInfo(String bsn) throws Exception {
+		Map<Version, ResourceInfo> map = index.get(bsn);
+		if (map == null)
+			return null;
+
+		return MapStream.of(map.entrySet())
```
bjhargrave, line 297: Since you only want the values, just do `map.values().stream()`

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -324,13 +324,11 @@ private boolean isMulti(String name) {
...
-		if (!hasIdentity) {
-			String name = archive.getWithoutVersion();
-			MavenVersion version = archive.revision.version;
-
-			BridgeRepository.addInformationCapability(rb, name, version.getOSGiVersion(), repo.toString(),
-				Constants.NOT_A_BUNDLE_S);
-		}
+		String name = archive.getWithoutVersion();
+		MavenVersion version = archive.revision.version;
+		BridgeRepository.addInformationCapability(rb, name, version.getOSGiVersion(), repo.toString(),
```
bjhargrave, line 329: Not sure this is wise. Did you check for any places which assume the presence of a bnd.info cap means the artifacts is NOT a bundle? You have changed things from a resource has either a osgi.identity or a bnd.info to where every resource has a bnd.info.

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -253,6 +258,14 @@ public static Workspace getWorkspace() throws Exception {
 					ws.addBasicPlugin(getEclipseWorkspaceRepository());
 					ws.addBasicPlugin(new JobProgress());
 
+					String tabulationChar = prefsService.getString("org.eclipse.jdt.core.prefs",
+						"org.eclipse.jdt.core.formatter.tabulation.char", "tab", null);
```
bjhargrave, line 262: Is this really the formatting properties to use for POMs?

bndtools.m2e/src/bndtools/m2e/MavenDependenciesRepository.java
```diff
@@ -43,9 +49,21 @@ public String getName() {
 		}
 
 		try {
-			MavenBndrunContainer mavenBndrunContainer = MavenBndrunContainer.getBndrunContainer(projectFacade, null, true,
-				true, allScopes, monitor);
-			mavenBndrunContainer.resolve()
+			MavenBndrunContainer mavenBndrunContainer = MavenBndrunContainer.getBndrunContainer(projectFacade, null,
+				true, true, allScopes, monitor);
+			MapStream.of(mavenBndrunContainer.resolve()
+				.entrySet())
```
bjhargrave, line 55: You don't need to say `.entrySet()`, MapStream.of also takes a map.

bndtools.m2e/src/bndtools/m2e/MavenFile.java
```diff
@@ -0,0 +1,299 @@
...
+import java.io.File;
+import java.io.FileFilter;
+import java.io.FilenameFilter;
+import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.nio.file.Path;
+
+@SuppressWarnings("serial")
+class MavenFile extends File {
```
bjhargrave, line 14: I would not extend File. That is super risky. I am sure it is not meant to be subclassed.

rotty3000:
> a) Isn't drag & drop easier?

Not really. Don't get me wrong, it would be cool, but it has limitations:
* can you do it from a menu? If not how to pick from multiple syntax choices?
* make the drop context sensitive? You need a drop handler for every editor, xml, m2e xml, buildship, groovy, egradle, etc.

I discussed D&D with @gamerson a couple weeks back and for all the possible drop points we need a new handler. With Copy&Paste anything that accepts text input is a valid target, plus you get to choose a syntax if multiple are available.

> b) Shouldn't you also support the bnd native format?

you mean the `-buildpath` format? sure I can add that easily.

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
rotty3000: :|

bndtools.core/src/bndtools/central/Central.java
rotty3000: fixed

bndtools.m2e/src/bndtools/m2e/MavenDependenciesRepository.java
rotty3000: fixed

bndtools.m2e/src/bndtools/m2e/MavenFile.java
rotty3000: fixed

rotty3000:
gonna revisit the d&d concept.

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
```diff
@@ -327,6 +369,16 @@ public static void addInformationCapability(ResourceBuilder rb, String name, Ver
 				c.addAttribute("error", error);
 			if (version != null)
 				c.addAttribute("version", version);
+			if (groupId != null)
+				c.addAttribute("groupId", groupId);
```
bjhargrave, line 373: In the work I did to get the maven coordinate info into Container, I defined all the maven attr names to start with `maven-`:

https://github.com/bndtools/bnd/blob/7a5b449d1469cedefc10ecd8c7ec6292e91e6def/biz.aQute.repository/src/aQute/maven/api/Program.java#L165-L166

https://github.com/bndtools/bnd/blob/7a5b449d1469cedefc10ecd8c7ec6292e91e6def/biz.aQute.repository/src/aQute/maven/api/Revision.java#L131

https://github.com/bndtools/bnd/blob/7a5b449d1469cedefc10ecd8c7ec6292e91e6def/biz.aQute.repository/src/aQute/maven/api/Archive.java#L248-L249

I think we should use the same attr names here. Then is very clear where the names come from.

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
```diff
@@ -318,6 +354,12 @@ public static void addInformationCapability(ResourceBuilder rb, String name, Str
 	public static void addInformationCapability(ResourceBuilder rb, String name, Version version, String from,
 		String error) {
 
+		addInformationCapability(rb, name, version, from, error, null, null, null, null, null);
+	}
+
+	public static void addInformationCapability(ResourceBuilder rb, String name, Version version, String from,
+		String error, String groupId, String artifactId, String mavenVersion, String extension, String classifier) {
```
bjhargrave, line 361: Instead of adding all these parameters for maven info, consider a map and passing `archive.attributes()`

https://github.com/bndtools/bnd/blob/7a5b449d1469cedefc10ecd8c7ec6292e91e6def/biz.aQute.repository/src/aQute/maven/api/Archive.java#L246

Then you can just add the map to the cap attributes.

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
rotty3000: Will this work via the proxy? Note that InfoCapability is created as a proxy, are you sure there is a mangling for those property names?

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
rotty3000: same question as above.

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
bjhargrave: Why not? You are adding attributes to a builder. You can do

```java
c.addAttributes(map);`
```

rotty3000: it's not the map usage I'm concerned with, but your recommendation to use maven-* naming. Will our proxy handle those? I guess we could fix that too I suppose...

rotty3000:
I've been playing around with SWT Drag&Drop, and it's a going to be painful to enable it the way we're thinking. There is no callback to the drag source about where it's being dropped as far as I can tell. I'm still digging but it appears that only the drop target can listen to drop events, which boils down to having to wire in a drop listener on every editor we'd want to support which would couple bndtools to those.

rotty3000:
@kriegfrj I've implemented the default of `version=snapshot` for the workspace repository (only affects dropping on bnd(run) files)

bndtools.core/src/bndtools/dnd/gav/GAVIPageListener.java
```diff
@@ -0,0 +1,112 @@
...
+	public void partOpened(IWorkbenchPart part) {
+		IDragAndDropService dndService = part.getSite()
+			.getService(IDragAndDropService.class);
+
+		if (dndService == null) {
+			return;
+		}
+
+		if (part instanceof BndEditor) {
+			part = ((BndEditor) part).getSourcePage();
```
gamerson, line 39: Perhaps we should modify the `BndEditor` so that you don't have to do this.  The `BndEditor` could override the `getAdapter` method and return the styledText where someone queries for `Control.class`

bndtools.core/src/bndtools/dnd/gav/GAVIPageListener.java
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	@Override
+	public void partClosed(IWorkbenchPart part) {
+		IDragAndDropService dndService = part.getSite()
+			.getService(IDragAndDropService.class);
+
+		IAdaptable adapter = part;
+
+		Control control = adapter.getAdapter(Control.class);
```
gamerson, line 96: BndEditor would need to add `getAdapter` support for Control.class for this to properly unmerge the dropTarget.

bndtools.core/src/bndtools/dnd/gav/GAVIPageListener.java
rotty3000: fixed!

bndtools.core/src/bndtools/dnd/gav/GAVIPageListener.java
rotty3000: fixed!

rotty3000:
@pkriens finally I did implement drag and drop as you suggested.

rotty3000:
I simplified one more time. I think this is phase 1 complete.

PR #4069, opened 2020-05-18
abstractindexingrepo: Refactor to generify File

bjhargrave:
This can allow subclasses to support indexing things other than File.


bjhargrave:
@rotty3000 I think this would make your #4067 PR nicer since you will not need to subclass File. MavenFile can wrap a File.

PR #4078, opened 2020-05-19
Allow to parameterize the basedir for bundles and bndruns.

### LINKED ISSUE #4077: bnd-resolver-maven-plugin: Allow to reference bndrun and bundle files outside ${project.basedir}
Currently looking up both `bndrun` files and `bundles` only works if they are located inside the `${project.basedir}` (https://github.com/bndtools/bnd/blob/19c7bf46b4ec2719fc4d4e72000cb495efd0dd35/maven/bnd-resolver-maven-plugin/src/main/java/aQute/bnd/maven/resolver/plugin/ResolverMojo.java#L106 and https://github.com/bndtools/bnd/blob/19c7bf46b4ec2719fc4d4e72000cb495efd0dd35/maven/bnd-resolver-maven-plugin/src/main/java/aQute/bnd/maven/resolver/plugin/ResolverMojo.java#L114). Sometimes it is desirable to reuse both for multiple Maven projects (and therefore reference them from outside the current Maven module's base directory).

Currently a config like this silently fails
```
<plugin>
                    <artifactId>bnd-resolver-maven-plugin</artifactId>
                    <version>5.0.1</version>
                    <executions>
                        <execution>
                            <goals>
                                <goal>resolve</goal>
                            </goals>
                            <phase>verify</phase>
                        </execution>
                    </executions>
                    <configuration>
                        <bndruns>
                            <bndrun>../target-osgi-environment/aem6.3.bndrun</bndrun>
                        </bndruns>
                        <bundles>
                            <includes>
                                <include>../target-osgi-environment/target/bundles/*.jar</include>
                            </includes>
                        </bundles>
```

kwin:
This closes #4077

Signed-off-by: Konrad Windszus <kwin@apache.org>

### DIFF
--- a/maven/bnd-resolver-maven-plugin/README.md
+++ b/maven/bnd-resolver-maven-plugin/README.md
@@ -65,10 +65,11 @@ An *implicit repository* containing the project artifact and project dependencie
 
 |Configuration Property       | Description |
 | ---                         | ---         |
-|`bndruns`                    | Can contain `bndrun` child elements naming a bndrun file to resolv. You can also specify `include` and `exclude` child elements using Ant-style globs to specify bndrun files. These are relative to the `${project.basedir}` directory. _Defaults to `<include>*.bndrun</include>`._|
+|`baseDir`                    | The base directory below which all referenced `bndruns` and `bundles` must be located. _Defaults to `${project.basedir}`._|
+|`bndruns`                    | Can contain `bndrun` child elements naming a bndrun file to resolve. You can also specify `include` and `exclude` child elements using Ant-style globs to specify bndrun files. These are relative to given `baseDir` directory. _Defaults to `<include>*.bndrun</include>`._|
 |`failOnChanges`              | Whether to fail the build if any change in the resolved `-runbundles` is discovered. _Defaults to `true`._|
 |`writeOnChanges`             | Whether to write the resolved run bundles back to the `-runbundles` property of the `bndrun` file. _Defaults to `true`._|
-|`bundles`                    | A collection of files to include in the *implicit repository*. Can contain `bundle` child elements specifying the path to a bundle. These can be absolute paths. You can also specify `include` and `exclude` child elements using Ant-style globs to specify bundles. These are relative to the `${project.basedir}` directory. _Defaults to dependencies in the scopes specified by the `scopes` property, plus the current artifact (if any and `useMavenDependencies` is `true`)._|
+|`bundles`                    | A collection of files to include in the *implicit repository*. Can contain `bundle` child elements specifying the path to a bundle. These can be absolute paths. You can also specify `include` and `exclude` child elements using Ant-style globs to specify bundles. These are relative to given `baseDir` directory. _Defaults to dependencies in the scopes specified by the `scopes` property, plus the current artifact (if any and `useMavenDependencies` is `true`)._|
 |`useMavenDependencies`       | If `true`, adds the project dependencies subject to `scopes` to the collection of files to include in the *implicit repository*. _Defaults to `true`._|
 |`reportOptional`             | If `true`, resolution failure reports will include optional requirements. _Defaults to `true`._|
 |`scopes`                     | Specify from which scopes to collect dependencies. _Defaults to `compile, runtime`._ Override with property `bnd.resolve.scopes`.|
--- a/maven/bnd-resolver-maven-plugin/src/main/java/aQute/bnd/maven/resolver/plugin/ResolverMojo.java
+++ b/maven/bnd-resolver-maven-plugin/src/main/java/aQute/bnd/maven/resolver/plugin/ResolverMojo.java
@@ -48,6 +48,9 @@ public class ResolverMojo extends AbstractMojo {
 	@Parameter(defaultValue = "${repositorySystemSession}", readonly = true, required = true)
 	private RepositorySystemSession								repositorySession;
 
+	@Parameter(defaultValue = "${project.basedir}", required = true)
+	private File												baseDir;
+
 	@Parameter
 	private Bndruns												bndruns	= new Bndruns();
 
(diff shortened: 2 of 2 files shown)


bjhargrave:
@rotty3000 do you have an opinion on this? In almost every place in Bnd, we required the bndrun files to be in the project folder. We use the project folder as the base for launching, etc. I am not sure how wise it is to support, in this one place, to allow bndrun files to come from elsewhere.

rotty3000:
I think the goal is to be able to place _bundles_ outside the project. @kwin can you confirm? (that's at least what I read from the code.) But if that is indeed the case I feel the variable name is confusing and should be called rather something like `bundleBaseDir`.

kwin:
Look at https://github.com/apache/jackrabbit-filevault/commit/6a1f74d21edcc9768c94392d47bcb3f9251364e0 how I use it. I really want to reuse a bndrun file among multiple different modules.

bjhargrave:
What you can do is put a bndrun file in each project and that bndrun file can `-include` a bndrun file from anywhere you want. This is how you can "share" a bndrun file across projects.

pkriens:
bndrun files are _independent_ of the project folder. They inherit _only_ from the workspace if not standalone. A bndrun is peer with  any project

rotty3000:
@pkriens we're talking about maven projects.

rotty3000:
@kwin before we add options can you try include as suggested by @bjhargrave ?

kwin:
`-include ../target-osgi-environment/somebndrun` works but I don't really get what speaks against adjusting the baseDir for bndrun files as well (maybe separate from the baseDir of bundles). In my case I use exactly the same bndrun file among several modules, so having the additional local bndrun with the same include feels like unnecessary overhead!

I can adjust the PR so that both baseDirs can be adjusted individually but I would like to hear what speaks against adjusting the baseDir of bndruns.

kwin:
@bjhargrave Thanks for the explanation but what is the workspace in case of the Maven Build? Just think about deeply nested multimodules like
```
root
+ module1
++ module1parta
+++ module1parta1
+++ module1parta2
+ module2
++ module2parta
...
```
Just taking the parent folder will not work here!

bjhargrave:
> but what is the workspace in case of the Maven Build?

A "standalone" workspace.

https://github.com/bndtools/bnd/blob/fcb0c366de7dd2e810ca392ac15093921c41f009/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java#L194

https://github.com/bndtools/bnd/blob/553ea60391b778fbbb100511966ef17812920738/biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java#L60-L61

kwin:
ok, I will rework the PR to only make only the bundle baseDir configurable for now.

bjhargrave:
> ok, I will rework the PR to only make only the bundle baseDir configurable for now.

But why? `<bundles>` already allows absolute files. So you can reference bundle files from anywhere already.

And if you were to allow a different base dir for `<bundles>`, we should support this in _all_ the bnd maven plugins which have a `<bundles>` configuration. And we will need tests to verify the function and prevent us from breaking it in the future.

kwin:
> But why? <bundles> already allows absolute files. So you can reference bundle files from anywhere already.

I cannot rely on absolute file names as those are different for every developer. The point is that the bundle files are always located beneath another Maven module and currently I cannot reference those as the relative files are only found below the current Maven project dir (https://github.com/bndtools/bnd/blob/fcb0c366de7dd2e810ca392ac15093921c41f009/maven/bnd-resolver-maven-plugin/src/main/java/aQute/bnd/maven/resolver/plugin/ResolverMojo.java#L106).

bjhargrave:
Your commit seems to show getting a bunch of dependencies in some folder. And then wanting to use that folder. Couldn't you use the `useMavenDependencies` and `includeDependencyManagement` configurations to build the necessary bundles for the implicit repository instead? Then you don't need some side project to do that.

I am no expert in the bnd maven plugins. Perhaps @rotty3000 can tell me I am all wrong here.

kwin:
I can't because for those projects the compile dependencies differ from the bundle dependencies, therefore I have to give the bundle versions for the bndrun container separately. But I want to do it only once (in the pom.xml for downloading the dependencies) not twice (also in addition for every reference), therefore I need to rely on wildcards in my bndrun.

rotty3000:
@bjhargrave I would guess the _bundles_ in question are binaries stored in the source repository.

That being the case, @kwin, why not index them and define an OSGiRepository in the included (shared) bndrun file?

```
-plugin.binary.bundles: \
    aQute.bnd.repository.osgi.OSGiRepository;\
    name="Binary Bundles";\
    poll_time=-1;\
    locations="${fileuri;${.}/../other/directory/index.xml}"
```

rotty3000:
so in the maven module that downloads the dependencies also use the bnd-indexer-maven-plugin and then use the above mechanism to access them in the bndrun

rotty3000:
as @bjhargrave points out you could even directly use:

`-standalone: ${fileuri;${.}/../other/directory/index.xml}`

rotty3000:
Just directly use the `bnd-indexer-maven-plugin` to index the project dependencies without manually downloading them. See the [index goal](https://github.com/bndtools/bnd/tree/master/maven/bnd-indexer-maven-plugin#building-indexes-with-index).
and pair that with `-standalone: ${fileuri;${.}/../other/directory/index.xml}`
This is exactly how enRoute works already :)

bjhargrave:
> Just directly use the `bnd-indexer-maven-plugin` to index the project dependencies without manually downloading them. See the [index goal](https://github.com/bndtools/bnd/tree/master/maven/bnd-indexer-maven-plugin#building-indexes-with-index).
> and pair that with `-standalone: ${fileuri;${.}/../other/directory/index.xml}`
> This is exactly how enRoute works already :)

Yes, this is the Bnd Maven Plugin way to do what you want to do and it is more correct. Using `<bundles>` will reindex the bundles for each bndrun execution which is wasteful since you can index them only once with the bnd-indexer-maven-plugin.

So I don't think we want to make a change for basedir here as there is a better way to do what you want to do.

kwin:
Thanks for the input, works with https://github.com/apache/jackrabbit-filevault/commit/b8e952da80cf3a3cabd8fda6f89d0ed1b797cfb5. I will come up with a documentation enhancement to make it clearer how to use those plugins.

bjhargrave:
I think we can close this PR.

> I will come up with a documentation enhancement to make it clearer how to use those plugins.

Please make another PR if you do any doc improvements.

PR #4081, opened 2020-05-19
Add a new Bndtools Java Model Dynamic working set that filters unwanted types

gamerson:
See commit message for full description, here is screencapture:

bndtoolsWorkingSet

With this dynamic working set you can add it to various places in the UI that query the default java model, you can enable the Bndtools Java Model working set and it will remove unwanted types.

### DIFF
--- a/bndtools.core/_plugin.xml
+++ b/bndtools.core/_plugin.xml
@@ -58,7 +58,7 @@
 		<startupParticipant
 			class="bndtools.central.Central" />
 		<startupParticipant
-			class="bndtools.javasearch.AddQueryListenerStartupParticipant" />
+			class="bndtools.javasearch.JavaSearchStartupParticipant" />
 	</extension>
 
    <!-- EDITORS -->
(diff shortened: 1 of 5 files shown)


bndtools.core/_plugin.xml
```diff
@@ -1199,4 +1199,13 @@
             id="bndtools.explorer.BndProjects">
       </filter>
    </extension>
+   <extension point="org.eclipse.ui.workingSets">
+      <workingSet
+           name="Bndtools Java Model"
+           description="Bndtools Java Model Working Set"
```
gamerson, line 1205: I'd like to hear from others what a good name for this working set should be.

gamerson:
I'm on a mission to remove these duplicate types from generated jars! :)

Last commit adds a generalized java element filter that can be applied to particular java lookup context.  In this screencap I'm filtering generated jars from the "open implementation" dialog.

bndtoolsFilter

gamerson:
I just found a deadlock with this code, where it restores the state of the workingsets, so this is gonna need some more work before it's ready.

PR #4084, opened 2020-05-20
[baseline] Lost baseline error

pkriens:
If you export a package form another bundle and that
package has a baseline error, then this error is lost.

The bndlib project exports a package aQute.bnd.annotation.plugin
from the biz.aQute.bnd.annotation project. 

I added a Capability header to the BndPlugin and then
suddenly the bndlib was no longer built, but no errors
were marked.

After debugging it became clear that biz.aQute.bnd.annotation
was not baselined but the aQute.bnd.annotation.plugin
package caused a baseline error in bndlib.

The BaselineErrorHandler could not find 
the package when it tried to create a marker. It
therefore returned zero markers. The MarkerSupport
did not take the case into account where the
Error Handler had zero markers after it indicated
that it had a handler and returned.

This fix will create a default error on bnd when
the handler creates no markers. 

Took some effort to run this in the debugger :-(

### DIFF
--- a/bndtools.builder/src/org/bndtools/builder/MarkerSupport.java
+++ b/bndtools.builder/src/org/bndtools/builder/MarkerSupport.java
@@ -126,34 +126,36 @@ class MarkerSupport {
 			BuildErrorDetailsHandler handler = BuildErrorDetailsHandlers.INSTANCE.findHandler(type);
 
 			List<MarkerData> markers = handler.generateMarkerData(project, model, location);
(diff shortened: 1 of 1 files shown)


pkriens:
We might want to consider this for the RC2?

bjhargrave:
> We might want to consider this for the RC2?

YES! I have seen this myself (but was too lazy to debug and fix :-/ )

PR #4093, opened 2020-05-22
[m2e] Ensure MavenWorkspaceRepository calls listener callbacks

### LINKED ISSUE #4092: [m2e] MavenWorkspaceRepository ignores DownloadListener parameter
Following up an issue I mentioned in #4089 and in the developers' Slack workspace.

While working on the osgi-test I was persistently hitting issues with trying to launch `org.osgi.test.junit5/test.bndrun` from within Bndtools - it seemed there was about a 50% chance that it would hang.

Using the debugger, I managed to find out that it was hanging in `DownloadBlocker.getFile()`:

https://github.com/bndtools/bnd/blob/b907eb38181892a2270785f4d88517b7a172b8cc/biz.aQute.bndlib/src/aQute/bnd/build/DownloadBlocker.java#L136

The corresponding `CountDownLatch.countDown()` wasn't being called.

Through judicious sprinkling of `System.err.println()` I managed to find out that it was a particular dependency in `-runbundles` that was causing the issue - `org.osgi.test.assertj`. I think that the fact that it is the only dependency (apart from the project itself) which resolves to another workspace project means that something different is happening. The fact that it was this also explains a couple of other things:

1. Why it only started happening recently (this dependency only recently became useful enough to add to the junit5 project)
2. Possibly also why it only happens intermittently - as the dependency is available in the MavenWorkspaceRepository and the MavenDependenciesRepository, it may resolve to one sometimes and the other on other occasions.

Digging deeper with more println()s, I managed to figure out that it was trying to call `MavenWorkspaceRepository.get()`:

https://github.com/bndtools/bnd/blob/238f6d148bd717926139ceb82d27ae8ee217c4a9/bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java#L119-L127

Note that this implementation completely ignores the `listeners` parameter. So any users of this repository which try to asynchronously initialize the `Container` object (as `Project.getBundle()` does here when called by `getRunbundles()`: https://github.com/bndtools/bnd/blob/b907eb38181892a2270785f4d88517b7a172b8cc/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L1363-L1364 ) will wait indefinitely on the call to `Container.getFile()`, causing all manner of issues throughout the system.

Fix should be simple (and will come shortly).

@bjhargrave, @rotty3000 - I think that this one might be a cherry-pick candidate for 5.1?

kriegfrj:
Fixes #4092

### DIFF
--- a/bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
+++ b/bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
@@ -121,7 +121,25 @@ public class MavenWorkspaceRepository extends
 		throws Exception {
 		File promise = get(bsn, version);
 		if (promise == null) {
+			if (listeners != null) {
+				for (DownloadListener dl : listeners) {
+					try {
+						dl.failure(null, "bundle " + bsn + ":" + version + " does not exist in this workspace");
+					} catch (Exception e) {
+						logger.warn("updating listener has error", e);
+					}
+				}
+			}
 			return null;
+		}
+		if (listeners != null) {
+			for (DownloadListener dl : listeners) {
+				try {
+					dl.success(promise);
+				} catch (Exception e) {
+					logger.warn("updating listener has error", e);
+				}
+			}
 		}
 		return promise;
 	}


rotty3000:
Since this is in reference to launching do you have a stacktrace that puts MavenWorkspaceRepository in the launcher code path? It shouldn't be there.

bjhargrave:
Good catch. If there are DownloadListeners, they _must_ be called!

rotty3000:
Sure, I don't mind the fix. We should merge it. But I still don't think this repo is in the launcher call path.

kriegfrj:
I think i more-or-less documented the stack trace in the issue report. The stack trace (from memory) is (roughly, in reverse)
* AbstractOSGiLaunchDelegate.finalLaunchCheck()
* AOLD.initialiseBndLauncher()
* At some point, a Run is created to model the bndrun file.
* Then it calls Run.getRunBundles() which builds the containers. 
* The Container initialization iterates through all available repos, calling get() to resolve them (asynchronously) to a File.
* One of them (the one that returned a match for the assertj bundle) was a MavenWorkspaceRepository instance.

I think there is still more to this bug. The fact that it was intermittent is still a bit concerning. I did note that when i disabled building completely (including the build before launch fix) it seemed to fail consistently.

rotty3000:
But how did the Run get hold of the MavenWorkspaceRepository? It's not provided to runs anywhere! Nor is it intended to be provided to Runs.

kriegfrj:
I think i may have discovered the source of @rotty3000 's confusion. "Launcher" is an overloaded term There is the Eclipse/Bndtools launcher (ILaunchDelegate) and the nmd launcher.  The former is responsible for invoking the latter. MavenWorkspaceRepository is being called from the former but not the latter. Does that make sense?

rotty3000:
Sorry, my bad it is... I need to think about it since it's not really intended to be there. I may have goofed on that. sigh...

rotty3000:
So, I was getting confused. It's the MavenDependenciesRepository that should not be included. the workspace one definitely should be. Ok... I'm with you know. Sorry for the confusion.

Issue #4100, opened 2020-05-23
When "refresh on access" is enabled, its easy to deadlock bndtools threads


gamerson:
I have "refresh on access" preference turned on in my Eclipse instances, which means I very often run into deadlock where there are 3 jobs that never finish.  However, it isn't locking the UI thread, it still prevents any workspace access so the first time you try to do an operation that requires workspace access (like editing a java file) then it will show the deadlock in the UI progress window.

Here is a stack outputs of the 3 stuck bnd threads:

```
"Worker-10: Building" #50 prio=5 os_prio=31 tid=0x00007facd107e000 nid=0x16c03 waiting for monitor entry [0x0000700012665000]
   java.lang.Thread.State: BLOCKED (on object monitor)
        at bndtools.central.Central.getWorkspace(Central.java:218)
        - waiting to lock <0x00000006c0e00000> (a org.osgi.util.promise.Deferred)
        at bndtools.central.Central.getProject(Central.java:579)
        at bndtools.central.Central$$Lambda$627/905283481.apply(Unknown Source)
        at aQute.lib.exceptions.FunctionWithException.lambda$orElseThrow$0(FunctionWithException.java:21)
        at aQute.lib.exceptions.FunctionWithException$$Lambda$628/1932888002.apply(Unknown Source)
        at java.util.Optional.map(Optional.java:215)
        at bndtools.central.Central.getProject(Central.java:585)
        at org.bndtools.builder.BndSourceGenerateCompilationParticipant.isActive(BndSourceGenerateCompilationParticipant.java:24)
        at org.eclipse.jdt.internal.core.JavaModelManager$CompilationParticipants.getCompilationParticipants(JavaModelManager.java:473)
        at org.eclipse.jdt.internal.core.builder.JavaBuilder.initializeBuilder(JavaBuilder.java:635)
        at org.eclipse.jdt.internal.core.builder.JavaBuilder.build(JavaBuilder.java:181)
        at org.eclipse.core.internal.events.BuildManager$2.run(BuildManager.java:832)
        at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
        at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:220)
        at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:263)
        at org.eclipse.core.internal.events.BuildManager$1.run(BuildManager.java:316)
[...]

PR #4101, opened 2020-05-23
[repositories view] process refresh in normal job instead of workspace job

gamerson:


bjhargrave:
> I am not sure what you are trying to accomplish.

I can see that this is for https://github.com/bndtools/bnd/issues/4100.

But I think my other comments still apply.

gamerson:
> It seems a WorkspaceJob is nice since it defers any builds until the job is done. Using a regular job means builds could start while we are processing the refresh.

If we use a workspace job, we must ensure that we will be able to get the Central.getWorkspace() without blocking.  We should try to avoid having a "UI" feature like this both hold a workspace lock and then potentially "block" for the Central.getWorkspace() that is the cause of the deadlock I was able to see in #4100 


If there a method that will try to "get" the Central.getWorkspace without blocking?  If so, we coudl do something like this

1. start a workspace job
2. nonblocking get() of workspace, if we get the workspace, then do our work
3. if we can't get the workspace without a block, then just end the workspace job and reschedule another one for a second later.

bjhargrave:
Isn't the answer to do "onWorkspace(ws -> run WorkspaceJob)"? Then you defer the job until there is a workspace and the job is a workspace job.

gamerson:
> Isn't the answer to do "onWorkspace(ws -> run WorkspaceJob)"? Then you defer the job until there is a workspace and the job is a workspace job.

I will try that.

PR #4124, opened 2020-05-31
m2e: refresh Maven repositories async

rotty3000:

### DIFF
--- a/bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
+++ b/bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
@@ -1,7 +1,5 @@
 package bndtools.m2e;
 
-import static aQute.lib.exceptions.Exceptions.unchecked;
-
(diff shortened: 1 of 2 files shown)


bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
```diff
@@ -130,8 +128,15 @@ public MavenWorkspaceRepository() {
 
 	@Activate
 	void activate() {
-		process();
-		list = memoize(this::list0);
+		try {
+			Central.bndCall(() -> {
+				process();
+				list = memoize(this::list0);
```
bjhargrave, line 134: This does not do what you want. It just creates the memoize object. You need to wrap the supplier. So move the bndCall invocation into list0.

rotty3000: changed the logic

PR #4135, opened 2020-06-04
Added documentation for p2, osgi and local repo

juergen-albert:
- Also removed an unused config property from the P2Config
- removed some odd characters in the wrapping md

With the LocalIndexRepo I would need some help. I can't make heads or tails out of some of the properties, 

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Config.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Config.java
@@ -36,6 +36,9 @@ public interface P2Config {
 	 * If not set or false, this assumes a P2 repository, i.e. the url points to
 	 * a P2 repository directory. If set to true, the url is assumed to point to
 	 * an Eclipse Target platform.
+	 * 
+	 * @deprecated
 	 */
+	@Deprecated
 	boolean targetPlatform();
 }
--- a/docs/_plugins/osgirepo.md
+++ b/docs/_plugins/osgirepo.md
@@ -1,0 +1,33 @@
+---
+title: OSGi Repository
+layout: default
+summary: A plugin to use OSGi repositories 
+---
+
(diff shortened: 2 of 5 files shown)


biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Config.java
```diff
@@ -32,10 +32,4 @@
 	 */
 	String location(String string);
 
-	/**
-	 * If not set or false, this assumes a P2 repository, i.e. the url points to
-	 * a P2 repository directory. If set to true, the url is assumed to point to
-	 * an Eclipse Target platform.
-	 */
-	boolean targetPlatform();
```
bjhargrave, line 40: Removing this hardly seems worth a major package bump. Perhaps just mark it deprecated instead. And you don't need to document it in the docs page.

docs/_plugins/locindexrepo.md
```diff
@@ -0,0 +1,36 @@
+---
```
bjhargrave, line 1: I would call the file`localindexrepo.md` That is spell out `local`.

docs/_plugins/osgirepo.md
```diff
@@ -0,0 +1,33 @@
+---
+title: OSGi Repository
+layout: default
+summary: A plugin to use OSGi repositories 
+---
+
+The OSGi Repository can read index files as specified by the [OSGi Repository Service Specification and is a read only repository.
```
bjhargrave, line 7: There is a `[` in front of the second `OSGi` in the sentence

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Config.java
juergen-albert: We could deprecate it, but for how long should it stay in there? I've needed to bump the Version this hard, because baselining complained about it.

bjhargrave: It is complaining because you deleted the method from an interface in an exported package :-). So don't delete it. Just mark is deprecated and comment that it is unused. There is little harm in leaving the method in place.

PR #4139, opened 2020-06-04
ee: update package/module list for upcoming java versions

rotty3000:


biz.aQute.bndlib/src/aQute/bnd/build/model/JavaSE_15.properties
```diff
@@ -463,10 +540,6 @@ jpms.modules = \
  jdk.nio.mapmode;\
   exports:List<String>='\
    jdk.nio.mapmode',\
- jdk.scripting.nashorn;\
```
bjhargrave, line 466: Can you also remove the modules with empty exports lists?

rotty3000: will do!

PR #4145, opened 2020-06-05
mainclass: Version attributes must be of type Version

bjhargrave:


bjhargrave:
> Was there a reason for this or just purity?

Yes. The purpose of the version attribute was for comparing against a version range. 

https://github.com/bndtools/bnd/blob/cff3f922c28d92cb7a4879a14ba78cc2db5a77ea/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/MainClassNamespace.java#L40-L54

If the version attribute was a plain string, then the version range filter would not properly compare.

There are not tests which confirm this, so the problem of the missing Version type was undetected. I noticed the problem when reindexing our Eclipse 2018-12 repo and saw the new bnd.mainclass capability has a version attribute that was not typed Version like the version attribute on osgi.wiring.bundle.

> We generally treat version as a string to allow error handling in these places and maybe weird use cases.

This is not true in any of the resource builder code. We always treat versions as typed Version.

> The verifier has to ensure that Bundle-Version is a proper version?

Perhaps, but that has nothing to do with proper filter comparison semantics for versions.

PR #4147, opened 2020-06-08
[plugins] Plugin management

pkriens:
Provides a menu entry that inserts a plugin with its properties. I.e. if you want
to create a MavenBndRepository, open the context menu and select the appropriate entry.

The plugins are tracked via the requirements and capabilities. The BndPlugin is creating capabilities that are tracked by the InternalPluginTracker


pkriens:
Actually the patch also has separate Insert menu entries for:

* Instructions
* Headers
* macros

This is based on information in ...help.Syntax

PR #4151, opened 2020-06-09
Insert Menu for plugins and Syntax based values

pkriens:
Provides a menu entry that inserts a plugin with its properties. I.e. if you want
to create a MavenBndRepository, open the context menu and select the appropriate entry.

The plugins are tracked via the requirements and capabilities. The BndPlugin is creating capabilities that are tracked by the InternalPluginTracker

The patch also has separate Insert menu entries for:

Instructions
Headers
macros

This is based on information in ...help.Syntax


bndtools.core/bnd.bnd
```diff
@@ -13,6 +13,8 @@ Import-Package: \
  ${eclipse.importpackage},\
  *
 
+Export-Package: org.osgi.service.metatype.annotations
```
bjhargrave, line 16: Why are we exporting a package of class retention annotations? Who would be importing them?

bndtools.core/src/bndtools/central/InternalPluginTracker.java
```diff
@@ -0,0 +1,135 @@
...
+				}
+			};
+		} catch (ClassNotFoundException e) {
+			return null;
+		}
+	}
+
+	static String getTemplate(String name, Class<?> implementation, Class<?> configuration) {
+		try (Formatter sb = new Formatter()) {
+			ObjectClassDefinition od = configuration != null ? configuration.getAnnotation(ObjectClassDefinition.class)
```
bjhargrave, line 95: Isn't `configuration.getAnnotation(ObjectClassDefinition.class)` always `null` since ObjectClassDefinition is class retention and thus not visible at runtime?

bndtools.core/src/bndtools/central/InternalPluginTracker.java
```diff
@@ -0,0 +1,135 @@
...
+			}
+
+			sb.format("%s", implementation.getName());
+
+			for (Method m : used.values()) {
+				String description = "";
+				String deflt = "";
+				boolean required = true;
+
+				AttributeDefinition ad = m.getAnnotation(AttributeDefinition.class);
```
bjhargrave, line 122: Same here, this would also always be null, no?

pkriens:
Yeah, you're right :-( I think the originals were runtime ....

It works without, but the annotations would give a better GUI.

pkriens:
Rock and a hard place. I guess I should replace them with Meta.OCD/AD?

bjhargrave:
Since this is for BndPlugins, why not just add the desired information to that annotation (new elements)? Instead of trying to repurpose OCD/AD?

pkriens:
It annotates the configiuration class, so it needs an annotation applied to the methods of the configuration class methods. I.e. exactly AD and OCD (or the OSGi counterparts)

Issue #4154, opened 2020-06-10
[bndtools] support cancel on resolve operation


juergen-albert:
This is a bit connected to #3201 . If we resolve something in the IDE, there is no possibility to stop the Job. You may already have experienced this yourself. If the possibilities for the resolver are too vast, it simply eats up all the memory and you can only restart eclipse or it runs in an OOM.

Eclipse Jobs support cancellation and I already took a look at the issue. At the moment I don't really see a way to tell the resolver to stop. The only idea I had was to use a `ResolutionCallback` to throw an Exception, when somebody triggers a cancel from the outside. This seems a bit dirty though and I have no clue, how regular this callbacks may be called when the resolver is in full swing.

Any Ideas?

kriegfrj:
It would be even nicer if the resolve operation could give some kind of useful feedback about the resolve operation as it is in progress - then you might get a clue early on if it's gone down a rabbit warren or else if it might finish in a reasonable time. But the ability to cancel would be a good start.

rotty3000:
I feel that a problem of resolving these days is originating from p2 repository. I rarely hear about resolve issues unless it's with P2.

However, I don't believe the issue is with the indexation or even the magnitude necessarily. My gut tells me that there is some logical construct within the bundles or with some characteristic of the bundles generally found in P2 repositories which causes much more stress on the resolver than elsewhere.

We've been resolving lately against some pretty massive repos and the cost seems so insignificant that it barely registers.

Perhaps it's related to `Require-Bundle` and maybe some stress this places on the class-space calculations the resolver has to compute?

kriegfrj:
Interesting thoughts, @rotty3000 - I've just hit what looks like a resolver hang when working in the Bnd workspace, which doesn't have any p2 repos. I am, however, trying to resolve an Eclipse launch (which has lots of bundles with `Require-Bundle`), so maybe it is related to `Require-Bundle`? On the other hand, I've also had Eclipse launches resolve in a reasonable timeframe too, so perhaps there's something else going on.

Whatever *is* going on with these long resolutions, however, it will be difficult to trace unless we can get some kind of progress output from the resolver. I'm sure if we had that progress information then the root cause would become immediately obvious in many cases.

rotty3000:
FYI, I didn't mean the P2 repositorty implementations in bnd (which from resolver perspective is irrelevant since all it sees are resources). Rather I meant the _characterics of bundles one commonly finds in p2 repositories_.

[...]

PR #4155, opened 2020-06-10
Implements cancel for resolve job #4154

juergen-albert:
- Adds the ability to keep the resolve log after it is canceled


bndtools.core/src/org/bndtools/core/resolve/CancelOperationCallback.java
```diff
@@ -0,0 +1,27 @@
...
+
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.osgi.resource.Capability;
+import org.osgi.resource.Requirement;
+
+import biz.aQute.resolve.ResolutionCallback;
+
+public class CancelOperationCallback implements ResolutionCallback {
+
+	private IProgressMonitor monitor;
```
bjhargrave, line 14: `private final`

biz.aQute.resolve/src/biz/aQute/resolve/RunResolution.java
```diff
@@ -73,10 +73,28 @@
 	 */
 	public static RunResolution resolve(Project project, Processor actualProperties,
 		Collection<ResolutionCallback> callbacks) {
+		return resolve(project, actualProperties, callbacks, new ResolverLogger());
```
bjhargrave, line 76: I think the last argument need to be `null`. Otherwise the ResolverLogger is not closed.

bndtools.core/src/bndtools/editor/BndEditor.java
```diff
@@ -328,8 +329,9 @@ public IStatus runInUIThread(IProgressMonitor monitor) {
 					sourcePage.commit(false);
 				}
 
+				ResolverLogger logger = new ResolverLogger();
```
bjhargrave, line 332: Who closes this ResolverLogger? I have the feeling there are several code paths where it is quite unclear who, if anyone, closes the ResolverLoggers.

juergen-albert: you are right, the implementation was a bit too messy. I need to keep the logger till the End, so I can instruct it not to delete the logfile if the resolver ran in circles and needed canceling. 

The logger will will now be created for each resolve operation and will be closed when the wizard is disposed or the when we reresolve after adding optional bundles

bjhargrave:
Thanks!

Issue #3980, opened 2020-04-27
publish the workspace instance as a service


rotty3000:
- entities depending on the workspace should track it
- "we will want a service tracker which lets people add callbacks to be called when the service appears. That way it is easy to init something later when the workspace is available"

bjhargrave:
Do we still need this given recent changes made to Central for anyWorkspace?

rotty3000:
I'd like to switch to this model since it will make several things much simpler.

bjhargrave:
I am not sure how. Not all places in the code will have ready access to a BundleContext to get services. You could use DS but that does not get you a callback model like the current on*Workspace methods. You would also need to distinguish between a default workspace and a cnf workspace.

But I am open to a proposal...

rotty3000:
We can close this for now. And if I ever come up with a proposal I'll just add another or reopen this issue.

kriegfrj:
I'm currently looking at #3175. Prior to #3175, I'd like to convert the builder bundle to use a form of the facade pattern to make the dev work easier. However, I am running into startup dependency issues similar to the ones that plagued the quick fix processor tests.

I would like to revisit the idea behind this issue:
* Publish the workspace as a DS component.
* Other components that 

Re the comment: "Not all places in the code will have ready access to a BundleContext to get services" - with the `ExtensionFacade`, the interface between non-DS and DS code is much easier, and the facade has callback methods for when the backing service arrives/disappears. Also, the facade pattern makes it possible to turn many of the clients of the Workspace into DS components themselves, in which case DS reference annotation mechanism can be used to express the dependency on the workspace and DS will automatically handle the callback mechanism.

Unfortunately, I don't think I'm going to have a full appreciation for how this might work without pressing ahead with a prototype, so there's a risk I'll be wasting my time if I can't make it work, but I think it's a risk worth taking.

bjhargrave:
> * Publish the workspace as a DS component.

See https://github.com/bndtools/bnd/blob/ec9368b9e7e2f71a86ac6a8835aa001a2495487b/bndtools.core/src/bndtools/central/Central.java#L311 where the Workspace is already registered as a service.

kriegfrj:
Yeah thanks I found that after I posted. What do you think of the idea of 
1. unregistering and re-registering when the workspace directory changes;
2. setting a boolean property "default=true/false" on the registration depending on whether it is the default workspace.

bjhargrave:
[...]

PR #4163, opened 2020-06-15
[quickfix] Integration tests for BuildpathQuickFixProcessor

### LINKED ISSUE #3881: [bndtools] Integration tests for BuildpathQuickFixProcessor
If you just could have a standalone test where I can give a Java source with a problem and get the quick fixes, that would be great. We could use this in other cases as well since this stuff is highly undertested.

_Originally posted by @pkriens in https://github.com/bndtools/bnd/pull/3833_

kriegfrj:
Ported from the old ImportPackageQuickFixProcessorTest. Fixes #3881.

As I suspected and reported in my review of the new implementation (#3833), while the new implementation of the quick fix processor has overall vastly better features, there are a few of the more esoteric cases with the old implementation used to handle but which the new one doesn't. I have left the tests for these in there but annotated as `@Disabled` so that they don't interfere with the CI output. If/when someone decides to have a go at adding support for them, they will be there for them.

The Eclipse side of things was interesting - unfortunately, it seems that `bndtools.core` has some dependencies at activation time which it maybe shouldn't, and so I couldn't get the quick fix class to load without loading the entire workbench. I used a custom launcher for this based on the existing applaunch bundle, but I might be able to reconfigure this to use the existing applaunch.


cnf/ext/repositories.bnd
```diff
@@ -7,6 +7,11 @@ mavencentral:           https://repo.maven.apache.org/maven2
         releaseUrl="${mavencentral}";\
         index="${.}/central.mvn";\
         readOnly=true,\
+    aQute.bnd.repository.maven.provider.MavenBndRepository;\
+        name="Sonatype";\
+        snapshotUrl="https://oss.sonatype.org/content/repositories/snapshots/";\
+        index="${.}/sonatype.mvn";\
+        readOnly=true,\
```
bjhargrave, line 14: Can we not fold this into "OSGi Snapshots"? After all the osgi-test artifacts are from OSGi.

kriegfrj:
Hey BJ, thanks for the review. 

> How to these tests run during the CI build? I did not see how the new bndtools.core.test project was connected to the gradle build.

They aren't. That was a bit outside my area of expertise and I was hoping you could help. :smile:

bjhargrave:
For OSGi integration tests, the `testOSGi` gradle task is the task that would be executed. So the project's `bnd.bnd` file needs to set the necessary `Test-Cases` and `-run*` properties so that the framework under test can be run and the test cases executed. See https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.gradle/testresources/workspaceplugin1/test.simple/bnd.bnd for a simple example which is used in the gradle plugin tests to test the `testOSGi` task (among other tests).

Once the project's `testOSGi` task is operating as desired, we can add the project to the `-dependson` list in `dist/bnd.bnd` to make sure it is part of the build.

cnf/ext/repositories.bnd
kriegfrj: I kept it separate for resolving - the OSGi Snapshot repo is not included in -runresolve, but Sonatype is. I suppose i could use -runblacklist instead to exclude just those we want to exclude.

bjhargrave: OK, that makes sense. But then "Sonatype" is not a descriptive name for the new repo.

kriegfrj: It is about as descriptive as "Maven Central"...

Im hoping that the release of osgi-test is not far away and i can add them to Maven Central. Actually, i think i settled on an implementation which didn't use osgi-test so maybe i can ditch it altogether.

bjhargrave: My point was that you segregated those artifacts into the separate repo so you can use that repo in resolving. I could not see that, so I asked why. Perhaps a name like "Test Resolve" or something would better convey the intent of the repo and stop someone else in the future adding something to the repo which we don't want in the test resolve.

kriegfrj: Sorry, I didn't explain it very well. My point was, as we have a generic "Maven Central" repo where we throw all of the bundles that live on Maven Central (and don't belong in another repo), I thought it might be nice to have a generic "Sonatype" repo where we throw all of the snapshots that don't live anywhere else. The purpose of segregating was more about keeping the OSGi Snapshot repo *out* of the resolve (because it seems to exclusively contain build-time-only bundles) than about having a specific repo for resolving the integration test.

bndtools.core.test/bnd.bnd
```diff
@@ -0,0 +1,70 @@
...
+	bndtools.builder;version=snapshot,\
+	org.junit.platform.commons;version=latest,\
+	org.junit.platform.engine;version=latest,\
+	assertj-core;version=latest,\
+	org.opentest4j;version=latest,\
+	org.apiguardian:apiguardian-api;version=latest,\
+	org.junit.jupiter.api;version=latest,\
+	org.junit.jupiter.engine;version=latest,\
+	org.junit.jupiter.params;version=latest,\
+	org.eclipse.jdt.core;version=3.16.0,\
```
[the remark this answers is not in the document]
kriegfrj: I agree with this - the reason these dependencies have dependencies is because they were added (mostly) using the quick fix processor. I think we need to consider changing the default behaviour of the quick fix processor on this point, and maybe making it configurable.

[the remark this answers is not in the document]
kriegfrj: Actually, it shouldn't be too awkward to upgrade Eclipse as the version specifiers don't give an upper bound on the version range. As long as we don't go try to backwards to support older versions of Eclipse it should still be able to find the build dependencies. I still agree that we don't need them though.

kriegfrj:
> For OSGi integration tests, the `testOSGi` gradle task is the task that would be executed. So the project's `bnd.bnd` file needs to set the necessary `Test-Cases` and `-run*` properties so that the framework under test can be run and the test cases executed. See https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.gradle/testresources/workspaceplugin1/test.simple/bnd.bnd for a simple example which is used in the gradle plugin tests to test the `testOSGi` task (among other tests).
> 
> Once the project's `testOSGi` task is operating as desired, we can add the project to the `-dependson` list in `dist/bnd.bnd` to make sure it is part of the build.

Hmm, is there anyway to use a bndrun in the Gradle workspace model? The problem is we need platform-specific run configurations for the tests. Alternatively, a way to handle the platform-specificity in the bnd.

(Btw, were you at least able to run the integration tests on Mac using `test.cocoa.macosx.x86_64.bndrun`?)

bndtools.core.test/bnd.bnd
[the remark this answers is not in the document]
kriegfrj: Fixed now.

cnf/ext/repositories.bnd
kriegfrj: I have left this one as it is for now. Longer term (once osgi-test goes into a proper release) we can either move osgi-test into Maven Central, or split Maven Central into "Maven Central" and "Maven Central Testing" for test-oriented dependencies and make osgi-test a member of that.

bjhargrave:
> Hmm, is there anyway to use a bndrun in the Gradle workspace model? The problem is we need platform-specific run configurations for the tests. Alternatively, a way to handle the platform-specificity in the bnd.

You can set the name of the bndrun to use with a `TestOSGi` type task. You can even create multiple `TestOSGi` type tasks. For an example, see

https://github.com/bndtools/bnd/blob/450c34de9d8c293235409ee0cd2bbbb8ed4a20cb/biz.aQute.bnd.gradle/testresources/testosgitask1/build.gradle#L26-L31

> (Btw, were you at least able to run the integration tests on Mac using `test.cocoa.macosx.x86_64.bndrun`?)

I did not pull the PR onto my system.

cnf/ext/repositories.bnd
kriegfrj: I found that the dependency that I was looking for in osgi-test has an equivalent in `aQute.lib.exceptions`, so the latest push has removed osgi-test (and the Sonatype repo) completely.

kriegfrj:
Ok, I've had mixed success with connecting it in to the CI build. See latest results here: https://github.com/kriegfrj/bnd/runs/794016113

It seems that my headless test is not quite headless enough. The Windows CI build seems to work properly, but the integration tests fail on Ubuntu with a gtk error which looks suspiciously like what one might get when trying to run a gtk application on a fully headless server:

```
Unable to init server: Could not connect: Connection refused
! Unexpected error in the run body: No more handles [gtk_init_check() failed]
org.eclipse.swt.SWTError: No more handles [gtk_init_check() failed]
	at org.eclipse.swt.SWT.error(SWT.java:4621)
```

I'll need to look into this further.

kriegfrj:
So close, yet so far... this is turning into a mini nightmare...

Here are the main constraints, which I am recording for posterity:

* `bndtools.builder` has a dependency during build on the `WorkingSetManager`. This in turn means that you can't use the bndtools builder without a fully running Workbench.
* I nearly got around the above problem by building using bnd directly, but then the introduction of the `JavaSearchStartupParticipant` introduced a similar dependency into `bndtools.core`, which means that you can't even start `bndtools.core` unless the `WorkingSetManager` (and hence a `Workbench`) is running.

I tried to create a `WorkingSetManager` service in `HeadlessLauncher` so that I could avoid starting the workbench. However, there is no portable way for the bndtools plugins to retrieve a reference to this service (at least, not in a way that is portable between live and test environments) without access to an `IEclipseContext`, which you can't get unless you fully adopt the Eclipse E4 dependency injection mechanism for your components. This problem is of a category that something like #3980 might help to solve.

These problems might be solvable, I'm not sure. Possible solutions:

1. Try and decouple enough of a subset of the other plugins from the Workbench enough so that we can run the quick fix integration tests against that subset without the Workbench. This is might be a solution for the quick fix processor tests perhaps, but it doesn't solve the wider problem of integration testing for other parts of bndtools.
2. Headless, pure-Java implementation of SWT for headless testing purposes. I attempted this but abandoned it as there's a lot of work involved, even for producing a minimal set of SWT functionality to support the current set of tests. It's made more difficult by the lack of a common set of API classes/interfaces for SWT that the implementations extend/implement.
3. Configure GitHub actions to allow GUI code to execute. I had a quick look at this as it would be the easiest solution, but it doesn't seem possible.
4. Leave it as-is for now. The tests run on the windows CI which is sufficient for the CI to verify that they are working. They also work (or at least, they should work - I'm not in a position to test the theory) when you're developing/running on your own (non-headless) machine regardless of OS, which is useful for development purposes.

I've tried to see what the Eclipse testing framework does, but as far as I can see it doesn't really address the headless problem - it simply requires you to run the test on a headed machine. Indeed, I read in the SWTBot documentation that you can't use the UI while a test is running.

This has taken up a lot more of my time than I would have liked, so I'm tempted to go for the fourth option for now, if the core team is supportive. We can revisit other more elegant solutions later.

Thoughts?

bjhargrave:
> Thoughts?

I guess if we can get some testing only on Windows, we should do it. We will need to make sure the code comments for the test make clear as to why it is only tested on Windows. There may also come a day when the tests no longer work on some future version of the Windows used for GitHub Actions.

kriegfrj:
Ok, thank God, I have found https://github.com/GabrielBB/xvfb-action which I think is the answer - it is a Github action that runs the script within xvfb, and this seems to do the trick (though I've had to use v1.0 rather than v1.2 because of  GabrielBB/xvfb-action#10). However, now it seems there are some intermittent startup failures involving JFace that need addressing.

kriegfrj:
Not much success so far in tackling this last issue. I've also seen it happen on gradle on my local machine (though not as frequently as it seems to happen in CI), which seems to point to some kind of race condition. For posterity (and in case I don't get around to fixing it), here's the stack trace:

```
java.lang.NoClassDefFoundError: Could not initialize class org.eclipse.jface.preference.PreferenceConverter
	at org.eclipse.ui.internal.themes.ThemeElementHelper.installFont(ThemeElementHelper.java:116)
	at org.eclipse.ui.internal.themes.ThemeElementHelper.populateRegistry(ThemeElementHelper.java:71)
	at org.eclipse.ui.internal.Workbench$22.runWithException(Workbench.java:1788)
	at org.eclipse.ui.internal.StartupThreading$StartupRunnable.run(StartupThreading.java:35)
	at org.eclipse.swt.widgets.Synchronizer.syncExec(Synchronizer.java:236)
	at org.eclipse.ui.internal.UISynchronizer.syncExec(UISynchronizer.java:147)
	at org.eclipse.swt.widgets.Display.syncExec(Display.java:5807)
	at org.eclipse.ui.internal.StartupThreading.runWithoutExceptions(StartupThreading.java:98)
	at org.eclipse.ui.internal.Workbench.initializeFonts(Workbench.java:1781)
	at org.eclipse.ui.internal.Workbench.init(Workbench.java:1632)
	at org.eclipse.ui.internal.Workbench.runUI(Workbench.java:2804)
	at org.eclipse.ui.internal.Workbench.lambda$3(Workbench.java:623)
	at org.eclipse.core.databinding.observable.Realm.runWithDefault(Realm.java:339)
	at org.eclipse.ui.internal.Workbench.createAndRunWorkbench(Workbench.java:563)
	at org.eclipse.ui.PlatformUI.createAndRunWorkbench(PlatformUI.java:151)
	at org.eclipse.ui.internal.ide.application.IDEApplication.start(IDEApplication.java:155)
	at org.eclipse.equinox.internal.app.EclipseAppHandle.run(EclipseAppHandle.java:203)
	at bndtools.core.test.launch.HeadlessLauncher.lambda$launch$0(HeadlessLauncher.java:30)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at aQute.launcher.Launcher.launch(Launcher.java:450)
```

Underlying cause of the above is this:
```
2020-06-23T05:25:35.4500349Z java.lang.ExceptionInInitializerError: null
2020-06-23T05:25:35.4500901Z 	at org.eclipse.jdt.ui.PreferenceConstants.initializeDefaultValues(PreferenceConstants.java:3966)
2020-06-23T05:25:35.4501114Z 	at org.eclipse.jdt.internal.ui.JavaUIPreferenceInitializer.initializeDefaultPreferences(JavaUIPreferenceInitializer.java:38)
2020-06-23T05:25:35.4501290Z 	at org.eclipse.core.internal.preferences.PreferenceServiceRegistryHelper$1.run(PreferenceServiceRegistryHelper.java:301)
2020-06-23T05:25:35.4501441Z 	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
2020-06-23T05:25:35.4501594Z 	at org.eclipse.core.internal.preferences.PreferenceServiceRegistryHelper.runInitializer(PreferenceServiceRegistryHelper.java:304)
2020-06-23T05:25:35.4501784Z 	at org.eclipse.core.internal.preferences.PreferenceServiceRegistryHelper.applyRuntimeDefaults(PreferenceServiceRegistryHelper.java:134)
2020-06-23T05:25:35.4501951Z 	at org.eclipse.core.internal.preferences.PreferencesService.applyRuntimeDefaults(PreferencesService.java:375)
2020-06-23T05:25:35.4502100Z 	at org.eclipse.core.internal.preferences.DefaultPreferences.applyRuntimeDefaults(DefaultPreferences.java:225)
2020-06-23T05:25:35.4502324Z 	at org.eclipse.core.internal.preferences.DefaultPreferences.load(DefaultPreferences.java:279)
2020-06-23T05:25:35.4502475Z 	at org.eclipse.core.internal.preferences.EclipsePreferences.create(EclipsePreferences.java:370)
2020-06-23T05:25:35.4502623Z 	at org.eclipse.core.internal.preferences.EclipsePreferences.internalNode(EclipsePreferences.java:623)
2020-06-23T05:25:35.4502769Z 	at org.eclipse.core.internal.preferences.EclipsePreferences.node(EclipsePreferences.java:766)
2020-06-23T05:25:35.4502918Z 	at org.eclipse.core.internal.preferences.AbstractScope.getNode(AbstractScope.java:41)
2020-06-23T05:25:35.4503044Z 	at org.eclipse.core.runtime.preferences.DefaultScope.getNode(DefaultScope.java:77)
2020-06-23T05:25:35.4503207Z 	at org.eclipse.jdt.internal.core.manipulation.MembersOrderPreferenceCacheCommon.install(MembersOrderPreferenceCacheCommon.java:112)
2020-06-23T05:25:35.4503371Z 	at org.eclipse.jdt.internal.ui.preferences.MembersOrderPreferenceCache.install(MembersOrderPreferenceCache.java:38)
2020-06-23T05:25:35.4503520Z 	at org.eclipse.jdt.internal.ui.JavaPlugin.start(JavaPlugin.java:408)
2020-06-23T05:25:35.4503662Z 	at org.eclipse.osgi.internal.framework.BundleContextImpl$3.run(BundleContextImpl.java:782)
2020-06-23T05:25:35.4503806Z 	at org.eclipse.osgi.internal.framework.BundleContextImpl$3.run(BundleContextImpl.java:1)
2020-06-23T05:25:35.4503941Z 	at java.security.AccessController.doPrivileged(Native Method)
2020-06-23T05:25:35.4504087Z 	at org.eclipse.osgi.internal.framework.BundleContextImpl.startActivator(BundleContextImpl.java:775)
2020-06-23T05:25:35.4504430Z 	at org.eclipse.osgi.internal.framework.BundleContextImpl.start(BundleContextImpl.java:732)
2020-06-23T05:25:35.4504649Z 	at org.eclipse.osgi.internal.framework.EquinoxBundle.startWorker0(EquinoxBundle.java:1005)
2020-06-23T05:25:35.4504800Z 	at org.eclipse.osgi.internal.framework.EquinoxBundle$EquinoxModule.startWorker(EquinoxBundle.java:357)
2020-06-23T05:25:35.4505206Z 	at org.eclipse.osgi.container.Module.doStart(Module.java:588)
2020-06-23T05:25:35.4505341Z 	at org.eclipse.osgi.container.Module.start(Module.java:456)
2020-06-23T05:25:35.4505473Z 	at org.eclipse.osgi.framework.util.SecureAction.start(SecureAction.java:471)
2020-06-23T05:25:35.4505618Z 	at org.eclipse.osgi.internal.hooks.EclipseLazyStarter.postFindLocalClass(EclipseLazyStarter.java:117)
2020-06-23T05:25:35.4505768Z 	at org.eclipse.osgi.internal.loader.classpath.ClasspathManager.findLocalClass(ClasspathManager.java:557)
2020-06-23T05:25:35.4505928Z 	at org.eclipse.osgi.internal.loader.ModuleClassLoader.findLocalClass(ModuleClassLoader.java:331)
2020-06-23T05:25:35.4506075Z 	at org.eclipse.osgi.internal.loader.BundleLoader.findLocalClass(BundleLoader.java:395)
2020-06-23T05:25:35.4506229Z 	at org.eclipse.osgi.internal.loader.sources.SingleSourcePackage.loadClass(SingleSourcePackage.java:39)
2020-06-23T05:25:35.4506373Z 	at org.eclipse.osgi.internal.loader.BundleLoader.findClassInternal(BundleLoader.java:457)
2020-06-23T05:25:35.4506517Z 	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:422)
2020-06-23T05:25:35.4506642Z 	at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:414)
2020-06-23T05:25:35.4506787Z 	at org.eclipse.osgi.internal.loader.ModuleClassLoader.loadClass(ModuleClassLoader.java:153)
2020-06-23T05:25:35.4506924Z 	at java.lang.ClassLoader.loadClass(ClassLoader.java:351)
2020-06-23T05:25:35.4507054Z 	at java.lang.Class.getDeclaredMethods0(Native Method)
2020-06-23T05:25:35.4507186Z 	at java.lang.Class.privateGetDeclaredMethods(Class.java:2701)
2020-06-23T05:25:35.4507319Z 	at java.lang.Class.getDeclaredMethods(Class.java:1975)
2020-06-23T05:25:35.4507471Z 	at aQute.tester.bundle.engine.discovery.BundleSelectorResolver.tryToResolveTestClass(BundleSelectorResolver.java:381)
2020-06-23T05:25:35.4507645Z 	at aQute.tester.bundle.engine.discovery.BundleSelectorResolver.lambda$getSelectorsFromTestCasesHeader$27(BundleSelectorResolver.java:401)
2020-06-23T05:25:35.4507803Z 	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
2020-06-23T05:25:35.4507943Z 	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
2020-06-23T05:25:35.4508080Z 	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
2020-06-23T05:25:35.4508201Z 	at aQute.libg.qtokens.QuotedTokenizer$1.tryAdvance(QuotedTokenizer.java:182)
2020-06-23T05:25:35.4508334Z 	at java.util.Spliterator.forEachRemaining(Spliterator.java:326)
2020-06-23T05:25:35.4508470Z 	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
2020-06-23T05:25:35.4508612Z 	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
2020-06-23T05:25:35.4508753Z 	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
2020-06-23T05:25:35.4508893Z 	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
2020-06-23T05:25:35.4509030Z 	at java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:566)
2020-06-23T05:25:35.4509183Z 	at aQute.tester.bundle.engine.discovery.BundleSelectorResolver.getSelectorsFromTestCasesHeader(BundleSelectorResolver.java:413)
2020-06-23T05:25:35.4509344Z 	at aQute.tester.bundle.engine.discovery.BundleSelectorResolver.computeChildren(BundleSelectorResolver.java:574)
2020-06-23T05:25:35.4509501Z 	at aQute.tester.bundle.engine.discovery.BundleSelectorResolver.lambda$resolve$18(BundleSelectorResolver.java:276)
2020-06-23T05:25:35.4509630Z 	at java.util.stream.ForEachOps$ForEachOp$OfRef.accept(ForEachOps.java:183)
2020-06-23T05:25:35.4509900Z 	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
2020-06-23T05:25:35.4510037Z 	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
2020-06-23T05:25:35.4510223Z 	at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
2020-06-23T05:25:35.4510366Z 	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:482)
2020-06-23T05:25:35.4510503Z 	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:472)
2020-06-23T05:25:35.4510639Z 	at java.util.stream.ForEachOps$ForEachOp.evaluateSequential(ForEachOps.java:150)
2020-06-23T05:25:35.4510775Z 	at java.util.stream.ForEachOps$ForEachOp$OfRef.evaluateSequential(ForEachOps.java:173)
2020-06-23T05:25:35.4510912Z 	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
2020-06-23T05:25:35.4511044Z 	at java.util.stream.ReferencePipeline.forEach(ReferencePipeline.java:485)
2020-06-23T05:25:35.4511180Z 	at aQute.tester.bundle.engine.discovery.BundleSelectorResolver.resolve(BundleSelectorResolver.java:270)
2020-06-23T05:25:35.4511339Z 	at aQute.tester.bundle.engine.discovery.BundleSelectorResolver.resolve(BundleSelectorResolver.java:103)
2020-06-23T05:25:35.4511487Z 	at aQute.tester.bundle.engine.BundleEngine.discover(BundleEngine.java:63)
2020-06-23T05:25:35.4511631Z 	at org.junit.platform.launcher.core.DefaultLauncher.discoverEngineRoot(DefaultLauncher.java:181)
2020-06-23T05:25:35.4511773Z 	at org.junit.platform.launcher.core.DefaultLauncher.discoverRoot(DefaultLauncher.java:168)
2020-06-23T05:25:35.4511914Z 	at org.junit.platform.launcher.core.DefaultLauncher.execute(DefaultLauncher.java:132)
2020-06-23T05:25:35.4512047Z 	at aQute.tester.junit.platform.Activator.test(Activator.java:352)
2020-06-23T05:25:35.4512178Z 	at aQute.tester.junit.platform.Activator.automatic(Activator.java:317)
2020-06-23T05:25:35.4512309Z 	at aQute.tester.junit.platform.Activator.run(Activator.java:217)
2020-06-23T05:25:35.4512437Z 	at java.lang.Thread.run(Thread.java:748)
2020-06-23T05:25:35.4512550Z Caused by: java.lang.NullPointerException: null
2020-06-23T05:25:35.4512678Z 	at org.eclipse.swt.widgets.Display.asyncExec(Display.java:856)
2020-06-23T05:25:35.4512818Z 	at org.eclipse.jface.preference.PreferenceConverter.<clinit>(PreferenceConverter.java:98)
2020-06-23T05:25:35.4512948Z 	... 75 common frames omitted
```

The offending code in the Gtk implementation of SWT:

```java
public void asyncExec (Runnable runnable) {
	synchronized (Device.class) {
		if (isDisposed ()) error (SWT.ERROR_DEVICE_DISPOSED);
		synchronized (idleLock) {  // <----NPE occurs here, I think
			if (idleNeeded && idleHandle == 0) {
	 			//NOTE: calling unlocked function in OS
				idleHandle = OS._g_idle_add (idleProc, 0);
			}
		}
		synchronizer.asyncExec (runnable);
	}
}
```

It's a bit odd that this could be null, unless the display has been disposed... but at least this gives us a clue.

kriegfrj:
Ok, I've identified the cause but I'm struggling for a solution.

The problem is that this is Eclipse :smile:, but more specifically, there are certain classes that will fail to load and resolve if the Workbench is not running yet. `BuildpathQuickFixProcessorTest` depends on some of these classes (unavoidably) to perform the testing.

`biz.aQute.tester.junit-platform` is started automatically by the bnd launcher. Because the SWT event dispatcher must run on the main thread, the tester is configured to run in a background thread, and it immediately starts listening for test bundles.

The bnd launcher then starts the test bundle, and the tester notices this and starts to try and run its tests. This involves attempting to resolve the test class `BuildpathQuickFixProcessorTest`.

If the above all happens too quickly, then it will happen before the Workbench is started, and so the attempt by the tester to load the test class fails.

I had some code in `BuildpathQuickFixProcessorTest` to spin and wait for the Workbench before continuing - however, this is too late because it requires the class to already be loaded, and the class load itself requires the Workbench to already be running.

I tried to move the code that spins waiting for the Workbench into the test bundle's activator - but this simply blocked the launcher from starting all the bundles - which meant that the `Runnable` method that actually starts the application won't get called.

Is there an easy way to configure the launcher so that it doesn't try to start certain specific bundles before proceeding? That way I could configure my custom HeadlessLauncher to only start the tester bundle once the Workbench is running, which would solve this problem. Alternatively, other suggestions welcomed!

pkriens:
Isn't this a case for the start levels? They work rather nicely now.

kriegfrj:
> Isn't this a case for the start levels? They work rather nicely now.

I didn't expect that suggestion to come from you... 😉 

I have been wrestling with start levels as a way around this, but I couldn't seem to hit on the right combination. Specifically, although I put the tester (`biz.aQute.tester.junit-platform`) and the test bundle (`bndtools.core.test.tests`) on high start levels hoping that they would start after the workbench was running, what seemed to happen is that the bnd launcher would keep going until it had reached the highest startlevel and all the bundles had been started *before* it would call the main `Runnable` (which starts the Workbench). By that stage the tester is already running no matter what startlevel I set it to originally, and my attempt to block until the workbench was running simply . For this approach to work, I need a way to tell the launcher to stop when it reaches a certain start level and invoke the main `Runnable` without proceeding to the higher startlevels - is there an easy way to do this?

In the meantime, I think I've just now hit on an alternative solution that works without relying on startlevels or relying on my code to manually start the tester or the bundle-under-test - I've still got the activator in my test bundle that spin-waits for the workbench to start, but I've set its activation policy to "lazy". This means that it will be started in the context of the *tester* thread (not the launcher/framework thread) so it won't prevent the workbench from starting. So now when my test bundle starts, instead of blocking the launcher thread (and blocking the workbench from starting), it blocks the tester thread until the Workbench is running. 

This seems to work (see https://github.com/kriegfrj/bnd/runs/802378717?check_suite_focus=true). Unless your experience can see an obvious flaw with this approach that I cannot, I will resubmit my PR with this fix shortly and hopefully the CI integration will be working (finally!)

pkriens:
Eclipse is a lost cause for putting proper OSGi rules on. That ship sailed 16 years ago :-(

- So you need to start the workbench on the main thread?
- Relying on lazy activation to be on the right thread sounds a tad dangerous to me
- Maybe an idea. You could add a jar to the -runpath in the plugin code with an Embedded-Activator and set it to AFTER_BUNDLES_START. This activator is run on the main thread after the bundles have been started but before the start levels code runs, so they should not be activated yet. If you use start levels, this should be before any bundle is started. In the activator you set start levels to 1 and, assuming Workbench is on there, start the bundle. Other start levels should be higher than 1.

kriegfrj:
> Eclipse is a lost cause for putting proper OSGi rules on. That ship sailed 16 years ago :-(
>

As a priest I'm in the business of lost causes. :wink:

> * So you need to start the workbench on the main thread?

Yes.

> * Relying on lazy activation to be on the right thread sounds a tad dangerous to me

Well, it's not so much that it *must* happen on the right thread - more that it has to *not* happen on the wrong thread, at the wrong time. I don't care what thread it happens on as long as it is a thread that I can safely block until the Workbench is running - this can basically be any thread except for the startup thread. This will allow the startup process to complete and the main `Runnable` (which starts the Workbench) to get executed.

> * Maybe an idea. You could add a jar to the -runpath in the plugin code with an Embedded-Activator and set it to AFTER_BUNDLES_START. This activator is run on the main thread after the bundles have been started but before the start levels code runs, so they should not be activated yet. If you use start levels, this should be before any bundle is started. In the activator you set start levels to 1 and, assuming Workbench is on there, start the bundle. Other start levels should be higher than 1.

Thanks for this - if my lazy activator approach proves unreliable, I'll have a look at this one.

Another solution I'm considering is to add a new feature to the tester (something I've been contemplating for a while) - just as Jupiter has various lifecycle callback hooks, I'd like to add similar lifecycle callbacks to `BundleEngine`. Things like "Before/AfterAllEngines", "Before/AfterEachEngine", "Before/AfterAllBundles", "Before/AfterEachBundle", "Before/AfterAllClasses", "Before/AfterEachClass". If such callbacks were available, I could implement the code to wait for the Workbench to start in one of them (probably in a "BeforeAllBundles" callback). Honestly I think would be the neatest/most transparent/most reliable solution, but it would require a fair bit more work on the `BundleEngine` to make it happen (which I'd like to defer if possible).

.github/workflows/cibuild.yml
```diff
@@ -53,9 +50,31 @@ jobs:
       with:
         java-version: ${{ matrix.java }}
     - name: Build
-      run: |
-        ./.github/scripts/${{ matrix.script }}
+       # Can't upgrade this until https://github.com/GabrielBB/xvfb-action/issues/10 is fixed
+      uses: GabrielBB/xvfb-action@v1.0
```
bjhargrave, line 54: I am not keen on using this action and these changed to the workflow. See https://github.com/bndtools/bnd/pull/4208 for my suggested alternate.

kriegfrj: I have reviewed and approved your alternate solution - agreed, it is better. I didn't realise that xvfb was part of the standard GitHub Actions offering.

I have also rebased this PR onto yours (minus the commit that changes the workflow).

kriegfrj:
I have removed the commit that updates the workflow and instead rebased my branch onto @bjhargrave's superior alternative (#4208). Still seems to work ok.

bjhargrave:
I have merged #4208.

kriegfrj:
> I have merged #4208.

Rebased onto master and re-pushed.

bndtools.core.test/.metadata/.plugins/org.eclipse.core.runtime/.settings/org.eclipse.jdt.ui.prefs
```diff
@@ -0,0 +1,9 @@
+content_assist_proposals_background=255,255,255
```
bjhargrave, line 1: We should not be checking in the `.metadata/` folder!?

bndtools.core.test/build.gradle
```diff
@@ -0,0 +1,37 @@
+import aQute.bnd.gradle.TestOSGi
+import org.apache.tools.ant.taskdefs.condition.Os
+
+task testosgi(type: TestOSGi) {
```
bjhargrave, line 4: There is already a `testOSGi` task defined by the Bnd Gradle plugin. Why are you not just configuring that task? And also you need to use the configuration avoidance style of writing gradle task for this build.

bndtools.core.test/build.gradle
```diff
@@ -0,0 +1,37 @@
...
+		'org.bndtools.templates.template',
+		'org.bndtools.templating.gitrepo',
+		'org.bndtools.templating',
+		'org.bndtools.versioncontrol.ignores.manager',
+		'org.bndtools.versioncontrol.ignores.plugin.git'
+	].each { dependsOn ":$it:jar" }
+   description "Bndtools Core Integration tests"
+   group 'test'
+   inputs.files jar
+   if (Os.isFamily(Os.FAMILY_WINDOWS)) {
```
bjhargrave, line 26: Is this the gradle way to do this?

bndtools.core.test/build.gradle
```diff
@@ -0,0 +1,37 @@
...
+      bndrun 'test.win32.x86_64.bndrun'
+   } else if (Os.isFamily(Os.FAMILY_UNIX)) {
+      bndrun 'test.gtk.linux.x86_64.bndrun'
+   } else if (Os.isFamily(Os.FAMILY_MAC)) {
+      bndrun 'test.cocoa.macosx.x86_64.bndrun'
+   }
+}
+
+check {
```
bjhargrave, line 35: The `check` task already depends on the `testOSGi` task mentioned above.

bndtools.core.test/bnd.bnd
```diff
@@ -0,0 +1,52 @@
+# Set javac settings from JDT prefs
+-include: ${workspace}/cnf/includes/jdt.bnd, ${workspace}/cnf/includes/bndtools.bnd
+-sub: *.bnd
```
bjhargrave, line 3: Since this is a test project and should not be released, you need to make sure it is not released.

```properties
-buildrepo:
-releaserepo:
```

bndtools.core.test/src/bndtools/core/test/launch/NullContextPresentationEngine.java
```diff
@@ -0,0 +1,154 @@
+/*******************************************************************************
+ * Copyright (c) 2009, 2015 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
```
bjhargrave, line 9: This whole repo is dual licensed under AL2 OR EPL2. Not sure why you have this license header here.

bndtools.core.test/src/simple/package-info.java
```diff
@@ -0,0 +1,3 @@
+@org.osgi.annotation.bundle.Export
```
bjhargrave, line 1: simple.bnd already has these packages as exported, so we can remove the duplicate Export annotation here.

bndtools.core.test/src/simple/pkg/package-info.java
```diff
@@ -0,0 +1,3 @@
+@org.osgi.annotation.bundle.Export
```
bjhargrave, line 1: simple.bnd already has these packages as exported, so we can remove the duplicate Export annotation here.

bndtools.core.test/build.gradle
kriegfrj: Im not a Gradle expert, so please excuse my ignorance, but what's the "configuration avoidance " style? I knew there was already a testOSGi task but I didn't know how to reconfigure it.

bndtools.core.test/build.gradle
kriegfrj: It's the best i found- if you know a better way,  let me know!

bndtools.core.test/.metadata/.plugins/org.eclipse.core.runtime/.settings/org.eclipse.jdt.ui.prefs
kriegfrj: Oops...

bndtools.core.test/build.gradle
bjhargrave: See https://github.com/bndtools/bnd/blob/35661c2dd92f56993c78cf6c984323af0f41d42f/biz.aQute.bnd.gradle/build.gradle#L22-L53 for some examples of configuring tasks that avoid executing the configuration when the task is not part of the task graph.

bndtools.core.test/build.gradle
bjhargrave: Apparently `Os.isFamily(Os.FAMILY_UNIX)` returns true on macOS. Which means this if/else construct is broken.

We do this elsewhere in the build:

https://github.com/bndtools/bnd/blob/35661c2dd92f56993c78cf6c984323af0f41d42f/maven/build.gradle#L40

bndtools.core.test/src/bndtools/core/test/launch/NullContextPresentationEngine.java
kriegfrj: This is a cut-and-paste error. NullContextPresentationEngine started off life as a clone of https://github.com/eclipse/eclipse.platform.ui/blob/master/tests/org.eclipse.e4.ui.tests/src/org/eclipse/e4/ui/tests/application/HeadlessContextPresentationEngine.java

bndtools.core.test/src/simple/pkg/package-info.java
kriegfrj: Fair enough, but just curious - what's best practice, to use the annotation or to mark it for export in the .bnd file?

bndtools.core.test/build.gradle
kriegfrj: Fixed, thanks for the pointer.

bndtools.core.test/build.gradle
kriegfrj: I'm not sure what "os.name" is going to throw at me on the various OSes. Using `Os.isFamily()` is voted the best solution on [this](https://stackoverflow.com/questions/11235614/how-to-detect-the-current-os-from-gradle) SO thread, though the answer is old. I think it's a bit of an oversight that Gradle doesn't have any built-in support for this.

I have reorded the if/else so that Mac is tested first and Linux is only used if it's not Mac. I think it's more compact. If you really prefer though, I can hand-craft something based on inspection of `os.name` (I just don't have a good way to verify it as I don't have access to a Mac to test it on).

kriegfrj:
Ok, let's try again!

bjhargrave:
Thanks!

PR #4165, opened 2020-06-15
improve naming of `access` values when defining `-jpms-module-info`

### LINKED ISSUE #4164: improve naming of `access` values when defining `-jpms-module-info`
Currently it only works if you use the based 10 integer value (32, 4096, 32768) instead of either a literal name of the hex values defined in the spec (0x0020, 0x1000, 0x8000).

rotty3000:
Fixes #4164

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/plugin/jpms/JPMSModuleInfoPlugin.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/plugin/jpms/JPMSModuleInfoPlugin.java
@@ -1,6 +1,8 @@
 package aQute.bnd.plugin.jpms;
 
+import static aQute.bnd.classfile.ModuleAttribute.ACC_MANDATED;
 import static aQute.bnd.classfile.ModuleAttribute.ACC_OPEN;
+import static aQute.bnd.classfile.ModuleAttribute.ACC_SYNTHETIC;
 import static aQute.bnd.osgi.Constants.ACCESS_ATTRIBUTE;
 import static aQute.bnd.osgi.Constants.AUTOMATIC_MODULE_NAME;
 import static aQute.bnd.osgi.Constants.DYNAMICIMPORT_PACKAGE;
@@ -34,6 +36,7 @@ import static java.util.stream.Collectors.toCollection;
 import static java.util.stream.Collectors.toList;
 import static org.osgi.namespace.extender.ExtenderNamespace.EXTENDER_NAMESPACE;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


biz.aQute.bndlib/src/aQute/bnd/plugin/jpms/JPMSModuleInfoPlugin.java
```diff
@@ -76,6 +78,46 @@
  */
 public class JPMSModuleInfoPlugin implements VerifierPlugin {
 
+	enum Access {
+		CLOSED(
```
bjhargrave, line 82: I hope to god the next Eclipse version fixes this idiotic code formatting bug...

biz.aQute.bndlib/src/aQute/bnd/plugin/jpms/JPMSModuleInfoPlugin.java
```diff
@@ -76,6 +78,46 @@
...
+			0),
+		OPEN(ACC_OPEN),
+		SYNTHETIC(ACC_SYNTHETIC),
+		MANDATED(ACC_MANDATED);
+
+		public static Access parse(String input) {
+			switch (input) {
+				case "OPEN" :
+				case "open" :
+				case "0x0020" :
```
bjhargrave, line 92: Isn't this a bit restrictive? `0x20` wont work. You probably need two paths: string and numbers.

rotty3000: k

rotty3000:
```
				default :
					int parsedValue = Integer.parseInt(input);
					return Arrays.stream(values())
						.filter(a -> a.getValue() == parsedValue)
						.findFirst()
						.orElse(CLOSED);

```
?

bjhargrave: I think you want `Integer.decode` which can handle hex strings and octal. But you should catch any exception from decode and map to an error.

bjhargrave: Also the orElse case should also be an error, no?

biz.aQute.bndlib/src/aQute/bnd/plugin/jpms/JPMSModuleInfoPlugin.java
rotty3000: :)

PR #4169, opened 2020-06-16
[reporter] Use Class instead of class name to register default plugins

### LINKED ISSUE #4150: [ReportGenerator] Short Plugin Names
I ran into a bug and then had to debug the ReportGenerator. 

I noticed: https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.reporter/src/biz/aQute/bnd/reporter/generator/ReportGeneratorBuilder.java#L215-L231

First, I think this is a nice feature, it would be nice if that was implemented in bndlib for everybody. (Actually now have support for this by using capabilities). 

Anyway, my problem is with the implementation. In the code you first get the class, then get the name, and then load the class in findName. This is a very roundabout way. Just use the class, and in this case since you have a selected set, I would use a constant for the name that is used in the registration and the BndPlugin name. Then you can use the same constant wherever you are referring to this thing.

cdelg:
Closes #4150 

Am I right? @stbischof, @pkriens.


pkriens:
LGTM

stbischof:
Lgtm

bjhargrave:
Thanks

PR #4178, opened 2020-06-22
[bndtools.core] Moved launch delegates into api package

### LINKED ISSUE #4177: Export Bndtools LaunchDelegate
When writing a profiling plugin it's needed to import the LaunchDelegate (e.g. OSGiRunLaunchDelegate) but currently that's not possible because the package is not exported. PDE is exporting the launch delegate bundle here: https://github.com/eclipse/eclipse.pde.ui/blob/master/ui/org.eclipse.pde.launching/META-INF/MANIFEST.MF#L22

To enable such use cases we should probably move the launch delegates into the [api package](https://github.com/bndtools/bnd/tree/master/bndtools.core/src/bndtools/launch/api) or add the launch package to the exported packages list.

swimmesberger:
Fixes #4177

### DIFF
--- a/bndtools.core/_plugin.xml
+++ b/bndtools.core/_plugin.xml
@@ -492,7 +492,7 @@
    <!-- DEFINE LAUNCH CONFIGURATION TYPES -->
    <extension point="org.eclipse.debug.core.launchConfigurationTypes">
       <launchConfigurationType
-        delegate="bndtools.launch.OSGiRunLaunchDelegate"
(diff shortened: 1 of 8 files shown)


swimmesberger:
It seems that since https://github.com/bndtools/bnd/pull/2920/https://github.com/bndtools/bnd/issues/2792 the bndtools.launch package is indeed exported and I'm simply using an older version of bnd.

PR #4208, opened 2020-07-01
ci: Use xvfb-run to build on linux

bjhargrave:
This is needed by UI test cases to provide a frame buffer.

The GitHub Action unbuntu-latest os already has xvfb installed and up to date.

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -24,20 +24,24 @@ jobs:
   Build_Matrix:
     strategy:
       matrix:
-        java: [ 11, 14, 15-ea ]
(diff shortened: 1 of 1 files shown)


bjhargrave:
> your branch label has a transposition error (xfvb-run instead of xvfb-run)

Oops. But not worth fixing. I have more egregious spelling mistakes in commit comments!

kriegfrj:
> Oops. But not worth fixing. I have more egregious spelling mistakes in commit comments!

:smile: The only reason I bothered mentioning this one at all is because it is a key word that someone *might* potentially want to search on in the future.  But the chances that this will be a problem are slim.

PR #4209, opened 2020-07-02
maven: Simplify junit5 enum value

bjhargrave:
We now use the marker annotation
org.junit.platform.commons.annotation.Testable.

### DIFF
--- a/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/TestCases.java
+++ b/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/TestCases.java
@@ -4,11 +4,7 @@ public enum TestCases {
 
 	junit3("${classes;EXTENDS;junit.framework.TestCase;CONCRETE}"),
 	junit4("${classes;HIERARCHY_ANNOTATED;org.junit.Test;CONCRETE}"),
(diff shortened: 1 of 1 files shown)


kriegfrj:
> We now use the marker annotation
> org.junit.platform.commons.annotation.Testable.

I think that this is a good idea, but technically incorrect. To include all classes with `@Testable` tests is not the same thing as including all JUnit5 tests, as there may be test engines other that Jupiter that meta-annotate using `@Testable`.

At the moment, I don't think there are any, so this may be only an issue in theory for now. But if we want to future-proof this implementation, we could consider leaving the "junit5" definition as it was and adding a "testable" enum value which includes all `@Testable` tests. "all" could then be JUnit 3 + JUnit 4 + testable.

However, we could simplify the existing JUnit5 enum implementation by leaving out `@RepeatedTest` and `@ParameterizedTest`, as these are meta-annotated by `@TestTemplate`.

kriegfrj:
Also need to update https://github.com/bndtools/bnd/blob/master/maven/bnd-maven-plugin/README.md

bjhargrave:
I think `Testable` is probably correct for most users of JUnit 5. If people have special needs, they can set their own test case names with the custom option.

bjhargrave:
> Also need to update https://github.com/bndtools/bnd/blob/master/maven/bnd-maven-plugin/README.md

Fixed. Thanks!

Issue #4213, opened 2020-07-04
Support for further test frameworks


gurkerl83:
The bnd-process-tests goal provided in the Bnd-Maven plugin currently only supports JUnit. Other test frameworks, such as TestNG are not supported.

A developer will face a difficult to track error message when using the new bnd-process-tests goal in addition to the standard Bnd-Maven plugin and the bnd-process goal.

If the function and effect of the bnd-process-tests goal are first elaborated on a small sample project, it is clear that the JUnit framework annotation "Test" must be used at least once in a test to avoid provoking the error message listed at the end.

If the new target is not elaborated on a sample project in advance, the principles of using the target will be questioned. I can only speak from our experience in using the goal, and unfortunately, the behavior of the goal was not tested like this in advance. We have assumed that the behavior is equivalent to the standard goal and is not bound to the use of a specific testing framework.

Those who switch from using the Maven Bundle plugin to the Maven Bnd plugin will not have the primary testing framework used in a project in view when using the additional goal.

How additional test frameworks can be supported, the question arises as to how this could be done in a way that is as independent as possible. Of course, an extension of the Enums TestCase with additional frameworks would be the easiest approach. 

I may be completely wrong here, but the goal itself does not interfere with the execution control of tests; this is still reserved for other Maven plugins? Is it not possible to determine if results are being generated by upstream test execution?

Error message when no JUnit framework test annotation is used.

```
biz.aQute.bnd:bnd-maven-plugin:5.1.1:bnd-process-tests (bnd-process-tests) on project xyz: ${sort}, is not understood. too few arguments, for cmd: sort, arguments; [sort]
```

kriegfrj:
Hi @gurkerl83 , thanks for raising this issue.

## Executive Summary:

The error message that you showed is quite cryptic - I think we could do better to more directly detect and report missing test cases. But to the main issue: while I've not directly tested it, I think that Bnd 5.0 and greater can already do what you are asking.

## Detail: 

[...]

Issue #4088, opened 2020-05-21
[maven/m2e] Warn when testCases has invalid data


kriegfrj:
I was trying to embed my Test-Cases header macro directly into the `<testCases>` element and wondering why it wasn't working.

Upon reading the docs more closely, I can see it's not designed for that - it only takes one of a few options (one of which is `useTestCasesHeader`, which I can use to do what I want). But instead of issuing an error, it silently ignored the error and used the default value (junit5).

It would be nice if it weren't so silent. Either warn and continue with the default, or else error and stop execution. I'm leaning towards the latter.

bjhargrave:
I think this may be a maven issue. It is trying to coerce the string value in the `<useTestCasesHeader>` XML element into an enum value. I don't know how is handles a string value which does not correspond to an enum value? It may just use the default (`junit5`) without allowing our plugin to know of the coercion failure.

kriegfrj:
I'd be surprised if Maven didn't have better handling of this but I don't have the time to investigate.

Another possibility (if Maven cannot be coerced into doing the right thing) is that we declare it as a String and do the conversion to enum in the plugin - this would give us the opportunity to do the sanity checking ourselves.

bjhargrave:
> I was trying to embed my Test-Cases header macro directly into the `<testCases>` element and wondering why it wasn't working.

I tried `<testCases>${classes;HIERARCHY_ANNOTATED;org.testng.annotations.Test;CONCRETE}</testCases>` with debug output showing.

```
<testCases default-value="junit5">${classes;HIERARCHY_ANNOTATED;org.testng.annotations.Test;CONCRETE}</testCases>
```

this becomes:

```
testCases = junit5
```

in the actual configuration.

Maven is trying to evaluate `${classes;HIERARCHY_ANNOTATED;org.testng.annotations.Test;CONCRETE}` as a Maven property and, not finding such a property, uses an empty string.

If you specify `<testCases>${some.missing.property.name}</testCases>` or `<testCases></testCases>`, you get the same result: the default value is used.

However, if I specify a bad value `<testCases>badvalue</testCases>`, then Maven properly complains:

[...]

PR #4216, opened 2020-07-06
Maven Test-Cases header

bjhargrave:
Stop erroring if sort list is empty. Warn is Test-Cases header is empty for bnd-process-tests.

See https://github.com/bndtools/bnd/issues/4213.


kriegfrj:
See also #4088

PR #4222, opened 2020-07-10
set equinox configuration app args

gamerson:


org.bndtools.applaunch/src/org/bndtools/applaunch/Activator.java
```diff
@@ -69,6 +73,38 @@ private static Filter createFilter() {
...
+							equinoxConfig.setAllArgs(parameters);
+							equinoxConfig.setAppArgs(parameters);
+						}
+					}
+
+					log.fine("configured program arguments " + parameters);
+				}
+			} else {
+				log.log(Level.SEVERE,
+					"service aQute.launcher.Launcher with props launcher.arguments could not be retireved");
```
rotty3000, line 101: Typo in `retrieved`

org.bndtools.applaunch/src/org/bndtools/applaunch/Activator.java
```diff
@@ -69,6 +75,39 @@ private static Filter createFilter() {
...
+					if (srvcRef != null) {
+						EnvironmentInfo service = context.getService(srvcRef);
+						if (service instanceof EquinoxConfiguration) {
+							EquinoxConfiguration equinoxConfig = (EquinoxConfiguration) service;
+							equinoxConfig.setAllArgs(args);
+							equinoxConfig.setAppArgs(args);
+						}
+					}
+
+					log.fine("configured program arguments " + Arrays.stream(args)
```
bjhargrave, line 99: I think `Arrays.toString(args)` is probably fine and simpler.

Why is this log `log.fine(` when the other logs are like `log.log(Level.`? It seems a bit inconsistent.

org.bndtools.applaunch/src/org/bndtools/applaunch/Activator.java
```diff
@@ -69,6 +75,39 @@ private static Filter createFilter() {
...
+				"(launcher.arguments=*)");
+			if (serviceReferences != null && serviceReferences.length > 0) {
+				Object property = serviceReferences[0].getProperty("launcher.arguments");
+
+				if (property instanceof String[]) {
+					args = (String[]) property;
+
+					ServiceReference<EnvironmentInfo> srvcRef = context.getServiceReference(EnvironmentInfo.class);
+					if (srvcRef != null) {
+						EnvironmentInfo service = context.getService(srvcRef);
```
bjhargrave, line 91: What about `unget`ing the service?

org.bndtools.applaunch/src/org/bndtools/applaunch/Activator.java
```diff
@@ -69,6 +75,39 @@ private static Filter createFilter() {
...
+						if (service instanceof EquinoxConfiguration) {
+							EquinoxConfiguration equinoxConfig = (EquinoxConfiguration) service;
+							equinoxConfig.setAllArgs(args);
+							equinoxConfig.setAppArgs(args);
```
bjhargrave, line 95: Why are the command line args (`setAllArgs`) and the non-framework args (`setAppArgs`) set to the same value?

gamerson: AppArgs is all I need so I'll remove the other.

org.bndtools.applaunch/src/org/bndtools/applaunch/Activator.java
gamerson: I'll add the unget

org.bndtools.applaunch/src/org/bndtools/applaunch/Activator.java
gamerson: I'll clean those up.

org.bndtools.applaunch/src/org/bndtools/applaunch/Activator.java
```diff
@@ -69,8 +75,40 @@ private static Filter createFilter() {
...
+					"service aQute.launcher.Launcher with props launcher.arguments could not be retrieved");
+			}
+		} catch (Exception e) {
+			log.log(Level.SEVERE, "command line parameters could not be configured.");
+		}
+	}
+
 	@Override
 	public void removedService(ServiceReference<Object> reference, ServiceRegistration<ApplicationLauncher> service) {
+		context.ungetService(envInfo);
```
bjhargrave, line 111: `envInfo` could be null here, no?

gamerson: ah yes

PR #4223, opened 2020-07-10
[junit-platform] Add pruning functionality

### LINKED ISSUE #4128: [junit-platform] Optionally prune singleton non-leaf nodes from test hierarchy
In a departure from the original bnd tester, the new junit-platform tester adds additional levels of hierarchy into the test output:

test bundle -> test fragment (optional) -> test engine -> regular JUnit hierarchy

I did this because I envisioned that large integration tests could have lots of tests from lots of different bundles (even potentially having tests with the same name from different bundles), and having some sort of visual representation of this would allow one to navigate it more easily.

@rotty3000 first mentioned in #3673 that he found this extra depth annoying. Having played with osgi-test for a while now I have come to appreciate this position a bit better - while the feature I added will be helpful for other use cases, I can see that a common use case will be "one test bundle, one engine" (and maybe one fragment). This is how osgi-test is structured, for example. In such cases, the representation of the bundle [, fragment] and test engine nodes do not add any value - in fact, they just get in the way.

As I mentioned in #3673 (and indeed in the original PR #3145), I think the best way to get the best of both worlds is to have a "prune" function that removes singleton non-leaf nodes (ie, those with no siblings) from the test hierarchy. This feature would be enabled by default but could be disabled if you really want to see the full hierarchy always. I guess this is similar in principle to how GitHub's file tree browser skips over empty directories.

One question that I wanted to put to the team for consideration: how far down the hierarchy should this extend? There are three options that make sense to me:

1. `BundleEngine` prunes only its part of the hierarchy and preserves the hierarchy generated by the sub-engine(s) as-is.
2. `BundleEngine` also prunes the hierarchy of sub-engines using the same algorithm.
3. Configurable so that you can do either 1 or 2 (or neither).

Thoughts?

kriegfrj:
Fixes #4128

This is a draft. I'd like some feedback on the syntax and functionality, once that's settled I'll write the doc and re-submit as a full PR.

There is a new tester directive called "tester.prune". It has the following syntax:

`tester.prune=<nodetype>=<prunemode>[,<nodetype>=<prunemode>]`

The supported node types are `bundle`, `fragment` and `sub-engine`.

The supported prune modes are `never` (default), `onlychild` (prune if the node has no siblings) and `always`.

Had to jump through a few hoops to make this work but this seems to do the trick.

### DIFF
--- a/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/BundleDescriptor.java
+++ b/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/BundleDescriptor.java
@@ -1,13 +1,5 @@
 package aQute.tester.bundle.engine;
 
-import java.util.HashMap;
-import java.util.Map;
-
-import org.junit.platform.engine.ConfigurationParameters;
-import org.junit.platform.engine.EngineExecutionListener;
-import org.junit.platform.engine.ExecutionRequest;
-import org.junit.platform.engine.TestDescriptor;
-import org.junit.platform.engine.TestEngine;
 import org.junit.platform.engine.UniqueId;
 import org.junit.platform.engine.support.descriptor.AbstractTestDescriptor;
 import org.osgi.framework.Bundle;
@@ -17,7 +9,6 @@ public class BundleDescriptor extends AbstractTestDescriptor {
 
 	final private Bundle							bundle;
 	final private BundleException					bundleException;
-	final private Map<TestDescriptor, TestEngine>	engineMap	= new HashMap<>();
 
 	public static String displayNameOf(Bundle bundle) {
 		return "[" + bundle.getBundleId() + "] " + bundle.getSymbolicName() + ';'
@@ -38,17 +29,6 @@ public class BundleDescriptor extends AbstractTestDescriptor {
 		return bundle;
 	}
 
-	public void addChild(TestDescriptor descriptor, TestEngine engine) {
-		engineMap.put(descriptor, engine);
-		addChild(descriptor);
-	}
-
-	public void executeChild(TestDescriptor descriptor, EngineExecutionListener listener, ConfigurationParameters params) {
-		TestEngine engine = engineMap.get(descriptor);
-		ExecutionRequest er = new ExecutionRequest(descriptor, listener, params);
-		engine.execute(er);
-
-	}
 	@Override
 	public Type getType() {
 		return bundleException == null ? Type.CONTAINER : Type.TEST;
--- a/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/BundleEngine.java
+++ b/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/BundleEngine.java
@@ -1,31 +1,27 @@
 package aQute.tester.bundle.engine;
 
 import java.util.Optional;
-import java.util.stream.Stream;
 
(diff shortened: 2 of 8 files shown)


bjhargrave:
I am not sure what pruning is here in this context. Is this means to avoid running certain tests (like vintage tests). Or to avoid displaying the output of certain tests? Or how the output of all the tests are organized? That is, the term prune is not meaningful enough here to tell me what you intend it to do.

I am not sure that how the tests are displayed is something that should be configurable by a user. We should decide what the best display format is and do that. 

Frankly, I am more than fine with things as they currently are. In the normal case, all test pass and I see green. If a few tests fail, I can easily see the few failures. If many tests fail, then I have a BIG problem to solve and how the many failed tests are displayed is the least of my issues :-)

kriegfrj:
> I am not sure what pruning is here in this context. Is this means to avoid running certain tests (like vintage tests). Or to avoid displaying the output of certain tests? Or how the output of all the tests are organized? That is, the term prune is not meaningful enough here to tell me what you intend it to do.
> 

Of the three options you've presented, I guess it is closest to the third: "how the output of all the tests are organised". Let me explain in a bit more detail.

It is based on @rotty3000 's original complaint, which I came to appreciate more when working on `osgi-test` (#4128). It is to (optionally) avoid *creating* the nodes higher up in the hierarchy. I'll continue to use the word "prune" to explain the operation, until we can settle on an more appropriate term (I agree, the term "prune" is technically not a 100% correct description of the operation).

By default, BundleEngine always outputs the hierarchy like this:

`bundle -> [fragment] -> sub-engine -> <sub-engine tests>`

If I "prune" the sub-engine node, then the output will look like this:

`bundle -> [fragment] -> <sub-engine tests>`

Note that the sub-engine's tests are still included in the output - it's just that they are included directly as children of the bundle (or fragment) rather than having the sub-engine appear as an intermediate node in the hierarchy. (Hence I agree that "prune" is not 100% accurate, as that would imply that the sub-engine is removed along with all its children.)

* If you set the sub-engine "prune" mode to "always", then the sub-engine nodes will always be "pruned".
* If you set the sub-engine "prune" mode to "never", then the sub-engine nodes will always be included.
* If you set the sub-engine "prune" mode to "onlychild", then the sub-engine nodes will be "pruned" iff there is only one sub-engine for that bundle [fragment]. If there is more than one then the sub-engine nodes will not be "pruned".

Similar applies to fragments and bundles (except that when a bundle is "pruned", its fragments will automatically get "pruned" also).

> I am not sure that how the tests are displayed is something that should be configurable by a user. We should decide what the best display format is and do that.
>

I guess that's a good point to debate. At the end of the day though, if the user doesn't want to configure it they don't have to.

> Frankly, I am more than fine with things as they currently are. In the normal case, all test pass and I see green. If a few tests fail, I can easily see the few failures. If many tests fail, then I have a BIG problem to solve and how the many failed tests are displayed is the least of my issues :-)

I can see your point, but I can also see the utility in simplifying the output for simpler test fixtures.

I think it would be nice if @rotty3000 could weigh in on this - see if the bundle/fragment hierarchy still bugs him or if he's come to terms with it over time. :smile:

rotty3000:
I don't think there is huge value in a lot of configuration here. Let's go with simplest solution.

djencks:
I have no idea if this is a good idea, but for a name ‘graft’ seems more appropriate than ‘prune’

PR #4231, opened 2020-07-16
[bndtools launch] Unify launch delegates' pre-launch validation logic

### LINKED ISSUE #3954: [launcher] OSGiJUnitLaunchDelegate silently skips project errors
If you have a bundle spec in `-runbundles` and the launcher doesn't find a matching bundle, it will silently skip over it without warning (even with `-runtrace` enabled). If this is by design, why? If not, then perhaps we should fix it? It helps alert the developer to a problem earlier than they may otherwise be alerted.

kriegfrj:
Move the pre-launch validation logic from `OSGiRunLaunchDelegate` into `AbstractOSGiLaunchDelegate`, and remove the custom logic from `OSGiJUnitLaunchDelegate`. Net effect is to have both delegates using the same pre-launch validation.

Fixes #3954.

### DIFF
--- a/bndtools.core/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
+++ b/bndtools.core/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
@@ -1,7 +1,9 @@
 package bndtools.launch;
 
 import java.io.File;
+import java.io.IOException;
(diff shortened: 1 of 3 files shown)


bjhargrave:
Nice!

PR #4228, opened 2020-07-13
[quickfix] Quickfix enhancements

kriegfrj:
* Improved support for quickfix in new situations (#4195).
* Reinstated some support for static imports that got broken in the new implementation.
* Implemented a facade so that the quick fix processor can be dynamically re-installed without restarting the workbench (#4227).

PR #4239, opened 2020-07-29
[generate] Typo fixes for generate command

kriegfrj:


bjhargrave:
https://github.com/bndtools/bnd/pull/4239/checks?check_run_id=923129733#step:5:549

We get intermittent failures in `bndtools.core.test.editors.quickfix.BuildpathQuickFixProcessorTest` in the CI builds. I don't think this is caused by this PR, but there is some timing issue in this test class which rears up from time to time. Can you look into this?

kriegfrj:
> https://github.com/bndtools/bnd/pull/4239/checks?check_run_id=923129733#step:5:549
> 
> We get intermittent failures in `bndtools.core.test.editors.quickfix.BuildpathQuickFixProcessorTest` in the CI builds. I don't think this is caused by this PR, but there is some timing issue in this test class which rears up from time to time. Can you look into this?

I've already made an attempt at fixing this in #4228, see: https://github.com/bndtools/bnd/blob/f603f1d3d0df2fb8bc2b3a93b008fe9dd619fa06/bndtools.core.test/src/bndtools/core/test/editors/quickfix/BuildpathQuickFixProcessorTest.java#L101-L114

If @pkriens (or someone else) can review #4228 and we can get it merged soon, we can see if that fixes the issue. If that merge is too far in the future then I will try and isolate that one change and make a separate PR for it.

bjhargrave:
> I've already made an attempt at fixing this in #4228,

That is a big PR. I think it make take a little while for @pkriens and @rotty3000 to eyeball it.

> I will try and isolate that one change and make a separate PR for it.

Yes (or even this PR).

kriegfrj:
> > I will try and isolate that one change and make a separate PR for it.
> 
> Yes (or even this PR).

I was afraid you'd say that.  :) hopefully tomorrow.

bjhargrave:
Thanks for the fix!

PR #4241, opened 2020-07-30
Add support for importing java.* packages for OSGi Core R7 

### LINKED ISSUE #2507: OSGi Java 9 support: Importing java.* packages
When building a Java 9 bundle, Bnd needs to import java.* packages.

bjhargrave:
This can be disabled with `-noimportjava: true`. The support allows
control over the specific java packages imported via the `Import-Package`
instruction. The support is only enabled if either the bundle imports
the `org.osgi.framework` package at version 1.9 (Core R7), or higher, or
if the bundle includes class files for Java 11, or higher.

Fixes #2507

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/ImportJavaTest.java
+++ b/biz.aQute.bndlib.tests/test/test/ImportJavaTest.java
@@ -1,0 +1,138 @@
+package test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.io.File;
+import java.lang.reflect.Method;
+import java.util.jar.Manifest;
+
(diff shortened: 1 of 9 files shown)


pkriens:
I had not noticed that we now import java. Was it discussed to so drastically change the default? 

In my case we compare bundles against a previous release and only update when needed. This is creating a massive change. 

I think the default should be as it always ways and then an opt-in.

bjhargrave:
Yes, it was discussed at length in #2507. I went with the consensus decision from that discussion albeit with some tweaks. 

This is in some way parallel to when we automatically added osgi.ee requirements which the `-noee` instruction to opt-out.

If we want to tweak how and when we want to import java.* packages, we should probably open a new issue to discuss. This PR is already merged.

PR #4244, opened 2020-08-01
[junit-platform] Pluggable test listeners

### LINKED ISSUE #3526: [junit-platform tester] Pluggable test listeners
Now that the JUnit Platform port is merged and being used, I'd like to propose a feature for consideration that I had been thinking about during development.

The current implementation of `biz.aQute.tester.junit-platform` (following in the footsteps of `biz.aQute.tester`) hard-codes the list of `TestListener`s during setup. This doesn't seem to be a particularly "OSGi" way of doing things.

I was thinking of changing this to use SCR to implement pluggable TestListeners. Then you can set up your system to add the listeners that you want. Some examples of how this might be useful:

* JUnit 5 has some nifty TestListener implementations that produce pretty output on a text console (including colours). Unfortunately they are hidden away in a package-private implementation in the `junit-platform-console` module at the moment, but there is an open feature request to make these publicly accessible (https://github.com/junit-team/junit5/issues/1963). It would be nice to be able to leverage these. I have had these working before in conjunction with `biz.aQute.tester.junit-platform` using forks of JUnit 5 and Bndtools.
* EclipseJUnitListener could be factored out and only deployed when actually needed.
* People might like to write their own custom test reporting tools. This gives them the ability to do so.

Thoughts?

kriegfrj:
Find any OSGi services registered under the TestExecutionListener interface and use them as listeners for the test execution.

Fixes #3526.

Hardest thing about this was coming up with a decent regression test.

I'm also interested in feedback from some of the OSGi hardcores about if & how `biz.aQute.tester.junit-platform` should handle service dynamics - what happens if a test listener disappears during the test run? Perhaps I can have an implementation of TestExecutionListener that maintains the service tracker, and getting a fresh list of services on every invocation to delegate the invocation to.


kriegfrj:
This needs a bit of doc as well. I'll add the doc once the implementation is bedded down.

biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/Activator.java
```diff
@@ -209,6 +210,16 @@ public void executionSkipped(TestIdentifier testIdentifier, String reason) {
...
+		ServiceTracker<TestExecutionListener, TestExecutionListener> track = new ServiceTracker<>(context,
+			TestExecutionListener.class, null);
+		track.open(true);
+		Optional.ofNullable(track.getServices())
+			.ifPresent(services -> Stream.of(services)
+				.map(TestExecutionListener.class::cast)
+				.forEach(listener -> {
+					listenerList.add(listener);
+				}));
+		track.close();
```
bjhargrave, line 222: This is problematic. When you close the tracker, you release the services. Which means you should no longer use the service objects. But you clearly keep using the service objects here. A bundle providing the service could use a ServiceFactory and disable the object in someway when the service is released. It is never good practice to use a service object after you release the service.

A tracker like this should be created and opened in the bundle's activator start method and closed in the stop method.

kriegfrj: Thanks for the clarification!

I thought about opening it in the activator start, but in the end I opted for opening it just before the test run starts. The reason for this is because I would like at some time to add the option of supplying a filter to the tracker constructor so that you can control which listeners are included in the test run on a per-run basis. If ever we get the Gogo commandlet working, then in a continuous testing scenario it might be nice to be able to dynamically change that filter and have it take effect straight away without having to restart the test bundle (restarting the test bundle will cause the framework to exit if it has been launched on the main thread). See if what I've done is ok and if not I'll change it.

biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/Activator.java
```diff
@@ -348,11 +344,21 @@ LauncherDiscoveryRequest buildRequest(List<? extends DiscoverySelector> selector
 	long test(LauncherDiscoveryRequest testRequest) {
 		trace("testing request %s", testRequest);
 		try {
+			ServiceTracker<TestExecutionListener, TestExecutionListener> track = new ServiceTracker<>(context,
+				TestExecutionListener.class, null);
+			track.open(true);
```
bjhargrave, line 349: You do not want `true` here. Using `true` means you want to track all services including services which are not type compatible with your bundle. So you could track a `TestExecutionListener` service which this bundle cannot cast to `TestExecutionListener`.

Just use `track.open();`

kriegfrj: Done, thanks (I obviously misunderstood the docs around `open(true)`.)

biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/Activator.java
```diff
@@ -209,6 +210,16 @@ public void executionSkipped(TestIdentifier testIdentifier, String reason) {
...
+		ServiceTracker<TestExecutionListener, TestExecutionListener> track = new ServiceTracker<>(context,
+			TestExecutionListener.class, null);
+		track.open();
+		Optional.ofNullable(track.getServices())
+			.ifPresent(services -> Stream.of(services)
+				.map(TestExecutionListener.class::cast)
+				.forEach(listener -> {
+					listenerList.add(listener);
+				}));
+		track.close();
```
bjhargrave, line 222: You revert back to the bad behavior of closing the tracker before using the tracked services.

kriegfrj: Oops, something went wrong with my git pull before i made the change. Will have to fox it tomorrow now... sorry!

kriegfrj: Fixed.

kriegfrj:
Latest push also adds some documentation for this new feature.

PR #4248, opened 2020-08-04
[m2e] MavenRunProvider to handle SOURCES mode

### LINKED ISSUE #4130: [debug m2e] Bnd Dependencies Container doesn't find bundles under m2e
Under m2e, the new debug source lookup mechanism (see #3993) doesn't seem to work. It only includes the ~~`-runvm`~~ `-runee` property and that imperfectly.

kriegfrj:
Fixes #4130


bjhargrave:
Thanks!

PR #4251, opened 2020-08-05
[Reporter] show component-description in readme

stbischof:
add a description section into the component. 'service.description' property is used.


bjhargrave:
Thanks!

PR #4254, opened 2020-08-07
[junit-platform] Added Gogo commandlet

### LINKED ISSUE #3812: [tester] Gogo commandlet for invoking tests
In a continuous testing environment, it would sometimes be nice to be able to invoke tests from the shell. This should be achievable with a simple gogo shell wrapper around the existing testing infrastructure.

Suggestions (including how useful an idea this might be) would be appreciated.

kriegfrj:
If Gogo is in the environment, then in automatic mode the tester will register a "runtests" command that allows you to interactively run tests from the shell.

scope: tester
command syntax: runtests :my.bsn.name my.fq.ClassName my.fqClassName#forMethod

Fixes #3812

### DIFF
--- a/biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/Activator.java
+++ b/biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/Activator.java
@@ -32,6 +32,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 
+import org.apache.felix.service.command.Descriptor;
 import org.junit.platform.engine.DiscoverySelector;
 import org.junit.platform.engine.TestExecutionResult;
 import org.junit.platform.engine.support.descriptor.ClassSource;
@@ -62,18 +63,19 @@ import aQute.tester.junit.platform.utils.BundleUtils;
 
 @Header(name = Constants.BUNDLE_ACTIVATOR, value = "${@class}")
 public class Activator implements BundleActivator, Runnable {
-	String								unresolved;
-	Launcher							launcher;
-	BundleContext						context;
-	volatile boolean					active;
-	boolean								continuous	= false;
-	boolean								trace		= false;
-	PrintStream							out			= System.out;
-	JUnitEclipseListener				jUnitEclipseListener;
-	volatile Thread						thread;
-	private File						reportDir;
-	private SummaryGeneratingListener	summary;
-	private List<TestExecutionListener>	listeners	= new ArrayList<>();
+	String									unresolved;
+	Launcher								launcher;
+	BundleContext							context;
+	volatile boolean						active;
+	boolean									continuous	= false;
+	boolean									trace		= false;
+	PrintStream								out			= System.out;
+	JUnitEclipseListener					jUnitEclipseListener;
+	volatile Thread							thread;
+	private File							reportDir;
+	private SummaryGeneratingListener		summary;
+	private List<TestExecutionListener>		listeners	= new ArrayList<>();
+	final BlockingDeque<DiscoverySelector>	queue		= new LinkedBlockingDeque<>();
 
 	public Activator() {}
 
@@ -127,6 +129,8 @@ public class Activator implements BundleActivator, Runnable {
 			.enableTestEngineAutoRegistration(false)
 			.addTestEngines(new BundleEngine())
 			.build());
+
+		List<TestExecutionListener> listenerList = new ArrayList<>();
 
 		String testcases = context.getProperty(TESTER_NAMES);
 		trace("test cases %s", testcases);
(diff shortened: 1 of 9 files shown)


bjhargrave:
@kriegfrj What is the status on this? For 5.2.0?

kriegfrj:
> @kriegfrj What is the status on this? For 5.2.0?

Should be able to go in 5.2.0. There doesn't seem to be a sensible reason for the lone failing test on JDK 14 and I couldn't see much output in the log. I have rebased onto master and re-pushed to see what happens.

bjhargrave:
Ooops. I think you need to rebase again. I created a conflict when I updated junit to 4.13.

kriegfrj:
Done!

biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/Activator.java
```diff
@@ -295,12 +297,26 @@ public void removedBundle(Bundle bundle, BundleEvent event, Bundle object) {
...
+				GogoCommandlet cmd = new GogoCommandlet();
+				Hashtable<String, Object> properties = new Hashtable<>();
+				properties.put("osgi.command.function", new String[] {
+					"runTests"
+				});
+				properties.put("osgi.command.scope", "tester");
+
+				trace("starting gogo commandlet");
+				context.registerService(Object.class, cmd, properties);
+			} catch (NoClassDefFoundError e) {}
```
bjhargrave, line 311: Can you even get an NCDFE here? The only gogo type is Description which is only used as an annotation and the class loader will be happy to load the GogoCommandlet class even if the annotations are not available. So I don't think you can ever get the error.

biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/Activator.java
```diff
@@ -295,12 +297,26 @@ public void removedBundle(Bundle bundle, BundleEvent event, Bundle object) {
...
+			try {
+				GogoCommandlet cmd = new GogoCommandlet();
+				Hashtable<String, Object> properties = new Hashtable<>();
+				properties.put("osgi.command.function", new String[] {
+					"runTests"
+				});
+				properties.put("osgi.command.scope", "tester");
+
+				trace("starting gogo commandlet");
+				context.registerService(Object.class, cmd, properties);
```
bjhargrave, line 310: Perhaps this should only be registered if in continuous mode? Otherwise there is no loop watching the queue. The test case sets continuous on to get the test to work I assume.

bjhargrave (changes requested):

biz.aQute.tester.test/bnd.bnd
```diff
@@ -5,11 +5,12 @@
...
 	org.apache.servicemix.bundles.junit;version=latest,\
-	org.junit.jupiter.api;version="${junit.jupiter.eclipse.version}",\
-	org.junit.jupiter.params;version="${junit.jupiter.eclipse.version}",\
+	org.junit.jupiter.api;version='${junit.jupiter.eclipse.version}',\
+	org.junit.jupiter.params;version='${junit.jupiter.eclipse.version}',\
 	org.opentest4j,\
 	org.apiguardian,\
-	osgi.core;version=latest
+	osgi.core;version=latest,\
+	junit-jupiter-api;version='5.6.2'
```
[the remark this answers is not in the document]
kriegfrj: Sorry, looks like I got a bit trigger-happy on the quick fix at one point (the presence of the version attribute is a dead giveaway!)

biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/Activator.java
kriegfrj: Well I've learned something. I didn't know that the classloader would ignore annotations if their type was missing. I've removed it and the tests pass. In fact, it didn't occur to me (or at least I never seriously investigated) that the gogo-specific tests were running and passing even though I hadn't added gogo to the command line.

And this is also a lesson to me, that this is the kind of thing that happens when you don't do proper TDD! I obviously never attempted (or at least never succeeded) to write that failing test to trigger the NCDFE.

biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/Activator.java
kriegfrj: > Perhaps this should only be registered if in continuous mode? Otherwise there is no loop watching the queue.

In auto, non-continuous mode, the tester can still it can still sit waiting if it doesn't find any test bundles. In theory, at least, the commandlet could still be used at this time. However if there are no test bundles in the environment it won't do much.

I'll change it to only register during continuous mode.

> The test case sets continuous on to get the test to work I assume.

Yes:
https://github.com/bndtools/bnd/blob/62dbaa2683a50c091393821995ad2a0ccb597bc7/biz.aQute.tester.test/test/aQute/tester/junit/platform/test/GogoShellTests.java#L76

bjhargrave (approved):

bjhargrave:
Thanks!

PR #4259, opened 2020-08-10
fixed # 4235 NPE on Resolve job cancel

### LINKED ISSUE #4235: NPE on Resolve job cancel
@scottslewis  reported: 
So...when I hit resolve with this bndrun...without one of the requirements:
screen1
after just a second or so I get:
screen2

Which is fine.

If, however I add one more requirement (org.eclipse.ecf.provider.jersey) in this bndrun, and save it...
screen3

*then* hit Resolve  I get  the above...i.e. the 'Resolving health.examples.grpc.server.bndrun' never completes on it's own.

If I hit the stop button on the right of the Progress view I get:
screen4

and there is nothing in the Errors view that indicates what went wrong...or any info about the NPE or previous hang in the workspace log.

What I would have expected:  Rather than having me cancel via the progress view (which may not be present)...is that a progress monitor dialog would appear after hitting Resolve...and that I could cancel the resolve from this progress monitor dialog.   A progress monitor dialog is common in Eclipse for cancellable long-running operations.

On hang+cancellation, I would expect something other than the Problem Occurred/NPE dialog as above...i.e. something that allows me to access the resolve log (or something that allows me to determine or at least investigate why the resolver was not completing).

Currently using this version of bndtools: 
screen5

If I can test or help please let me know.   If you desire I can give you steps to run this example (from the ECF remote services bndworkspace here:

https://github.com/ECF/bndtools.workspace

Thanksinadvance,

Scott

juergen-albert:
Fixes #4235

### DIFF
--- a/bndtools.core/src/org/bndtools/core/resolve/ResolveOperation.java
+++ b/bndtools.core/src/org/bndtools/core/resolve/ResolveOperation.java
@@ -17,6 +17,7 @@ import org.osgi.service.coordinator.Coordinator;
 import org.osgi.service.resolver.ResolutionException;
 
 import aQute.bnd.build.model.BndEditModel;
+import aQute.lib.exceptions.Exceptions;
 import aQute.lib.exceptions.RunnableWithException;
 import biz.aQute.resolve.ResolutionCallback;
 import biz.aQute.resolve.ResolverLogger;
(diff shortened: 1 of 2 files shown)


bjhargrave:
Thanks!

PR #4260, opened 2020-08-11
[template] Add parameter support for workspace templates

kriegfrj:
I had a need for custom parameters in a workspace template I was trying to set up.

Also had a need to configure a password in the template, so added support for password attributes.

### DIFF
--- a/bndtools.core/src/org/bndtools/core/ui/wizards/shared/TemplateParamsWizardPage.java
+++ b/bndtools.core/src/org/bndtools/core/ui/wizards/shared/TemplateParamsWizardPage.java
@@ -156,10 +156,14 @@ public class TemplateParamsWizardPage extends WizardPage implements ISkippableWi
 	}
 
 	private Control createFieldControl(Composite parent, final AttributeDefinition ad) {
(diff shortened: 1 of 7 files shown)


bndtools.core/src/org/bndtools/core/ui/wizards/shared/TemplateParamsWizardPage.java
```diff
@@ -156,10 +156,13 @@ private void updateValidation() {
 	}
 
 	private Control createFieldControl(Composite parent, final AttributeDefinition ad) {
+		int textType = SWT.BORDER;
 		switch (ad.getType()) {
+			case AttributeDefinition.PASSWORD :
+				textType |= SWT.PASSWORD;
```
bjhargrave, line 162: You should put in a comment to document the intended FALL-THROUGH here.

kriegfrj: Done

bjhargrave:
Thanks!

Issue #4253, opened 2020-08-07
[quickfix] Integration test is flaky


kriegfrj:
There appears to be a race condition in the startup of the integration test that causes the `beforeAll()` method of the integration test to fail to get a local repo:

https://github.com/bndtools/bnd/blob/b4cddedfe8f233e1005828d00f3c83a1345611bc/bndtools.core.test/src/bndtools/core/test/editors/quickfix/BuildpathQuickFixProcessorTest.java#L77-L91

This needs to be fixed.

bjhargrave:
This is still happening. https://github.com/bndtools/bnd/pull/4278/checks?check_run_id=1020131734#step:5:587

bjhargrave:
https://github.com/bndtools/bnd/runs/1031734002?check_suite_focus=true#step:5:743

kriegfrj:
Yes, I've been keeping an eye on it while working on #4279. Recording some observations to aid further investigation:

1. It (or similar problems) happened occasionally even while continuous testing. When continuous testing the quick fix processor, the core Eclipse and Bndtools don't restart. So whatever the problem is, it probably doesn't have to do with Bndtools/JDT startup issues.
2. I managed to dump a snapshot of the workspace resource hierarchy at one point when it happened - it looked normal.

So I think the issue lies somewhere in the code that initialises the Eclipse projects. The WorkbenchExtension creates a temp directory to use as an empty workspace at startup, and then the WorkspaceImporter is used to import the template workspace from the resources folder into the temporary workspace to work on. It seems that sometimes, for some reason, Bndtools is not picking up the changes to the workspace. Sometimes this results in `bndProject` coming back `null`, and sometimes it results in the LocalIndexedRepo coming back `null`.

One theory is that despite the synchronisation efforts, the attempt to fetch the Bnd config is a bit too early and the workspace isn't ready. Bndtools caches the config and then doesn't notice when the workspace is updated.

Any other suggested avenues of investigation welcomed.

bjhargrave:
https://github.com/bndtools/bnd/pull/4285/checks?check_run_id=1049644892#step:5:621

bjhargrave:
https://github.com/bndtools/bnd/runs/1102242709#step:5:613

kriegfrj:
Has this disappeared? Or have you simply given up reporting when it happens?

bjhargrave:
> Has this disappeared?

No.

> Or have you simply given up reporting when it happens?

[...]

PR #4268, opened 2020-08-14
[quickfix] More enhancements

kriegfrj:
- Disambiguate cases where there is no ambiguity (eg, my.package.name<type>.something will not attempt to look in package "my.package.name" as it is clear from that context that ".name" is a class, not a package. Enables us to reinstate most of the original tests that had not already been reinstated.
- I noticed while working with Bnd that certain types of class literal expression for missing types would not produce suggestions, even though they were literals of classes that Bnd should have been able to find. This has now been fixed.
- If multiple problem markers produce multiple suggestions for the same bundle, the proposal for that bundle will now appear only once (previously, the proposal was repeated for each problem marker). If the proposals related to distinct types, then each of those distinct types will be listed in the proposal summary in the "found..." section.
- I have re-instated the integration test with some extra debugging checks to produce more diagnostics. Hopefully if/when we get another failure I'll have more information to go on.


kriegfrj:
> -I have re-instated the integration test with some extra debugging checks to produce more diagnostics. Hopefully if/when we get another failure I'll have more information to go on.

Cross-referencing the integration test flakiness issue for reference: #4253.

bjhargrave:
@kriegfrj Rebase this on master for force push. Something changed in the github actions build env which caused the builds to fail. I pushed a change to master to address this.

kriegfrj:
> * Disambiguate cases where there is no ambiguity (eg, my.package.name.something will not attempt to look in package "my.package.name" as it is clear from that context that ".name" is a class, not a package. Enables us to reinstate most of the original tests that had not already been reinstated.

Sorry, `my.package.name.something` should have read `my.package.name<type>.something` - the type tag got mis-interpreted as Github markdown because I forgot to escape it. Hopefully this explains to the reader why it ought to be clear that ".name" is a class and not a package...

> @kriegfrj Rebase this on master for force push.

Done.

biz.aQute.bndlib.tests/test/aQute/bnd/build/classindex/ClassIndexTest.java
```diff
@@ -20,50 +18,97 @@
...
-	public final TestName		testName	= new TestName();
-	private File				testDir;
-	private File				home		= IO.getFile("testresources/classindex");
+	public final JUnitSoftAssertions	softly	= new JUnitSoftAssertions();
 
-	@Before
-	public void setUp() throws IOException {
-		testDir = new File(TMPDIR, getClass().getName() + "/" + testName.getMethodName());
+	@BeforeClass
+	public static void setUp() throws Exception {
```
bjhargrave, line 30: Each test method should have its own tmp dir so there is no chance of side effects between tests. This should be `@BeforeEach`.

bndtools.core.test/src/bndtools/core/test/editors/quickfix/BuildpathQuickFixProcessorTest.java
```diff
@@ -826,8 +853,25 @@ void withFQClassLiteral_asAnnotationParameter_suggestsBundles(SoftAssertions sof
 			suggestsBundle("bndtools.core.test.fodder.iface", "1.0.0", "iface.bundle.MyInterface"));
 	}
 
+	@Disabled("Can't seem to make this work - binding information seems to be missing from class literal, I think this is"
```
bjhargrave, line 856: If you need parameter names, then you need to tell the compiler (javac and eclipse) to compile the parameter names in. So you probably need to make a change to this test project similar to:

https://github.com/bndtools/bnd/commit/26f649a722fcd0dcf3d334972e887a6453e9ce4f

biz.aQute.bndlib.tests/test/aQute/bnd/build/classindex/ClassIndexTest.java
kriegfrj: Setting up the workspace afresh for each test comes at performance penalty. I believe that this would be necessary if the tests were attempting to modify the workspace in some way, but as the workspace is read-only in these tests it seems like overkill to me.

bndtools.core.test/src/bndtools/core/test/editors/quickfix/BuildpathQuickFixProcessorTest.java
kriegfrj: Thanks - but I don't think that will help. The quick fix processor doesn't need to know the variable names, only their types.

It seemed that even though `ClassWithInterfaceExtendingMyInterface` is on the classpath and I have bindings recovery turned on, Eclipse's compiler wasn't properly generating the type binding information for the hierarchy of `ClassWithInterfaceExtendingMyInterface` so that I could find the missing type.

I got to the point where I felt that this was actually a bug in Eclipse that it wasn't fully recovering the incomplete type. However, I just had another idea that I will retry.

biz.aQute.bndlib.tests/test/aQute/bnd/build/classindex/ClassIndexTest.java
bjhargrave: Lets take the perf hit. I want to test cases to be side effect free and that is how we guarantee that. We do this in many other test cases. Also, a workspace is not read-only. Many things can be going on since there is a lot of code involved.

kriegfrj: Done.

bndtools.core.test/src/bndtools/core/test/editors/quickfix/BuildpathQuickFixProcessorTest.java
kriegfrj: I retried the idea and re-confirmed my original diagnosis. This time I documented the issue more precisely in the test code itself. I left in there in case the Eclipse issue is ever resolved, then this test case can be re-instated.

I'm able to get the binding information for `ClassWithInterfaceExtendingMyInterface`. However, when I call `getInterfaces()` on the type binding it returns an empty array. What it *should* be doing is returning a recovered type binding for its implemented interface `InterfaceExtendingMyInterface` (which the quick fix processor would then know how to fix).

The annoying thing is that if you print out the type binding for `ClassWithInterfaceExtendingMyInterface` or examine its private state variables, you can see that Eclipse knows it implements `InterfaceExtendingMyInterface` and has a recovered binding for it. It's just not returning it in `getInterfaces()`.

bjhargrave: Hmm. Is there an Eclipse bug filed for this, I wonder?

kriegfrj: There is now! https://bugs.eclipse.org/bugs/show_bug.cgi?id=566145

bjhargrave:
Thanks!

PR #4271, opened 2020-08-15
[bndtools] Un-export bndtools.wizards.workspace

### LINKED ISSUE #4267: [bndtools] Bnd warning for bndtools.wizards.workspace package
In the wake of #4260, @bjhargrave alerted me to the fact that the following Bnd warning has appeared:

```
Export bndtools.wizards.workspace,  has 1,  private references [org.bndtools.core.ui.wizards.shared]	bnd.bnd	/bndtools.core	
```

The offending interface is `ISkippingWizard`, which is implemented by the workspace wizard.

There are two ways this could be fixed:
1. export org.bndtools.core.ui.wizards.shared; or
2. stop exporting `bndtools.wizards.workspace`.

 The wizards in the other sub-packages (`bndtools.wizards.project`, `bndtools.wizards.bndfile`) also implement `ISkippingWizard` but they are not generating the same warning because these packages are not exported. There does not appear to be a good reason why the workspace wizard package needs to be exported either. I have tried removing the export declaration locally, and it seems to work as there are no dependencies on this package within the Bnd workspace. So from a cleanliness perspective, it seems that 2 is the better option.

Theoretically there is a risk of breaking any 3rd-party dependencies that depend on this package being exported. In reality though, I can't imagine there are any, nor could I see too many possible use cases for such.

Before pressing ahead though, I was wondering if any of the more experienced heads could foresee any issues with this approach, or if they had any tips on how to do it right? (Eg, do we still need the `@Version` tag if it's no longer exported?)

kriegfrj:
Fixes #4267.

### DIFF
--- a/bndtools.core/src/bndtools/wizards/workspace/package-info.java
+++ b/bndtools.core/src/bndtools/wizards/workspace/package-info.java
@@ -1,3 +1,2 @@
-@org.osgi.annotation.bundle.Export
 @org.osgi.annotation.versioning.Version("2.2.3")
 package bndtools.wizards.workspace;


bjhargrave:
Thanks!

PR #4281, opened 2020-08-26
[quickfix] Filters any proposals for types that are already on the bu…

pkriens:
…ild or test path


pkriens:
Sorry for accidentally combining commits.

PR #4293, opened 2020-09-04
[quickfix] Add support for static references to undefined types

kriegfrj:
This PR was motivated by me trying to convert a build which contained a statement `FrameworkUtil.getBundleContext()`.

In this context the compiler can't tell if the name `FrameworkUtil` refers to a missing type or a missing variable, and so the internal error it generates is not an "UnknownType" but an "UnknownName". The quickfix processor was not set up to handle this.

Also in this PR: courtesy of #4291, many of the test cases for nested classes now work and can be re-instated. The only remaining case that doesn't work is the bare simple name referring to a nested type (which doesn't have any index support at the moment).


bjhargrave (changes requested):
I think this needs rebasing on master since it seems to contain a duplicate commit.

kriegfrj:
> I think this needs rebasing on master since it seems to contain a duplicate commit.

Sorry - yes, it was based off of #4291.

kriegfrj:
Rebased onto master now that #4291 is merged.

bjhargrave (approved):

bjhargrave:
Thanks!

PR #4301, opened 2020-09-08
Refactoring Maven Repository

pkriens:
The design of the MavenBndRepository is layered. A MavenBndRepository uses an IMavenRepository to access the repositories. This IMavenRepository is constructed from a _local_ repository (.m2) and a set of _backing_ repositories. The purpose of the backing repositories is to abstract the source. Either a file repository or remote repository.

For unknown reasons, the backing repositories implemented the metadata handling, making them unnecessarily complex. This wasn't a big deal because the implementations inherited from a base class. 

However, trying to properly handle the caching problems of snapshots, it became too hairy. Although this looks like an easy problem, it interacts with the metadata caching & archive stale time & HTTP client caching in mysterious ways. I.e. too much caching was going on.

I, therefore, moved the metadata methods to the MavenRepository implementation, actually simplifying things considerably. All caching is now in the MavenRepository class. I also moved some of the validation methods to the MavenRepository as well, simplifying the callers. I also unrolled the promises in the MavenRemoteRepository fetch method. This is a private method and was called locally once where there was a wait on the promise. The use of promises made it impossible to single step and verify while using more threads. So this has now become a simple method instead of promise acrobatics.

In the end the cache handling of snapshots is now a lot clearer. 

A MavenBndRepository reads the index and then fetches all files at startup or when there is an addition. Since the init must be done before the first use of the repository, it can assume there is a file present. I.e. the MavenBndRepository 'get' method never fetched content, the content was already there.

When the MavenBndRepository get method is called, it will 'find' the archive in the index. The archive is now checked to be stale via the IMavenRepo. The time for snapshots is an hour and a release a week. (We could make it infinite.) If the file is stale, an update is scheduled.

When the refresh is called it clears the metadata cache map and starts an update of all current snapshot archives.

For each updated archive, it is read from the IMavenRepo.
Optimizations in the IndexFile have been removed.

If the archive is a snapshot, it is first necessary to read the Revision Metadata to find the remote path since each snapshot release creates a new path. A refresh will remove this metadata, which forces a read for snapshots. Releases cannot be refreshed.

The MavenRemoteRepository will use the rewritten fetch method to get this file from the Http Client. This used
to another stale time during which there was no check of the remote status. This has been changed. The fetch method will now _always_ check the remote. However, it will use the etag to check if there was a refresh so no download will actually take place if the file is still the same.

This results in either an updated or unmodified file. In both cases, we parse the file and create a new  RevisionMetadata.

The snapshot is then resolved against the new metadata. 

This results in a potentially new remote path. We'll use the fetch method to get this file and then parse it.

This change will result that snapshots can be refreshed. However, snapshots are also fetched at startup now.


biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -366,14 +375,31 @@ private boolean isMulti(String name) {
...
+				.stream()
+				.filter(Archive::isSnapshot)
+				.collect(Collectors.toSet());
+			if (!snapshots.isEmpty()) {
+				result = serialize(() -> update(snapshots));
+			} else
+				return false;
 		}
-		return false;
+		Promise<Boolean> serializer = result.onResolve(refreshAction);
```
bjhargrave, line 400: Since you use `onResolve`, `serializer` is the same as `result` which means the refresh action will not happen before `sync(serializer)` returns. That is, `onResolve` is not chained. Just making sure this is your intention.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -64,7 +68,11 @@
  * results without having to wait for the operation to be done.
  */
 class IndexFile {
-	private final static Logger					logger		= LoggerFactory.getLogger(IndexFile.class);
+	private static final String[]				ZIP_EXTENSIONS	= new String[] {
```
bjhargrave, line 71: You may wish to use the new `Sets.of` method to create an immutable set of the extensions?

biz.aQute.repository/test/aQute/bnd/repository/maven/provider/MavenBndRepoTest.java
```diff
@@ -81,6 +81,7 @@
 
 	@BeforeEach
 	public void setUp(TestInfo testInfo) throws Exception {
+		System.setProperty("org.slf4j.simpleLogger.defaultLogLevel", "DEBUG");
```
bjhargrave, line 84: Is this just something left over from your debugging? Seems like it should not be part of the PR.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
pkriens: The promise was already serialized so this is ok. I've removed the serializer variable to make this more clear.

PR #4302, opened 2020-09-08
[m2e] more safely refresh entire targetDir

gamerson:


bndtools.m2e/src/bndtools/m2e/BndConfigurator.java
```diff
@@ -145,20 +150,15 @@ public AbstractBuildParticipant getBuildParticipant(final IMavenProjectFacade pr
...
-				// plugin by reading the classesDir property.
-				final File classesDir = maven.getMojoParameterValue(projectFacade.getMavenProject(), execution,
-					"classesDir", File.class, monitor);
+
+				Set<IPath> deltaFilePaths = collectResourceDeltas(getDelta(project)).map(IResourceDelta::getResource)
+					.filter(IFile.class::isInstance)
+					.map(IResource::getFullPath)
+					.collect(Collectors.toSet());
+
+				final String targetDirectory = projectFacade.getMavenProject()
```
bjhargrave, line 159: Doesn't this lose checking executions for different target directories as the removed comment says?

gamerson: Yes. I discussed with Ray and he agreed this more "course grained" approach would be ok provided we didn't his the "infinite build" loop problem, which I have tried to address with the new resource delta check.

rotty3000: @bjhargrave you might be right, but @gamerson and I discussed this and this change solves some problems that @kriegfrj reported and it's on me to test if this breaks the use case I documented in the comment.

bjhargrave: OK, as long as you got this! :-)

rotty3000 (approved):
LGTM (tested with osgi-test, aries-cdi, aries-jaxrs-whiteboard)

PR #4307, opened 2020-09-10
BndrunContainer incorrectly assumes artifactId == osgi.identity

### LINKED ISSUE #4189: BndrunContainer incorrectly assumes artifactId == osgi.identity
During resolving with the Maven plugin a `-runrequires` is automatically calculated from the artifact id (https://github.com/bndtools/bnd/blob/3691481b9bde119f06deb3531e2c892a2d00d68c/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java#L266). This must be the bundle symbolic name instead (https://docs.osgi.org/specification/osgi.core/7.0.0/framework.namespaces.html#i1776750)!

Otherwise you run into errors like
```
[INFO] --- bnd-resolver-maven-plugin:5.1.1:resolve (default) @ <artifact-id> ---
[INFO] Bnd inferred -runrequires: osgi.identity;filter:='(osgi.identity=<artifact-id>)'
[ERROR] Resolution failed. Capabilities satisfying the following requirements could not be found:
    [<<INITIAL>>]
      ⇒ osgi.identity: (osgi.identity=<artifact-id>)
```

Although this behaviour is documented at https://github.com/bndtools/bnd/tree/master/maven/bnd-resolver-maven-plugin#bndrun-details-inferred-from-maven it would be nice to improve that behaviour to evaluate the previously generated manifest instead.

rotty3000:
fixes #4189

### DIFF
--- a/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
+++ b/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
@@ -1,9 +1,12 @@
 package aQute.bnd.maven.lib.resolve;
 
+import static aQute.lib.exceptions.FunctionWithException.asFunction;
+
 import java.io.File;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
@@ -26,6 +29,7 @@ import aQute.bnd.build.Workspace;
 import aQute.bnd.build.model.EE;
 import aQute.bnd.maven.lib.configuration.BeanProperties;
 import aQute.bnd.osgi.Constants;
+import aQute.bnd.osgi.Domain;
 import aQute.bnd.osgi.Processor;
 import aQute.bnd.repository.fileset.FileSetRepository;
 import aQute.bnd.service.RepositoryPlugin;
@@ -263,8 +267,22 @@ public class BndrunContainer {
 
 			Artifact artifact = project.getArtifact();
 
+			String bsn = artifact.getArtifactId();
+
+			try {
+				bsn = Optional.ofNullable(artifact.getFile())
+					.map(asFunction(Domain::domain))
+					.map(Domain::getBundleSymbolicName)
+					.map(Map.Entry::getKey)
+					.orElseGet(artifact::getArtifactId);
+			} catch (Exception e) {
+				logger.warn(
+					"Could not get the Bundle-SymbolicName from the project artifact {}, falling back to artifactId {}",
+					artifact, bsn, e);
+			}
+
 			run.setProperty(Constants.RUNREQUIRES,
-				String.format("osgi.identity;filter:='(osgi.identity=%s)'", artifact.getArtifactId()));
+				String.format("osgi.identity;filter:='(osgi.identity=%s)'", bsn));
 
 			logger.info("Bnd inferred {}: {}", Constants.RUNREQUIRES, run.getProperty(Constants.RUNREQUIRES));
 		}
--- a/maven/bnd-resolver-maven-plugin/README.md
+++ b/maven/bnd-resolver-maven-plugin/README.md
@@ -55,7 +55,7 @@ mvn bnd-resolver:resolve
 The `-runee` and `-runrequires` values can be inferred from the maven project as follows:
 
   * `-runee`, if omitted from the bndrun file, will be inferred from the `<target>` configuration of `maven-compiler-plugin`
-  * `-runrequires`, if omitted from the bndrun file, will be inferred from the project's `artifactId` and applied as `osgi.identity;filter:='(osgi.identity=<artifactId>)'`, if the project packaging is `jar` or `war` and the project has the `bnd-maven-plugin`
+  * `-runrequires`, if omitted from the bndrun file, will be inferred by attempting to get the `Bundle-SymbolicName` (bsn) from the project's main artifact and if not found will use `artifactId`. The value will be applied as `osgi.identity;filter:='(osgi.identity=<bsn|artifactId>)'`, if the project packaging is `jar` or `war` and the project has the `bnd-maven-plugin`
 
 ## Implicit Repository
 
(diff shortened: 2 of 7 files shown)


biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
```diff
@@ -262,9 +263,19 @@ public void setRunrequiresFromProjectArtifact(Run run) {
 			&& (project.getPlugin("biz.aQute.bnd:bnd-maven-plugin") != null)) {
 
 			Artifact artifact = project.getArtifact();
+			String bsn = artifact.getArtifactId();
+
+			try (Jar jar = new Jar(artifact.getFile())) {
```
bjhargrave, line 268: This is kind of expensive as it will read the entire jar when all you need is the manifest. It is less costly to use Bnd Domain class (`Domain.domain(File)`) to process the manifest and get the BSN.

https://github.com/bndtools/bnd/blob/aea111f779c5dfd2a91a82d698abb28bb25aaf0b/biz.aQute.bndlib/src/aQute/bnd/osgi/Domain.java#L509-L519

rotty3000: right right, I'll fix.

rotty3000: fixed

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
```diff
@@ -263,8 +265,22 @@ public void setRunrequiresFromProjectArtifact(Run run) {
 
 			Artifact artifact = project.getArtifact();
 
+			String bsn = artifact.getArtifactId();
+
+			try {
+				Domain domain = Domain.domain(artifact.getFile());
+
+				bsn = Optional.of(domain.getBundleSymbolicName())
```
bjhargrave, line 273: getBundleSymbolicName can return null. Are you counting on the NPE from `of` ending up in the catch block?

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
```diff
@@ -263,8 +265,22 @@ public void setRunrequiresFromProjectArtifact(Run run) {
...
+					.map(Map.Entry::getKey)
+					.orElseGet(() -> artifact.getArtifactId());
```
bjhargrave, line 275: `.orElseGet(artifact::getArtifactId)`

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
```diff
@@ -263,8 +265,22 @@ public void setRunrequiresFromProjectArtifact(Run run) {
 
 			Artifact artifact = project.getArtifact();
 
+			String bsn = artifact.getArtifactId();
+
+			try {
+				Domain domain = Domain.domain(artifact.getFile());
```
bjhargrave, line 271: domain can be null. Are you counting on the NPE dereferencing domain ending up in the catch block?

bjhargrave (changes requested):
The problem with allowing NPEs to be thrown and then caught is that you end up with an unhelpful exception in the log message which looks like a programming error on our part. I think we need to look for the nulls.

```java
bsn = Optional.of(artifact.getFile())
   .map(Domain::domain)
   .map(Domain::getBundleSymbolicName)
   .map(Map.Entry::getKey)
   .orElseGet(() -> {
        logger.warn("No bsn in artifact", ...);
        return artifact.getArtifactId();
    });
``` 

Then in the catch block we log a real exception such as the jar is malformed.

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
rotty3000: how about:
```
			Artifact artifact = project.getArtifact();

			String bsn = artifact.getArtifactId();

			try {
				bsn = Optional.ofNullable(artifact.getFile())
					.map(asFunction(Domain::domain))
					.map(Domain::getBundleSymbolicName)
					.map(Map.Entry::getKey)
					.orElseGet(artifact::getArtifactId);
			} catch (Exception e) {
				logger.warn(
					"Could not get the Bundle-SymbolicName from the project artifact {}, falling back to artifactId {}",
					artifact, artifact.getArtifactId(), e);
			}
```

bjhargrave: That's fine, but you will not log any warning if the jar has no manifest or no BSN which I though you wanted to log.

rotty3000: not precisely, if the project HAS the bnd-maven-plugin (already checked) and the artifact IS NOT a bundle, then something went wrong, no? Isn't that a different problem entirely? the resolver would fail to find the bundle since it couldn't be indexed as a proper OSGi bundle. It was not my intention to capture all those scenarios from this specific logic whose only job is to find a suitable value for inferred `-runrequires`.

rotty3000: so the question is... should it be capturing all those cases?

rotty3000: I've updated with new logic... let's go from that.

bjhargrave: I don't think so. I was just pointing out that your original commit logged for all those case. I think we only need to log for the exception case (as long as we don't NPE).

bjhargrave (approved):

PR #4310, opened 2020-09-11
[junit-platform] Serialise JUnitEclipseListener

### LINKED ISSUE #4308: [junit-platform] JUnitEclipseListener is not thread-safe
Following #4296 and a discussion with @bjhargrave started to get me wondering - I had a bit of a look at our current implementation of `JUnitEclipseListener`. It doesn't look very thread-safe. A quick test seems to confirm this: I just tried running osgi-test junit5 tests with parallel enabled and the Eclipse display gets messed up, with several random tests reporting that they haven't run, or that they started and didn't finish:
ParallelFail

Note that `BundleContextExtension_CleanupTest#cleansUpBundleListenersMultiLevel()` (and a few others in that class) never ran, and that `cleansUpRegisteredServicesMultieLevel()->afterClass` is still marked as running (as are a few other tests further down) even though the entire test suite has finished.

Getting this working properly will require some attention. For now, if anyone wants to run the tests with the Eclipse IDE support enabled, I recommend that they don't try and use parallelism.

kriegfrj:
This allows biz.aQute.tester.junit-platform to be used with concurrency and have the Eclipse JUnit display still work properly.

Fixes #4308.

Seems to work properly, as evidenced by the following clip:

osgi-old-workspace-Eclipse-SDK-2020-09-11-11-43-09

Of course, many of the tests fail because the tests themselves are not thread-safe. However, you can see that Eclipse is correctly reporting the test results - unlike the screenshot in  #4308 none of the tests are left displayed as unfinished or unstarted. You can also see from the clip that there are points where the GUI is displaying results from multiple tests simultaneously. 

I've been too lazy to write regression tests for this though - maybe one day...

### DIFF
--- a/biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/JUnitEclipseListener.java
+++ b/biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/JUnitEclipseListener.java
@@ -68,16 +68,18 @@ public class JUnitEclipseListener implements TestExecutionListener, Closeable {
 		// Reporting them all as assumption failures triggers the GUI to display
 		// the skip reason in the failure trace, which the Eclipse
 		// implementation doesn't do.
-		if (!testIdentifier.isContainer()) {
-			message("%TESTS  ", testIdentifier);
-		}
-		message("%FAILED ", testIdentifier, "@AssumptionFailure: ");
-		message("%TRACES ");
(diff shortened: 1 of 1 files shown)


kriegfrj:
Failing tests need some investigation...

kriegfrj:
I'm thankful for the existing regression tests. I let a bug slip in unnoticed which they picked up...

bjhargrave (approved):
Well done!

PR #4311, opened 2020-09-11
[quickfix] Suggestions for overloaded methods

kriegfrj:
Sometimes you get a "no such method" or "parameter mismatch" because the version of the method that you're trying to call is defined in a superclass that is not on the classpath. This patch will catch such occurrences and suggest adding the missing supertype.


bjhargrave (approved):

bjhargrave:
Thanks!

PR #4314, opened 2020-09-15
Add Maven version prerequisite for all Maven plugins

### LINKED ISSUE #4313: The Maven version prerequisite should be part of the pom
The prerequisite listed in https://github.com/bndtools/bnd/tree/master/maven#maven-plugins should be reflected in the `pom.xml` as well. This is used by Maven and some of its plugins to enforce this.

kwin:
This closes #4313

### DIFF
--- a/maven/bnd-baseline-maven-plugin/pom.xml
+++ b/maven/bnd-baseline-maven-plugin/pom.xml
@@ -21,6 +21,10 @@
 		<developerConnection>scm:git:git@github.com:bndtools/bnd.git</developerConnection>
 	</scm>
 
+	<prerequisites>
+		<maven>${maven.target.version}</maven>
+	</prerequisites>
+
 	<dependencies>
 		<dependency>
 			<groupId>org.apache.maven</groupId>
--- a/maven/bnd-export-maven-plugin/pom.xml
+++ b/maven/bnd-export-maven-plugin/pom.xml
@@ -19,6 +19,10 @@
 		<developerConnection>scm:git:git@github.com:bndtools/bnd.git</developerConnection>
 	</scm>
 
+	<prerequisites>
+		<maven>${maven.target.version}</maven>
+	</prerequisites>
+
 	<dependencies>
 		<dependency>
 			<groupId>org.apache.maven</groupId>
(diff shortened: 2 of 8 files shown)


maven/pom.xml
```diff
@@ -11,6 +11,10 @@
 	<packaging>pom</packaging>
 	<version>${revision}</version>
 
+	<prerequisites>
```
bjhargrave, line 14: We already use maven 3.6.3 (see mvnw) to build the maven plugins. This is also sort of wrong. I don't know what actual version of maven we need to _build_ the plugins. We don't test building against an older version of maven, so this prerequisite is not in anyway known to be true.

We do however set the version of maven to compile against

https://github.com/bndtools/bnd/blob/417811e3183520aa0977404b01aaa832bbf0ddf2/maven/pom.xml#L18

so that our plugins can be used with versions of maven back to 3.1.1. That is, we don't rely upon new maven API added since 3.1.1. It is our intention that the maven plugins can be used with older versions of maven.

It is not our intention to be able to build the maven plugins with older versions of maven. So this prerequisite is not accurate.

bjhargrave (changes requested):
I don't think this PR is correct.

bjhargrave:
Can you resubmit the PR with 

```xml
	<prerequisites>
		<maven>3.1.1</maven>
	</prerequisites>
```

not in the parent pom, but in each child pom. Since we flatten the poms, the flattening process does not include the prerequisite element from the parent pom.

kwin:
It is not the fault of the flatten plugin. This is due to 

> By default, every model field is inherited as-is from parent, with a few exceptions that are intentionally not inherited: modelVersion, artifactId, profiles (injected in phase 1) and prerequisites.

(http://maven.apache.org/ref/3.6.3/maven-model-builder/index.html#Inheritance_Assembly)

So indeed prerequisites have to be added to each plugin individually.

bjhargrave:
> Can you resubmit the PR with

@kwin Will you be able to update this PR?

kwin:
Yes, but before I can update I need to clarify https://github.com/bndtools/bnd/issues/4327. Any inputs highly appreciated.

kwin:
Once the underlying issue with requiring 3.5.3 by mistake is fixed, this can be merged from my POV now.

bjhargrave (approved):

bjhargrave:
Thanks!

Issue #3709, opened 2020-01-13
[tester] Support test re-run interface of Eclipse


kriegfrj:
Eclipse has support for re-running the test suite (or one of the nodes in the hierarchy of tests) from the GUI. The current Bndtools tester implementations (both the legacy and junit-platform implementations) do not support this functionality. It might be nice if they did.

If implemented, this functionality would probably slo resolve #3559 in a more robust manner.

bjhargrave:
This can only be useful for OSGi JUnit in continuous mode. Otherwise the remote process will terminate at the end of the test run.

kriegfrj:
As it currently stands, yes. I was going to explore the possibility of changing this behaviour so that the remote process doesn't exit until Eclipse closes the connection, which would fix this as well as fixing #3559 in a more robust manner.

kriegfrj:
I haven't fully implemented this yet...?

bjhargrave:
@kriegfrj Can you fix by Sep 15 (or defer to 6.1)?

kriegfrj:
I'll try to get this one done.

kriegfrj:
I don't think I'm quite going to get this one done.

I have it about 80% working. The trick is, for it to be useful, you need a configuration option to suppress the tester's built-in automatic re-run of all tests. The reason for this is: There's usually not much point re-running an individual test unless you've changed something in either the test code or the code-under-test. But at the moment, as soon as you change something it will end the current test session and start afresh with a new one, re-running the test anyway. For the re-run feature to be useful, you need an option to suppress the automatic re-run of tests (at least while a test session is still open). This complication is what led me to put in on the backburner.

I think that the most pressing usability issue is the related one that I fixed in #4813.

PR #4317, opened 2020-09-17
[junit-platform] Make tester.names dynamic and work in continuous mode

kriegfrj:
The current state of the tester is that if a `tester.names` filter is set the tester will run once and exit - the setting of `tester.continuous` is ignored. With this fix, the tester will instead continue to run if `tester.continuous` is set to true. It will respond to bundle updates, but it will only trigger a re-run of tests that match `tester.names`.

I've been wanting this feature when working on large and relatively slow integration tests (eg, `bndtools.core.test`). Often when you're working in a TDD scenario your focus is on a particular test method - you just want to quickly re-run that test while you're working on it to see the results. While working on the quick fix processor I found that the continuous testing really helped overcome the slow startup time of Eclipse, but this was counteracted somewhat as the number of test methods grew. This feature would allow you to combine continuous testing with also filtering the test suites being run using `tester.names`.

Additionally, on a related note, the gogo commandlet has been enhanced as follows:
* `set/getTesterNames()` commands have been added to allow you to set/get the current value of tester.names. So you can change the filter dynamically at runtime while testing.
* `runTests()` will now run all test bundles (filtered by `tester.names`) if it is invoked with no arguments. 

I'm anticipating that the addition of these features will also make it easier to implement #3709


bjhargrave:
Nice work!

PR #4318, opened 2020-09-21
Support adding bnd instructions to BundleTaskConvention as Gradle providers.

chrisr3:
Expose `instructions.add(Provider<? extends CharSequence>)` on `BundleTaskConvention`. This allows us to generate instructions using Gradle providers without forcing those providers to be evaluated. In particular, it resolves this Gradle warning:
```
Querying the mapped value of task ':jar' property 'archiveFile' before task ':jar' has completed has been deprecated. This will fail with an error in Gradle 7.0.
```
I am currently working around this problem by attaching my lazy Gradle provider to the project as a property, and then invoking this property as a bnd macro via `\${project.myProvider}`. However, this is _considerably_ less obvious (from a Gradle perspective) than just adding the `Provider` to the bnd instructions directly.


chrisr3:
Actually, it would probably be useful to provide "lazy" equivalents for _all_ of the `bnd` property accessors.

rotty3000:
hey @chrisr3 just wanted to say keep it coming. We're just going to wait for @bjhargrave to be available to review this since he's the gradle guy :)

chrisr3:
> hey @chrisr3 just wanted to say keep it coming. We're just going to wait for @bjhargrave to be available to review this since he's the gradle guy :)

No worries. I've discovered that I can also use Gradle providers as bnd macros to ensure that they are evaluated lazily (which is what I need). However, this does require some subtle thinking about using `$` vs `\$` in a Gradle `build.script`, and is therefore hideously non-obvious, Especially when it's so trivial to expose the `add(Provider)` method for the `ListProperty` instead :-).

biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BundleTaskConvention.groovy
```diff
@@ -28,9 +28,7 @@ import static aQute.bnd.gradle.BndUtils.jarLibraryElements
 import static aQute.bnd.gradle.BndUtils.logReport
 import static aQute.bnd.gradle.BndUtils.unwrap
 
-import java.util.Properties
```
bjhargrave, line 31: The Properties type is used in the `buildBundles` method, so this line should not be removed.

bjhargrave (changes requested):
Thanks! One minor nit. I am not sure why you removed an import for the `Properties` type which is used in the class. Can you please update to avoid removing that line?

biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BundleTaskConvention.groovy
chrisr3: Err, it can only be because IntelliJ flagged it as "unused" - for some reason known only to IntelliJ. I'll add it back...

chrisr3: Ah, the explanation! [Groovy imports these packages by default](https://groovy-lang.org/structure.html):
```
import java.lang.*
import java.util.*
import java.io.*
import java.net.*
```

bjhargrave (approved):

bjhargrave:
Thanks for the PR!

rotty3000:
does this change warrant README changes?

bjhargrave:
> does this change warrant README changes?

I don't think so. The README does not mention the argument types. This PR just overloads the methods to support Provider types.

PR #4342, opened 2020-10-06
[junit-platform] Handle subclasses of comparison failure assertions

### LINKED ISSUE #4338: [junit-platform] Comparison failures doesn't handle subclasses
I've noticed while using AssertJ soft assertions that the GUI comparison feature doesn't seem to work when using the junit-platform tester.

Upon inspection I think I know why this is: the code in JUnitEclipseListener looks for exact matches on the class name and doesn't consider inheritance:
https://github.com/bndtools/bnd/blob/f03d29ab2740eec4308eef02675c54204506a441/biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/JUnitEclipseListener.java#L303-L305

AssertJ's soft assertions throw an `AssertJMultipleFailuresError`, which extends `org.opentest4j.MultipleFailuresError`.  But because none of the cases in the switch above exactly matches AssertJMultipleFailuresError, it doesn't get handled and so no actual/expected information is sent.

This should be a fairly simple fix.

kriegfrj:
Fixes #4338.

### DIFF
--- a/biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/JUnitEclipseListener.java
+++ b/biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/JUnitEclipseListener.java
@@ -20,11 +20,10 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Optional;
 import java.util.Set;
-import java.util.StringJoiner;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.BooleanSupplier;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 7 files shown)


biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/JUnitEclipseListener.java
```diff
@@ -41,6 +40,109 @@
 import org.opentest4j.MultipleFailuresError;
 
 public class JUnitEclipseListener implements TestExecutionListener, Closeable {
+
+	static final Class<? extends AssertionError>	JUNIT3_COMPARISON_FAILURE;
```
bjhargrave, line 44: I think these should be `Predicate<Object>`.

biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/JUnitEclipseListener.java
```diff
@@ -41,6 +40,109 @@
...
+	static final Class<? extends AssertionError>	JUNIT4_COMPARISON_FAILURE;
+	static final Class<? extends AssertionError>	ASSERTION_FAILED_ERROR;
+	static final Class<? extends AssertionError>	MULTIPLE_FAILURES_ERROR;
+
+	@SuppressWarnings("unchecked")
+	static Class<? extends AssertionError> tryLoad(String name) {
+		try {
```
bjhargrave, line 51: We then always make a non-null predicate:

```java
static Predicate<Object> tryLoad(String name) {
		try {
			return Class.forName(name)::isInstanceOf;
		} catch (ClassNotFoundException cnfe) {
			return obj -> false;
		}
}
```

biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/JUnitEclipseListener.java
```diff
@@ -41,6 +40,109 @@
...
+			// the block, so that by that stage it knows that the class
+			// is available and that NCDFE will not be thrown. On the other
+			// hand, if the class is not on the classpath then the code
+			// will never be executed and the thread context classloader
+			// will not attempt to load the class, so again no NCDFE.
+			//
+			// This is particularly important for the JUnit 3/4 comparison
+			// failure assertions, because the OpenTest4J ones will be on the
+			// classpath courtesy of junit-platform-launcher's dependencies.
+			if (ASSERTION_FAILED_ERROR != null && ASSERTION_FAILED_ERROR.isAssignableFrom(exception.getClass())) {
```
bjhargrave, line 97: With the above change to `Predicate<Object>`, the if test becomes more simple and we avoid repeating the test logic:

```java
if (ASSERTION_FAILED_ERROR.test(exception)) {
```

bjhargrave (changes requested):

biz.aQute.tester.junit-platform/src/aQute/tester/junit/platform/JUnitEclipseListener.java
kriegfrj: I made them `Predicate<Throwable>` instead, but basically used your suggestion. Thanks for the idea, it does neaten up the code a bit.

bjhargrave (approved):

PR #4346, opened 2020-10-09
[launcher] The launcher will now be registered as soon as possible

juergen-albert:
The serviceregistration of the launcher is currently the only way, to retrieve the launcher arguments in the Framework. Especially for a lot of the Eclipse Bundles, it is necessary to have this information as early as possible. Thus the launcher needs to be registered right after the framework is started and not after all is done.. 

### DIFF
--- a/biz.aQute.launcher/src/aQute/launcher/Launcher.java
+++ b/biz.aQute.launcher/src/aQute/launcher/Launcher.java
@@ -40,6 +40,7 @@ import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Dictionary;
 import java.util.Enumeration;
 import java.util.Formatter;
 import java.util.Hashtable;
@@ -74,6 +75,7 @@ import org.osgi.framework.FrameworkEvent;
 import org.osgi.framework.FrameworkListener;
 import org.osgi.framework.ServiceEvent;
 import org.osgi.framework.ServiceListener;
+import org.osgi.framework.ServiceRegistration;
 import org.osgi.framework.connect.ConnectFrameworkFactory;
 import org.osgi.framework.connect.ModuleConnector;
 import org.osgi.framework.launch.Framework;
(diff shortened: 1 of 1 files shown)


juergen-albert:
@pkriens would you mind having a look? I have a customer that would be happy to use this. Thx

pkriens:
I think it might be slightly better not to set the ready status to false, just don't set the property.

And after that LGTM

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -121,6 +122,9 @@
 	private boolean							restart					= false;
 	private boolean							frameworkInited			= false;
 
+	private ServiceRegistration<?>			launcherServiceRegistraion;
+	private Hashtable<String, Object>		launcherServiceProperties;
```
bjhargrave, line 126: This should not be an instance variable. You should not reuse the Hashtable across service calls. Create a new Hashtable and populated it at each place you make a service call.

bjhargrave (changes requested):

biz.aQute.launcher/src/aQute/launcher/Launcher.java
juergen-albert: @bjhargrave I'll change it, but I'm unsure for the reason for it. I know that there had been issues in the past, because of service props that contained certain kinds of lists. besides this, we use this pattern in felix and equinox for years and it works like a charm. Can you elaborate some more?

bjhargrave: It is not clear in the registerService method who owns the Dictionary after the call. So the safest course is to assume the Dictionary cannot be reused and is owned by the framework. I recall that Tom Watson encountered some code where they reused the Dictionary and encountered side effects (probably because of mutated values such as collections). So we should always be safe and not reuse the Dictionary and any values.

bjhargrave (approved):

bjhargrave:
Thanks!

PR #4363, opened 2020-10-21
Improved classloading for Plugins

juergen-albert:
- ActivelyClosingClassLoader will try to figure out its own bundle in
OSGi and tries to load classes from there if non can be found before
refering to the system bundle.


bjhargrave (changes requested):
Please use a good commit comment "title". Just "fixes issues 999" is not helpful when reading the git shortlog later to understand what the commit does. You need to use a properly descriptive commit comment so that anyone can look at the comment title and understand what the commit's purpose is.

juergen-albert:
> Please use a good commit comment "title". Just "fixes issues 999" is not helpful when reading the git shortlog later to understand what the commit does. You need to use a properly descriptive commit comment so that anyone can look at the comment title and understand what the commit's purpose is.

I hope that is better!?

bjhargrave (approved):

PR #4370, opened 2020-10-24
[quickfix] Prevent StackOverflow on recursive types

### LINKED ISSUE #4369: [quickfix] Problems on self-referential types cause StackOverflow
If you have a problem on a type whose definition is self-referential (eg, using the "self type" generic pattern `class C<T extends C<T>>`), this can cause the quick fix processor to get into an infinite loop and cause a `StackOverflow`.

Fix to come.

kriegfrj:
Track types that have been examined in a `Set` and skip any if they've already been examined. Main reason is to prevent recursion (fixes #4369).


kriegfrj:
@bjhargrave, this could be a cherry-pick candidate for 5.2.1 if we do one (though probably not worth one on its own).

bjhargrave (approved):

PR #4383, opened 2020-10-30
[shader] Rename packages

pkriens:
Adding a function to rename packages.

	-shade (<package-spec> ( ';prefix:=' <prefix> )? )+

I've added a rename method to the awesome ClassFile,
and a Shader in the osgi package, although it is not
much code now.

I did not add the code to the ClassFile to keep it clean,
a ClassFileRenamer holds the nitty gritty details. It
contains a SignatureParser that is not creating
objects and is extremely fast, especially when
no renaming takes place. I ran it against 350.000
signatures.

Testing uses a new generator that can generate type
names based on a regex. I think that can be a very
useful tool.


biz.aQute.bndlib/src/aQute/bnd/classfile/ClassFile.java
```diff
@@ -153,4 +166,28 @@ static ByteBuffer slice(DataInput in, int length) throws IOException {
 		in.readFully(array, 0, length);
 		return ByteBuffer.wrap(array, 0, length);
 	}
+
+	public byte[] toBytes() throws IOException {
+		ByteArrayOutputStream bout = new ByteArrayOutputStream();
```
bjhargrave, line 171: We have tried very hard to never use ByteArrayInputStream in Bnd due to performance issues with over synchronization. Instead we use ByteBufferInputStream. But in this case simply use ByteBufferDataOutput since you need a DataOutput anyway.

biz.aQute.bndlib/src/aQute/bnd/classfile/ClassFile.java
```diff
@@ -153,4 +166,28 @@ static ByteBuffer slice(DataInput in, int length) throws IOException {
...
+		DataOutputStream dout = new DataOutputStream(bout);
+		write(dout);
+		return bout.toByteArray();
+	}
+
+	public static ClassFile parseFile(File file) throws FileNotFoundException, IOException {
+		return parseInputStream(new FileInputStream(file));
```
bjhargrave, line 178: We try very hard to never use FileInputStream in Bnd for performance reasons. We generally use IO.stream.

biz.aQute.bndlib/src/aQute/bnd/classfile/ClassFile.java
```diff
@@ -153,4 +166,28 @@ static ByteBuffer slice(DataInput in, int length) throws IOException {
...
+
+	/**
+	 * Rename classes referenced in this class file by their binary name. This
+	 * will change all attributes, and references that refer to class names. The
+	 * mapper must also be useful for the optional `.class` suffix used for
+	 * class paths.
+	 *
+	 * @param mapper maps a binary class name to either null or a different name
+	 */
+	public Optional<ClassFile> rename(Function<String, String> mapper) {
```
bjhargrave, line 189: The classfile package is carefully structure for read-only operations. All mutation operation are kept outside of this package. I would prefer to continue this. So we should move all mutation code to a subpackage such as aQute.bnd.classfile.builder or some other subpackage. I want to be clear that the classfile package is not about mutation.

biz.aQute.bndlib/src/aQute/bnd/classfile/ClassFileRenamer.java
```diff
@@ -0,0 +1,588 @@
+package aQute.bnd.classfile;
```
bjhargrave, line 1: The classfile package is carefully structure for read-only operations. All mutation operation are kept outside of this package. I would prefer to continue this. So we should move all mutation code to a subpackage such as aQute.bnd.classfile.builder or some other subpackage. I want to be clear that the classfile package is not about mutation.

biz.aQute.bndlib/src/aQute/bnd/classfile/ClassFileRenamer.java
```diff
@@ -0,0 +1,588 @@
...
+	final ClassFileBuilder			builder;
+	final SignatureRenamer			sp;
+	boolean							changed	= false;
+
+	static Optional<ClassFile> rename(ClassFile cf, Function<String, String> mapper) {
+		ClassFileRenamer classFileRenamer = new ClassFileRenamer(cf, mapper);
+		return classFileRenamer.rename();
+	}
+
+	static class SignatureParser {
```
bjhargrave, line 36: Why was the signature package not used? We now have to signature parsers and bug sources?

biz.aQute.bndlib/src/aQute/bnd/classfile/ConstantPool.java
```diff
@@ -5,32 +5,42 @@
...
-	public static final int	CONSTANT_NameAndType		= 12;
-	public static final int	CONSTANT_MethodHandle		= 15;
-	public static final int	CONSTANT_MethodType			= 16;
-	public static final int	CONSTANT_Dynamic			= 17;
-	public static final int	CONSTANT_InvokeDynamic		= 18;
-	public static final int	CONSTANT_Module				= 19;
-	public static final int	CONSTANT_Package			= 20;
-
-	final Object[]			pool;
+import aQute.bnd.classfile.preview.PermittedSubclassesAttribute;
```
bjhargrave, line 19: Why do these need to be referenced in the constant pool?

biz.aQute.bndlib/src/aQute/bnd/classfile/ConstantPool.java
```diff
@@ -888,4 +898,316 @@ public void write(DataOutput out) throws IOException {
...
 		}
 	}
+
+	/**
+	 * A visitor of each entry with the proper types. Saves a lot of
+	 * 'instanceof' checking. See #
+	 */
+
+	@ProviderType
+	public interface EntryVisitor {
```
bjhargrave, line 908: I tried very hard to avoid visitor patterns since it is the main problem I have with asm. If you need this, cannot it not be done in a subclass of ConstantPool in a subpackage of classfile?

bjhargrave (commented):
Some general comments. I did not look deeply into the actual shading code yet.

biz.aQute.bndlib/src/aQute/bnd/classfile/ConstantPool.java
```diff
@@ -888,4 +898,316 @@ public void write(DataOutput out) throws IOException {
...
+
+		default Attribute visit(RuntimeInvisibleParameterAnnotationsAttribute attr) {
+			return null;
+		}
+
+		default Attribute visit(RuntimeVisibleParameterAnnotationsAttribute attr) {
+			return null;
+		}
+
+		default Attribute visit(PermittedSubclassesAttribute attr) {
```
bjhargrave, line 1075: We cannot have preview package type exposed in the classfile package API. The classfile API is currently safe in that the API never references the preview package. We do not want to have to make a major change because a preview package type changes before going final in the JVMS.

biz.aQute.bndlib/src/aQute/bnd/classfile/ConstantPool.java
bjhargrave: You don't need to use instanceof in nested ifs. Do `switch (contant_pool.tag(index))` and the cases are the `CONSTANT_` values. See https://github.com/bndtools/bnd/blob/0eca95e0e2c99bfd22d495fa368d355eed624d1c/biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java#L955-L957

Similarly for attributes. Do `switch (attribute.name())` and the cases are the attribute NAMEs. See https://github.com/bndtools/bnd/blob/0eca95e0e2c99bfd22d495fa368d355eed624d1c/biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java#L976-L981

Both of these avoid `instanceof` tests and also avoid the future compatibility problems of visitor interfaces which need extending whenever a new thing is added. See asm :-(

biz.aQute.bndlib/src/aQute/bnd/classfile/ClassFile.java
pkriens: not sure I agree ... but ok

biz.aQute.bndlib/src/aQute/bnd/classfile/Attribute.java
```diff
@@ -12,6 +12,54 @@
...
+
+		public Class<? extends Attribute> clazz;
+
+		AttributeTag(Class<? extends Attribute> clazz) {
+			this.clazz = clazz;
+		}
+
+		@SuppressWarnings("unchecked")
+		public <T extends Attribute> T cast(Attribute attr) {
+			return (T) clazz.cast(attr);
```
bjhargrave, line 57: This is not really a safe case since T is not bound to anything. For example the compiler will accept 

```java
SignatureAttribute foo = AttributeTag.Preview.cast(new RecordAttribute(...));
```
So it is a lot of work to get something not very safe.

If you want to make an easy (but perhaps not perfectly safe cast), then just add something like

```java
<A extends Attribute> A cast();
```

to `Attribute` and then each attribute implementation will do something like:

```java
	@Override
	public SignatureAttribute cast() {
		return this;
	}
```

Then if you have `Attribute attribute`, you can do `SignatureAttribute sigAttr = attribute.cast();`

biz.aQute.bndlib/src/aQute/bnd/classfile/ClassFile.java
```diff
@@ -108,6 +112,10 @@ public static ClassFile parseClassFile(DataInput in) throws IOException {
 		return class_file;
 	}
 
+	public static ClassFile parseInputStream(InputStream in) throws IOException {
+		return parseClassFile(new DataInputStream(in));
```
bjhargrave, line 116: I think you need to do something here to close the input stream. DataInput does not have a close so parseClassFile(DataInput) will not be able to close the input stream. 

Properly closing the input stream will be especially important for your new parseFile method so we don't lock files open on Windows.

biz.aQute.bndlib/src/aQute/bnd/classfile/ConstantPool.java
```diff
@@ -30,7 +33,102 @@
...
+		InterfaceMethodRef(CONSTANT_InterfaceMethodref, InterfaceMethodrefInfo.class), //
+		NameAndType(CONSTANT_NameAndType, NameAndTypeInfo.class), //
+		MethodHandle(CONSTANT_MethodHandle, MethodHandleInfo.class), //
+		MethodType(CONSTANT_MethodType, MethodTypeInfo.class),
+		Dynamic(CONSTANT_Dynamic, DynamicInfo.class), //
+		InvokeDynamic(CONSTANT_InvokeDynamic, InvokeDynamicInfo.class), //
+		Module(CONSTANT_Module, ModuleInfo.class), //
+		Package(CONSTANT_Package, PackageInfo.class);
+
+		public int					tag;
```
bjhargrave, line 56: These should be final and also probably not public instead having accessor methods.

biz.aQute.bndlib/src/aQute/bnd/classfile/renamer/ClassFileRenamer.java
```diff
@@ -0,0 +1,686 @@
...
+import aQute.bnd.classfile.LocalVariableTableAttribute;
+import aQute.bnd.classfile.LocalVariableTableAttribute.LocalVariable;
+import aQute.bnd.classfile.LocalVariableTypeTableAttribute;
+import aQute.bnd.classfile.LocalVariableTypeTableAttribute.LocalVariableType;
+import aQute.bnd.classfile.MethodInfo;
+import aQute.bnd.classfile.SignatureAttribute;
+import aQute.bnd.classfile.builder.ClassFileBuilder;
+import aQute.bnd.classfile.builder.MutableConstantPool;
+import aQute.bnd.classfile.preview.RecordAttribute;
+import aQute.bnd.classfile.preview.RecordAttribute.RecordComponent;
```
bjhargrave, line 28: Don't you also need to handle PermittedSubclassesAttribute?

biz.aQute.bndlib/src/aQute/bnd/classfile/renamer/ClassFileRenamer.java
```diff
@@ -0,0 +1,686 @@
...
+			String renamed = renameBinary(interf);
+			builder.interfaces(renamed);
+		}
+
+		String thisClass = renameBinary(builder.this_class());
+		builder.this_class(thisClass);
+
+		builder.super_class(renameBinary(classFile.super_class));
+
+		for (int index = 0; index < builder.constant_pool()
```
bjhargrave, line 397: What does this loop do? The switch statement is empty.

biz.aQute.bndlib/src/aQute/bnd/classfile/ConstantPool.java
```diff
@@ -888,4 +986,14 @@ public void write(DataOutput out) throws IOException {
 			}
 		}
 	}
+
+	@Override
+	public Iterator<Object> iterator() {
+		return Arrays.asList(pool)
+			.iterator();
```
bjhargrave, line 993: `return stream().iterator();` which uses the (updated) stream method that avoids returning the invisible 0-index entry.

biz.aQute.bndlib/src/aQute/bnd/classfile/ConstantPool.java
```diff
@@ -888,4 +986,14 @@ public void write(DataOutput out) throws IOException {
...
+	}
+
+	public Stream<Object> stream() {
+		return Stream.of(pool);
```
bjhargrave, line 997: `return Arrays.stream(pool, 1, pool.length);` to avoid returning the invisible 0-index entry.

biz.aQute.bndlib/test/aQute/bnd/classfile/renamer/SignatureParserTest.java
```diff
@@ -0,0 +1,133 @@
...
+		AtomicInteger i = new AtomicInteger(0);
+
+		Files.lines(IO.getFile("testresources/shade/signatures.txt")
+			.toPath())
+			.forEach(l -> {
+				sp.parse(l);
+				i.incrementAndGet();
+			});
+
+		System.out.println(i);
```
bjhargrave, line 38: There are no test assertions in this method.

biz.aQute.bndlib/test/aQute/bnd/classfile/renamer/SignatureParserTest.java
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	@Test
+	public void testSpecial() {
+		sp.parse(
+			"(I)Lorg/springframework/security/config/annotation/web/configurers/openid/OpenIDLoginConfigurer<TH;>.AttributeExchangeConfigurer.AttributeConfigurer;");
+	}
```
bjhargrave, line 45: There are no test assertions in this method.

biz.aQute.bndlib/test/aQute/bnd/classfile/renamer/SignatureParserTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+	@Test
+	public void testRenamer() {
+		String s = "(I)Lorg/springframework/security/config/annotation/web/configurers/openid/OpenIDLoginConfigurer<TH;>.AttributeExchangeConfigurer.AttributeConfigurer;";
+		Map<String, String> map = new HashMap<>();
+		map.put("org/springframework/security/config/annotation/web/configurers/openid/OpenIDLoginConfigurer", "foo");
+		SignatureRenamer sr = new SignatureRenamer(map::get);
+		String rename = sr.rename(s);
+		System.out.println(rename);
```
bjhargrave, line 54: There are no test assertions in this method.

biz.aQute.bndlib/src/aQute/bnd/classfile/ClassFile.java
```diff
@@ -153,4 +161,15 @@ static ByteBuffer slice(DataInput in, int length) throws IOException {
...
 		return ByteBuffer.wrap(array, 0, length);
 	}
+
+	public byte[] toBytes() throws IOException {
+		ByteBufferDataOutput dout = new ByteBufferDataOutput();
+		write(dout);
+		return dout.toByteArray();
+	}
+
+	public static ClassFile parseFile(File file) throws FileNotFoundException, IOException {
```
bjhargrave, line 171: Don't need the FileNotFoundException since it is a subclass of the already listed IOException.

bjhargrave (changes requested):
Next round of feedback. Could we talk about the enum idea and how to possibly improve it?

biz.aQute.bndlib/src/aQute/bnd/classfile/Attribute.java
```diff
@@ -12,6 +12,54 @@
 
 @ProviderType
 public interface Attribute {
+
+	@ProviderType
+	enum AttributeTag {
```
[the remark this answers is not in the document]
pkriens: The problem they solve is that when you introduce a new type, the places where you depend on types will fire a warning. i.e. if you add a new Attribute, the methods where I depend on analyzing the attributes need to be inspected.

I understand that you want to have a preview phase, but that as a good example where this will help finding the places where there are dependencies.

biz.aQute.bndlib/src/aQute/bnd/classfile/Attribute.java
```diff
@@ -12,6 +12,54 @@
...
+		RuntimeVisibleAnnotations(RuntimeVisibleAnnotationsAttribute.class), //
+		RuntimeVisibleParameterAnnotations(RuntimeVisibleParameterAnnotationsAttribute.class), //
+		RuntimeVisibleTypeAnnotations(RuntimeVisibleTypeAnnotationsAttribute.class), //
+		Signature(SignatureAttribute.class), //
+		SourceDebugExtension(SourceDebugExtensionAttribute.class), //
+		SourceFile(SourceFileAttribute.class), //
+		StackMapTable(StackMapTableAttribute.class), //
+		Synthetic(SyntheticAttribute.class);
+
+		public Class<? extends Attribute> clazz;
```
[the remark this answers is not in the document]
pkriens: I thought enums could only have final fields but that might actually not be true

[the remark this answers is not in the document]
pkriens: I do not think we need an accessor when it is final. Just busy work.

biz.aQute.bndlib/test/aQute/bnd/classfile/renamer/SignatureParserTest.java
pkriens: It tests for exceptions not being thrown

biz.aQute.bndlib/test/aQute/bnd/classfile/renamer/SignatureParserTest.java
pkriens: It tests for exceptions not being thrown

biz.aQute.bndlib/test/aQute/bnd/classfile/renamer/SignatureParserTest.java
pkriens: ok

bjhargrave:
Closing as the new Transformer plugin for Bnd will fill this role.

PR #4385, opened 2020-11-01
Fix Domain.getBundleDocURL to return correct manifest header

maho7791:
- getBundleDocURL not returns the BUNDLE_DOCURL Manifest header

Signed-off-by: Mark Hoffmann <m.hoffmann@data-in-motion.biz>


bjhargrave (approved):

bjhargrave:
Thanks!

PR #4386, opened 2020-11-01
Fixed Windows paths not properly being sanitized

terminalsin:
This bug occurs when sourceSets is used in Gradle. This unfortunately cannot be fixed on an exterior basis and requires the following transformation to be made. Since this specific method exclusively calls paths, it should not contain and escaping characters anyways. Henceforth, we can replace such by a proper regex-friendly name.

The following stacktrace occurs on Gradle:

> Caused by: java.util.regex.PatternSyntaxException: Illegal/unsupported escape sequence near index 3
C:\Users\sanja\Documents\GitHub\ASM-9-Fork\asm\build\classes\java\main
   ^
	at aQute.libg.glob.Glob.toPattern(Glob.java:157)
	at aQute.libg.glob.Glob.<init>(Glob.java:30)
	at aQute.libg.glob.Glob.<init>(Glob.java:26)
	at aQute.bnd.osgi.Analyzer.getJarFromName(Analyzer.java:1763)
	at aQute.bnd.osgi.Analyzer.getClasspath(Analyzer.java:2522)
	at aQute.bnd.osgi.Analyzer$getClasspath$0.call(Unknown Source)
	[omitted]
* Get more help at https://help.gradle.org

After further inspection in the project (OW2 ASM), the following was found to be causing the issue:
`sourceSets.main.java.srcDirs += project(':asm').sourceSets.main.java.srcDirs`

Unfortunately, this issue cannot be fixed internally within ASM. Upon investigating the stacktrace, we can find that the origin of this issue comes from the following:

```java
String cp = getProperty(CLASSPATH);
if (cp != null)
	for (String s : split(cp)) {
		Jar jar = getJarFromName(s, "getting classpath");
```

Hence, the property classpath on all windows device starts with you guessed it: C:**\**, an escaping character. This issue **will persist on every function calling getJarFromName**, hence the changes needed.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -1759,7 +1759,7 @@ public class Analyzer extends Processor {
 	@Override
 	public Jar getJarFromName(String name, String from) {
 		Jar j = super.getJarFromName(name, from);
-		Glob g = new Glob(name);
+		Glob g = new Glob(name.replaceAll("\\\\", "\\\\"));
 		if (j == null) {
 			for (Jar entry : getClasspath()) {
 				if (entry.getSource() == null)
@@ -1780,7 +1780,7 @@ public class Analyzer extends Processor {
 		if (j != null)
 			return Collections.singletonList(j);
 
-		Glob g = new Glob(name);
+		Glob g = new Glob(name.replaceAll("\\\\", "\\\\"));
 		List<Jar> result = new ArrayList<>();
 		for (Jar entry : getClasspath()) {
 			if (entry.getSource() == null)


bjhargrave:
What happened here? @terminalsin why are you pushing unrelated commits to this PR?

terminalsin:
> 
> 
> What happened here? @terminalsin why are you pushing unrelated commits to this PR?

Oh lord I had one hell of a bad experience with the signoff not properly working with my site's email. Now the check seems to work properly. It's all good.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -1759,7 +1759,7 @@ protected void begin() {
 	@Override
 	public Jar getJarFromName(String name, String from) {
 		Jar j = super.getJarFromName(name, from);
-		Glob g = new Glob(name);
+		Glob g = new Glob(name.replaceAll("\\\\", "\\\\"));
```
bjhargrave, line 1762: I don't think this is correct. You do not describe the actual problem you are trying to solve (nor include a test case to demonstrate the fix solves the problem) so it is hard to know what your fix is supposed to fix. This method is about compare the file _name_ not the file path (the method is getJarFromName rather than getJarFromPath). The file name thus does not include file separator characters and should not need escaping. 

Can you please support more information in the specific problem you think you are solving?

bjhargrave (changes requested):
And please remove the other commit you added to this PR.

bjhargrave:
> It's all good.

It is not good in that your PR includes someone else's commit in addition to your own. Please only include your commit in this PR.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
terminalsin: > 
> 
> > It's all good.
> 
> It is not good in that your PR includes someone else's commit in addition to your own. Please only include your commit in this PR.

I've got no idea how that happened. Correcting it right away.

terminalsin: As for the issue I'm trying to solve, it's in the commit description:

> This bug occurs when sourceSets is used in Gradle. This unfortunately cannot be fixed on an exterior basis and requires the following transformation to be made. Since this specific method exclusively calls paths, it should not contain and escaping characters anyways. Henceforth, we can replace such by a proper regex-friendly name.

bjhargrave: The description does not really help me understand the problem you are trying to fix. SourceSets are always used in gradle and things are working for many. The getJarFromName method is not connected directly to gradle source sets. Any `name` argument to getJarFromName will come from a Bnd instruction such as `-includeresource`. So I will need a more concrete example of your problem. Such as a small git repo which demonstrates it.

And the getJarFromName method compares the name argument to file names and file names don't contain file separator characters.

bjhargrave:
Also, do not include the DCO text in your commit message. That just makes the commit message unnecessarily large.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
terminalsin: The following stacktrace occurs on Gradle:

> Caused by: java.util.regex.PatternSyntaxException: Illegal/unsupported escape sequence near index 3
C:\Users\sanja\Documents\GitHub\ASM-9-Fork\asm\build\classes\java\main
   ^
	at aQute.libg.glob.Glob.toPattern(Glob.java:157)
	at aQute.libg.glob.Glob.<init>(Glob.java:30)
	at aQute.libg.glob.Glob.<init>(Glob.java:26)
	at aQute.bnd.osgi.Analyzer.getJarFromName(Analyzer.java:1763)
	at aQute.bnd.osgi.Analyzer.getClasspath(Analyzer.java:2522)
	at aQute.bnd.osgi.Analyzer$getClasspath$0.call(Unknown Source)
	[omitted]
* Get more help at https://help.gradle.org

After further inspection in the project (OW2 ASM), the following was found to be causing the issue:
`sourceSets.main.java.srcDirs += project(':asm').sourceSets.main.java.srcDirs`

Unfortunately, this issue cannot be fixed internally within ASM. Upon investigating the stacktrace, we can find that the origin of this issue comes from the following:

```java
String cp = getProperty(CLASSPATH);
if (cp != null)
	for (String s : split(cp)) {
		Jar jar = getJarFromName(s, "getting classpath");
```

Hence, the property classpath on all windows device starts with you guessed it: C:**\**, an escaping character. This issue **will persist on every function calling getJarFromName**, hence the changes needed.

bjhargrave:
The updated description helps. I am working on a better fix. PR soon.

PR #4387, opened 2020-11-02
analyzer: Avoid regex errors on windows

bjhargrave:
On windows, entries on `-classpath` instruction can be passed to
`getJarFromName` which may attempt to make a `Glob` from the entry. On
Windows, the entry values can have backslashes which upset regex. So
we change to normalize the entry before passing to `Glob` which will
convert any backslash to forward slash. Since we only match file names
with the `Glob`, a pattern with forward slashes or backslashes wont match
the file name.

We also refactor to move shared logic into a private method.

Closes #4386


terminalsin:
Great job on this better fix!

PR #4390, opened 2020-11-03
fixed gpg signing for MavenBndRepository release

juergen-albert:

### DIFF
--- a/biz.aQute.repository/src/aQute/maven/nexus/provider/Signer.java
+++ b/biz.aQute.repository/src/aQute/maven/nexus/provider/Signer.java
@@ -70,9 +70,8 @@ public class Signer {
 	 */
 	public static int sign(File f, String cmdName, String key, String passphrase, File output) throws Exception {
 
-		Command cmd = new Command();
+		Command cmd = new Command(cmdName);
 
-		cmd.add(cmdName);
 		cmd.add("--batch");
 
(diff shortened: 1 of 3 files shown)


biz.aQute.repository/src/aQute/maven/nexus/provider/Signer.java
```diff
@@ -71,8 +71,7 @@ public static int sign(File f, String cmdName, String passphrase, File output) t
 	public static int sign(File f, String cmdName, String key, String passphrase, File output) throws Exception {
 
 		Command cmd = new Command();
-
-		cmd.add(cmdName);
+		cmd.add(cmdName.split(" "));
```
bjhargrave, line 74: This seems dangerous since the command could have quoted elements. You  want:

```java
cmd.full(cmdName);
```

`full` will safely split the argument.

bjhargrave (changes requested):

biz.aQute.repository/src/aQute/maven/nexus/provider/Signer.java
juergen-albert: Thanks for the advice! I used the constructor of Command, because it uses full internally.

bjhargrave (approved):

bjhargrave:
Thanks!

Issue #4082, opened 2020-05-20
Export of bndrun fails but existing and calculated runbundles are identical


njbartlett:
Bnd 5.0.1 cannot build the enRoute Quickstart project.

I created a sample project as follows:

```
$ mvn -B org.apache.maven.plugins:maven-archetype-plugin:3.0.1:generate \
    -DarchetypeGroupId=org.osgi.enroute.archetype \
    -DarchetypeArtifactId=project \
    -DarchetypeVersion=7.0.0 \
    -DgroupId=org.example \
    -DartifactId=example \
    -Dversion=1.0.0-SNAPSHOT \
    -Dpackage=org.example
$ cd example
```

I edited `pom.xml` to change `<bnd.version>4.1.0</bnd.version>` to `5.0.1` and then ran `mvn clean verify` and it failed on the export goal:

```
[INFO] --- bnd-export-maven-plugin:5.0.1:export (default) @ app ---
[ERROR] Error   : Fail on changes set to true (--xchange,-x) and there are changes
[ERROR] Error   :    Existing runbundles   [org.apache.felix.scr;version='[2.1.10,2.1.11)', org.example.impl;version='[1.0.0,1.0.1)']
[ERROR] Error   :    Calculated runbundles [org.apache.felix.scr;version='[2.1.10,2.1.11)', org.example.impl;version='[1.0.0,1.0.1)']
```

Note that the existing and calculated runbundles list are identical.

njbartlett:
NB the same bug exists in 5.1.0-SNAPSHOT and 5.0.0.

njbartlett:
Sorry the bug is not quite reproducible as described, I missed the resolve step (though the error *message* from the export is very misleading).

I still think there is a problem but it manifests in a more complex project, I will try again to produce a reproducible example.

jmini:
Even without changing the version from 4.1.0 to 5.2.0, I get an error when running the `mvn clean install` on the project generated by the archetype-generator:

```
[INFO] --- bnd-export-maven-plugin:4.1.0:export (default) @ app ---
[ERROR] Error   : The runbundles have changed. Failing the build!
Was: null
Is: \
	org.apache.felix.scr;version='[2.1.10,2.1.11)',\
	org.example.impl;version='[1.0.0,1.0.1)'
```

Tested with (`mvn -version`):

```
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
[...]

PR #4392, opened 2020-11-04
resolve: improve error message

jmini:
As discussed in #4082, the error message when `failOnChanges` is true is misleading.

The `Existing runbundles` line should display the current status in the `*.bndrun` file and not something modified.

Fixes: https://github.com/bndtools/bnd/issues/4082


bjhargrave (approved):

bjhargrave:
Thanks!

PR #4397, opened 2020-11-05
allow the use of macros in the gpg property

juergen-albert:
the gpg command can be extended and replaced, by putting a property gpg somewhere in the workspace. Till now, it was not possible to use macros and variables here. This commit chances this.

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
@@ -194,7 +194,7 @@ public class MavenBndRepository extends BaseRepository implements RepositoryPlug
 				}
 
 				logger.debug("Put release {}", pom.getRevision());
-				try (Release releaser = storage.release(pom.getRevision(), options.context.getProperties())) {
+				try (Release releaser = storage.release(pom.getRevision(), options.context.getFlattenedProperties())) {
 					if (releaser == null) {
 						logger.debug("Already released {}", pom.getRevision());
 						return result;


juergen-albert:
@bjhargrave It seems I'm to stupid to find in the CodeQL build log what test failed. Additionally the build seem to be a bit unstable in recent days. Do you know the cause of this?

bjhargrave:
> It seems I'm to stupid to find in the CodeQL build log what test failed.

The CodeQL build passed. It is the others that failed.



> Additionally the build seem to be a bit unstable in recent days. Do you know the cause of this?

Yes, we now seem to have two unstable test buckets. `:biz.aQute.bndall.tests:test` and `:bndtools.core.test:testOSGi`. The latter one already has an open issue. I don't know anything about the former one (sigh).

biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
```diff
@@ -593,6 +593,7 @@ private boolean isMaster(InfoRepository repo, String bsn, Version v) throws Exce
 		return descriptor.phase == Phase.MASTER;
 	}
 
+	// TODO: Needs to be removed. We allready have this method
```
bjhargrave, line 596: So why not remove it instead of passing the buck to the future with a comment? :-)

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
```diff
@@ -194,7 +194,7 @@ public PutResult put(InputStream stream, PutOptions options) throws Exception {
 				}
 
 				logger.debug("Put release {}", pom.getRevision());
-				try (Release releaser = storage.release(pom.getRevision(), options.context.getProperties())) {
```
bjhargrave, line 197: Why not just use flattened properties? It is big deal to change all these API from getting a Properties object to get a Processor object. I am not sure that is a great idea.

If flattened properties do not work, then I would suggest we implement a Properties subclass which wraps a Processor and pass that.

bjhargrave (changes requested):

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
juergen-albert: true that. Much easier solution!

biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
juergen-albert: Right you are. I accidentally committed the comment. Fixed it. I hope you don't mind that I did it in this PR.

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -1146,7 +1146,7 @@ private URI releaseRepo(RepositoryPlugin releaseRepo, Processor context, String
 		}
 	}
 
-	private List<RepositoryPlugin> getReleaseRepos(String names) {
+	public List<RepositoryPlugin> getReleaseRepos(String names) {
```
bjhargrave, line 1149: I don't think this needs to be public, just not private. So use default (package-private access).

biz.aQute.bndlib/src/aQute/bnd/build/package-info.java
```diff
@@ -1,6 +1,6 @@
 /**
  */
-@Version("3.5.0")
+@Version("3.6.0")
```
bjhargrave, line 3: Don't need this since we don't make the method public.

bjhargrave (changes requested):

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -1146,7 +1146,7 @@ private URI releaseRepo(RepositoryPlugin releaseRepo, Processor context, String
 		}
 	}
 
-	private List<RepositoryPlugin> getReleaseRepos(String names) {
+	protected List<RepositoryPlugin> getReleaseRepos(String names) {
```
bjhargrave, line 1149: Not protected. Protected is still API. Just default access (package-private).

bjhargrave (changes requested):

biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
```diff
@@ -594,29 +594,32 @@ private boolean isMaster(InfoRepository repo, String bsn, Version v) throws Exce
...
-					.equals(repoName)) {
-					return r;
-				}
-			}
-		}
-		if (repoName == null)
+		String repoNames = getProperty(Constants.RELEASEREPO);
+
+		if (repoNames == null) {
 			error("Could not find a writable repo for the release repo (-releaserepo is not set)");
```
bjhargrave, line 601: The error message seems a bit wrong since it mentions writeable repos when no one has even looked for a writeable repo. I also thought we decided not to change behavior and just warn if -releaserepo was not set but still release to the first writeable repo.

biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
```diff
@@ -594,29 +594,32 @@ private boolean isMaster(InfoRepository repo, String bsn, Version v) throws Exce
...
-		else
-			error("No such -releaserepo %s found", repoName);
+			return null;
+		}
+
+		List<RepositoryPlugin> releaseRepos = project.getReleaseRepos(repoNames);
 
+		if (!releaseRepos.isEmpty()) {
+			warning("Found multiple release repositories [%s], so we will use the first one", repoNames);
```
bjhargrave, line 608: Not empty can also mean 1. So the warning will trigger when there is only one repo in the list.

biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
```diff
@@ -594,29 +594,32 @@ private boolean isMaster(InfoRepository repo, String bsn, Version v) throws Exce
...
+			return releaseRepos.get(0);
+		}
+
+		error("No such -releaserepo %s found", repoNames);
```
bjhargrave, line 612: The message should be something like "no repo found to release into" since we still allow the old behavior of releasing into a writeable rep.

biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
```diff
@@ -594,29 +594,32 @@ private boolean isMaster(InfoRepository repo, String bsn, Version v) throws Exce
...
 		return null;
 	}
 
 	private RepositoryPlugin getBaselineRepo() {
 		String repoName = getProperty(Constants.BASELINEREPO);
-		if (repoName == null)
+		if (repoName == null) {
+			warning("baselining is active, but no %s is set. Will fall back to release repositories",
```
bjhargrave, line 619: For these messages, we should use proper grammar. So we should capitalize "Baselining".

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
```diff
@@ -194,7 +194,7 @@ public PutResult put(InputStream stream, PutOptions options) throws Exception {
...
+				try (Release releaser = storage.release(pom.getRevision(), options.context.getFlattenedProperties())) {
```
bjhargrave, line 197: I think this change should be in a separate commit from the other changes to repo finding since they are totally unrelated. This enable proper cherry-picking if we want.

bjhargrave (changes requested):

bjhargrave (approved):

bjhargrave:
Thanks!

PR #4404, opened 2020-11-10
[bndtools] Gracefully handle missing cnf

### LINKED ISSUE #4203: [bndtools] Bndtools should mark error for projects outside of the Bnd workspace
I usually have my Eclipse workspace (containing the workspace metadata) in a different directory to my Bnd workspace. I'm happy to take advice if people feel that this is not the best approach, but it has some advantages. One drawback of this approach, however, is that when you copy a project in your Eclipse workspace, by default it copies it into the Eclipse workspace and not the Bnd workspace. 

Things don't work properly when you do this - in particular, the Bndtools builder won't generate the artifacts. But there is no direct mention of the fact that this is an issue.

In my case I did have a few errors come up like this one:
```
Error in Bnd resolution analysis.
java.lang.IllegalArgumentException: A Jar can only accept a file or directory that exists: C:\workspace\idempiere-test-workspace\org.idempiere.test.assertj\generated\org.idempiere.test.jar
	at aQute.bnd.osgi.Jar.<init>(Jar.java:138)
	at aQute.bnd.osgi.Jar.<init>(Jar.java:202)
	at bndtools.tasks.JarFileCapReqLoader.getBuilder(JarFileCapReqLoader.java:21)
	at bndtools.tasks.BndBuilderCapReqLoader.load(BndBuilderCapReqLoader.java:56)
	at bndtools.tasks.BndBuilderCapReqLoader.loadCapabilities(BndBuilderCapReqLoader.java:92)
	at bndtools.tasks.AnalyseBundleResolutionJob.run(AnalyseBundleResolutionJob.java:61)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```
However, this error does not directly mention the root cause of the problem. Because my eyes are normally drawn to the end of the path (which also had an issue), it was a while before I noticed that it was in the `C:\workspace\idempiere-test-workspace` directory when it should have been in the `C:\workspace\idempiere-test` directory (where the Bnd workspace is). 

It should be a fairly simple matter for Bndtools to place some kind of error marker on a project when it finds that it is not in the Bnd workspace - that would have saved me a good hour or so trying to track down the cause of this issue.

### LINKED ISSUE #4401: [bndtools] Bndtools doesn't handle cnf project created after starting
* Start with a normal Bndtools workspace in Eclipse.
* Delete the cnf project (without deleting the contents on disk). The Bndtools Builder will get into an annoying infinite loop with build errors, with a Build Error dialog popping up stating that the cnf project is missing. (While this could be handled more gracefully, that is a separate issue to the one I'm describing in this issue.)
* Re-import the cnf project. Annoying infinite loop and Build Error dialog stops.
* However, now every project's build is failing with an error marker (among others) stating "Bndtools: Error in setup, likely the cnf folder is not ../cnf relative to the project folder '&lt;path to project&gt;'. The workspace is in '&lt;path to workspace&gt;'.
* This situation seems to persists until you restart Eclipse, then the build starts normally.

Admittedly, this is not a common operation in ordinary development - you normally create your cnf project at the start and it remains for the life of the workspace. However, I believe it may be the root cause of #4253, or at the very least closely related - the integration test, in the `@BeforeAll` phase, clears the workspace of all projects (including cnf) and then re-imports them. My experience through the above demonstrates that there is a race condition in Bndtools such that it doesn't always react perfectly to the disappearance and reappearance of the cnf project, and it ends up remembering that the cnf project is missing. The code in the Bndtools builder that generates that error generates it when `Central` returns `null` when trying to get the Bnd `Project` object:

https://github.com/bndtools/bnd/blob/a45a1820b17b0604474f85bc7aeaaf9b40880ae6/bndtools.builder/src/org/bndtools/builder/validate/ProjectPathsValidator.java#L77-L89

This is the same code path that triggers when the integration test fails.

I believe that this may also have been the root cause of #4202. I think at the time that that happened, I was messing around with my cnf and may have deleted/reimported it. It looks like a similar error.

kriegfrj:
Without this fix, Bndtools generates an NPE during building if the cnf project is missing and continues to fail building even after the cnf project is added. This fixes #4401.

As a side effect, also introduces an error marker if the Bnd project is not inside the Bnd workspace. This fixes #4203.

This seems to work in my testing and not to break anything else, but it's hard to be sure without a decent set of regression tests. It would be good for us to smoke-test it on dev for a while before release.

### DIFF
--- a/bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
+++ b/bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
@@ -1,6 +1,8 @@
 package org.bndtools.builder;
 
 import java.io.File;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -25,6 +27,7 @@ import org.eclipse.core.resources.IWorkspaceRoot;
 import org.eclipse.core.resources.IncrementalProjectBuilder;
 import org.eclipse.core.resources.ProjectScope;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Status;
@@ -33,6 +36,7 @@ import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.ui.preferences.ScopedPreferenceStore;
 
 import aQute.bnd.build.Project;
+import aQute.bnd.build.Workspace;
 import aQute.bnd.osgi.Constants;
 import aQute.lib.io.IO;
 import bndtools.central.Central;
@@ -71,7 +75,6 @@ public class BndtoolsBuilder extends IncrementalProjectBuilder {
 		CnfWatcher.install();
 	}
 
-	private Project		model;
 	private BuildLogger	buildLog;
 	private IProject[]	dependsOn;
 	private boolean		postponed;
@@ -113,15 +116,46 @@ public class BndtoolsBuilder extends IncrementalProjectBuilder {
 					.getDynamicReferences();
 			}
 
-			if (model == null) {
-				try {
-					model = Central.getProject(myProject);
-				} catch (Exception e) {
-					markers.deleteMarkers("*");
-				}
-				if (model == null)
-					return noreport();
+			Project ourModel = null;
+			try {
+				ourModel = Central.getProject(myProject);
+			} catch (Exception e) {
+				markers.deleteMarkers("*");
+				logger.logError("Exception while trying to fetch bnd project for " + myProject.getName(), e);
+				markers.createMarker(null, IMarker.SEVERITY_ERROR, "Exception while trying to fetch bnd project: " + e,
+					BndtoolsConstants.MARKER_BND_PATH_PROBLEM);
+				return noreport();
 			}
+			if (ourModel == null) {
+				markers.deleteMarkers("*");
+				// Do a bit of digging to see if we can find the reason
+				Workspace ws = Central.getWorkspaceIfPresent();
+				String msg = "Cannot find bnd project: check that it is actually a bnd project and is part of the bnd workspace";
+				if (ws == null) {
+					msg = "No Bnd workspace is available";
+				} else if (ws.isDefaultWorkspace()) {
+					msg = "The Bnd cnf directory is not part of the Eclipse workspace";
+				} else {
+					IPath projPath = myProject.getLocation();
+
+					if (projPath == null) {
+						msg = "The project does not exist on the local file system";
+					} else {
+						Path javaPath = projPath.toFile()
+							.toPath();
+						Path wsPath = ws.getBase()
+							.toPath();
+						if (!Files.isSameFile(javaPath.getParent(), wsPath)) {
+							msg = "The project directory: " + javaPath + " is not a subdirectory of the bnd workspace "
+								+ wsPath;
+						}
+					}
+				}
+				markers.createMarker(null, IMarker.SEVERITY_ERROR, msg, BndtoolsConstants.MARKER_BND_PATH_PROBLEM);
+				return noreport();
+			}
+
+			final Project model = ourModel;
 
 			try {
 				return Central.bndCall(() -> {
@@ -222,7 +256,7 @@ public class BndtoolsBuilder extends IncrementalProjectBuilder {
 					if (model.isNoBundles()) {
 						buildLog.basic("-nobundles was set, so no build");
 						buildLog.setFiles(0);
-						return report(markers);
+						return report(model, markers);
 					}
 
 					if (markers.hasBlockingErrors(delta)) {
@@ -240,7 +274,7 @@ public class BndtoolsBuilder extends IncrementalProjectBuilder {
 									"Will not build project %s until the compilation and/or path problems are fixed, output files are kept.",
 									myProject.getName());
 							}
-							return report(markers);
+							return report(model, markers);
 						}
 						buildLog.basic("Blocking errors, continuing anyway");
 						model.warning("Project %s has blocking errors but requested to continue anyway",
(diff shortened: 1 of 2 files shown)


bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
```diff
@@ -113,15 +112,21 @@
...
-				if (model == null)
-					return noreport();
+			Project ourModel = null;
+			try {
+				ourModel = Central.getProject(myProject);
+			} catch (Exception e) {
+				markers.deleteMarkers("*");
 			}
+			if (ourModel == null) {
+				markers.deleteMarkers("*");
```
bjhargrave, line 122: You now call `markers.deleteMarkers("*")` twice if you get an exception on `Central.getProject(myProject)`.

bndtools.builder/src/org/bndtools/builder/MarkerSupport.java
```diff
@@ -98,7 +98,7 @@ private void deleteCnfMarkersForThisProject(String markerType) throws CoreExcept
 		IProject cnf = ResourcesPlugin.getWorkspace()
 			.getRoot()
 			.getProject("cnf");
-		if (cnf != null) {
+		if (cnf != null && cnf.exists()) {
```
bjhargrave, line 101: Should this not be `cnf.isOpen()`?

kriegfrj: I don't know enough about Eclipse's marker handling to know the answer to this question. Checking for existence was sufficient to stop the NPE from happening in the enclosing code if the cnf project was missing. But checking for openness is probably not a bad idea.

bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
kriegfrj: Fixed.

kriegfrj:
Ok, latest push fixes the issues raised by @bjhargrave and adds a number of other enhancements:

1. If there is an error at this level for a particular project, then not only are all the Bnd markers deleted, but all the Java markers as well. The reason for this: If your cnf is missing/misconfigured in some way, there will inevitably be build path problems that will lead to dozens or hundreds of compile errors in your project. Not only are these errors not particularly helpful (as they don't point directly to the root cause), there are so many of them that the root cause can get lost in the noise. So now we remove them to make the root cause more prominent.
2. The builder now performs some more precise diagnostic tests once the missing `Project` is detected, in order to give a more detailed and specific report as to the possible reason for the failure.

@bjhargrave, I'm also hopeful that this will fix #4253. If you can keep an eye on the integration test after this PR is merged and if that failure stops popping up maybe we can re-instate the integration test. I'll also look at adding regression tests for the fixes in this PR if that's the case.

bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
```diff
@@ -113,16 +118,57 @@
...
+			Project ourModel = null;
+			try {
+				ourModel = Central.getProject(myProject);
+			} catch (Exception e) {
+				markers.deleteMarkers("*");
+				// If the cnf project is missing there will likely be a myriad
+				// of Java problems. These do nothing but clutter the error
+				// output and make it hard to find the real problem in the
+				// problem log.
+				myProject.deleteMarkers(IJavaModelMarker.JAVA_MODEL_PROBLEM_MARKER, true, IResource.DEPTH_INFINITE);
```
bjhargrave, line 130: It does seem a bit presumptuous of us (Bndtools) to remove the error markers of another builder, no? I am not sure we should do this.

bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
```diff
@@ -113,16 +118,57 @@
...
+					BndtoolsConstants.MARKER_BND_PATH_PROBLEM);
+				return noreport();
+			}
+			if (ourModel == null) {
+				markers.deleteMarkers("*");
+				// If the cnf project is missing there will likely be a myriad
+				// of Java problems. These do nothing but clutter the error
+				// output and make it hard to find the real problem in the
+				// problem log.
+				myProject.deleteMarkers(IJavaModelMarker.JAVA_MODEL_PROBLEM_MARKER, true, IResource.DEPTH_INFINITE);
```
bjhargrave, line 142: It does seem a bit presumptuous of us (Bndtools) to remove the error markers of another builder, no? I am not sure we should do this.

bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
kriegfrj: I suppose i could try and see if we can coerce the JavaBuilder to do it itself under these conditions, eg by failing the BndClasspathContainer somehow.

bjhargrave: I don't think that is a good idea or possible either. The Java builder's errors are real. They are compilation failures. And they are the responsibility of the Java builder to manage.

kriegfrj: I disagree, but rather than hold up this PR over this issue (which is not directly related to the original problem)  I have removed the offending lines for now so that we can get it merged. We can discuss this further at another time.

pkriens:
Just a comment in this area ... why not use the Bndtools Explorer to report about this case? The Bndtools Explorer is really about the Workspace model and it can easily provide a lot of information that cnf is missing.

kriegfrj:
@bjhargrave, are you still waiting for me to do something before meeting this?

bjhargrave:
The build is bad

kriegfrj:
The failure is in bndlib tests. I can't see how that was caused by this PR?

bjhargrave:
yes, but it is in another test you authored which is now regularly failing CI builds like this one.

https://github.com/bndtools/bnd/pull/4404/checks?check_run_id=1389155797#step:5:449

Can you look into this? I would like to stabilize the CI builds.

bjhargrave (approved):

PR #4423, opened 2020-11-18
[projecttracker] Missing bnd files

pkriens:
The project tracker (suddenly?) needs bnd files
to be present in the project. However, if you
create this file, the error is not cleared until the project tracker updates again.

This patch will force an update if the bnd file exist but it is not in the project tracker.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/ProjectTracker.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/ProjectTracker.java
@@ -57,6 +57,11 @@ class ProjectTracker implements AutoCloseable {
 	 */
 	synchronized Optional<Project> getProject(String name) {
 		update();
+		if (!models.containsKey(name) && workspace.getFile(name + "/" + Project.BNDFILE)
+			.isFile()) {
+			changed = true;
+			update();
+		}
 		return Optional.ofNullable(models.get(name));
 	}
 


bjhargrave:
> The project tracker (suddenly?) needs bnd files
> to be present in the project.

This has always been the case.

biz.aQute.bndlib/src/aQute/bnd/build/ProjectTracker.java
```diff
@@ -57,6 +57,11 @@ synchronized void refresh() {
 	 */
 	synchronized Optional<Project> getProject(String name) {
 		update();
+		if (!models.containsKey(name) && workspace.getFile(name + "/" + Project.BNDFILE)
```
bjhargrave, line 60: The ProjectTracker was designed to require an external call to `refresh()` (from the workspace), to force an update. What has changed to need this new behavior? If there is a change needed, perhaps it should be in the the workspace rather than here?

pkriens: Well, we're not having a tracker on the files, do we?

I'd like to take this patch and then the people that worked in this area can look. It is terribly annoying since I now have to restart eclipse to get rid of the error.

bjhargrave: The ProjectTracker has not really changed since 4.3.0, so I am wondering what has changed elsewhere to trigger whatever issue you are seeing.

The genesis of the ProjectTracker was to not always go to disk to see what the projects where. Originall, the workspace always went to disc to see the set of project and you replaced that with the ProjectTracker. A change like this is heading back to checking the file system again.



> hen the people that worked in this area can look

Whom do you mean here? The people who worked on the ProjectTracker class (you and me)? Or someone else?

>  have to restart eclipse to get rid of the error.

So this is a Bndtool side error? Perhaps your probelm was caused by https://github.com/bndtools/bnd/pull/4404 which changed how the BndtoolsBuilder looked for projects and did error markers? That maybe were we need a fix?

pkriens: Well, the problem is genuine, the bnd.bnd file did not exist. So the marker, albeit a bit opaque, is good. However, the fix is to cure this when someone adds the file. It is a PT problem that it does not listen to changes on disk. 

The project tracker gets a free opportunity to check if it needs to refresh, what on earth is against that?

bjhargrave: > It is a PT problem that it does not listen to changes on disk.

That was the whole point of ProjectTracker. It was not supposed to listen to the disk. It was supposed to be told externally (via a call to refresh) to recheck the disk. If we want ProjectTracker to follow the disk, we should probably do a more thorough job :-)

pkriens: You mind if I commit this?

bjhargrave: It seems the wrong place to fix things and it can double update, but I don't think it will break anything.

pkriens: Why is the wrong place? Bernie calls you if the mail has arrived. You have no mail on your desk so you tell her "NO!". Bernie pissed because that mail she is expecting is important to her. Now, since Bernie asked, isn't that an indication that it actually might have arrived? 

Yes, a better solution would be to have a dog that barks at the postman arriving but since we don't have a dog, I would look out of the window if the flag is up on the mailbox before saying NO? 

Anyway, I am going to commit this. I'll add some info to the error and move it to abeyance.

bjhargrave: Why are you making this personal? Please remove personal information from this discussion that has zero to do with the technical issue. Adding personal information is not professional.

pkriens: ok, removed personal, but it was a good illustration

kriegfrj: I think that in the context of Bndtools, the correct program flow is for Bndtools to do the file watching (using a `ResourceChangeListener`)  and for this listener to call `ProjectTracker.refresh()` if necessary. That is how Bndtools reacts to most other changes on the file system, so I don't see why this should be any different. You might have to manually refresh the project if you create the bnd.bnd file outside of Eclipse, but other than that it should work. Just my 2c.

pkriens: @kriegfrj that is why I moved it to abyeance. I agree that is the gold standard but I do not have time and this is at least handling an important use case.

Issue #4396, opened 2020-11-05
${.} macro processing issue when resolving in Bndtools (IDE)


amitjoy:
I did setup this small [repo](https://github.com/amitjoy/bndtools-macro-bug) where I am using `-distro` instruction to specify a bundle comprising all the distro specific capabilities. To specify the distro, I did use `{.}` to locate the jar in the current working directory but I cannot resolve the bndrun anymore since the distro capabilities cannot be found by the resolver. On the contrary, when I execute `./gradlew :com.amitinside.sample.product:resolve.product`, it works perfectly. If I change `${.}` to use the location using `${workspace}`, it works. Somehow `${.}` doesn't work in IDE anymore but few months earlier, it worked in IDE too.

PR #4424, opened 2020-11-18
bndtools: Use model Project as the properties for resolving

bjhargrave:
The model properties are not properly resolved for the special here
`${.}` macro. Using the model Project as the resolve properties is
consistent with all other callers to RunResolution.resolve.

Fixes https://github.com/bndtools/bnd/issues/4396


pkriens:
I do not think this works. You need the BndEditModel because this contains temporary changes to the bnd file. Now you always get what is on disk.

I guess the BndEditModel needs to be fixed to understand what ${.} means.

bjhargrave:
Ohh. Here is (hopefully) a better fix: https://github.com/bndtools/bnd/pull/4425

PR #4425, opened 2020-11-18
bndtools: Fixup model properties for resolving

bjhargrave:
The model properties are not properly resolved for the special here
`${.}` macro. So make a copy and call replaceHere on the copy.

Fixes https://github.com/bndtools/bnd/issues/4396

### DIFF
--- a/bndtools.core/src/org/bndtools/core/resolve/ResolveOperation.java
+++ b/bndtools.core/src/org/bndtools/core/resolve/ResolveOperation.java
@@ -1,8 +1,10 @@
 package org.bndtools.core.resolve;
 
+import java.io.File;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Properties;
 
 import org.bndtools.core.resolve.ResolutionResult.Outcome;
 import org.eclipse.core.runtime.IProgressMonitor;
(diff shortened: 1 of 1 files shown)


pkriens:
Should the code to fixup `${.}` to the owning file not be part of the BndEditModel? To fixup on the place where you use it means we have the same bug everywhere we use the BndEditoModel properties.

bjhargrave:
> Should the code to fixup `${.}` to the owning file not be part of the BndEditModel?

I guess BndEditModel would need a getProperties method which resolved the `${.}` here macro since you would not want it resolved in the underlying properties of the model which would be written back to the file being edited. But I am not super familiar with BndEditModel, so I would need to leave such an improvement to you.

> To fixup on the place where you use it means we have the same bug everywhere we use the BndEditoModel properties.

How many other places do you think there would be? It would need to be something which uses the properties to then possible mutate the properties like resolving does.

pkriens:
The purpose of the bndedit model is to keep the changes from the file. Every bit of GUI works from this.

As far as I can see, you can just move the code change you did for the RunResolution into the properties you get from the BndEditModel. I do not think there is anyone that would have a problem with fixed up `${.}`, on the contrary. Not doing this can create very subtle bugs like this one.

As far as I can see, the -runbundles update has to go to the BndEditModel and not the file. So you can return a copy from getProperties(), as you do in your fix.

bjhargrave:
OK, I moved the here macro resolution into BndEditModel.getProperties. See https://github.com/bndtools/bnd/pull/4426

PR #4437, opened 2020-12-03
[maven] inferring the EE from maven build resolves to 1.8 regardless …

### LINKED ISSUE #4435: [maven] inferring the EE from maven build resolves to 1.8 regardless of Java versioning running
Up to and including 5.1.2 when I change JAVA_HOME and rerun a maven build the bnd-*-maven-plugins that infer the EE from the build java would properly detect the correct EE to use. With bnd 5.2.0 this always resolves to 1.8 instead.

This causes builds to fail when something like jaxb is used, which in Java 8 is part of the JDK, but after 10 is no longer included.

There is no error, however the info emitted from the mojos does show the incorrect EE:
```
]$ mvn -version
Apache Maven 3.6.3 (cecedd343002696d0abb50b32b541b8a6ba2883f)
Maven home: /home/rotty/bin/maven/apache-maven-latest
Java version: 11.0.8, vendor: Azul Systems, Inc., runtime: /usr/lib/jvm/zulu-11-amd64
Default locale: en_CA, platform encoding: UTF-8
OS name: "linux", version: "5.4.0-55-generic", arch: "amd64", family: "unix"
...
]$ mvn clean verify
...
[INFO] --- bnd-resolver-maven-plugin:5.2.0:resolve (owb-resolve) @ org.apache.aries.cdi.executable ---
[INFO] Bnd inferred -runee: JavaSE-1.8
...
```
As indicated reverting to bnd 5.1.2 produces the correct EE.

rotty3000:
…of Java versioning running

fixes #4435

### DIFF
--- a/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
+++ b/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
@@ -323,12 +323,14 @@ public class BndrunContainer {
 						return property;
 					}
 					// so fallback to the currently running Java version
-					return System.getProperty("java.version");
+					return System.getProperty("java.specification.version");
 				})
 				.flatMap(EE::highestFromTargetVersion)
-				// if that all fails at least we know bnd needs at least
-				// Java 8 at this point
-				.orElse(EE.JavaSE_1_8);
+				.orElseGet(() -> Optional.ofNullable(System.getProperty("java.specification.version"))
+					.flatMap(EE::highestFromTargetVersion)
+					// if that all fails at least we know bnd needs at least
+					// Java 8 at this point
+					.orElse(EE.JavaSE_1_8));
 
 			run.setProperty(Constants.RUNEE, ee.getEEName());
 
--- a/biz.aQute.bndlib.tests/test/test/model/EETest.java
+++ b/biz.aQute.bndlib.tests/test/test/model/EETest.java
@@ -49,7 +49,7 @@ public class EETest {
 
 	@Test
 	public void getEEFromJvm() throws Exception {
-		String java_version = System.getProperty("java.version");
+		String java_version = System.getProperty("java.specification.version");
 		Version version = MavenVersion.parseMavenString(java_version)
 			.getOSGiVersion();
 		assertThat(EE.highestFromTargetVersion(java_version)).hasValueSatisfying( //


biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
```diff
@@ -323,12 +323,13 @@ public void setEEfromBuild(Run run) {
 						return property;
 					}
 					// so fallback to the currently running Java version
-					return System.getProperty("java.version");
+					return System.getProperty("java.vm.specification.version");
```
bjhargrave, line 326: I think this should be `java.specification.version` rather than `java.vm.specification.version`. See https://docs.oracle.com/javase/8/docs/technotes/guides/versioning/spec/versioning2.html#wp90020.

We use `java.specification.version` elsewhere in Bnd. https://github.com/bndtools/bnd/blob/a92691029d5db01863dfc73f570f1839c9783411/biz.aQute.bndlib/src/aQute/bnd/build/defaults.bnd#L30-L31

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
```diff
@@ -323,12 +323,13 @@ public void setEEfromBuild(Run run) {
...
 				})
 				.flatMap(EE::highestFromTargetVersion)
-				// if that all fails at least we know bnd needs at least
-				// Java 8 at this point
-				.orElse(EE.JavaSE_1_8);
+				.orElseGet(() -> EE.highestFromTargetVersion(System.getProperty("java.vm.specification.version"))
```
bjhargrave, line 329: I think this should be `java.specification.version` rather than `java.vm.specification.version`.

There is a minuscule possibility that the System.getProperty call could return null and then the EE.highestFromTargetVersion call will NPE. So this should probably be:

```java
.orElseGet(() -> Optional.ofNullable(System.getProperty("java.specification.version"))
    .flatMap(EE::highestFromTargetVersion)
    // if that all fails at least we know bnd needs at least
    // Java 8 at this point
    .orElse(EE.JavaSE_1_8));
```

bjhargrave (changes requested):

bjhargrave:
You should also include this fix (while we are fixing usages of the `java.version` system property for EEs).

```diff
diff --git a/biz.aQute.bndlib.tests/test/test/model/EETest.java b/biz.aQute.bndlib.tests/test/test/model/EETest.java
index 436f7bcd5..621bfe353 100644
--- a/biz.aQute.bndlib.tests/test/test/model/EETest.java
+++ b/biz.aQute.bndlib.tests/test/test/model/EETest.java
@@ -49,7 +49,7 @@ public class EETest {
 
        @Test
        public void getEEFromJvm() throws Exception {
-               String java_version = System.getProperty("java.version");
+               String java_version = System.getProperty("java.specification.version");
                Version version = MavenVersion.parseMavenString(java_version)
                        .getOSGiVersion();
                assertThat(EE.highestFromTargetVersion(java_version)).hasValueSatisfying( //
```

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
rotty3000: done!

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
rotty3000: done!

bjhargrave (approved):

PR #4448, opened 2020-12-09
removed duplicate getReleaseRepo method

juergen-albert:

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
@@ -594,29 +594,29 @@ public class ProjectBuilder extends Builder {
 	}
 
 	private RepositoryPlugin getReleaseRepo() {
-		String repoName = getProperty(Constants.RELEASEREPO);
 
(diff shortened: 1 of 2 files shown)


biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java
```diff
@@ -594,29 +594,34 @@ private boolean isMaster(InfoRepository repo, String bsn, Version v) throws Exce
...
-		for (RepositoryPlugin r : repos) {
-			if (r.canWrite()) {
-				if (repoName == null || r.getName()
-					.equals(repoName)) {
-					return r;
-				}
+		String repoNames = getProperty(Constants.RELEASEREPO);
+
+		if (repoNames == null) {
+			error("Could not find a release reposiotry (-releaserepo is not set)");
```
bjhargrave, line 601: There is a misspelling here. Also, we agreed that missing `-releaserepo` was a warning and not an error. And if `-releaserepo` was not set, we still release to the first writable repo for backwards compatibility.

bjhargrave (changes requested):

bjhargrave (approved):

PR #4457, opened 2020-12-18
[resource] Main-Class attribute not trimmed

pkriens:
The Main-Class attribute was not trimmed. This
caused extraneous spaces to be at the end sometimes.

This trims the Main-Class and Bundle-Version now.

Also unrolled a stream loop because the check for
capabilities was 20 levels deep in a stream. This makes it impossible to single step or debug.

Unless there  is a demonstrable _and_ significant speed gain we should stop replacing plain old loops by streaming. Not only because it is hard to debug, it also looses the earlier context when you want to report something. Maybe if Java had true closures and no checked exception it might have been useful.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceRepositoryDynamic.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceRepositoryDynamic.java
@@ -4,6 +4,7 @@ import static java.util.function.Function.identity;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
 
+import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceRepositoryDynamic.java
```diff
@@ -39,11 +40,16 @@
...
 	private List<Capability> findProvider(Collection<? extends Resource> resources, Requirement requirement) {
+		List<Capability> l = new ArrayList<>();
 		String namespace = requirement.getNamespace();
-		return resources.stream()
-			.flatMap(resource -> ResourceUtils.capabilityStream(resource, namespace))
-			.filter(ResourceUtils.matcher(requirement))
-			.collect(ResourceUtils.toCapabilities());
+		for (Resource r : resources) {
+			for (Capability c : r.getCapabilities(namespace)) {
+				if (ResourceUtils.matches(requirement, c)) {
```
bjhargrave, line 47: Well you lost an nice performance fix of using `ResourceUtils.matcher(requirement)`. `requirement` never changes in the loop, but `ResourceUtils.matches(requirement, c)` keeps recomputing the filter object every time through the O(n^2) loop. So there was reason for using stream and blindly changing them back can loose things.

pkriens: The code had become completely unreadable. I'll check this optimization. The optimization has got nothing to do with streams.

bjhargrave: I already have the fix in a commit ready to push.

PR #4463, opened 2020-12-18
identify apiguardian on any type it supports

rotty3000:
Enable a future where apiguardian supports packages.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -564,8 +564,9 @@ public class Analyzer extends Processor {
 
 		MapStream.of(getClassspace().values()
 			.stream()
-			.filter(c -> !c.isModule() && !c.isInnerClass() && !c.isSynthetic() && c.annotations()
-				.contains(apiAnnotation))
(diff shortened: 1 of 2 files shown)


bjhargrave:
The point of Bnd's support of apiguardian was to allow it to act as an `Export` annotation so that packages holding the type annotated with `API` are exported.

So removing `!c.isModule()` is wrong since even if you could apply `API` to a module-info, what package is Bnd supposed to export? Modules are always in the default package which cannot be exported in OSGi.

We had `!c.isSynthetic()` as a performance boost since no synthetic code generated by the compiler will be marked `API` so it is better to filter the class out early in the stream processing. I see in https://github.com/apiguardian-team/apiguardian/issues/19 that they want to allow applying `API` to package-info which is a synthetic class. So the better fix is to open up this test to allow package-info to proceed.

The removal of `!c.isInnerClass()` seems odd in that the outer class holding the inner class is not already marked `API`. That is, if an inner class is API, the outer class must also be API, no? It would then be a programming mistake to not also mark the outer class as `API`.

So I updated the PR to add an `isPackageInfo()` method to `Clazz` and then use that method in the filter to include package-info classes.

rotty3000:
Yes you're right of course. I tried to take the easy road and that was wrong. Thanks for the correct fix. You didn't have to go that far 🙂

PR #4467, opened 2020-12-21
[classindex] Add hashes when memoizing direct from Jar

### LINKED ISSUE #4288: [quickfix] Workspace repository doesn't match class names
It appears that the quick fix processor is unable to find classes that are declared in other projects in the workspace. It matches on exported packages, but not classes. Suspect that bnd.hashes is not being generated/used for the workspace repository?

kriegfrj:
Fixes #4288.

Note that the pre-existing code tried hard not to keep a reference to the Jar object. In order to honour this, it is necessary to calculate the hashes before the Jar object is discarded, which means that the hash calculation cannot be deferred into the memoization object. This could be a potential performance hit at build time. If this is an issue, the only sensible way around it (that I can see) is to get rid of this direct-from-the-Jar object optimisation altogether and simply use `parseBuildResources()`. This will make the memo creation slower when it happens.


bjhargrave (approved):
Well done!

PR #4470, opened 2020-12-21
running tests with remote agent hangs forever #4469

rotty3000:


biz.aQute.tester/src/aQute/tester/plugin/ProjectTesterImpl.java
```diff
@@ -83,6 +83,14 @@ protected void updateFromProject() throws Exception {
 
 		launcher.addRunBundle(me.getFile()
 			.getAbsolutePath());
+
+		//
+		// We also add it to the project to handle the case of the remote
+		// launcher that wants to sync the bundles with the runtime dynamically
+		//
+
+		getProject().getRunbundles()
```
bjhargrave, line 92: This seems an unwanted side effect. We already add `me` to the launcher above. Jamming into the project seems wrong. Every time you launch, you will add it again, no? I think some other solution is needed.

biz.aQute.remote/src/aQute/remote/plugin/RunSessionImpl.java
```diff
@@ -96,8 +97,13 @@ public int launch() throws Exception {
 
 			Agent agent = supervisor.getAgent();
 
-			if (agent.isEnvoy())
+			if (agent.isEnvoy()) {
 				installFramework(agent, dto, properties);
+			} else {
+				for (Entry<String, String> entry : properties.entrySet()) {
+					agent.setSystemProperty(entry.getKey(), entry.getValue());
```
bjhargrave, line 104: Doesn't this just stack up side effects since the system properties are not restored when done?

rotty3000: I'm open to other solutions. We'd have to rewrite a testers to get information via some other means. I can't see that happening in short order. This at least is a short term yet pretty safe fix. No one else cares about these properties... and the ONLY way to hand over `-runproperties` in the remote use case.

biz.aQute.tester/src/aQute/tester/plugin/ProjectTesterImpl.java
rotty3000: There's no other way. the remote launcher is blowing away the runbundles set using `launcher.addRunBundle`.
 I'm open to other options.

rotty3000:
Please see the following for an example https://github.com/rotty3000/liferay-workspace-integration-test-examples

biz.aQute.tester/src/aQute/tester/plugin/ProjectTesterImpl.java
rotty3000: - the tester gets the launcher instance
- the tester calls prepare() where it adds the tester bundle to launcher runbundles
- the tester then calls getProjectLauncher().launch(); which calls launcher.prepare() which calls launcher.updateFromProject(); which calls runbundles.clear();

Also, RemoteProjectLauncherPlugin inherits from aQute.bnd.build.ProjectLauncher which has https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/build/ProjectLauncher.java#L148

biz.aQute.remote/src/aQute/remote/plugin/RunSessionImpl.java
rotty3000: I guess we could track the original value if any and then reset... I can add that.

rotty3000: fixed

biz.aQute.tester/src/aQute/tester/plugin/ProjectTesterImpl.java
rotty3000: fixed this

rotty3000:
Please note that due to the nature of how the bnd build utilizes the current snapshot build of bnd to build itself in CI the test API which is provided by the SNAPSHOT must always be compatible with the implementation which comes from the current build. Hence why I have to first commit the API and later commit the impl. Otherwise CI will fail.

PR #4474, opened 2020-12-23
[quickfix] Configurable version handling

### LINKED ISSUE #4410: Quickfix -buildpath/-testpath addition
Currently an addition via the Quickfix to the -buildpath/-testpath adds the version.

I do not like these versions and want to skip them when possible. In a curated repo there is only a single version. If you add a version, you need to update it everywhere when you update the repo.

I would therefore like to remove the version when the repositories contain only one candidate for a BSN.

Thoughts?

kriegfrj:
Includes the following changes to quickfix handling:
* BSNs are only listed once even if multiple versions are available
* There is a new Bndtools preference that specifies the version handling policy:
** No version attribute (lowest version)
** "version=latest" (latest version)
* `AddBundleCompletionProposal` will add the appropriate version attribute (or omit
  it entirely) according to the setting of that preference.

Fixes #4410.


pkriens:
Looks very good to me ...

PR #4475, opened 2020-12-26
[builder] Improve error message for missing bnd.bnd

### LINKED ISSUE #4422: Bndtools can't find model because project has no bnd.bnd file
I suddenly get errors on projects that have no bnd.bnd file.

Now I do think they should have a bnd.bnd file but this was not the case before, was it? At least I've never seen this error.

There error also does not go away when you create a bnd.bnd file.

kriegfrj:
Fixes #4422.

### DIFF
--- a/bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
+++ b/bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
@@ -148,13 +148,28 @@ public class BndtoolsBuilder extends IncrementalProjectBuilder {
 						if (!Files.isSameFile(javaPath.getParent(), wsPath)) {
 							msg = "The project directory: " + javaPath + " is not a subdirectory of the bnd workspace "
 								+ wsPath;
+						} else {
+							Path bndPath = javaPath.resolve(Project.BNDFILE);
+
+							if (!Files.exists(bndPath)) {
+								msg = "The project does not have a " + Project.BNDFILE + " file";
+							} else if (!Files.isReadable(bndPath)) {
+								msg = Project.BNDFILE + " is not readable";
+							} else if (Files.isDirectory(bndPath)) {
+								msg = Project.BNDFILE + " is a directory";
+							} else if (!Files.isRegularFile(bndPath)) {
+								// If it exists but it's not a directory, it
+								// still might not be a file - eg, could be a
+								// pipe or a device if the user has done
+								// something silly.
+								msg = Project.BNDFILE + " is not a regular file";
+							}
 						}
 					}
 				}
 				markers.createMarker(null, IMarker.SEVERITY_ERROR, msg, BndtoolsConstants.MARKER_BND_PATH_PROBLEM);
 				return noreport();
 			}
-
 			final Project model = ourModel;
 
 			try {


bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
```diff
@@ -148,6 +148,10 @@
 						if (!Files.isSameFile(javaPath.getParent(), wsPath)) {
 							msg = "The project directory: " + javaPath + " is not a subdirectory of the bnd workspace "
 								+ wsPath;
+						} else if (!projPath.append(Project.BNDFILE)
```
bjhargrave, line 151: Shouldn't this be:

```java
} else if (!Files.isRegularFile(javaPath.resolve(Project.BNDFILE))) {
```

kriegfrj: Yes, that would probably be more efficient and more compact, thanks for the tip.

I've done this as well as added a few more different checks for weird error-causing conditions that could occur so that I can make the error message even more specific.

bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
```diff
@@ -148,13 +148,24 @@
...
+							Path bndPath = javaPath.resolve(Project.BNDFILE);
+
+							if (!Files.exists(bndPath)) {
+								msg = "The project does not have a " + Project.BNDFILE + " file";
+							} else if (!Files.isReadable(bndPath)) {
+								msg = Project.BNDFILE + " is not readable";
+							} else if (Files.isDirectory(bndPath)) {
+								msg = Project.BNDFILE + " is a directory";
+							} else if (!Files.isRegularFile(bndPath)) {
+								msg = Project.BNDFILE + " is not a regular file";
```
bjhargrave, line 161: This is unreachable. If the file exists and is not a directory, then it must be a regular file. There are no other choices since you follow symlinks.

kriegfrj: It could be a pipe or a device or some other special file?

bjhargrave: How could the `bnd.bnd` path be a special file? I don't see how that could ever be the case. I think you want to test for the normal case first (isRegularFile) and it that fails, then try to make a nice error message.

kriegfrj: > How could the `bnd.bnd` path be a special file? I don't see how that could ever be the case. I think you want to test for the normal case first (isRegularFile) and it that fails, then try to make a nice error message.

It could be a special file if the user does something a bit silly or tries to be a bit smart. Admittedly, this is not an expected scenario. But I think it's a good rule of UI design to be able to handle the user doing something silly and/or unexpected and to give them the most useful feedback possible to help them fix the problem, especially if it is relatively simple and inexpensive to do so. In this case, the cost of the extra test is low (both in execution overhead and code size) and it's not in a performance-critical part of the code, so I'm struggling to see the downside in including it. But if you insist, I will remove the specific check for a directory.

There is already a "nice" error message (or at least, as nice as it can be) as the default:
https://github.com/bndtools/bnd/blob/0a3846f7f550ad0e2014c6f444bd760e5105993b/bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java#L133

If none of the other diagnostic checks finds any more specific cause, this is what will be shown. If you can think of a better generic message, or think of more diagnostic tests that could be done, then I'm all ears!

bjhargrave (approved):

PR #4479, opened 2021-01-02
[bndtools] Add progress reporting for resolver

kriegfrj:
With this fix, the resolver will attempt to report progress of the current resolve operation.

It shows the current requirement being processed as well as incrementing the progress bar as progress continues.

Due to the nature of the resolve problem, it is difficult to report the progress bar properly because you don't know how much is left to resolve. The total number of requirements pending to be resolved is an implementation detail hidden away in the resolver - the callback hook we have for reporting progress only gets the current requirement and the set of already-resolved wires (and capabilities for the current wires). So I used the recommendation [here](https://www.eclipse.org/articles/Article-Progress-Monitors/article.html) for indeterminate work remaining (section 3.6). I also changed it to follow the policy mentioned in that article, which is to let `split()` do the cancelation checks rather than checking explicitly (apparently this is more efficient).

I have already found this feature useful in debugging a long-running resolve (in fact, that's why I implemented it!) - I could tell when it got stuck and waited until then to hit cancel. It could be further improved but this is a massive improvement over what we have.

I would be interested in feedback about performance. I didn't notice a perceptible decrease in resolution performance as a result of updating the progress bar, but I don't have many long-running resolve operations to test it on.


bjhargrave (approved):
Nice!

Issue #4485, opened 2021-01-05
Create macro to give the current running BND Version


juergen-albert:
@bjhargrave I would give it a shot in the coming days. Would care to assign it to me?

bjhargrave:
It should go in the Macro class.

juergen-albert:
> And make sure to document the macro with a new doc page.

The macro index list page is created automatically or do I need to add it somewhere?

bjhargrave:
> The macro index list page is created automatically or do I need to add it somewhere?

Automatic.

PR #4486, opened 2021-01-05
[#4485] adds macro to return the currently running bnd version

juergen-albert:

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/MacroTest.java
+++ b/biz.aQute.bndlib.tests/test/test/MacroTest.java
@@ -7,6 +7,7 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNotSame;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
 import static org.junit.jupiter.api.condition.OS.WINDOWS;
 
 import java.io.File;
@@ -23,12 +24,14 @@ import org.junit.jupiter.api.condition.EnabledForJreRange;
 import org.junit.jupiter.api.condition.EnabledOnOs;
 import org.junit.jupiter.api.condition.JRE;
 
+import aQute.bnd.osgi.About;
 import aQute.bnd.osgi.Analyzer;
 import aQute.bnd.osgi.Builder;
 import aQute.bnd.osgi.Constants;
 import aQute.bnd.osgi.Jar;
 import aQute.bnd.osgi.Macro;
 import aQute.bnd.osgi.Processor;
+import aQute.bnd.version.Version;
 import aQute.lib.io.IO;
 import aQute.lib.strings.Strings;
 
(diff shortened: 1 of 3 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
```diff
@@ -1617,6 +1617,15 @@ public String _basenameext(String[] args) throws Exception {
 		return result;
 	}
 
+	static final String _bndversionHelp = "${bndversion}, returns the currently running bnd version";
+
+	public String _bndversion(String[] args) throws Exception {
```
bjhargrave, line 1622: You should probably add a `verifyCommand` call here to validate the input.

biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
```diff
@@ -1617,6 +1617,15 @@ public String _basenameext(String[] args) throws Exception {
...
+		try (Analyzer a = new Analyzer()) {
+			String v = a.getBndVersion();
```
bjhargrave, line 1624: You don't need to make an Analyzer. Just return `About.CURRENT.toString()`.

biz.aQute.bndlib.tests/test/test/MacroTest.java
```diff
@@ -1780,6 +1781,14 @@ public void testBase64() {
 		assertEquals(b64, b.getProperty("b64"));
 	}
 
+	@Test
+	public void testBndVersion() {
+		Processor b = new Processor();
+		b.setProperty("thebndversion", "${bndversion}");
+		String version = b.getProperty("thebndversion");
+		assertTrue(Version.isVersion(version));
```
bjhargrave, line 1789: This does not test that the macro returns the current bnd version. You need to compare with `About.CURRENT` which is the current bnd version.

docs/_macros/bndversion.md
```diff
@@ -0,0 +1,8 @@
+---
+layout: default
+class: Macro
+title: bndversion
+summary: Returns the current running bndversion
+---
+
+	Returns the currently running bnd version. E.g 5.3.0
```
bjhargrave, line 8: You should also mention the returned version includes the micro. So it could be 5.3.1 for example.

biz.aQute.bndlib.tests/test/test/MacroTest.java
```diff
@@ -1780,6 +1781,14 @@ public void testBase64() {
 		assertEquals(b64, b.getProperty("b64"));
 	}
 
+	@Test
+	public void testBndVersion() {
+		Processor b = new Processor();
```
bjhargrave, line 1786: Best to put the Processor in a try-with-resources.

bjhargrave (changes requested):

biz.aQute.bndlib/src/aQute/bnd/osgi/Macro.java
juergen-albert: I've just realized that myself. 

I wonder why the analizer is using this mechnism with the `bnd.info` file to determine the version. It looks to me as if the whole file and the mechanisms around it are obsolete.

bjhargrave: Probably :-)

bjhargrave (approved):

bjhargrave:
Thanks!

PR #4487, opened 2021-01-05
gradle: Add -dependson to test task inputs

### LINKED ISSUE #4484: [gradle] Gradle seems to not obey -dependson for parallel tests
I have a test project where I specify that it depends on all other projects:

     -dependson *

This works when I run gradlew without the `--parallel` but with parallel enabled the tests are run _before_ the other projects have been build, and thus the test fails.

If I add the projects literally to `-dependson`, it still fails. However, when I add them to `-buildpath` these dependencies are properly build ahead of the test execution. 

Using `bnd debug` in the test project directory, I can see that in all cases the dependencies are listed in the DependsOn section. It seems bnd does provide the proper information. It almost feels as if the Gradle parallel execution does not take `-dependson` into account?

I can consistently repeat this with this workspace that I cannot share. If needed I can make a test workspace. However, since this depend on the scheduling rules of the parallel tasks, I expect there is non-determinism is involved. Maybe we can first take a look at the source code if the scheduling takes `-dependson` into account.

bjhargrave:
The test task was already dependent on the -buildpath and -testpath
through the test task's dependencies on the compileJava and
compileTestJava tasks, respectively. We further ensure that the test
task is also dependent on the jars of any additional projects specified
through -dependson.

So like Eclipse, we now ensure all dependency jars (-buildpath,
-testpath, and -dependson) are built before the test task runs.

Fixes #4484

### DIFF
--- a/biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BndPlugin.groovy
+++ b/biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BndPlugin.groovy
@@ -371,6 +371,8 @@ public class BndPlugin implements Plugin<Project> {
 
       def test = tasks.named('test') { t ->
         t.enabled !bndis(Constants.NOJUNIT) && !bndis('no.junit')
+        /* tests can depend upon jars from -dependson */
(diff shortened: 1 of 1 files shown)


pkriens:
Thanks!

PR #4494, opened 2021-01-09
adds findexternalproviders macro

juergen-albert:
The current findproviders macro can be used in bndrun files. If used in
a bnd file, it causes a circular build because it wants to search the
projects as well. This version of the macro excludes the workspace.

### DIFF
--- a/biz.aQute.bndlib.tests/test/aQute/bnd/build/repo/WorkspaceRepositoryTest.java
+++ b/biz.aQute.bndlib.tests/test/aQute/bnd/build/repo/WorkspaceRepositoryTest.java
@@ -1,6 +1,8 @@
 package aQute.bnd.build.repo;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
 
 import java.io.File;
 import java.io.IOException;
@@ -10,6 +12,7 @@ import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TestName;
 
+import aQute.bnd.build.Project;
 import aQute.bnd.build.Run;
 import aQute.bnd.build.Workspace;
 import aQute.bnd.repository.maven.provider.MavenBndRepository;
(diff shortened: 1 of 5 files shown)


biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -1530,7 +1545,11 @@ Repository getResourceRepository() throws Exception {
 
 	public String _findproviders(String[] args) throws Exception {
 		Macro.verifyCommand(args, _findprovidersHelp, null, 2, 3);
-		return findProviders(args[1], args.length > 2 ? args[2] : null).map(Capability::getResource)
+		return findProvidersMacro(true, args);
+	}
+
+	public String findProvidersMacro(boolean includeWorkspace, String[] args) throws Exception {
```
bjhargrave, line 1551: This method should be private. We don't need to make this API.

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -1541,18 +1560,38 @@ public String _findproviders(String[] args) throws Exception {
 
 	static final String _findprovidersHelp = "${findproviders;<namespace>;<filter>}";
 
+	/**
+	 * A macro that returns a set of resources in bundle selection format from
+	 * the repositories, but excludes the workspace. For example:
+	 *
+	 * <pre>
+	 * ${findproviders;osgi.wiring.package;(osgi.wiring.package=aQute.bnd.build)}
```
bjhargrave, line 1568: This comment uses the wrong macro name.

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -1541,18 +1560,38 @@ public String _findproviders(String[] args) throws Exception {
...
+	 * </pre>
+	 */
+
+	public String _findexternalproviders(String[] args) throws Exception {
```
bjhargrave, line 1572: Not sure `external` is the right name here. Do we even need a new macro? Couldn't we just add another optional argument to the existing macro?

`${findproviders;<namespace>;<filter>[;<excludeWorkspace>]}`

Where `excludeWorkspace` defaults to false and behaves like the current impl. When set to `true` it does not search the workspace.

I think this is much better and avoids a new doc page too.

bjhargrave (changes requested):
I think we want to just modify the existing macro to add a new optional argument to exclude the workspace.

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
bjhargrave: > Where `excludeWorkspace` defaults to false and behaves like the current impl. When set to `true` it does not search the workspace.

Actually, instead of a boolean argument, we should do an "enum" like STRATEGY in the repo macro: <https://bnd.bndtools.org/macros/repo.html>.  Valid values for the enum could be: `REPOS` and `WORKSPACE` where `WORKSPACE` is the default.

juergen-albert: Good Idea. I ended up with 3 Strategies: ALL, WORKSPACE and REPOS. WORKSPACE will only look in the Workspace. Might be useful as well.

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -1506,11 +1507,37 @@ public void refresh(RepositoryPlugin repo) {
...
+	 * Return an aggregate repository of all the workspace's repository and the
+	 * projects repository. This workspace must be gotten for each operation, it
+	 * might become stale over time.
+	 *
+	 * @param includeWorkspace will ignore the workspace to prevent build loops
+	 * @return an aggregate repository
+	 * @throws Exception
+	 */
+	Repository getResourceRepository(STRATEGY strategy) throws Exception {
+		List<Repository> plugins = new LinkedList<>();
```
bjhargrave, line 1523: There is no reason to use a linked list. An ArrayList is fine.

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -1506,11 +1507,37 @@ public void refresh(RepositoryPlugin repo) {
...
+		switch (strategy) {
+			case WORKSPACE :
+				plugins.add(new WorkspaceRepositoryDynamic(this));
+				break;
+			case REPOS :
+				plugins.addAll(getPlugins(Repository.class));
+				// replace any WorkspaceRepositoryMarker plugin
+				plugins.removeIf(WorkspaceRepositoryMarker.class::isInstance);
+				break;
+			default :
```
bjhargrave, line 1534: This should be `ALL` so that if we add more enum values later, we get a compile error in the switch statement if we fail to add the case for the new enum value.

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -1521,16 +1548,29 @@ Repository getResourceRepository() throws Exception {
...
 	 */
 
 	public String _findproviders(String[] args) throws Exception {
-		Macro.verifyCommand(args, _findprovidersHelp, null, 2, 3);
-		return findProviders(args[1], args.length > 2 ? args[2] : null).map(Capability::getResource)
+		Macro.verifyCommand(args, _findprovidersHelp, null, 2, 4);
+		return findProvidersMacro(args);
+	}
+
+	private enum STRATEGY {
```
bjhargrave, line 1563: Perhaps we should call the enum class `ResourceRepositoryStrategy`? Since there are many things for which we want a strategy.

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -1521,16 +1548,29 @@ Repository getResourceRepository() throws Exception {
...
+		ALL,
+		REPOS,
+		WORKSPACE
+	}
+
+	public String findProvidersMacro(String[] args) throws Exception {
+		String namespace = args[1];
+		String filter = args.length > 2 ? args[2] : null;
+		STRATEGY strategy = args.length > 3 ? STRATEGY.valueOf(args[3]) : STRATEGY.ALL;
```
bjhargrave, line 1572: Using `valueOf` here is a bit problematic since it does not allow the macro writer to use lower case and on misspellings we get an exception. See https://github.com/bndtools/bnd/blob/df01bd872b34497d9cae0a7a7eb8fd4a3c417dab/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L1637-L1644

biz.aQute.bndlib.tests/test/aQute/bnd/build/repo/WorkspaceRepositoryTest.java
```diff
@@ -46,6 +47,77 @@ public void findprovidersMacroTest() throws Exception {
...
+	public void findprovidersMacroTestWithStrategy() throws Exception {
+
+		test(ws -> {
+			Project project = ws.getProject("p1");
+			assertThat(project).isNotNull();
+			File[] build = project.build();
+			assertThat(build).isNotNull()
+				.isNotEmpty();
+			assertThat(build[0]).hasName("p1.jar");
+			String services = ws._findproviders(new String[] {
```
bjhargrave, line 60: I don't think we want tests to directly call the macro method. That bypasses actually testing the use of the macro and the parsing of parameters, etc.

bjhargrave (changes requested):

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
juergen-albert: Done. I've given the messages Interface a shot as well. I hoe I did it correctly.

PR #4495, opened 2021-01-11
docs: `-dsannotations-options: inherit` in FAQ

stbischof:


docs/_chapters/920-faq.md
```diff
@@ -120,7 +120,7 @@ Inspired by the osgi-dev thread [Why default version for Export-Package is 0.0.0
...
 
 ## Why are superclass not inspected for Component annotations? <a name="supercomps"/>
-Annotations are not inherited form the component's super classes. The problem is that bnd cannot in all cases can see super classes (and interfaces) because the super class does not have to be available in the build. In runtime you are sure the super class is there, at build time you do not have that luxury. So to make sure, always define your annotations on the actual component class.
+Annotations are not inherited form the component's super classes by default. The problem is that bnd cannot in all cases can see super classes (and interfaces) because the super class does not have to be available in the build. In runtime you are sure the super class is there, at build time you do not have that luxury. So to make sure, always define your annotations on the actual component class. Alternatively you can use the instruction `-dsannotations-options: inherit`. Then bnd use DS annotations found in the class hierarchy of the component class.
```
rotty3000, line 123: > .. is that bnd cannot in all cases can see super classes ..

Can you remove the last `can`? ;) it was there before you modified but since you're here.

It should perhaps be noted that this is not true for the @Component annotation itself. This will not be inherited from super classes; causing a subclass to suddenly be a component.

stbischof: sure, done! Thx.

bjhargrave (changes requested):
While I think we should document the inherit option, I don't think this change is sufficiently scary. We need to be clear that inheriting is dangerous and outside of the OSGi DS specification. So some more warning is needed here to caution the user to think very carefully before using inherit.

djencks:
IIRC, it’s perfectly safe ___IF___ all the annotated super classes end up in the bundle with the “leaf” class (and are not importable). I suggest advising to use bnd packaging magic to assure this is the case when using this option.  Problems arise if at runtime you get a superclass that doesn’t correspond to the one used at build time; if the hierarchy is all in one bundle this situation isn’t going to happen.

David Jencks

stbischof:
Maybe the inherit option should only work (without warning) if the sources are in same project and build into the same bundle?

bjhargrave:
> Maybe the inherit option should only work (without warning) if the sources are in same project and build into the same bundle?

I don't know that there is a warning emitted by the code. I am just talking about this documentation here. You are updated it to mention the inherit option and I would like this update to include a proper warning of the risks of the inherit option and why it is not the default behavior and not in the OSGi spec.

docs/_chapters/920-faq.md
```diff
@@ -120,7 +120,7 @@ Inspired by the osgi-dev thread [Why default version for Export-Package is 0.0.0
...
 
 ## Why are superclass not inspected for Component annotations? <a name="supercomps"/>
-Annotations are not inherited form the component's super classes. The problem is that bnd cannot in all cases can see super classes (and interfaces) because the super class does not have to be available in the build. In runtime you are sure the super class is there, at build time you do not have that luxury. So to make sure, always define your annotations on the actual component class.
+Annotations are not inherited form the component's super classes by default. The problem is that bnd cannot in all cases see super classes (and interfaces) because the super class does not have to be available in the build. In runtime you are sure the super class is there, at build time you do not have that luxury. So to make sure, always define your annotations on the actual component class. Alternatively you can use the instruction `-dsannotations-options: inherit`. Then bnd use DS annotations found in the class hierarchy of the component class. This will not work for the `@Component` annotation itself. This will not be inherited from super classes; causing a subclass to suddenly be a component.
```
bjhargrave, line 123: Old grammar error we should fix: not inherited _from_

docs/_instructions/dsannotations-options.md
```diff
@@ -18,7 +18,7 @@ The following options are supported:
 
 |option||
 |-|-|
-|inherit|use DS annotations found in the class hierarchy of the component class|
+|inherit|use DS annotations found in the class hierarchy of the component class. *WARNING:* Annotations are not inherited form the component's super classes by default. The problem is that bnd cannot in all cases see super classes (and interfaces) because the super class does not have to be available in the build. In runtime you are sure the super class is there, at build time you do not have that luxury. So to make sure, always define your annotations on the actual component class. |
```
bjhargrave, line 21: Same grammar fix needed here: not inherited _from_

bjhargrave: and if the super classes are from imported packages, then the actual super classes at runtime can differ from the super classes seen at build time.

bjhargrave (changes requested):

docs/_chapters/920-faq.md
djencks: Also... `Then bnd _will_ use`

docs/_instructions/dsannotations-options.md
djencks: That's a much bigger problem than "not available at build time", which is in my experience not likely with a correctly set up build.

bjhargrave:
@stbischof  Are you able to address these comments soon?

bjhargrave (approved):

PR #4496, opened 2021-01-12
[repoviewer] Suppress bundles under GAV pseudo-BSNs

### LINKED ISSUE #3383: Bndtools gets confused with Maven coordinates
With the MvnBndRepository, if a coordinate points to a bundle the Bndtools uses the BSN as the name in the repository tree. If it does not, then it uses the pseudo-BSN (of the form is G:A) followed by an exclamation mark indicating a potential problem.

The issue I discovered by accident is when you have two versions of the same artifact, and one is a bundle and the other is not. The one that is a bundle gets listed twice - once under its BSN, and once under the pseudo-BSN:

Duplicate Maven Bundle

I have verified that both of the 1.5.1 versions point to the same file/jar.

Should probably suppress the 1.5.1 from appearing under the pseudo-BSN node.

kriegfrj:
Prior to this fix, if there were bundle and non-bundle versions of a jar
in (eg) a MavenBndRepository, there would be two root entries - one
with the full BSN, and the other with a pseudo-BSN of the form group:artifact.
The plain jars would appear under the pseudo-BSN while the true bundles
would appear under both. With this fix, the bundles will be filtered from the
G:A node so that they only appear once under their actual BSN.

Fixes #3383.

### DIFF
--- a/bndtools.core/src/bndtools/model/repo/RepositoryTreeContentProvider.java
+++ b/bndtools.core/src/bndtools/model/repo/RepositoryTreeContentProvider.java
@@ -13,6 +13,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
 import java.util.SortedSet;
+import java.util.stream.Stream;
 
 import org.bndtools.api.ILogger;
 import org.bndtools.api.Logger;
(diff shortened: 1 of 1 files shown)


bjhargrave (approved):

bjhargrave:
Thanks!

PR #4499, opened 2021-01-13
fixed plugin refresh on workspace load

juergen-albert:
I stumbled upon this issue as I had  `-include` with a http source in one of my cnf/ext files. The moment the include happened, we internally used `getPlugin(HttpClient.class)`. This initialized the plugins with the properties that has been read up to this point. After all ext files have been loaded, the plugin definitions that came after this would not have been recognized. 

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
@@ -417,7 +417,10 @@ public class Workspace extends Processor {
 				}
 			}
 		}
-
+		// getPlugins may have been called during the load of the properties.
+		// E.g. there is an include of a http url. So we have to clear the
+		// plugins to be sure
+		clearPlugins();
 		super.propertiesChanged();
 	}
 
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
@@ -966,13 +966,7 @@ public class Processor extends Domain implements Reporter, Registry, Constants, 
 		for (Closeable c : toBeClosed) {
 			IO.close(c);
 		}
-		synchronized (this) {
-			plugins = null;
-		}
-		if (pluginLoader != null) {
-			IO.close(pluginLoader);
-			pluginLoader = null;
-		}
+		clearPlugins();
 
 		toBeClosed.clear();
 	}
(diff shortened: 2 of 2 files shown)


bjhargrave (changes requested):
I am not sure I care for this. I think a safer solution would be to have the code create an HttpClient object if needed and then close when done as happens in:

https://github.com/bndtools/bnd/blob/e2d6795be54d7513322e7df6a213cbe874dc8f59/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java#L745-L747

I think this is a much better fix:

```diff
diff --git a/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java b/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
index 7b5cdf814..8771f29c0 100644
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
@@ -1166,16 +1166,20 @@ public class Processor extends Domain implements Reporter, Registry, Constants,
                                                        if (n >= 0)
                                                                ext = value.substring(n);
 
-                                                       Path tmp = Files.createTempFile("url", ext);
-                                                       try (Resource resource = Resource.fromURL(url, getPlugin(HttpClient.class))) {
-                                                               try (OutputStream out = IO.outputStream(tmp)) {
-                                                                       resource.write(out);
+                                                       try (HttpClient client = new HttpClient()) {
+                                                               client.setRegistry(this);
+                                                               client.readSettings(this);
+                                                               Path tmp = Files.createTempFile("url", ext);
+                                                               try (Resource resource = Resource.fromURL(url, client)) {
+                                                                       try (OutputStream out = IO.outputStream(tmp)) {
+                                                                               resource.write(out);
+                                                                       }
+                                                                       Files.setLastModifiedTime(tmp, FileTime.fromMillis(resource.lastModified()));
+                                                                       doIncludeFile(tmp.toFile(), overwrite, p);
+                                                               } finally {
+                                                                       removeIncluded(tmp.toFile());
+                                                                       IO.delete(tmp);
                                                                }
-                                                               Files.setLastModifiedTime(tmp, FileTime.fromMillis(resource.lastModified()));
-                                                               doIncludeFile(tmp.toFile(), overwrite, p);
-                                                       } finally {
-                                                               removeIncluded(tmp.toFile());
-                                                               IO.delete(tmp);
                                                        }
                                                } catch (MalformedURLException mue) {
                                                        if (fileMustExist)
```

This is probably a mistake on my part to have not done this in the first place.

juergen-albert:
Yes, this would be a solution as well. But can we sure, in the future, that nobody will call getPlugin during read? I might have done the same, without realizing what I have done.

bjhargrave:
> But can we sure, in the future, that nobody will call getPlugin during read?

Yes, this was a mistake on my part to rely upon the partially configured Processor to configure the Processor. That is read the include file. And my diff avoids that mistake.

juergen-albert:
okay. 
I'm a bit confused about what I should do now. Will you push your change you seemingly already made or should I do it instead of my  current fix?

bjhargrave:
I will push my change to this PR.

bjhargrave:
Replaced by #4501.

PR #4501, opened 2021-01-13
Clear plugins on properties changed

bjhargrave:
@juergen-albert:

> I stumbled upon this issue as I had -include with a http source in one of my cnf/ext files. The moment the include happened, we internally used getPlugin(HttpClient.class). This initialized the plugins with the properties that has been read up to this point. After all ext files have been loaded, the plugin definitions that came after this would not have been recognized.

Closes #4499 

We also fix the workspace to load the `cnf/ext/*.bnd` files in sorted order rather than however they are presented by the file system.


juergen-albert:
Thanks a lot! Works like charm.

PR #4503, opened 2021-01-14
[quickfix] Test to wait for build to complete before running proposals

kriegfrj:
@bjhargrave  reported that simpleReference_toWorkspaceClass_suggestsBundles() was failing intermittently. I was able to reproduce in my workspace. Doing a clean before building seemed to make it pass.

When I tried it in my own workspace I had the same problem both in Gradle and through Bndtools. It would fail very consistently. When I tried to hook up the debugger and set a breakpoint on the test in question, it would pass consistently even with no other code changes.

I suspected a race condition: `local-proj` was not finished building/being indexed before the code was run. Added a wait in `proposalsFor()` (just before building the AST that generates the `IProblem` markers) to make sure that building was finished (so that the indexing would be finished) before running the test. This seems to have stabilised it in my testing.

### DIFF
--- a/bndtools.core.test/src/bndtools/core/test/editors/quickfix/AbstractBuildpathQuickFixProcessorTest.java
+++ b/bndtools.core.test/src/bndtools/core/test/editors/quickfix/AbstractBuildpathQuickFixProcessorTest.java
@@ -43,6 +43,7 @@ import org.eclipse.core.resources.IResourceVisitor;
 import org.eclipse.core.resources.IWorkspace;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.jobs.Job;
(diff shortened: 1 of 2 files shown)


bjhargrave (approved):
Worked on my system. So lets merge and see how things go in the longer term,

kriegfrj:
When @bjhargrave tried to re-introduce the quickfix tests to the CI build (#4538), this failure re-appeared. I suspect that the re-appearance was coincidental and that I never really fixed this issue 100%.

Thinking about the code, although I put in the change to wait until the build is finished before proceeding, there is still a race condition (albeit a more infrequent one) as the wait assumes that the build is already in progress - it won't wait if the build hasn't started yet. I'll try and put some more synchronization in there and see if I can cover that case too.

kriegfrj:
> When @bjhargrave tried to re-introduce the quickfix tests to the CI build (#4538), this failure re-appeared. 

This is now being tracked in #4553.

PR #4523, opened 2021-02-02
analyzer: Allow resolution directive on java imports

bjhargrave:
Per request from @tjwatson, we now allow the resolution directive on
java imports to support optional resolution of a java import. This is
interesting to the [Apache Felix Atomos project][1].

[1]: https://github.com/apache/felix-atomos

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -1969,10 +1969,14 @@ public class Analyzer extends Processor {
 				Attrs importAttributes = imports.get(packageRef);
 				//
 				// Check if we have a java.* package. If so, we should remove
-				// all attributes to have a plain import.
+				// all attributes (except resolution:) to have a plain import.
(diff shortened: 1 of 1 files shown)


pkriens:
Hmm, I would have not hard coded this but added a list of directives allowed on Java? I think we already do this in other places?

bjhargrave:
> added a list of directives allowed on Java? I think we already do this in other places?

Oh. Can you provide a pointer to an example I could potentially emulate?

pkriens:
I saw that the printing skips them: aQute.bnd.osgi.Processor.skipPrint(String).

Processor is not the right level I guess, on the other hand it is sadly in the osgi package. So we could handle it there?

bjhargrave:
> I saw that the printing skips them: aQute.bnd.osgi.Processor.skipPrint(String).

This is just identifying Bnd-specific attributes/directives to remove. This is different than determining which attributes/directives should be retained/removed on java.* imports. If that becomes more complicated in the future, we may want a more complicated code. But what is there works for now and we can improve it in the future when the need arises.

PR #4528, opened 2021-02-04
[bndtools test] JUnit launching fixes

### LINKED ISSUE #4524: [bndtools] Tidy-up context-sensitive JUnit launching
In trying out #4521, I stumbled across a number of other other issues with the context-sensitive launching, which has never really had a good going-over since the JUnit 5 support was added. None of these issues is super difficult to fix so it would be nice to get the fixes in for 5.3.

1. The context-sensitive launch filter in plugin.xml uses the `org.eclipse.jdt.junit.canLaunchAsJUnit` filter. For JUnit 5, this filter looks for `org.junit.platform.commons.annotation.Testable`, so if `junit-platform-commons` is not on the classpath, the presence of (say) `org.junit.jupiter.api.Test` on your method is insufficient to trigger the option on the context menu. Ironically, not having commons on the classpath is not uncommon in a Bndtools project (in standard Eclipse Java, JUnit libraries get added to the classpath), so it would be nice to fix this. I think we can dispense with the canLaunchAsJUnit filter altogether as it duplicates the functinoality of the following `hasTypeOnClasspath` tests.
2. `JUnitShortcut.isJUnit4()` is JUnit 4-specific - it won't work for JUnit 3 and won't always work for JUnit 5. I'm not super-concerned about JUnit 3 (though it would be a "nice to fix" thing), but with JUnit 5 methods need not always be public (and increasingly people are using package-private as it is less verbose), so it would be nice to fix that.
3. It generates the wrong tester.names parameter for test methods that take parameters, so you can't invoke these methods from the context menu.

kriegfrj:
Fixes #4524:

* Context-sensitive launch for JUnit was not available unless junit-platform-commons was on the path or JUnit 4 was on the path, even if you are using Jupiter annotations.
* Context-sensitive launch for test methods with parameters.
* Context-sensitive launch works better with various different types of Jupiter test class.

### DIFF
--- a/bndtools.core/src/bndtools/launch/JUnitShortcut.java
+++ b/bndtools.core/src/bndtools/launch/JUnitShortcut.java
@@ -4,6 +4,8 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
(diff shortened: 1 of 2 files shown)


bndtools.core/src/bndtools/launch/JUnitShortcut.java
```diff
@@ -223,18 +287,27 @@ private static boolean isTestable(IType type) throws JavaModelException {
...
-						if (names != null) {
-							for (String[] pair : names) {
-								if (pair[0].contains("org.junit"))
+			for (IAnnotation annotation : m.getAnnotations()) {
+				IType declaringType = m.getDeclaringType();
+				if (declaringType != null) {
+					String[][] names = declaringType.resolveType(annotation.getElementName());
+					if (names != null) {
+						for (String[] pair : names) {
+							if (pair[0].startsWith("org.junit")) {
```
bjhargrave, line 302: I think this should be `"org.junit."` (trailing `.`) like as is done for `"junit."` above.

bndtools.core/src/bndtools/launch/JUnitShortcut.java
```diff
@@ -223,18 +287,27 @@ private static boolean isTestable(IType type) throws JavaModelException {
...
+								final int flags = m.getFlags();
+								// Jupiter allows public, protected &
+								// package-protected methods & classes,
+								// JUnit 4 only public
+								if ((pair[0].contains("jupiter") && !Flags.isPrivate(flags))
```
bjhargrave, line 307: Should this not be `contains(".jupiter.")` or `startsWith("org.junit.jupiter.")`?

bndtools.core/src/bndtools/launch/JUnitShortcut.java
kriegfrj: Done.

bndtools.core/src/bndtools/launch/JUnitShortcut.java
kriegfrj: Done.

bjhargrave (approved):

Issue #4414, opened 2020-11-14
[bndlib] AlsoLauncherTest is flaky


kriegfrj:
yes, but it is in another test you [@kriegfrj] authored which is now regularly failing CI builds like this one.

https://github.com/bndtools/bnd/pull/4404/checks?check_run_id=1389155797#step:5:449

Can you look into this? I would like to stabilize the CI builds.

_Originally posted by @bjhargrave in https://github.com/bndtools/bnd/issues/4404#issuecomment-727048734_

This test parses STDERR looking for the right values. As such, it is sensitive to other bits of the code spewing garbage into STDERR at just the wrong time. Usually this indicates a bug elsewhere in the code (as was the case in #3558), but unfortunately the test as it is currently constructed doesn't give the full stderr output - it tries to truncate it to make it more readable, but in the process it seems (in this case) to be truncating the useful information.

First step will be to try and improve the diagnostics in the existing test. Hopefully then we can capture enough information to find out what the root cause is.

kriegfrj:
@bjhargrave , has this error happened again recently? Now that #4415 has been merged it should produce more useful diagnostics when it happens so I can debug the root cause.

It would be interesting if this stopped when you disabled the testOSGi task. Perhaps stray stderr output from that test failure was interfering with the other test somehow?

bjhargrave:
> has this error happened again recently?

I have not seen it since the change to capture more diagnostics. I'll keep my eye open.



> It would be interesting if this stopped when you disabled the testOSGi task. Perhaps stray stderr output from that test failure was interfering with the other test somehow?

I would be shocked if this were the case. They are two distinct gradle tasks so I don't seen any real way the outputs of the two gradle tasks could interleave.

kriegfrj:
> > has this error happened again recently?
> 
> I have not seen it since the change to capture more diagnostics. I'll keep my eye open.
> 

Anything?

> > It would be interesting if this stopped when you disabled the testOSGi task. Perhaps stray stderr output from that test failure was interfering with the other test somehow?
> 
> I would be shocked if this were the case. They are two distinct gradle tasks so I don't seen any real way the outputs of the two gradle tasks could interleave.

I concur... hence why I said it would be "interesting" (to be fair, perhaps "interesting" would be a bit of an understatement...).

bjhargrave:
Lets close for now. We can reopen if this issue rears its head again.

PR #4533, opened 2021-02-05
[AlsoLauncherTest] fix failure

rotty3000:
The check appears to occasionally contain an additional number of digits

    [System Bundle]
    Expecting:
      "Id    Levl State Modified      Location
    0     # Found bundle without a bsn [7], ignoring
    0   ACTIV <>            System Bundle
    1     10  ACTIV 201810132223  file:/home/runner/work/bnd/bnd/maven/target/m2/org/apache/felix/org.apache.felix.scr/2.1.12/org.apache.felix.scr-2.1.12.jar
    2     11  ACTIV 201810132026  file:/home/runner/work/bnd/bnd/maven/target/m2/org/apache/felix/org.apache.felix.configadmin/1.9.10/org.apache.felix.configadmin-1.9.10.jar
    3     12  ACTIV 202009131314  file:/home/runner/work/bnd/bnd/maven/target/m2/org/junit/platform/junit-platform-commons/1.7.0/junit-platform-commons-1.7.0.jar
    4     12  ACTIV 202009131314  file:/home/runner/work/bnd/bnd/maven/target/m2/org/junit/platform/junit-platform-engine/1.7.0/junit-platform-engine-1.7.0.jar
    5     12  ACTIV 202011110524  file:/home/runner/work/bnd/bnd/maven/target/m2/org/assertj/assertj-core/3.18.1/assertj-core-3.18.1.jar
    6     12  ACTIV 201906061924  file:/home/runner/work/bnd/bnd/maven/target/m2/org/opentest4j/opentest4j/1.2.0/opentest4j-1.2.0.jar
    7     12  ACTIV 201906061907  file:/home/runner/work/bnd/bnd/maven/target/m2/org/apiguardian/apiguardian-api/1.1.0/apiguardian-api-1.1.0.jar
    8     12  ACTIV 202009131314  file:/home/runner/work/bnd/bnd/maven/target/m2/org/junit/jupiter/junit-jupiter-api/5.7.0/junit-jupiter-api-5.7.0.jar
    9     12  ACTIV 202009131315  file:/home/runner/work/bnd/bnd/maven/target/m2/org/junit/jupiter/junit-jupiter-engine/5.7.0/junit-jupiter-engine-5.7.0.jar
    10    12  ACTIV 202009131315  file:/home/runner/work/bnd/bnd/maven/target/m2/org/junit/jupiter/junit-jupiter-params/5.7.0/junit-jupiter-params-5.7.0.jar
    11    12  ACTIV 202011041016  file:/home/runner/work/bnd/bnd/maven/target/m2/org/apache/servicemix/bundles/org.apache.servicemix.bundles.junit/4.13.1_1/org.apache.servicemix.bundles.junit-4.13.1_1.jar
    12    12  ACTIV 202009131315  file:/home/runner/work/bnd/bnd/maven/target/m2/org/junit/vintage/junit-vintage-engine/5.7.0/junit-vintage-engine-5.7.0.jar
    13    12  ACTIV 202101042206  file:/home/runner/work/bnd/bnd/biz.aQute.bndall.tests/generated/tmp/test/biz.aQute.launcher.AlsoLauncherTest/testReporting_notUsingReferences/test%20ws/demo/generated/demo.jar
    # framework=org.apache.felix.framework [0]
    # setting launcher service registration to launcher.ready=true
    # will call main
    Running in main
    before joining
    Stopping framework
    After stopping framework, sleeping
    Goodbye world
    # framework event org.osgi.framework.FrameworkEvent[source=org.apache.felix.framework [0]] 64
    # framework event stopped
    "
    to contain pattern:
      "\n0\s+\d+\s*ACTIV\s+[<][>]\s+System Bundle"

### DIFF
--- a/biz.aQute.bndall.tests/test/biz/aQute/launcher/AlsoLauncherTest.java
+++ b/biz.aQute.bndall.tests/test/biz/aQute/launcher/AlsoLauncherTest.java
@@ -772,7 +772,7 @@ public class AlsoLauncherTest {
 
 			softly.assertThat(report)
 				.as("System Bundle")
-				.containsPattern("\\n0\\s+\\d+\\s*ACTIV\\s+[<][>]\\s+System Bundle");
+				.containsPattern("\\n0\\s+\\d*\\s*ACTIV\\s+[<][>]\\s+System Bundle");
 
 			final DateTimeFormatter f = DateTimeFormatter.ofPattern("YYYYMMddHHmm")
 				.withZone(ZoneOffset.UTC);


bjhargrave (approved):

kriegfrj:
I think this fix is incorrect and I don't think it actually fixed the problem that you saw.

The reason that the RE failed is because of the stray comment in the line that reports the system bundle. This sort of thing happens usually due to a bug elsewhere in the code, and is intermitted due to startup/logging race conditions.

The RE is trying to match a line like this:

```
0     0   ACTIV <>            System Bundle
```

The log output that appeared before the runlevel column "# Found bundle without a bsn [7], ignoring" is most likely what caused the RE to fail. It is therefore unlikely that the change in this PR actually fixed the issue, as the stray comment would also cause the revised RE to fail. Most likely, it succeeded when this patch was applied simply because the stray log appeared in a different spot this time. It is likely to fail again at some random time in the future unless the root cause is fixed.

It looks like the root cause is complaining about API Guardian not having a BSN. That's because it's including the Maven Central version (which is not a bundle) in `-runbundles`.

I've had a quick look through git history and it looks like this particular land mine has been sitting there for a while without having detonated:
* I updated the `junit` property to reference the Maven Central version of API guardian back in 2c68b7828f on 2019-09-26. This laid the mine.
* @bjhargrave added the log message to warn on bundles without bsns in 820ddffcdb on 2020-10-20. This uncovered this mine.

I think the root fix would be to change `demo/bnd.bnd`'s `-runbundles` definition so that it doesn't rely on the junit property (which is really intended for compile dependencies) and to hand-craft a suitable `-runbundles` that doesn't have this issue.

bjhargrave:
Isn't it just easier to fix the regex to not be sensitive to warnings in the output? We should in general not be too sensitive to the output since it will change and tests should not then fail unless the purpose of the test is to specifically test that output. I fixed a similar issue on osgi-test where a change in the output from assertj broke a test because it incorrectly relied on the specific output text.

kriegfrj:
@kriegfrj wrote:
> I've had a quick look through git history and it looks like this particular land mine has been sitting there for a while without having detonated:
> 
> * I updated the `junit` property to reference the Maven Central version of API guardian back in [2c68b78](https://github.com/bndtools/bnd/commit/2c68b7828ff78c7686024a70f2ee79faf981db67) on 2019-09-26. This laid the mine.
> * @bjhargrave added the log message to warn on bundles without bsns in [820ddff](https://github.com/bndtools/bnd/commit/820ddffcdb11bca0418cb2f98a334f753521567b) on 2020-10-20. This uncovered this mine.

The timing of [820ddff] explains the timing for when @bjhargrave started noticing flakiness in AlsoLauncherTest and reported it in #4414. I'm guessing this is the root cause of #4414 that the improved diagnostics have now helped us capture.

@bjhargrave wrote:
> Isn't it just easier to fix the regex to not be sensitive to warnings in the output? We should in general not be too sensitive to the output since it will change and tests should not then fail unless the purpose of the test is to specifically test that output. 

I agree - and the purpose of this test *is* specifically to test the output. It is a regression test for #3442 and #3451.

I tried to make the RE as non-specific as I could while still testing what needed to be tested. I don't see how I could write it efficiently to handle the case where some random warning string could appear at any random point in the output that I am trying to test.

At any rate, the warning in the output indicates a genuine problem with our test setup which still ought to be fixed regardless.

PR #4534, opened 2021-02-05
[resolver] Add tester dependencies to << INITIAL >>

### LINKED ISSUE #3795: [tester] Resolver to optionally resolve tester dependencies
When you run a test using `-tester`, it dynamically adds the specified tester bundle to the end of `-runbundles`. However, it doesn't add the tester's dependencies.

For `biz.aQute.tester`, it only has one dependency and normally that dependency will also be required by the test bundle itself, so the resolver will already have included it.

However, with `biz.aQute.tester.junit-platform`, it depends on the junit-platform bundles which are ordinarily not dependencies of individual test bundles. So you need to manually add all of these bundles to `-runrequires`.

It would be nice if there was a "tester-resolve" mode/command, that added `-tester` to `-runrequires` prior to running the resolution, so that the resolver will also find and add the tester's dependencies.

kriegfrj:
Looks for the -tester directive and if present makes sure that the chosen tester's requirements are added to the list of initial
requirements.

Fixes #3795.

This is a draft - no test cases at the moment, plus the GUI feedback could be better (user won't necessarily know why the tester's requirements have been added to the initial requirements). But thought to elicit some feedback on the concept.

### DIFF
--- a/biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java
+++ b/biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java
@@ -498,6 +498,26 @@ public class BndrunResolveContext extends AbstractResolveContext {
 			resBuilder.addRequirements(requires);
 		}
 
+		String tester = properties.mergeProperties(Constants.TESTER);
+		if (tester != null) {
+			tester = tester.trim();
+			if (!tester.isEmpty()) {
+				List<Container> containers = project.getBundles(Strategy.HIGHEST, tester, null);
+
+				if (containers == null || containers.isEmpty()) {
+					CapReqBuilder testerReq = new CapReqBuilder(IdentityNamespace.IDENTITY_NAMESPACE)
+						.addDirective("filter", "(osgi.identity=" + tester.trim() + ")");
+					resBuilder.addRequirement(testerReq);
+				} else {
+					Manifest m = containers.get(0)
+						.getManifest();
+					ResourceBuilder testerRes = new ResourceBuilder();
+					testerRes.addManifest(Domain.domain(m));
+					resBuilder.addRequirements(testerRes.getRequirements());
+				}
+			}
+		}
+
 		return resBuilder.build();
 	}
 


bjhargrave (commented):
I don't think we want to change the resolver for this. Why not just put the tester in the -runrequires in the bndrun file?

```properties
-runrequires.tester: ${tester}
```

rotty3000:
The two instructions have incompatible syntax. But I think the idea is a good one.

However, since in order to make it work we'd have to mangle the value of tester into a requirement, why not go so far as to automatically include it into the merged `-runrequires` property?

bjhargrave:
`-runrequires.tester: ${if;${tester};bnd.identity;id='${tester}'}`

should work

rotty3000:
just like I said tester is always set, it has a default value, so your if is pointless

The issue is not detecting tester, rather it's detecting if this is a "test" run.

bjhargrave:
> just like I said tester is always set, it has a default value, so your if is pointless

`tester` is a property name I made up. Perhaps you are thinking of the `-tester` instruction which is not always set but does have a default:

https://github.com/bndtools/bnd/blob/edff3f65ed0b1d1c89d8649cae96ea7c1a35855f/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L2962-L2967

> The issue is not detecting tester, rather it's detecting if this is a "test" run.

Do you mean a resolution for testing? Because running tests has nothing to do with resolution. For a project using OSGi Junit Testing, what is the point of resolving without the tester?

rotty3000:
yeah, I did miss the part about `tester` as a property name. However a single bnd file can be used both in a testing scenario and a non-testing scenario (although I wouldn't personally do it nor recommend it) but it is entire possible. I like where you're going I just think if we think about it a bit more we could probably come up with an automated way of handling it. I think if we detected that there are tests in `-runrequires*` to start, then we'd know we need to resolve the tester as well.

kriegfrj:
> I don't think we want to change the resolver for this. Why not just put the tester in the -runrequires in the bndrun file?
> 
> ```ini
> -runrequires.tester: ${tester}
> ```

Because then the tester has to be explicitly available in one of your workspace repos.

Normally, the launcher looks for a tester in your workspace repos,  bit if it can't find one it will look in the embedded repo. So most of the time, for testing, you don't need to explicitly have the tester bundle in your workspace. But if you put it in -runrequires, it must be explicitly in your workspace. 

With my fix, it will also look in the embedded repo for the tester. So you don't need to add it to your workspace.

kriegfrj:
Perhaps an alternative solution would be to allow the resolver to use the embedded repo during resolution.

bjhargrave:
> Perhaps an alternative solution would be to allow the resolver to use the embedded repo during resolution.

We need to make sure the `Workspace.CachedFileRepo` is also available as a `Repository` to resolutions. I can look into this after 5.3.

kriegfrj:
> just like I said tester is always set, it has a default value, so your if is pointless
> 
> The issue is not detecting tester, rather it's detecting if this is a "test" run.

Unfortunately, there's no way at resolution time to know whether it's being asked to resolve a test run or if it's being asked to resolve a normal run. The nearest flag I could come up with is if someone has explicitly set `-tester` in the bndrun file, then assume that that bndrun is intended to be run with tests.

The only other alternative I could come up with is a flag that the user can set to tell the resolver that it is intended to be used for testing (eg, a "bnd testresolve" on the command line, or a checkbox in the GUI, or perhaps a `-runrequirestester` directive). But these are not too different from BJ's suggestion.

kriegfrj:
> > Perhaps an alternative solution would be to allow the resolver to use the embedded repo during resolution.
> 
> We need to make sure the `Workspace.CachedFileRepo` is also available as a `Repository` to resolutions. I can look into this after 5.3.

Actually, it occurred to me that this approach would still have a problem: explicitly adding `biz.aQute.tester.junit-platform` will also result in it being explicitly added to the computed `-runbundles`. Then we're back where we started - the tester bundle needs to be available in one of the explicit repositories. Either you'd need to make `Workspace.CachedFileRepo` available to the launcher as well, or else remove any bundle from `-runbundles` that was found in the `CachedFileRepo`.

pkriens:
Why not make a require annotation that you put on the test classes?

kriegfrj:
> > > Perhaps an alternative solution would be to allow the resolver to use the embedded repo during resolution.
> > 
> > 
> > We need to make sure the `Workspace.CachedFileRepo` is also available as a `Repository` to resolutions. I can look into this after 5.3.
> 
> Actually, it occurred to me that this approach would still have a problem: explicitly adding `biz.aQute.tester.junit-platform` will also result in it being explicitly added to the computed `-runbundles`. Then we're back where we started - the tester bundle needs to be available in one of the explicit repositories. Either you'd need to make `Workspace.CachedFileRepo` available to the launcher as well, or else remove any bundle from `-runbundles` that was found in the `CachedFileRepo`.

With JUnit 5.8 around the corner, which includes the new Provide-Capability headers for the test engines (https://github.com/junit-team/junit5/pull/2544), and with the next Bnd/tools release being a major release, my mind returned to this issue because if we're going to do it and we have concerns about it from a backward compatibility perspective, now's the best time.

With further thought, it occurred to me that making the CachedFileRepo generally available as a `Repository` to resolutions will still end up with the bundle being added to the computed `-runbundles`, which will mean that the CachedFileRepo will also need to be available OSGi runtime's resolver. and I'm not sure that this is what we want? At the moment, bundles (only the launcher and tester bundles) from the embedded repo are specially loaded & instantiated by the launcher, and they directly add themselves to the runbundles just before the framework launch. Given that these bundles are handled as special-cases by the launcher, I think it is appropriate that they are handled as special cases by the resolver (along the vein of the original PR).

Thoughts?

kriegfrj:
> Why not make a require annotation that you put on the test classes?

The resolver still needs to be able to find the required bundle. The launcher already knows that the bundle is required because of the `-tester` directive and it knows how to find it; the problem is that the resolve task doesn't know that it is required nor where to find it.

kriegfrj:
I don't remember closing this issue - I certainly didn't do so on purpose.

I'm leaning towards an implementation along the vein of the original PR as I don't think the proposed alternatives will really work:
* Changing runrequires to add the tester places the burden on the developer to make sure that the tester bundle is present in one of the workspace's repos (which is precisely what this PR is aimed at avoiding).
* Making the embedded repo available as-is to the resolver is not prudent, because at runtime the embedded repo will not be available and the assembled set of runbundles may fail to resolve at runtime. The embedded tester bundles only become available at runtime due to some special-case handling in the launcher, and so it makes sense that they should only find themselves available to the resolver through special-case handling too.
* Adding an annotation on the class files could allow you to declare your dependency on the tester bundle, but it doesn't solve the issue of how the resolver finds the tester and *its* dependencies. While it would work if the test bundle declared the transitive dependencies of the tester in its own manifest, that seems wrong.

Do we concur?

kriegfrj:
Another issue that I forgot: if `-tester` declaration is included as a global setting, then all resolves will always include the tester's dependencies, which is probably not what we want.

A couple of tweaks to this PR could fix that:
1. First resolve as per normal (without adding the tester bundle's dependencies to the set of initial requirements).
2. Examine the set of resolved bundles - if (and only if) any of them contain a non-empty `Test-Cases` header, *then* find the declared `-tester`, add its requirements, and run another round of resolution to bring the tester's dependencies in.

Another possibility: the test launcher handles this extra resolution step at test launch time. Disadvantage of this is that you'd never get to see the full final set of `-runbundles`.

PR #4538, opened 2021-02-09
bndtools.core.test: Re-enable in CI build

bjhargrave:
Let's see if the tests are stable again.

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -32,6 +32,7 @@ defaults:
 jobs:
   Build_Matrix:
     strategy:
+      fail-fast: false
       matrix:
         os: [ ubuntu-latest ]
         java: [ 11, 15 ]
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -24,7 +24,6 @@ env:
     -Dhttp.keepAlive=false
     -Dmaven.wagon.http.pool=false
     -Dmaven.wagon.http.retryHandler.count=3
-  BNDTOOLS_CORE_TEST_NOJUNITOSGI: true
 
 defaults:
   run:


bjhargrave:
@kriegfrj I rebased on the latest updates to try again...

bjhargrave:
> @kriegfrj I rebased on the latest updates to try again...

Interestingly, it seems to work on Linux but failed on Windows.

kriegfrj:
> > @kriegfrj I rebased on the latest updates to try again...
> 
> Interestingly, it seems to work on Linux but failed on Windows.

There was a failure on both Linux and Windows - timed out waiting for import.

It is possible that we simply need to increase the timeout. Now that I have implemented locking on the import job, if the `FlushJob` happens to start just before the import job, the FlushJob will end up synchronously initializing the Workspace and the import job will have to wait until it is finished. That could take some time. Maybe I'll try bumping it up to 30s and see how it goes.

kriegfrj:
> > > @kriegfrj I rebased on the latest updates to try again...
> > 
> > 
> > Interestingly, it seems to work on Linux but failed on Windows.
> 
> There was a failure on both Linux and Windows - timed out waiting for import.
> 
> It is possible that we simply need to increase the timeout. Now that I have implemented locking on the import job, if the `FlushJob` happens to start just before the import job, the FlushJob will end up synchronously initializing the Workspace and the import job will have to wait until it is finished. That could take some time. Maybe I'll try bumping it up to 30s and see how it goes.

Ok, it was more complicated that simply increasing the timeout. The timeout was happening because the call to `Central.getWorkspace()` in the import job was mutually deadlocking with some other job. Moral: it doesn't seem to be safe to call `getWorkspace()` from within a job that holds an Eclipse lock.

I have moved the call to getWorkspace() out of the Job, and I used your tip here to wait for the initialized workspace instead: https://github.com/bndtools/bnd/issues/4573#issuecomment-790944584. This seems to have stabilised this issue (ref #4579).

bjhargrave:
Closing. See https://github.com/bndtools/bnd/pull/4579.

PR #4546, opened 2021-02-10
-maven-release now supports merge properties

juergen-albert:
I don't see, where I can set the maint-candidate lable. Maybe I don't have the proper rights for it. 

Regardless. According to the the instruction doc of -maven-release it uses merged properties. Well, now it actually does.


bjhargrave:
I am not entirely sure that making `-maven-release` a merged property is wise. It is not clear how we would expect the property combination to work. For example, `javadoc` is specified multiple times. Does the "last" one win? Do we combine all the attributes? Some thinking is needed here.

juergen-albert:
True that. It should be handled. We can throw this out.

PR #4548, opened 2021-02-11
Snapshot viewer improvements

chrisrueger:
- added a new tab "About"
- this tab contains a new section with description and explanation about the tool
- mention and link the Snapshot Viewer in the bnd manual / [resolving chapter](https://bnd.bndtools.org/chapters/250-resolving.html) ... Not sure this is the ideal place, but since it was about debugging, I think it fits.

Preview:
image

### DIFF
--- a/docs/_chapters/250-resolving.md
+++ b/docs/_chapters/250-resolving.md
@@ -286,6 +286,11 @@ However, often you find that even on its own it does not resolve. In most case t
 
 It is still a mystery, try checking the `Run Blacklist` list. If it is not there, it might be time to raise a bug.
 
+## Snapshot Viewer
+
+Another tool for diagnosing potential issues in your OSGI framework (bundle, packages, services), the SCR info, configuration, log, and custom extensions is the [bnd Snapshot Viewer](https://bnd.bndtools.org/snapshot.html).
+
+
 ## The Source View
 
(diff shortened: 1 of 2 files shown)


pkriens:
You need to sign the commit ... https://github.com/bndtools/bnd/pull/4548/checks?check_run_id=1882616037

pkriens:
I might be missing something but the information about the environment is static, isn't it? It is not updated by the json file?

chrisrueger:
@pkriens ok I will take care of the signing later. 

> I might be missing something but the information about the environment is static, isn't it? It is not updated by the json file?

Not sure I follow. What are you referring to? Any sentence which is misleading?

pkriens:
On the About tab you show system info like the amount of free memory, file name, etc. However, as far as I can see this information is static, it does not come from the JSON file?

chrisrueger:
> On the About tab you show system info like the amount of free memory, file name, etc. However, as far as I can see this information is static, it does not come from the JSON file?

Ah ok I see. This is just the output of the console command `bnd:snapshot`. Yes this information is from my system. I just wanted to show a full example of the output, so that people know what this command outputs.

But I can shorten it if you think this is too much noise.

pkriens:
Ok. Leave it as is

PR #4554, opened 2021-02-15
Added some more text & links to doc & extended help for command

pkriens:
---


maven/.classpath
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bjhargrave, line 1: This file is wrong (this project is not a Bndtools nature project) and should not be in git. The .classpath file is made by the m2e Update Maven Project command.

bjhargrave (changes requested):

Issue #4549, opened 2021-02-12
[launcher] When launching a test in Bndtools properties is empty


kriegfrj:
I updated to the latest Bndtools snapshot, tried to run `bndtools.core.test/test.win32_x86_64.bndrun` using the Bnd OSGi Test Launcher, and I got the following error on the console:

```
! Unexpected error in the run body: null
java.lang.NullPointerException
	at java.util.StringTokenizer.<init>(StringTokenizer.java:199)
	at java.util.StringTokenizer.<init>(StringTokenizer.java:221)
	at aQute.launcher.constants.LauncherConstants.split(LauncherConstants.java:159)
	at aQute.launcher.constants.LauncherConstants.<init>(LauncherConstants.java:137)
	at aQute.launcher.Launcher.launch(Launcher.java:392)
	at aQute.launcher.Launcher.run(Launcher.java:185)
	at aQute.launcher.Launcher.main(Launcher.java:161)
	at aQute.launcher.pre.EmbeddedLauncher.executeWithRunPath(EmbeddedLauncher.java:170)
	at aQute.launcher.pre.EmbeddedLauncher.findAndExecute(EmbeddedLauncher.java:135)
	at aQute.launcher.pre.EmbeddedLauncher.main(EmbeddedLauncher.java:52)
java.lang.NullPointerException
	at aQute.launcher.Launcher.trace(Launcher.java:1562)
	at aQute.launcher.Launcher.launch(Launcher.java:460)
	at aQute.launcher.Launcher.run(Launcher.java:185)
	at aQute.launcher.Launcher.main(Launcher.java:161)
	at aQute.launcher.pre.EmbeddedLauncher.executeWithRunPath(EmbeddedLauncher.java:170)
	at aQute.launcher.pre.EmbeddedLauncher.findAndExecute(EmbeddedLauncher.java:135)
	at aQute.launcher.pre.EmbeddedLauncher.main(EmbeddedLauncher.java:52)
```

I have traced this through and it seems that the launch.properties file is completely empty. It is like the properties are not loaded before `ProjectLauncherImpl.writeProperties()` is called. Makes it hard to do testing.

It also seems that this error doesn't happen if I launch the same .bndrun file with the regular Bnd OSGi Launcher (not the test launcher).

I suspect this has something to do with #4470 or #4480 (or both). It isn't happening when I run from https://github.com/bjhargrave/bnd/tree/enable-test.

bjhargrave:
[...]

PR #4555, opened 2021-02-16
Revert

rotty3000:
This broke upstream so in the interest of stability reverting until I can figure out what I did wrong. Sorry for the inconvenience.


rotty3000:
This should fix #4549

bjhargrave:
Note, 9edd9b00c2af190bb76dcf879e7e4efe7b3be62e is still in the branch. If this fix is not replaced, we should revert 9edd9b00c2af190bb76dcf879e7e4efe7b3be62e to remove the unimplemented API. I just reverted 9edd9b00c2af190bb76dcf879e7e4efe7b3be62e in next for [5.3.0.RC2](https://github.com/bndtools/bnd/tree/5.3.0.RC2) so the unimplemented API is not in 5.3.0.

PR #4561, opened 2021-02-19
Update classes macro to add STATIC and INNER queries

bjhargrave:
`STATIC` and `INNER` are opposites.

Documentation for the classes macro is also updated to document some query types that were previously undocumented.


rotty3000:
nice!

PR #4569, opened 2021-02-25
deps: Update Eclipse dependency to 4.16 2020-06

bjhargrave:
Include fixes for `:bndtools.core.test:testOSGi` from @kriegfrj.


bjhargrave:
@kriegfrj Can you look at this PR? There are numerous failures in the biz.aQute.tester.test tests. There is some oddity I could not figure out but perhaps you can! Let me know.

kriegfrj:
> @kriegfrj Can you look at this PR? There are numerous failures in the biz.aQute.tester.test tests. There is some oddity I could not figure out but perhaps you can! Let me know.

Currently investigating.

Have you looked at the Launchpad failures (visible in the CI output)? The tester tests use Launchpad extensively, if there are issues with Launchpad it could cause the tester tests to fail unexpectedly.

kriegfrj:
I created a PR to merge in a fix into your branch: https://github.com/bjhargrave/bnd/pull/23 Give it a go and see if that works for you (hopefully it should...).

bjhargrave:
@kriegfrj Thanks for the PR. Your debugging and analysis were spot on. I incorporated your fix into this PR and made more progress. The launchpad test failures are solved.

I think everything works now except `:bndtools.core.test:testOSGi` where I need your help again.

bjhargrave:
@kriegfrj BTW, you can now resolve the bndrun files in `bndtools.core.test`. You run resolve on the os/processor specific bndruns (and not the shared bndrun). Each os/processor specific bndrun holds the complete computed `-runbundles` for that os/processor. With the `-runblacklist` I updated, it resolves fairly quickly.

kriegfrj:
> @kriegfrj Thanks for the PR. Your debugging and analysis were spot on.
 :+1: 

The most finnicky part of those Launchpad tests was always trying to match the versions between the build/testpath and the runbundles, so that was always the first place to look.

I'm not exactly sure what it was about my changes that fixed it, but given that it worked I didn't want to waste too much time investigating in order to answer that question. :smile: A lot of what I had was arrived at by trial-and-error.

The original implementation was complicated by Eclipse using a too-old version of the `junit-platform-launcher` module (and that's where the `junit.platform.tester.version` property became important). The need for that split version should be gone now that we've moved to a more recent version of Eclipse, which simplifies things significantly.

> I think everything works now except `:bndtools.core.test:testOSGi` where I need your help again.

Yikes... I've had a quick look and it is dredging up some bad memories... it's going to take a bit of digging, I think. Either we're missing an Eclipse bundle, we've got the start order wrong, or there's a startup race condition...

> @kriegfrj BTW, you can now resolve the bndrun files in `bndtools.core.test`. You run resolve on the os/processor specific bndruns (and not the shared bndrun). Each os/processor specific bndrun holds the complete computed `-runbundles` for that os/processor. With the `-runblacklist` I updated, it resolves fairly quickly.
> 
> This also applies to the bndruns in `bndtools.core`.

:+1: This is fantastic, great work!

bjhargrave:
> Yikes... I've had a quick look and it is dredging up some bad memories... it's going to take a bit of digging, I think. Either we're missing an Eclipse bundle, we've got the start order wrong, or there's a startup race condition...

The issue seems to be that https://github.com/bndtools/bnd/blob/b53c3b8568d7bc6800a74a05685c523c303d20c1/bndtools.core.test/src/bndtools/core/test/utils/WorkspaceImporter.java#L82 never calls the countdown monitor. I debugged through the import operation and the countdown monitor is wrapped several times in submonitors but I don't fully grok submonitors.

kriegfrj:
> > Yikes... I've had a quick look and it is dredging up some bad memories... it's going to take a bit of digging, I think. Either we're missing an Eclipse bundle, we've got the start order wrong, or there's a startup race condition...
> 
> The issue seems to be that
> 
> https://github.com/bndtools/bnd/blob/b53c3b8568d7bc6800a74a05685c523c303d20c1/bndtools.core.test/src/bndtools/core/test/utils/WorkspaceImporter.java#L82
> 
> never calls the countdown monitor. I debugged through the import operation and the countdown monitor is wrapped several times in submonitors but I don't fully grok submonitors.

Yes, that is the conclusion that I came to in my brief look at this last night. Unfortunately (for this situation) there is nothing in the `IProgressMonitor` spec that requires the callee to call `done()` when it is finished (see [here](https://www.eclipse.org/articles/Article-Progress-Monitors/article.html), section 4). This could be the source of other timing issues we're having with this test; I'll need to see if I can think of another way to synchronize with the background tasks...

gamerson:
> > > Yikes... I've had a quick look and it is dredging up some bad memories... it's going to take a bit of digging, I think. Either we're missing an Eclipse bundle, we've got the start order wrong, or there's a startup race condition...
> > 
> > 
> > The issue seems to be that
> > https://github.com/bndtools/bnd/blob/b53c3b8568d7bc6800a74a05685c523c303d20c1/bndtools.core.test/src/bndtools/core/test/utils/WorkspaceImporter.java#L82
> > 
> > never calls the countdown monitor. I debugged through the import operation and the countdown monitor is wrapped several times in submonitors but I don't fully grok submonitors.
> 
> Yes, that is the conclusion that I came to in my brief look at this last night. Unfortunately (for this situation) there is nothing in the `IProgressMonitor` spec that requires the callee to call `done()` when it is finished (see [here](https://www.eclipse.org/articles/Article-Progress-Monitors/article.html), section 4). This could be the source of other timing issues we're having with this test; I'll need to see if I can think of another way to synchronize with the background tasks...

Hey @kriegfrj  so the team I work on at Liferay, for functional tests we have for our Eclipse plugins, we have code like this:

```
 protected void waitForBuildAndValidation() {
                IWorkspaceRoot root = null;

                IJobManager manager = Job.getJobManager();

                IWorkspace workspace = ResourcesPlugin.getWorkspace();

                try {
                        workspace.checkpoint(true);

                        manager.join(ResourcesPlugin.FAMILY_AUTO_BUILD, new NullProgressMonitor());
                        manager.join(ResourcesPlugin.FAMILY_MANUAL_BUILD, new NullProgressMonitor());
                        manager.join(ResourcesPlugin.FAMILY_AUTO_BUILD, new NullProgressMonitor());
                        manager.join(ValidatorManager.VALIDATOR_JOB_FAMILY, new NullProgressMonitor());

                        IJobManager jobManager = Job.getJobManager();

                        jobManager.join("org.eclipse.buildship.core.jobs", new NullProgressMonitor());

                        JobUtil.waitForLiferayProjectJob();

                        needJobsToBuild(manager);

                        Thread.sleep(200);

                        manager.beginRule(root = workspace.getRoot(), null);
                }
                catch (InterruptedException ie) {
                        failTest(ie);
                }
                catch (IllegalArgumentException iae) {
                        failTest(iae);
                }
                catch (OperationCanceledException oce) {
                        failTest(oce);
                }
                finally {
                        if (root != null) {
                                manager.endRule(root);
                        }
                }
        }
```

So this will try to create a rule that needs exclusive access to the workspace and wait.  I can dig in more to get more info if you want to try something like creating a rule and waiting on it.

Issue #4553, opened 2021-02-15
[quickfix tests] simpleReference_toWorkspaceClass_suggestsBundles() flaky


kriegfrj:
`bndtools.core.test.editors.quickfix.BuildpathQuickFixProcessor_WithEmptyBuildpath_Test#simpleReference_toWorkspaceClass_suggestsBundles()` is current failing intermittently.

As the original flakiness issue (#4253) seems to have been solved by #4404, this sole flaky test is the only thing stopping it from being re-integrated into the CI build. So I thought to disable this flaky test only and raise a new issue to track this sole remaining flaky test, allowing us to re-instate the bulk of the tests.

Current state of play with this test:

* I can't seem to reproduce it simply by running Gradle on my workstation with JDK 1.8, nor from within Bndtools.
* It seems to happen fairly regularly in CI, so at the moment I have to try things out by pushing to the CI.
* Maybe it's related to a particular JDK version (JDK 15 seems to be failing).

PR #4579, opened 2021-03-05
[bndtools test] Wait for cnf workspace before proceeding

kriegfrj:
In order to try and overcome race conditions at startup, put in a callback that causes the tester to wait for the Workspace to be ready before proceeding.

This is a bit of a hack, but it seems to stabilise the tests - I ran it in my own CI instance on GitHub a few times and it seems to have succeeded on all platforms each time.

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -24,7 +24,6 @@ env:
     -Dhttp.keepAlive=false
     -Dmaven.wagon.http.pool=false
     -Dmaven.wagon.http.retryHandler.count=3
-  BNDTOOLS_CORE_TEST_NOJUNITOSGI: true
 
 defaults:
   run:
--- a/bndtools.core.test/src/bndtools/core/test/editors/quickfix/AbstractBuildpathQuickFixProcessorTest.java
+++ b/bndtools.core.test/src/bndtools/core/test/editors/quickfix/AbstractBuildpathQuickFixProcessorTest.java
@@ -230,6 +230,7 @@ abstract class AbstractBuildpathQuickFixProcessorTest {
 			if (buildPath != null && !buildPath.isEmpty()) {
 				model.setBuildPath(Collections.emptyList());
 				model.saveChanges();
+				Central.refresh(bndProject);
 				waitForClasspathUpdate("clearBuildpath()");
 			} else {
 				log("buildpath was not set; not trying to clear it");
@@ -248,6 +249,7 @@ abstract class AbstractBuildpathQuickFixProcessorTest {
 				model.addPath(new VersionedClause(bundleName, null), Constants.BUILDPATH);
 			}
 			model.saveChanges();
+			Central.refresh(bndProject);
 			waitForClasspathUpdate("addBundleToBuildpath");
 		} catch (Exception e) {
 			throw Exceptions.duck(e);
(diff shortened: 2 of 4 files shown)


kriegfrj:
Looking at the failure, it seems unrelated to the previous. It seems that BuildPathQuickFixProcessorTest_WithSimpleOnClasspath failed to add Bndtools.core.test.fodder.simple to the -buildpath. I've never seen this before and when I reran it it passed. 

Perhaps when it attempted to install the simple bundle onto -buildpath, the local repository hadn't finished updating yet?

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -240,9 +240,10 @@ public static Workspace getWorkspace() throws Exception {
 				// one
 				File workspaceDirectory = getWorkspaceDirectory();
 				// Check to see if we need to convert it...
-				if (workspaceDirectory != null && ws.isDefaultWorkspace()) {
```
bjhargrave, line 243: You should pull this commit out into its own PR so it does not languish here...

kriegfrj: Done; see #4582.

kriegfrj:
> Looking at the failure, it seems unrelated to the previous. It seems that BuildPathQuickFixProcessorTest_WithSimpleOnClasspath failed to add Bndtools.core.test.fodder.simple to the -buildpath. I've never seen this before and when I reran it it passed.
> 
> Perhaps when it attempted to install the simple bundle onto -buildpath, the local repository hadn't finished updating yet?

I revisited this stray failure - I noted that the bnd.bnd file is updated direct and not through the Eclipse resources API, meaning that it is possible Eclipse may not have noticed. Added an explicit call to `Central.refresh()` in order to try and prevent this.

bjhargrave:
@kriegfrj You feeling this is ready? If so, please remove the fail-fast commit and rebase.

kriegfrj:
> @kriegfrj You feeling this is ready? If so, please remove the fail-fast commit and rebase.

I've learned by painful experience not to trust my feelings on this one... but let's give it a go...

I *think* this will solve #4253. I'm hopeful (but less confident) that it has solved #4553.

I think longer term I need to try and simplify the test a little bit to isolate the code-under-test from the other bndtools components. The test as it currently stands relies on other core Bndtools components like the builder and the cnf watcher. My original attempt at this was to build by calling Bnd's `Project.build()` method directly, but I couldn't get this to work. I now know that this was because I needed to refresh the build results. With more experience I might be able to do better. But I think what we have will work now, even if it is not 100% ideal.

bjhargrave:
OK, its merged. Lets hope it remains stable :-)

kriegfrj:
So far so good? I haven't seen any complaints?

bjhargrave:
https://github.com/bndtools/bnd/runs/2399867371?check_suite_focus=true#step:5:4479 seems to be a random failure

bjhargrave:
And another https://github.com/bndtools/bnd/runs/2404153924?check_suite_focus=true#step:5:4119

I think I want to disable this test in the CI build again.

kriegfrj:
These two failure logs give me enough to go by to revisit and try and sort it out. This is unlikely to be soon so feel free to disable them in the meantime. The good thing is that the tests are still available to run any time someone is working on this code, which is when it is most important - having them run in CI would be icing on the cake.

The first failure seems to be a new failure mode. It is possible that Maven Central happened to be unreachable at the time and so it couldn't download the JUnit bundle. I could fix this by creating synthetic fodder classes that re-create the situation it is testing for and ditching the reliance on the external bundle.

The second seems like a variant of the original #4253. I now know that "Local Index" being null means that the workspace object hasn't been properly initialised. I'll have to have a closer look and see how this might be able to happen. I'll put my findings in the issue ticket.

bjhargrave:
> It is possible that Maven Central happened to be unreachable

This happens more often than I like but I think we cannot be trying to work around this. I would not bother to do anything special here since some other place in the build will probably also fail.

PR #4585, opened 2021-03-11
[resolve] Allow inclusion of the system resource

pkriens:
I had a use case where I also needed access to
the system resource. However, this is unfortunately
explicitly removed from the result.

This introduces a instruction -resolve.excludesystem
that by default excludes it but can be set to false
to include it in the result.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
@@ -223,6 +223,12 @@ public interface Constants {
 
 	String				RESOLVE										= "-resolve";
 
+	/**
+	 * Exclude the system resource from the resulting wiring in resolve. The
+	 * default is true
+	 */
+	String				RESOLVE_EXCLUDESYSTEM						= "-resolve.excludesystem";
+
 	String				RUNNOREFERENCES								= "-runnoreferences";
 	String				RUNFRAMEWORKRESTART							= "-runframeworkrestart";
 	String				RUNOPTIONS									= "-runoptions";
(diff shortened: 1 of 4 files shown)


amitjoy:
Thanks @pkriens for the quick addition. Just tested the branch and it works like a charm 👍

biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
```diff
@@ -223,6 +223,12 @@
 
 	String				RESOLVE										= "-resolve";
 
+	/**
+	 * Exclude the system resource from the resulting wiring in resolve. The
+	 * default is true
+	 */
+	String				RESOLVE_EXCLUDESYSTEM						= "-resolve.excludesystem";
```
bjhargrave, line 230: I think this needs to be called `-resolve-excludesystem`. Otherwise it looks like a merge property for `-resolve`.

docs/_instructions/resolve.excludesystem.md
```diff
@@ -0,0 +1,10 @@
+---
+layout: default
+class: Runtime
+title: -resolve.excludesystem true|false
```
bjhargrave, line 4: Also need to fix the name here (in the text and the file name).

biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
pkriens: Resolve is not a merge property,it can only be a single value.

bjhargrave: Today. But this will preclude it ever being a merge property. And it also looks like a merge property.  No other instruction uses a `.` to separate words. They all use `-`. e.g. `-dsannotations-options`. I think using a `.` here is a bad precedent when we can easily use `-`.

PR #4603, opened 2021-03-31
PomRepository should read maven GAV with Type and Classifier

stbischof:
At the moment the PomRepository  can only read a maven-revision, a G:A:V without Type and Classifier.
When reading an Archive instead of an Revision , like BndPomRepository does, this could be fixed.


stbischof:
Thank you @bjhargrave for the fast review. These first build was just that all tests works. (could not building it local)

I picked up your wishes and does the changes.

bjhargrave (approved):

bjhargrave:
Nicely done!

PR #4608, opened 2021-04-01
Introduced OBR Exporter Plugin

amitjoy:


biz.aQute.resolve/test/biz/aQute/resolve/obr/plugin/ObrExporterTest.java
```diff
@@ -0,0 +1,150 @@
...
+    @Test
+    public void test_obr_generation_without_system_bundle_capabilities() throws Exception {
+        final Map<String, String> options = new HashMap<>();
+
+        underTest.export(TYPE, project, options);
+
+		final File obr = getFile("target/test.xml");
+        assertThat(obr).exists();
+
+        final XQuery xq = XQuery.parse(new FileReader(obr));
```
timothyjward, line 76: Is there really any reason to use XPath here? The XML could be read back by bnd into a repository and queried using existing API. That would also better validate that the XML wasn't mangled.

timothyjward (changes requested):

biz.aQute.resolve/test/biz/aQute/resolve/obr/plugin/ObrExporterTest.java
timothyjward: In fact if the XQuery usage here is removed then the whole dependency becomes unnecessary.

amitjoy: I actually thought to make use of an external library for XML parsing as it would make the code more readable, otherwise, the code will be very cluttered with SAXParser.

pkriens: I agree that we should not use a dependency unless it is paramount. You can also parse it with XPath, which is built in.

timothyjward: > I actually thought to make use of an external library for XML parsing as it would make the code more readable, otherwise, the code will be very cluttered with SAXParser.

The bnd code already includes a class for parsing an XML into a repository. There would be no need to create a parser for the test.

amitjoy: > The bnd code already includes a class for parsing an XML into a repository. There would be no need to create a parser for the test.

Ah that would be awesome. I didn't know about it. Could you kindly point me to the API?

amitjoy: I have updated the code to use `XPathFactory` instead which doesn't require use of any external library as well.

timothyjward: > Ah that would be awesome. I didn't know about it. Could you kindly point me to the API?

[XMLResourceParser](https://github.com/bndtools/bnd/blob/8b7d203e5d87026651d7cbf88a75903ef59d0de6/biz.aQute.bndlib/src/aQute/bnd/osgi/repository/XMLResourceParser.java) usage visible in [RepositoryTest](https://github.com/bndtools/bnd/blob/8b7d203e5d87026651d7cbf88a75903ef59d0de6/biz.aQute.bndlib.tests/test/test/RepositoryTest.java)

This would be better than XPath as it demonstrates round tripping of the XML.

amitjoy: @timothyjward Thanks a lot for the support! I have updated the tests to use `XMLResourceParser` instead. It looks pretty clean now.

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
```diff
@@ -0,0 +1,85 @@
...
+import aQute.bnd.osgi.repository.XMLResourceGenerator;
+import aQute.bnd.service.export.Exporter;
+import aQute.lib.exceptions.Exceptions;
+import biz.aQute.resolve.Bndrun;
+import biz.aQute.resolve.RunResolution;
+
+@BndPlugin(name = "OBR Exporter")
+public final class ObrExporter implements Exporter {
+
+    static final String TYPE = "exporter.obr";
```
bjhargrave, line 27: I think the name should start with "bnd." like the other Bnd exporter names.

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
```diff
@@ -0,0 +1,85 @@
...
+
+        final String outputdir       = options.get("outputdir");
+        final String name            = options.getOrDefault("name", removeExtension(project.getName()) + ".xml");
+        final String excludesystem   = options.getOrDefault("excludesystem", "true");
+
+        File outputDirectory = null;
+        if (outputdir == null) {
+            outputDirectory = project.getTargetDir();
+        } else {
+            outputDirectory = new File(project.getBase(), outputdir);
```
bjhargrave, line 48: I think you want `project.getFile(outputdir)`. This will use the project's base.

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        project.setProperty(RESOLVE_EXCLUDESYSTEM, excludesystem);
```
bjhargrave, line 51: Is this is a good idea? To modify the input project?

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
```diff
@@ -0,0 +1,85 @@
...
+
+        final File obrIndex             =  new File(outputDirectory, name);
+        final RunResolution resolution  =  ((Bndrun) project).resolve();
```
bjhargrave, line 54: This seems risky? What if the input project is not a Bndrun? Perhaps you should confirm the input is an instanceof before casting and set a proper message if not.

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    private static void exportOBR(
+            final Project project,
+            final Collection<Resource> resources,
+            final File output) {
+
+        final XMLResourceGenerator xmlResourceGenerator = new XMLResourceGenerator();
+
+        try (final ByteArrayOutputStream buffer = new ByteArrayOutputStream()) {
```
bjhargrave, line 73: In general we avoid ByteArrayOutputStream in the code. Please use ByteBufferOutputStream (from aQute.libg) instead.

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
```diff
@@ -0,0 +1,85 @@
...
+            xmlResourceGenerator.resources(resources);
+            xmlResourceGenerator.save(output);
+        } catch (final Exception e) {
+            project.error("Failed to generate OBR from %s due to %s", project, Exceptions.toString(e));
```
bjhargrave, line 77: I think you want to use the exception method here instead of error.

biz.aQute.resolve/test.bndrun
```diff
@@ -0,0 +1,13 @@
+-standalone: true
```
bjhargrave, line 1: Since this is a file for testing, it should not live in the root of the project but inside the `testdata` folder.

biz.aQute.resolve/test/biz/aQute/resolve/obr/plugin/ObrExporterTest.java
```diff
@@ -0,0 +1,149 @@
...
+import biz.aQute.resolve.Bndrun;
+
+public final class ObrExporterTest {
+
+	private static Bndrun		project;
+	private static ObrExporter	underTest;
+
+	@BeforeClass
+	public static void setup() throws Exception {
+           project = createBndrun(findWorkspace(IO.work), getFile("test.bndrun"));
```
bjhargrave, line 31: You don't want to use static setup and teardown since there is the problem of side effects of one test method affecting other test methods. 

Also, you do not want to use the actual Bnd workspace as the workspace for your test. You need to construct a specific workspace in testdata to use. And then copy this workspace into the test output folder and use it from there. This also prevents side effects from one test method from influencing other test methods.

bjhargrave (changes requested):

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
amitjoy: Makes sense to be consistent with the current naming convention. will do.

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
amitjoy: Good to know. will do.

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
amitjoy: Actually the requirement to have such a configuration exists in our project so that we can include system resource too and that's I thought it might be a good idea to have this configurable in the project.

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
amitjoy: Yeah, it can lead to a `ClassCastException` if the project is not a type of `Bndrun`. will fix it.

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
amitjoy: Didn't know about it. will fix it.

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
amitjoy: Didn't know about it. I have seen one instance where `error` was used instead of `exception` and that's why I used the same.

biz.aQute.resolve/test.bndrun
amitjoy: Yeah, makes sense. Will fix it.

biz.aQute.resolve/test/biz/aQute/resolve/obr/plugin/ObrExporterTest.java
amitjoy: Yeah, you are right. Will move to method level setup and teardown.

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
baishalidutta: DONE ✔️

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
baishalidutta: DONE ✔️

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
amitjoy: DONE ✔️

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
amitjoy: DONE ✔️

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
amitjoy: DONE ✔️

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
amitjoy: DONE ✔️

biz.aQute.resolve/src/biz/aQute/resolve/obr/plugin/ObrExporter.java
amitjoy: DONE ✔️

biz.aQute.resolve/test.bndrun
amitjoy: DONE ✔️

biz.aQute.resolve/test/biz/aQute/resolve/obr/plugin/ObrExporterTest.java
amitjoy: DONE ✔️

biz.aQute.resolve/test/biz/aQute/resolve/obr/plugin/ObrExporterTest.java
```diff
@@ -0,0 +1,149 @@
...
+import biz.aQute.resolve.Bndrun;
+
+public final class ObrExporterTest {
+
+    private Bndrun      project;
+    private ObrExporter	underTest;
+
+    @Before
+    public void setup() throws Exception {
+        project = createBndrun(findWorkspace(IO.work), getFile("testdata/obr-exporter/test.bndrun"));
```
bjhargrave, line 31: This still seems wrong in that you are using the actual Bnd workspace as the workspace for the test. You need to create a test workspace in an output folder and the new Workspace on that folder.

biz.aQute.resolve/test/biz/aQute/resolve/obr/plugin/ObrExporterTest.java
```diff
@@ -0,0 +1,149 @@
...
+        underTest = new ObrExporter();
+    }
+
+    @After
+    public void teardown() throws IOException {
+        final File testResource1 = IO.getFile("testdata/obr-exporter/target/test.xml");
```
bjhargrave, line 37: This shows the problem with using the real Bnd workspace for testing. You make files in the real workspace which must be cleaned up. This should not be necessary since your test workspace should be in an output folder.

biz.aQute.resolve/testdata/obr-exporter/obr/.gitignore
```diff
@@ -0,0 +1 @@
+*.xml*
```
bjhargrave, line 1: Another symptom of incorrectly using the real Bnd workspace as the test workspace.

biz.aQute.resolve/testdata/obr-exporter/target/.gitignore
```diff
@@ -0,0 +1 @@
+*.xml*
```
bjhargrave, line 1: Another symptom of incorrectly using the real Bnd workspace as the test workspace.

bjhargrave (changes requested):

biz.aQute.resolve/testdata/obr-exporter/target/.gitignore
amitjoy: DONE ✔️

biz.aQute.resolve/testdata/obr-exporter/obr/.gitignore
amitjoy: DONE ✔️

biz.aQute.resolve/test/biz/aQute/resolve/obr/plugin/ObrExporterTest.java
amitjoy: DONE ✔️

biz.aQute.resolve/test/biz/aQute/resolve/obr/plugin/ObrExporterTest.java
amitjoy: DONE ✔️

pkriens:
Thanks Amit. Maybe you could also provide a description in `docs/plugins`? And maybe a post in the bndtools.discourse.group?

Open source sucks! Too much work :-)

baishalidutta:
@pkriens Yeah makes sense. Will do asap.

amitjoy:
@pkriens Makes sense. Will do asap.

PR #4611, opened 2021-04-07
[gradle] support filtering executable tasks

stbischof:
The goal is to control the testrun, resolve and export task by using context properties like
```
bnd {
  properties {
    property "bnd.testrun.filter", "test(.*)?\.bndrun" 
    property "bnd.export.filter", "export(.*)?\.bndrun"
    property "bnd.resolve.filter", "(test|export)(.*)?\.bndrun"
}}
```

and an option to exclude bnd.bnd files from the `testOSGi` task.

the commands then should run only filterd bndruns
```
./gradlew resolve
./gradlew export
./gradlew testrun
```

This is my first time with groovy and gradle..... Help is welcome!!!

### DIFF
--- a/biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BndPlugin.groovy
+++ b/biz.aQute.bnd.gradle/src/aQute/bnd/gradle/BndPlugin.groovy
@@ -50,12 +50,16 @@ public class BndPlugin implements Plugin<Project> {
   private ProjectLayout layout
   private ObjectFactory objects
   private aQute.bnd.build.Project bndProject
+  private BndPluginExtension extension
 
   /**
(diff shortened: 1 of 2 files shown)


bjhargrave:
This is the wrong way to go about this. You should not modify the Bnd gradle plugin for this. Just write your own gradle task. For example:

```groovy
tasks.register('resolveMatching') {
  dependsOn tasks.matching { t -> t.name ~= '(test|export)(.*)?\.bndrun')
}
```

This is untested and the specific language syntax may be wrong, but you can see how easy it is to just write a simple gradle task to do what you want.

I do not care for modifying the gradle plugin for this.

stbischof:
thank you!

PR #4613, opened 2021-04-09
[quickfix] Handle bound mismatches

kriegfrj:
Handles case where a type bound doesn't pass because of a missing common supertype.

Based on a real-world case that I encountered recently.


bndtools.core.test/src/bndtools/core/test/editors/quickfix/AbstractBuildpathQuickFixProcessorTest.java
```diff
@@ -331,7 +332,8 @@ void before() throws Exception {
 	private static final String CLASS_FOOTER = " var};";
 	protected static final Set<Integer> SUPPORTED = Stream
```
bjhargrave, line 333: This should probably be `Sets.of`. This uses the immutable collections support in aQute.libg.

kriegfrj: Done.

bjhargrave (approved):

PR #4647, opened 2021-05-01
[sync] Should be more careful in deleting

pkriens:


pkriens:
Immediately merged because deleted projetcts

PR #4662, opened 2021-05-12
ImmutableEntry - remove requireNonNull for value

stbischof:
a value of a Entry could be null.


bjhargrave:
No. This violated the design point which is that null keys and null values are not allowed in the unmodifiable map.

See AbstractMap.SimpleImmutableEntry if you need an immutable entry which allows for null key and null value.

stbischof:
Thx , did not heard about before the reasons. But now its clear.

PR #4674, opened 2021-05-17
[PomRepository] respect classifier when using artifact-gav as pseudo-bsn

stbischof:
`g:a:jar` and `g:a:jar:all` are different artifacts. We should respect the classifier when using artifact-gav as pseudo-bsn.

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
@@ -29,6 +29,7 @@ import org.osgi.util.promise.Promise;
 
 import aQute.bnd.annotation.plugin.BndPlugin;
 import aQute.bnd.build.Workspace;
+import aQute.bnd.exceptions.Exceptions;
 import aQute.bnd.http.HttpClient;
 import aQute.bnd.osgi.Constants;
 import aQute.bnd.osgi.Processor;
@@ -47,7 +48,6 @@ import aQute.bnd.service.repository.Prepare;
 import aQute.bnd.util.repository.DownloadListenerPromise;
 import aQute.bnd.version.Version;
 import aQute.lib.converter.Converter;
-import aQute.bnd.exceptions.Exceptions;
 import aQute.lib.io.IO;
 import aQute.lib.strings.Strings;
 import aQute.libg.reporter.slf4j.Slf4jReporter;
@@ -286,7 +286,8 @@ public class BndPomRepository extends BaseRepository
 		} else {
 
 			String name = resource.getInfo()
-				.name();
+				.name() + ":" + version;
+
 			archive = Archive.valueOf(name);
 		}
 
@@ -415,7 +416,7 @@ public class BndPomRepository extends BaseRepository
 			return null;
 
 		return Archive.valueOf(resource.getInfo()
-			.name())
+			.name() + ":" + version)
 			.getOther("jar", Archive.SOURCES_CLASSIFIER);
 	}
 
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/Traverser.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/Traverser.java
@@ -27,11 +27,11 @@ import org.osgi.util.promise.PromiseFactory;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import aQute.bnd.exceptions.Exceptions;
 import aQute.bnd.http.HttpClient;
 import aQute.bnd.maven.MavenCapability;
 import aQute.bnd.osgi.resource.CapabilityBuilder;
 import aQute.bnd.osgi.resource.ResourceBuilder;
-import aQute.bnd.exceptions.Exceptions;
 import aQute.maven.api.Archive;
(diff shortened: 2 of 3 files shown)


biz.aQute.repository/test/aQute/bnd/repository/maven/pom/provider/PomRepositoryTest.java
```diff
@@ -714,12 +714,17 @@ public void testMultipleRevisions() throws Exception {
...
 			assertThat(mcsr.list("*")).contains("biz.aQute.remote.launcher",
-				"biz.aQute.launcher", "biz.aQute.junit", "biz.aQute.tester", //
-				"org.jacoco:org.jacoco.agent", // with classifier
+				"biz.aQute.launcher", "biz.aQute.junit", "biz.aQute.tester",
+				// "org.jacoco:org.jacoco.agent:0.8.6"
+				// a bundle so normal bsn
+				"org.jacoco.agent",
+				// "org.jacoco:org.jacoco.agent:jar:runtime:0.8.6"
+				// with classifier & not a bundle
+				"org.jacoco:org.jacoco.agent:jar:runtime",
```
bjhargrave, line 723: This is changing the pseudo-bsn to include extension and classifier. It currently is groupId:artifactId only.

By changing the pseudo-bsn to include extension and classifier, you will break some builds which don't include this. You can see this here in the needed change in the test case. The pseudo-bsn of `org.jacoco:org.jacoco.agent` had to be changed to `org.jacoco:org.jacoco.agent:jar:runtime`.

I am not sure such a breaking change is a good idea.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/Traverser.java
```diff
@@ -207,7 +207,9 @@ private void parseResource(Archive archive, String parent) throws Exception {
 		ResourceBuilder rb = new ResourceBuilder();
 
 		Version frameworkVersion = toFrameworkVersion(archive.revision.version.getOSGiVersion());
-		String bsn = archive.revision.program.toString();
+
+		// g:a:jar and g:a:jar:all are different artifacts respect classifier
+		String bsn = archive.getWithoutVersion();
```
bjhargrave, line 212: Actually, this seems to be what MavenBndRepository does:

https://github.com/bndtools/bnd/blob/400cd7557624ba4428f074a9b3e0023ec220108c/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java#L328-L334

However, MavenBndRepository does not set a osgi.identity (which `addReserveIdentity` does). Instead it adds an information capability which the BridgeRepository understands. We should change BndPomRepository to match the MavenBndRepository behavior here.

biz.aQute.repository/test/aQute/bnd/repository/maven/pom/provider/PomRepositoryTest.java
stbischof: Afaik we did bit release a version where a gav with an classifier will be processed from pomrepository.

So this is mor a fix. 
Though that this is now the same behavior like the bndmaven plugin and the pom plugin when it uses a pom and not a revision.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/Traverser.java
stbischof: We should. Let both work same as much as possible.

Could i address this in an extra pr or would you have it in this?

stbischof: I added a new commit.
MavenBndRepository handles name and version seperate.
```
String name = archive.getWithoutVersion();  // groupId:artifactId
addInformationCapability(rb, name, version.getOSGiVersion(), repo.toString(),
```
while BndPomRepository holds the version in the name.
```
addInformationCapability(rb, archive.toString(), parent); //groupId:artifactId:version
```
because of this I had  to change some Archive name calculations.`Archive.valueOf(a + ":" + version);`


@bjhargrave This is what i can do with my very limited knowledge about the repository.  Hope this is the right direction.

biz.aQute.repository/src/aQute/bnd/repository/maven/pom/provider/BndPomRepository.java
```diff
@@ -286,7 +286,8 @@ public File get(String bsn, Version version, Map<String, String> properties, Dow
 		} else {
 
 			String name = resource.getInfo()
-				.name();
+				.name() + ":" + version;
```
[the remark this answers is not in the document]
stbischof: I agree that before this PR `resource.getInfo().name()` set by `BndPomRepository` also holds the version.

But as in your last comment described the MavenBndRepository uses as name `String name = archive.getWithoutVersion();` 

https://github.com/bndtools/bnd/blob/400cd7557624ba4428f074a9b3e0023ec220108c/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java#L328-L334

so i changed to this behavior an run into exceptions because of the missing version. that is why i added the concatenatioC.

Could you please tell me which unit-test you had debugged?

bjhargrave:
See #4684 which is a similar but more complete fix. Just adding `":" + version` is not correct when the bundle version != artifact version which is not uncommon.

stbischof:
Thank You!!!!

PR #4679, opened 2021-05-19
WIP: Workspace extension

juergen-albert:
Wokspace Extensions are the result of the discussion I had with @pkriens quite some time ago on [discourse](https://bnd.discourse.group/t/add-release-extra-bundles/70).

They are a mechanism to add additional configurations to a Workspace, Project or bndrun. The idea is to add an instruction like `-workspace-extensions: test;version=1.2.3` to any bnd file. The result will be that, looks through any configured repository for a bundle that provides a Capability like 
```
 bnd.workspace.extension;\
   bnd.workspace.extension="test";\
   version="1.2.3";\
   description="Something we can later use in the UI";\
   include="workspace.bnd"
``` 
unpacks it and includes the named bnd file.
This can be used to add complex configurations and setups to any build in a stable and formalized fashion. This can be e.g. used with our junit5 and continuous testing. We can provide one extension, that comes along with a repo config that provides all the dependencies usually used with junit5. This can bring another extension with it, that if used configures the  tester and any necessary properties needed. We can even create a extension for the jdt include we have in all our projects.

@bjhargrave @pkriens @kriegfrj @stbischof @rotty3000: I open this PR relatively early to get your input. I hock in the init phase of a Workspace/Project and I'm not sure if this will cause issues. Furthermore, I need some input on the name, because the extensions are not only for the workspace, but for projects and bndruns as well. Extensions are already taken, so I'm looking for a better name.

Known ToDos:
- rebase/merge with the current master to see where it might clash with or benefit from @pkriens recent changes
- obviously more tests
- squash the commits
- documentation


stbischof:
Here my first thoughts.
- At the moment i am often using `-include:` on remote files like https://github.com/geckoprojects-org/org.geckoprojects.bndincludes this does not work offline so it is a good idea to have this in a bundle.
- to define capability is a good way, because it opens the way to not only use bundles. I think feature-files would also be a good way to define this extensions. this is something i will test in the next weeks.
- in maven build, at the moment, we do not have a workspace and no repository. This we will need to properly use the workspace-extensions in maven layout.
- if you are using conditions in a bnd file e.g. via `${if;)` and a variables it would be good to have a description/ in the bundle that tells me which variables i have so set and what then happened.

bjhargrave:
Is this related to [`-extension`](https://bnd.bndtools.org/instructions/extension.html) which already exists to add workspace extensions? Since `-extension` already seems in the same general area, anything new should be related to this.

Also, you do need to rebase on master as much had changed and this PR cannot currently build.

kriegfrj:
Just thinking out loud here a little bit...

Just to make sure that I've got this straight: the goal here is to be able to include certain `.bnd` files via a bundle in the workspace.

I have faced a similar problem with iDempiere development - there is a substantial amount of setup which is common for any iDempiere plugin that you wish to develop. If you copy that setup into each plugin repo, it's a nightmare when you want to tidy up the setup as you have to repeat in each plugin repo.

Remote `-include` doesn't work for the reasons that @stbischof pointed out - if you're offline, it's a problem.

My workaround at the moment is to add subdirectories in cnf/ext which are git submodules that refer to the remote git idempiere-test library. Git caches the files locally, so that solves the "offline" problem. Git also takes care of the versioning - your submodule can link to a particular version in the remote repo, so you don't get bitten by upstream changes when you're not ready for them.

Just thinking out loud: Are there some use cases that your proposal will take care of that aren't covered by these approaches? One advantage that I can see is semantic versioning - my git approach is versioned, but the versioning is not semantic. I'm not sure how important semantic versioning will be in this context.

juergen-albert:
@bjhargrave 
> Is this related to [`-extension`](https://bnd.bndtools.org/instructions/extension.html) which already exists to add workspace extensions? Since `-extension` already seems in the same general area, anything new should be related to this.
> 
> Also, you do need to rebase on master as much had changed and this PR cannot currently build.
Yes they are kind of related but different. The `-extension` works with a BSN for one and expects a `ExtensionActivator` in it, that is then instantiated and called. With this on then can programmatically modify the Workspace.
My implementation is different in the following ways:
* One Bundle contain multiple extensions.
* It will work on workspace, project and bndrun level
* No magic reflective code invocation.

juergen-albert:
@kriegfrj This is exactly the issue I want to solve. A Capability can have a Version, so this should be taken care of.

juergen-albert:
> But it seems wrong to just ignore this and develop something else which appears to closely overlap. It makes more sense to expand upon the existing workspace extensions model and add more feature that you want.

I'm not sure how to do that without becoming weird. As my concept mainly differs in the fact that on bundle can have multiple extensions, is declared via a capability, does not require the execution of code and do not only work on the workspace level. Currently an extensions is declared as follows: `-extension: <bsn>;version=1.0`. As I'm not requiring a bsn we would need introduce a type, so I know what kind to handle. Besides the reuse of the instruction I don't know what other part of the model is worth keeping.

biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
```diff
@@ -698,7 +698,7 @@ public void close() throws IOException {
 		toBeClosed.clear();
 	}
 
-	private void clearPlugins() {
+	public void clearPlugins() {
```
bjhargrave, line 701: I do not think this is a good idea to make this public. Please keep private for internal use by Processor.

biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceExtensionHandler.java
```diff
@@ -0,0 +1,164 @@
...
+						doneSomething = true;
+					}
+				} catch (Exception e) {
+					processor.getLogger()
+						.error("Could not apply workspace extension " + extensionName + " in version " + versionString,
+							e);
+				}
+			}
+		}
+		processor.clearPlugins();
```
bjhargrave, line 148: I think you need to call a different method on Processor to clean the plugins. Like `propertiesChanged`.

biz.aQute.bndlib/src/aQute/bnd/workspace/extension/WorkspaceExtensionNamespace.java
```diff
@@ -0,0 +1,121 @@
...
+ * <pre>
+ * bnd.workspace.extension      name of the workspace extension
+ * version                      the version of the extension
+ * include                      declares the bnd file to be included
+ * description                  A useful description for the extension
+ * </pre>
+ */
+public interface WorkspaceExtensionNamespace {
+
+	Comparator<Capability>	VERSION_COMPARATOR					= new Comparator<Capability>() {
```
bjhargrave, line 31: Lambda?

biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceExtensionHandler.java
```diff
@@ -0,0 +1,164 @@
...
+	private Map<Capability, File>	enabledWorkspaceExtensions		= new HashMap<>();
+	private Workspace				ws;
+	private Processor				processor;
+
+	public WorkspaceExtensionHandler(Workspace ws) {
+		this.ws = ws;
+		processor = ws;
+	}
+
+	public WorkspaceExtensionHandler(Project project) {
```
bjhargrave, line 35: I don't understand how workspace extensions apply to projects. I can see you want to get some canned bnd files from a repository resource and apply it to the workspace. But how do Projects factor in here?

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -1801,4 +1817,12 @@ public OnWorkspace on(String ownerName) {
 		return notifier.on(ownerName);
 	}
 
+	public Collection<? extends Capability> getAvailableWorkspaceExtensions() {
```
bjhargrave, line 1820: Do these really need to be public API? For the most part we should try hard to just use private or default accessibility for new methods. Making then public or protected means we are making API we need to support in the long term.

bjhargrave:
> -extension: <bsn>;version=1.0

Is a requirement on a osgi.identity capability. So the current `-extension` also works on capabilities :-) Just osgi.identity capabilities.

The issue for me is that we have two disjoint workplace extensions mechanisms with this PR. The existing one and the one being proposed here. It is confusing to me, and I expect it will be to others, to have 2 things which are very similar in purpose and using the same "workspace extensions" terminology.

Your proposal introduces a new `bnd.workspace.extension` capability, so one could imaging doing

`-extension: test;namespace=bnd.workspace.extension;version=1.2.3`

To use the `bnd.workspace.extension` capability instead of the `osgi.identity` capability.

biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceExtensionHandler.java
juergen-albert: As I initially wrote, the name workspace extensions is a bit misleading. In comparison to the current extension, the new version can extend the properties for the workspace, a project or a bndrun. A better name would be PropertiesExtension and is basically only a mechanism to include properties that are wrapped in a bundle.

biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceExtensionHandler.java
juergen-albert: `propertiesChanged` might be a good idea, because this will solve another problem I had. I have to call it at least once though, because my mechanism initializes the plugins in a relatively early stage. This caused a couple of tests to fail. So, we will to the properties changed at least twice for each project/workspace initialization.

biz.aQute.bndlib/src/aQute/bnd/workspace/extension/WorkspaceExtensionNamespace.java
juergen-albert: thx, will do.

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
juergen-albert: you are right, this is unnecessary, as every run will go through the repositories anyway an.

biz.aQute.bndlib.tests/test/test/WorkspaceExtensionsTest.java
```diff
@@ -0,0 +1,51 @@
...
+	@Before
+	public void setUp() throws IOException {
+		testDir = new File(TMPDIR, getClass().getName() + "/" + testName.getMethodName());
+		IO.delete(testDir);
+		IO.mkdirs(testDir);
+	}
+
+	@Test
+	public void testSimpleExtension() throws Exception {
+		IO.copy(IO.getFile("testresources/ws-workspaceextensions-simple"), testDir);
```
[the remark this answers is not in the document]
juergen-albert: Will not work, because I need to copy different setups for different tests methods.

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -185,11 +185,13 @@ public void close() {
 	public Map<String, Container>								unreferencedClasspathEntries	= new HashMap<>();
 	public ProjectInstructions									instructions					= getInstructions(
 		ProjectInstructions.class);
+	private WorkspaceExtensionHandler							workspaceExtensionHandler;
 
 	public Project(Workspace workspace, File unused, File buildFile) {
 		super(workspace);
 		this.workspace = workspace;
 		setFileMustExist(false);
+		workspaceExtensionHandler = new WorkspaceExtensionHandler(this);
```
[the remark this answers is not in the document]
juergen-albert: This is addressed [here](https://github.com/bndtools/bnd/pull/4679#discussion_r639716891)

juergen-albert:
I accidentally clicked on ready to review. How do I revert this?

juergen-albert:
@pkriens implemented this as libaries via #4796

PR #4697, opened 2021-06-04
[sync] Remove rebuild from workspace sync

### LINKED ISSUE #4695: Suspected deadlock with workspace sync
At startup, I had to wait a long time for the new workspace sync to complete.

At some point, a Load Repositories/Refresh Repositories was initiated. I suspect it was initiated when the first project in the workspace was built which had dependencies on other projects in the workspace, as it appeared that the refresh repositories was waiting to refresh the Workspace repository.

I suspect that these two tasks deadlocked each other - the build of that first project initiated (and was waiting for) the repository refresh, but the repository refresh couldn't proceed because the sync workspace task had the entire workspace locked.

Fortunately, the manual cancel for the "refresh repositories" task actually worked. This allowed the build (part of sync-workspace) to proceed (albeit with compile errors).

It appears that the sync workspace is directly invoking (and waiting for) a rebuild upon completion. I think it might be better if it didn't, but rather let Eclipse take care of that via its usual mechanisms, after the basic sync is completed. This would avoid the deadlock.

Others have complained about this auto-building as part of the sync, because it does not honour the setting of the "build automatically" flag. Removing the explicit rebuild from the sync task and allowing Eclipse's mechanisms to take care of it would also resolve this complaint.

kriegfrj:
Fixes #4695.

Does the following:

* Moved the sync component into `bndtools.core.services`. Doesn't change the functionality, but makes it more convenient to do live coding/debugging as the `bndtools.core.services` bundle can be restarted while Eclipse is still running.
* Made sure that the event listener callback is de-registered in `@Deactivate` (only important because of the previous point).
* Removed code that disables the autobuild.
* Removed code that does the refresh and build.
* Added cancellation checks to the bndCall `after` callbacks` and short-circuit their execution if it has been cancelled..

There still seems to be an issue on  newly-loaded workspace. as you get a "resource out of sync" problem if (eg) you try delete a project before refreshing it. It is possible I might need to put the refresh back in.

Seems to work and the autobuild takes care of rebuilding (i necessary).

### DIFF
--- a/bndtools.core.services/src/org/bndtools/core/sync/SynchronizeWorkspaceWithEclipse.java
+++ b/bndtools.core.services/src/org/bndtools/core/sync/SynchronizeWorkspaceWithEclipse.java
@@ -1,8 +1,10 @@
-package bndtools.central.sync;
+package org.bndtools.core.sync;
 
 import java.io.File;
 import java.io.IOException;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
@@ -10,14 +12,12 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.eclipse.core.resources.IProject;
-import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IWorkspace;
-import org.eclipse.core.resources.IWorkspaceDescription;
 import org.eclipse.core.resources.IWorkspaceRoot;
-import org.eclipse.core.resources.IncrementalProjectBuilder;
 import org.eclipse.core.resources.ResourcesPlugin;
-import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.OperationCanceledException;
+import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.core.runtime.jobs.Job;
 import org.osgi.service.component.annotations.Activate;
 import org.osgi.service.component.annotations.Component;
@@ -26,12 +26,13 @@ import org.osgi.service.component.annotations.Reference;
 
 import aQute.bnd.build.Project;
 import aQute.bnd.build.Workspace;
+import aQute.bnd.build.api.OnWorkspace;
 import aQute.bnd.osgi.Processor;
 import aQute.lib.concurrent.serial.TriggerRepeat;
 import aQute.lib.io.IO;
 import aQute.lib.watcher.FileWatcher;
-import bndtools.Plugin;
 import bndtools.central.Central;
+import bndtools.central.sync.WorkspaceSynchronizer;
 
 /**
  * Synchronizes changes in the bnd workspace with Eclipse. This will use a Java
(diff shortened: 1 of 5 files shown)


bndtools.core.services/src/org/bndtools/core/sync/SynchronizeWorkspaceWithEclipse.java
```diff
@@ -92,15 +92,15 @@ private void sync(Collection<Project> doNotUse) {
 					Thread.sleep(500);
 
 					Central.bndCall(after -> {
-						boolean refresh = false;
```
bjhargrave, line 95: Leave this.

bndtools.core.services/src/org/bndtools/core/sync/SynchronizeWorkspaceWithEclipse.java
```diff
@@ -92,15 +92,15 @@ private void sync(Collection<Project> doNotUse) {
...
 						for (Project model : workspace.getAllProjects()) {
 							IProject project = projects.remove(model.getName());
 							if (project == null) {
-								System.out.println("create " + model);
-								refresh = true;
```
bjhargrave, line 100: Leave setting refresh.

bndtools.core.services/src/org/bndtools/core/sync/SynchronizeWorkspaceWithEclipse.java
```diff
@@ -114,32 +114,16 @@ private void sync(Collection<Project> doNotUse) {
 							if (!project.hasNature(Plugin.BNDTOOLS_NATURE))
 								continue;
 
-							System.out.println("remove " + project);
-							refresh = true;
```
bjhargrave, line 118: Leave setting refresh.

bndtools.core.services/src/org/bndtools/core/sync/SynchronizeWorkspaceWithEclipse.java
```diff
@@ -114,32 +114,16 @@ private void sync(Collection<Project> doNotUse) {
...
+								if (!monitor.isCanceled()) {
+									WorkspaceSynchronizer.removeProject(project, monitor);
+								}
+							});
 						}
 
 						if (monitor.isCanceled())
 							return null;
 
-						if (refresh) {
```
bjhargrave, line 125: Leave this if but don't call clean on the projects.

bndtools.core.services/src/org/bndtools/core/sync/SynchronizeWorkspaceWithEclipse.java
```diff
@@ -114,32 +114,16 @@ private void sync(Collection<Project> doNotUse) {
...
-							for (Project p : workspace.getAllProjects()) {
-
-								if (monitor.isCanceled())
-									return null;
-
-								p.clean();
-							}
-
-							after.accept("refresh", () -> {
-								root.refreshLocal(IResource.DEPTH_INFINITE, monitor);
```
bjhargrave, line 136: I think the disabling of autobuilding and the forcing of a clean and then full build are the big issues. Leave the refreshLocal call here.

bjhargrave (commented):
Some suggested changes. Basically get rid of the cleaning and full building, but leaving the refreshLocal still. This should just have the auto builder do the right thing.

kriegfrj:
I tried leaving the refresh in - it does fix the issue with the projects complaining about being out-of-sync with the filesystem; however it slows the sync down noticeably. I thought we could do better.

Instead of indiscriminately refreshing the entire workspace at the end, the current implementation refreshes project-by-project. This allowed the refresh to move into `WorkspaceSynchronizer.createProject()`. The advantage of this was that `createProject()` already had a condition in it to check if the (Eclipse) project already exists, and makes no change to the (Eclipse) workspace in that case. If no change has been made to the Eclipse workspace, then it doesn't need to do the refresh. This optimisation saves a lot of time, because in common usage most of the bnd projects have already been imported into the Eclipse workspace.

I also fixed up the monitor usage to bring it into line with the best practices described here: https://www.eclipse.org/articles/Article-Progress-Monitors/article.html

Without automated regression tests it is hard to know for sure if I broke something. But it seemed to work on my system (no warnings about projects been out-of-sync). Let me know how you go.

bjhargrave (approved):
This is way better on both the Bnd and OSGi workspaces!

PR #4698, opened 2021-06-07
Allow defaultProjectName to be set from bnd_build

Sczuka:
The defaultProjectName is always overwritten with the current folder name, even if bnd_build is specified. 

Only overwrite defaultProjectName with current folder, if defaultProjectName is empty.


bjhargrave (changes requested):
I am not sure of the problem you are trying to solve. The purpose of defaultProjectName is when when no project name is specified. If you are in a subfolder of the rootProject, then that project become the default project. So I don't think I agree with this change.

Also, your PR must not contain merge commits. There are 10 commits in the PR when you only are looking to change a few lines in one file.

Sczuka:
Actually we got problems with directory names shortened from our Jenkins in a multibranch pipeline. The shortening of the directory resulted in a directory name starting with a '.' character, which is not a good name for the gradle _rootProject.name_.

Setting the gradle _rootProject.name_ by adding a _bnd_build_ line in the _gradle.properties_ should be possible in order to avoid build failures that are caused by "incorrect" folder names.

May be there is an other way to specify the gradle _rootProject.name_?

bjhargrave:
> May be there is an other way to specify the gradle _rootProject.name_?

You can set it in your settings.gradle file.

Sczuka:
I'm afraid, but I didn't manage to set the project name successfully, but the same problem will occur, I you checkout/ rename this bnd repository to a folder that starts with a '.' character e.g. '.bnd'.

Where would you set the _gradle.rootProject.name_?

bjhargrave:
I am sorry, but I still do not understand the actual failure. Can you put the error message and/or exception in here?



> Where would you set the _gradle.rootProject.name_?

As I stated above, you can edit settings.gradle to set the rootProject.name. See <https://docs.gradle.org/current/userguide/multi_project_builds.html> for an example.

Sczuka:
.. sure.

I checked this repository 'bntools/bnd' to a local folder named '.bndxxx' and called './gradlew clean'. The result is:

`Downloading https://services.gradle.org/distributions/gradle-7.0.2-bin.zip
..........10%...........20%...........30%..........40%...........50%...........60%..........70%...........80%...........90%..........100%

Welcome to Gradle 7.0.2!

Here are the highlights of this release:
 - File system watching enabled by default
 - Support for running with and building Java 16 projects
 - Native support for Apple Silicon processors
 - Dependency catalog feature preview

For more details see https://docs.gradle.org/7.0.2/release-notes.html

Starting a Gradle Daemon (subsequent builds will be faster)

FAILURE: Build failed with an exception.

* What went wrong:
A problem occurred configuring root project '.bndxxx'.
> The project name '.bndxxx' must not start or end with a '.'. Set the 'rootProject.name' or adjust the 'include' statement (see https://docs.gradle.org/7.0.2/dsl/org.gradle.api.initialization.Settings.html#org.gradle.api.initialization.Settings:include(java.lang.String[]) for more details).

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 4m 8s
`

bjhargrave:
OK.

So if I change the settings.gradle in the Bnd build to:

```diff
diff --git a/settings.gradle b/settings.gradle
index b686f960e..b6e17bc02 100644
--- a/settings.gradle
+++ b/settings.gradle
@@ -53,4 +53,6 @@ gradle.ext.bndWorkspaceConfigure = { workspace ->
        }
 }
 
+rootProject.name = '.bnd'
+
 apply plugin: 'biz.aQute.bnd.workspace'
```

I get this error:

```
FAILURE: Build failed with an exception.

* Where:
Settings file '/Users/hargrave/git/bnd/settings.gradle' line: 56

* What went wrong:
> The project name '.bnd' must not start or end with a '.'. Set the 'rootProject.name' or adjust the 'include' statement (see https://docs.gradle.org/7.0.2/dsl/org.gradle.api.initialization.Settings.html#org.gradle.api.initialization.Settings:include(java.lang.String[]) for more details).
```

So I change the line in settings.gradle to:

```diff
diff --git a/settings.gradle b/settings.gradle
index b686f960e..7b8974ac1 100644
--- a/settings.gradle
+++ b/settings.gradle
@@ -53,4 +53,6 @@ gradle.ext.bndWorkspaceConfigure = { workspace ->
        }
 }
 
+rootProject.name = 'bnd'
+
 apply plugin: 'biz.aQute.bnd.workspace'
```

And it then works. So you just  need to set `rootProject.name` in your settings.gradle file.

Sczuka:
Thank you very much BJ,

that was actually what I was looking for. 

I see there is much room for improving my gradle skills, and sorry for the disturbance.

PR #4717, opened 2021-06-25
[bndtools tester] Stop console closing prematurely

### LINKED ISSUE #4716: [bndtools tester] MAX_TEST_RUNS limit causing tester console to close
One of the most annoying things about continuous testing is that the console seems to randomly drop out while the process keeps running. This is annoying for two reasons:
1. It means you can't see the output anymore (the obvious);
2. You can't terminate the process. You have to go to a OS, find the right java process to kill, and kill it with a `kill` or equivalent (and hope that you've killed the right one so that you don't add to your pain).

Up until now I've been living with it, because I had assumed (based on discussions with other core members who had experienced similar things) that it was a problem with the core Eclipse. But on reflection it is suspicious that the dropout always seems to happen right at the time that the tester (in continuous mode) starts a new test run, so I decided to dig further.

My first attempt to track this down, I added code into the `TerminationListener` to trace when the `TERMINATED` process debug event was fired, launched Eclipse from the Bndtools workspace, and re-ran a test until the console dropped out. But I found that the console was dropping out without any terminate event being triggered - when I killed the process from the OS, the terminate event triggered as expected. So the issue was specifically with the console somehow rather than the process management.

So I added a `ConsoleListener` that dumps a stack trace when `consolesRemoved()` is called and repeated. I managed to capture this stack trace:

```
java.lang.Exception: Stack trace
	at java.lang.Thread.dumpStack(Thread.java:1336)
	at bndtools.launch.AbstractOSGiLaunchDelegate$1.consolesRemoved(AbstractOSGiLaunchDelegate.java:262)
	at org.eclipse.ui.internal.console.ConsoleManager$ConsoleNotifier.run(ConsoleManager.java:156)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.ui.internal.console.ConsoleManager$ConsoleNotifier.notify(ConsoleManager.java:174)
	at org.eclipse.ui.internal.console.ConsoleManager.fireUpdate(ConsoleManager.java:251)
	at org.eclipse.ui.internal.console.ConsoleManager.removeConsoles(ConsoleManager.java:233)
	at org.eclipse.debug.internal.ui.views.console.ProcessConsoleManager.removeProcess(ProcessConsoleManager.java:159)
	at org.eclipse.debug.internal.ui.views.console.ProcessConsoleManager.removeLaunch(ProcessConsoleManager.java:142)
	at org.eclipse.debug.internal.ui.views.console.ProcessConsoleManager.launchRemoved(ProcessConsoleManager.java:137)
	at org.eclipse.debug.internal.core.LaunchManager$LaunchNotifier.run(LaunchManager.java:429)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.debug.internal.core.LaunchManager$LaunchNotifier.notify(LaunchManager.java:416)
	at org.eclipse.debug.internal.core.LaunchManager.fireUpdate(LaunchManager.java:1027)
	at org.eclipse.debug.internal.core.LaunchManager.removeLaunch(LaunchManager.java:2202)
	at org.eclipse.jdt.internal.junit.model.JUnitModel.notifyTestRunSessionRemoved(JUnitModel.java:586)
	at org.eclipse.jdt.internal.junit.model.JUnitModel.addTestRunSession(JUnitModel.java:374)
	at bndtools.launch.OSGiJUnitLaunchDelegate$TestRunSessionAndPort.<init>(OSGiJUnitLaunchDelegate.java:154)
	at bndtools.launch.OSGiJUnitLaunchDelegate$TestRunSessionAndPort.<init>(OSGiJUnitLaunchDelegate.java:147)
	at bndtools.launch.OSGiJUnitLaunchDelegate$ControlThread.run(OSGiJUnitLaunchDelegate.java:199)
	at java.lang.Thread.run(Thread.java:748)
```

So it is the call to `JUnitModel.addTestRunSession()` that is ultimately removing the console.

On closer inspection, it seems that what `addTestRunSession()` is designed to do this: There is a JUnit plugin preference called `org.eclipse.jdt.junit.max_test_runs`, which defaults to 10. Once this is reached, it will remove any test sessions that are older than that, along with their corresponding console (but not their process).

https://github.com/eclipse/eclipse.jdt.ui/blob/2a65ddcc0089b55fa3b848076b5a7cf8652034f8/org.eclipse.jdt.junit.core/src/org/eclipse/jdt/internal/junit/model/JUnitModel.java#L351-L377

This kind of thing is happening because Eclipse's JUnit design did not anticipate this kind of use case - assuming that every test run was a freshly launched process.

Having identified this problem, it should be possible to find a workaround.

kriegfrj:
Prior to this fix, Eclipse's JUnit plugin would close the console after MAX_TEST_RUNS was reached. The default is 10, but this is user-configurable in the "History..." menu item of the JUnit viewpart. 

Losing the console during continuous testing was often a pain...

Fixes #4716.

### DIFF
--- a/bndtools.core/src/bndtools/launch/OSGiJUnitLaunchDelegate.java
+++ b/bndtools.core/src/bndtools/launch/OSGiJUnitLaunchDelegate.java
@@ -7,6 +7,9 @@ import java.net.ServerSocket;
 import java.net.Socket;
 import java.net.SocketException;
 import java.text.MessageFormat;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -20,6 +23,7 @@ import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.debug.core.DebugPlugin;
@@ -29,6 +33,7 @@ import org.eclipse.debug.core.ILaunchConfigurationWorkingCopy;
 import org.eclipse.debug.core.IStatusHandler;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.internal.junit.JUnitCorePlugin;
+import org.eclipse.jdt.internal.junit.JUnitPreferencesConstants;
 import org.eclipse.jdt.internal.junit.launcher.JUnitLaunchConfigurationConstants;
 import org.eclipse.jdt.internal.junit.model.TestRunSession;
 import org.eclipse.jdt.junit.TestRunListener;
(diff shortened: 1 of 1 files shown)


bjhargrave:
Thanks!

PR #4722, opened 2021-06-29
builder: Update githead macro to understand git worktrees

bjhargrave:


bjhargrave:
This enables the tests to pass when building in a git worktree.

Issue #4724, opened 2021-07-01
Find the NPE in TestTask.java


containerAnalyzer:
Hello,
Our static analyzer found a following potential NPE. We have checked the feasibility of this execution trace. It is necessary to defend this vulnerability to improve the code quality.

1. Return **null** to caller
https://github.com/bndtools/bnd/blob/bf1d4b6a0f1ba1f34aef5e534751af7a798d676f/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java#L445

2. Function **getProject** executes and stores the return value to **project** (**project** can be **null**)
https://github.com/bndtools/bnd/blob/bf1d4b6a0f1ba1f34aef5e534751af7a798d676f/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java#L200

3. Function **getProject** executes and stores the return value to **baseProject** (**baseProject** can be null)
https://github.com/bndtools/bnd/blob/bf1d4b6a0f1ba1f34aef5e534751af7a798d676f/biz.aQute.bnd.ant/src/aQute/bnd/ant/TestTask.java#L28

4. Assign **baseProject** to **runProject**.(**runProject** can be **null**)
https://github.com/bndtools/bnd/blob/bf1d4b6a0f1ba1f34aef5e534751af7a798d676f/biz.aQute.bnd.ant/src/aQute/bnd/ant/TestTask.java#L39

4. Function **add** executes and **projects** contains **null**
https://github.com/bndtools/bnd/blob/bf1d4b6a0f1ba1f34aef5e534751af7a798d676f/biz.aQute.bnd.ant/src/aQute/bnd/ant/TestTask.java#L48

5. Function **next** executes and stores the return value to **project** (**project** can be null)
https://github.com/bndtools/bnd/blob/bf1d4b6a0f1ba1f34aef5e534751af7a798d676f/biz.aQute.bnd.ant/src/aQute/bnd/ant/TestTask.java#L54

6. **project** is used as the 1st parameter in function **clear** (**project** can be **null**), which will leak to null pointer dereference
https://github.com/bndtools/bnd/blob/bf1d4b6a0f1ba1f34aef5e534751af7a798d676f/biz.aQute.bnd.ant/src/aQute/bnd/ant/TestTask.java#L55

ContainerAnalyzer

containerAnalyzer:
thank you for fixing this NPE

PR #4735, opened 2021-07-05
Fix one NPE in TestTask.java

containerAnalyzer:
Fix the NPE in Issue #4724


bjhargrave:
This does not seem a useful fix in that it ignore the larger issue where the basedir is not a Bnd project.

PR #4746, opened 2021-07-09
gradle: Kotlin dsl for BundleTaskExtension

bjhargrave:
Fixes #4699

PR #4747, opened 2021-07-11
Add Gradle Enterprise plugin to make --scan work

marcphilipp:
Due to how settings.gradle is written, using `--scan` does not work out
of the box. While reviewing #4746, I found build scans to be immensely
helpful and hence applied the Gradle Enterprise plugin manually. Since I
think this might help others working on the build, this commit persists
that configuration.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -11,9 +11,11 @@ buildscript {
 		maven {
 			url = uri(bnd_repourl)
 		}
+		gradlePluginPortal()
 	}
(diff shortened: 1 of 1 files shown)


bjhargrave:
Thanks for this. I did not know about this issue. I fixed it in a different way by avoiding early resolution of the classpath configuration and this then allows --scan to add the Gradle Enterprise plugin to the classpath configuration.

PR #4754, opened 2021-07-16
Introduced XML Resource Repository Differ

### LINKED ISSUE #4711: [Feature Request] Extend Differ to compare OBR XMLs
In many scenarios, people require to compare OBR XMLs directly but the current implementation supports only comparison between JARs comprising OBR XMLs. It would be really useful to extend the current implementation of `aQute.bnd.differ.DiffPluginImpl` for the support of OBR XML comparison.

amitjoy:
Fixes #4711

### DIFF
--- a/biz.aQute.bnd/src/aQute/bnd/main/XmlRepoDiffCommand.java
+++ b/biz.aQute.bnd/src/aQute/bnd/main/XmlRepoDiffCommand.java
@@ -1,0 +1,131 @@
+package aQute.bnd.main;
+
+import static aQute.bnd.service.diff.Delta.IGNORED;
+import static aQute.bnd.service.diff.Delta.UNCHANGED;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toMap;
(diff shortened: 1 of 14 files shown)


biz.aQute.bnd.util/src/aQute/bnd/unmodifiable/Maps.java
```diff
@@ -62,6 +62,12 @@ private Maps() {}
 			entry(k6, v6), entry(k7, v7), entry(k8, v8), entry(k9, v9), entry(k10, v10));
 	}
 
+	public static <K, V> Map<K, V> of(K k1, V v1, K k2, V v2, K k3, V v3, K k4, V v4, K k5, V v5, K k6, V v6, K k7,
```
bjhargrave, line 65: Lets not do this. Just use the existing methods like ofEntries.

amitjoy: DONE

pkriens:
@amitjoy will you fix this baseline error?

amitjoy:
@pkriens I have just bumped the versions

PR #4755, opened 2021-07-16
gradle: Configure mavenLocal publishing if maven.repo.local prop set

bjhargrave:


bjhargrave:
Not actually needed. The mavenLocal configuration in the main repositories section is used.

PR #4762, opened 2021-07-22
Tester logback

kriegfrj:
This a quick proof-of-concept to look at implementing a logging listener for the test output that logs the output over SLF4J/OSGi Log Service.

Comments welcome.


bjhargrave:
Seems ok to me. You'll need to rebase to resolve conflicts.

kriegfrj:
Which implementation is better - over SLF4J direct, or over the OSGi log service? Or should I just do two sub-bundles, one for each?

bjhargrave:
> Which implementation is better - over SLF4J direct, or over the OSGi log service? Or should I just do two sub-bundles, one for each?

I don't have a strong opinion. Perhaps you have one bundle which can be configured (perhaps via system properties) to activate either or even both. Would this bundle be part of the normal bndtools install? If so, then you probably do want only one bundle which defaults to having neither activated.

kriegfrj:
> > Which implementation is better - over SLF4J direct, or over the OSGi log service? Or should I just do two sub-bundles, one for each?
> 
> I don't have a strong opinion.

On reflection, I think that both could be useful in different circumstances. For example, the OSGi log service implementation requires v 1.4 of the log service, so if you're stuck on an earlier version of the log service you'd need to use the "SLF4J direct" implementation.

> Perhaps you have one bundle which can be configured (perhaps via system properties) to activate either or even both.

If they are in separate bundles, you can easily configure your application to activate one or the other or both simply by selecting which bundle(s) you deploy in your application. I don't see what we gain by combining them into one bundle? In fact, we add complexity because the uber-bundle would need additional code to understand & respond to the configuration options.

> Would this bundle be part of the normal bndtools install? If so, then you probably do want only one bundle which defaults to having neither activated.

That was not what I intended. The intent was that they would be available for people to include them (from Maven Central) in their own test deployments if they want them.

bjhargrave:
> That was not what I intended. The intent was that they would be available for people to include them (from Maven Central) in their own test deployments if they want them.

Interesting. So why aren't these bundles in the osgi-test project then? They would seem to be a much better fit there than in this project since they are not tooling but runtime testing support.

kriegfrj:
> Interesting. So why aren't these bundles in the osgi-test project then? They would seem to be a much better fit there than in this project since they are not tooling but runtime testing support.

This is a good question. I was in two minds myself. I settled for this repo as they have a close affinity with the Bnd junit-platform tester. But I won't take much convincing.

bjhargrave:
> But I won't take much convincing.

Then let me convince you :-) I would much prefer to see this in the osgi-test project.

PR #4770, opened 2021-07-26
fix overwrite option on -include instruction

juergen-albert:
According to the documentation in the code and in the general doc, an
-include starting with ~ must overwrite the original value. It does not
however does not. 

Added some documentation around it as well.


docs/_chapters/150-build.md
```diff
@@ -50,8 +50,9 @@ For example, the Maven plugin that is built-in to bndlib has an extension file c
 	# Plugin maven setup
 	#
 	-plugin.maven = aQute.bnd.plugin.maven.MavenPlugin
-	
-	
+
+
+​
```
stbischof, line 55: `  `some whitespaces here.  this block must not change

docs/_instructions/include.md
```diff
@@ -25,13 +25,13 @@ There are two possible options. The path/URL starts with a:
 
 	# Read an optional file in the user's home directory
 	-include ${java.home}/.xyz/base.bnd
-
+
```
stbischof, line 28: useless whitespaces also here

bjhargrave:
I would suggest the code is correct (since it has been this way forever) and that the doc is wrong. Thus only the doc needs to be fixed to match the behavior everyone has been using.

juergen-albert:
> I would suggest the code is correct (since it has been this way forever) and that the doc is wrong. Thus only the doc needs to be fixed to match the behavior everyone has been using.

Done

docs/_chapters/150-build.md
juergen-albert: hope this fixes this

docs/_instructions/include.md
juergen-albert: hope this fixes this

docs/_chapters/150-build.md
```diff
@@ -63,14 +63,18 @@ For example, the Maven plugin that is built-in to bndlib has an extension file c
 	testbin=target/test-classes
 	target-dir=target
 
-We will not explain this plugin here (you can find it in the plugin sections), it only illustrates here how it is possible to setup the environment for a specific optional functionality like a plugin.
+We will not explain this plugin here (you can find it in the [plugin sections}(/chapters/870-plugins.html)), it only illustrates here how it is possible to setup the environment for a specific optional functionality like a plugin.
```
bjhargrave, line 66: I think the markdown is wrong.  `[]` section ends in `}`.

docs/_chapters/150-build.md
```diff
@@ -63,14 +63,18 @@ For example, the Maven plugin that is built-in to bndlib has an extension file c
...
 
 If you create local extension files then you should use a prefix to identify this is your file, like:
 
 	com.acme-local.bnd
 
 It is possible to use file links to maintain these files in one place when you have many workspaces.
 
+### Extension File loading mechanism and order
+
+Extension files will be loaded in alphabetical order. If a files defines a property that was already defined in another previously loaded file or the `build.bnd`, the property will be asigned a namespace. `-plugin` in `/cnf/ext/test.bnd` will become `-plugin.test`. Please note that this mechansim is just a best guess curtesy and has its limits. If `-plugin.test` was already defined previously, the old value will be overwritten.
```
bjhargrave, line 76: spell check

PR #4772, opened 2021-07-28
[quickfix] Suggestions for CannotThrowType and UnhandledException

kriegfrj:
This error can cause if one of the exception's superclasses is not on the classpath and so the compiler cannot tell that it is an exception.


kriegfrj:
Also added checks for unbundled exception errors, which can be caused if the exception is a subclass of a missing unchecked exception.

Issue #4778, opened 2021-08-03
Bnd $permissions macro generates deprecated "export" instead of "exportonly"


chrisr3:
Observed that [Bnd 5.3.0](https://bnd.bndtools.org/releases/5.3.0/macros/permissions.html) still generates `export`, although `export` is apparently deprecated in favour of `exportonly` [according to the Javadocs](http://docs.osgi.org/javadoc/osgi.core/8.0.0/).

I also noticed that OSGi defines:
```
(org.osgi.framework.PackagePermission "java.*" "import")
```
as an "implied permission" that the framework is supposed to apply automatically. This _would_ explain why `${packages}` doesn't include it too... except that I still seem to need to add it myself anyway to get things working :thinking:.

bjhargrave:
I am going to close this issue. We discussed and decided it is better to continue to use the old "export" action as it is supported by old OSGi framework implementations and, while deprecated, supported by all new OSGi implementations because the code that cares is in the OSGi PackagePermission class which will never remove support for the "export" action.

PR #4780, opened 2021-08-04
permissions: Update macro to use exportonly action

bjhargrave:
This means the generated permissions clause for exports will not work
on OSGi Core R4V4.1 or earlier.

Fixes https://github.com/bndtools/bnd/issues/4778


bjhargrave:
I am wondering if this change is really worth it. It adds zero functionality and means the generated permissions don't work on OSGi Core R4V41 and earlier. `export` action may be deprecated but is supported and still works just fine. I am leaning towards not changing things.

bjhargrave:
I am going to abandon this change. We discussed and decided it is better to continue to use the old "export" action as it is supported by old OSGi framework implementation and, while deprecated, supported by all new OSGi implementations because the code that cares is in the OSGi PackagePermission class which will never remove support for the "export" action.

PR #4782, opened 2021-08-04
Changed StartLevelRuntimeHandler to use a SynchronousBundleListener t…

### LINKED ISSUE #4781: Launcher: Asynchronous nature of StartLevelRuntimeHandler can cause issues when used together with Felix File Install
I found that when I export an executable jar in which Felix File Install is deployed that in turn loads and starts bundles, sometimes the default startlevel from my bndrun file unexpectedly overrules the startlevel Felix File Install tries to set. 

I traced the issue to StartLevelRuntimeHandler using a regular BundleListener, which is notified asynchronously of the INSTALLED event triggered by the loading of a bundle by Felix File Install. When Felix File Install immediately after installing the bundle tries to set the startlevel, its effort of setting is _sometimes_ overridden by the StartLevelRuntimeHandler.

I have prepared and tested a fix that changes the BundleListener to a SynchronousBundleListener, which ensures that the startlevel set by StartLevelRuntimeHandler happens immediately upon a bundle being installed, and no longer inadvertently interferes with others influencing startlevel after bundle installation, such as Felix File Install.

glimmerveen:
…o prevent races between it, and others setting startlevel for new bundles

Fixes #4781 

Signed-off-by: Arnoud Glimmerveen <arnoud@glimmerveen.org>

### DIFF
--- a/aQute.libg/src/aQute/lib/startlevel/StartLevelRuntimeHandler.java
+++ b/aQute.libg/src/aQute/lib/startlevel/StartLevelRuntimeHandler.java
@@ -10,6 +10,7 @@ import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleEvent;
 import org.osgi.framework.Constants;
 import org.osgi.framework.FrameworkListener;
+import org.osgi.framework.SynchronousBundleListener;
 import org.osgi.framework.launch.Framework;
 import org.osgi.framework.startlevel.BundleStartLevel;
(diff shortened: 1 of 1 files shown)


pkriens:
I do not have an issue with this because it will make the behavior more consistent. However, I think having more than one controller of the bundle's start level is begging on your knees for problems :-)

glimmerveen:
Thanks for merging and the warning :)

PR #4785, opened 2021-08-09
IncludeResouce handles extras in jars paths

stbischof:
When using `-includeresource` Instructions to process files inside a jar, no extras Attributes had been processed.

Now it should handle:
`flatten:=BOOLEAN` - puts all files in the file-tree into one folder

`rename:=RENAME`  - maps the Path using a given recipe , Regex groups allowed on Folders and Files

@juergen-albert 

For your need: `-includeresource: newFoldername=@fromFolder/zipFile.zip!/InnerFolder1/InnerFolder2/*;rebase:=/");`

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
@@ -22,6 +22,7 @@ import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 import java.util.jar.Manifest;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -946,7 +947,7 @@ public class Builder extends Analyzer {
 		}
 
 		if (source.startsWith("@")) {
-			extractFromJar(jar, source.substring(1), parts.length == 1 ? "" : destination, absentIsOk);
+			extractFromJar(jar, source.substring(1), parts.length == 1 ? "" : destination, absentIsOk, extra);
 		} else if (extra.containsKey("cmd")) {
 			doCommand(jar, source, destination, extra, preprocess, absentIsOk);
 		} else if (extra.containsKey(LITERAL_ATTRIBUTE)) {
@@ -1274,8 +1275,11 @@ public class Builder extends Analyzer {
 
 	/**
 	 * Extra resources from a Jar and add them to the given jar.
+	 *
+	 * @param extra
 	 */
-	private void extractFromJar(Jar jar, String source, String destination, boolean absentIsOk)
+	private void extractFromJar(Jar jar, String source, String destination, boolean absentIsOk,
+		Map<String, String> extra)
 		throws ZipException, IOException {
 		// Inline all resources and classes from another jar
 		// optionally appended with a modified regular expression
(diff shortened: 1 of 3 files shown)


bjhargrave (commented):
Can you also update the doc file for includeresource?

stbischof:
@bjhargrave I added docs as requested

biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
```diff
@@ -1294,11 +1299,29 @@ private void extractFromJar(Jar jar, String source, String destination, boolean
 
 			error("Can not find JAR file '%s'", source);
 		} else {
+
+			Function<String, String> f = v -> v;
```
bjhargrave, line 1303: Perhaps we should use a more descriptive variable name than `f`.  Maybe `nameMapper`?

bjhargrave (changes requested):

docs/_instructions/includeresource.md
```diff
@@ -70,6 +70,20 @@ Wrapping often requires access to a JAR from the repository. It is therefore com
 
     -includeresource    @${repo;biz.aQute.bndlib}!/about.html
 
+### Unrolling with extras
+
+`flatten:=BOOLEAN` - puts all files in the file-tree into one folder
+
+    -includeresource new.package=@jar/file.jar!/META-INF/cxf/**;flatten:=true
+
+`rebase:=PATH` - rebases the path to the file to the given value, regex groups allowed on folders
```
bjhargrave, line 79: They are not really regex, they are instruction selectors. https://bnd.bndtools.org/chapters/820-instructions.html#selector

docs/_instructions/includeresource.md
```diff
@@ -70,6 +70,20 @@ Wrapping often requires access to a JAR from the repository. It is therefore com
...
+
+    -includeresource new.package=@jar/file.jar!/META-INF/cxf/*;rebase:=/a/
+
+`recipe:=RECIPE` - maps the path using a given recipe, regex groups allowed on folders and files
```
bjhargrave, line 83: They are not really regex, they are instruction selectors. https://bnd.bndtools.org/chapters/820-instructions.html#selector

I don't think `**` is appropriate in the example as it implies ant-style path matching but selectors are just globbing. That is, `*` can match `/`. So `(M*TA-I*F)` can match `Monkey/Foo/TA-IamOK/POOF`

bjhargrave (commented):
A couple of comments on the doc changes.

docs/_instructions/includeresource.md
stbischof: i hope that the description and test is now more clear.

biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
```diff
@@ -1294,11 +1298,33 @@ private void extractFromJar(Jar jar, String source, String destination, boolean
 
 			error("Can not find JAR file '%s'", source);
 		} else {
+			Function<String, String> nameMapper = v -> v;
+			if (isTrue(extra.get("flatten:"))) {
+				nameMapper = path -> Strings.getLastSegment(path, '/');
+			} else if (instr != null) {
+				if (extra.containsKey("rebase:")) {
+					/*
+					 * TODO This is broken. The string cannot be both a glob,
```
bjhargrave, line 1307: I rebased on master, made some changes and found that the `rebase` directive is broken.  See the comment here.

bjhargrave (changes requested):

stbischof:
I removed the `rebase`. That could be done with the `recipe` as well.

bjhargrave:
> I removed the `rebase`. That could be done with the `recipe` as well.

Thanks.

On the naming, could we pick a better name than `recipe`? Perhaps `rename`? I think `rename` would more concisely explain the purpose of the directive.

stbischof:
I remaned `recipe` to `recipe`

bjhargrave:
Thanks for all the work!

Issue #4087, opened 2020-05-21
[bndtools m2e] Jars continuously rebuilt when "refresh on access" set


kriegfrj:
This bug has been floating around for a while but it just got annoying enough for me to report it. @rotty3000 has seen it in action on my machine, but I've just managed to narrow down the root cause significantly.

It seems that Bndtools + m2e wants to continously re-run the "jar" goal for the project as soon as the jars become visible in the Project Explorer. It stops again once they are hidden. The attached screen capture demonstrates this.
![osgi-test-workspace-Eclipse-IDE-2020-05-21-11-19-07](https://user-images.githubusercontent.com/7224087/82515769-37928d00-9b58-11ea-9402-04016678686f.gif)

rotty3000:
Is there a project for me to test? I have fixed bugs in this area but have
not experienced this with recent bndtools versions.

kriegfrj:
The screen capture is from the `osgi-test` project.

If you're not already seeing it, then perhaps it's a Windows-specific thing?

gamerson:
hey @kriegfrj   do you have "refresh on access" preference turned on in `window > preferencese > general > workspace` ?

kriegfrj:
> hey @kriegfrj do you have "refresh on access" preference turned on in `window > preferencese > general > workspace` ?

@gamerson , yes I did - and when I turned it off, this continuous rebuilding behaviour stopped! Thanks for the pointer.

This is still a bug in some form or another though, isn't it? I'm not exactly sure what "refresh on access" is supposed to do but surely it should cause this?

gamerson:
Yes, there is still some threading of the needle that the bnd-m2e-configurator needs to do in order to prevent the endless loops.  See here: https://github.com/bndtools/bnd/blob/master/bndtools.m2e/src/bndtools/m2e/BndConfigurator.java#L179-L192

gamerson:
I have not been able to reproduce this issue on MacOS, even with the `osgi-test` workspace and `refresh on access` enabled.  There must be some other way to trigger this, but until we have a reproducible way to trigger it, there doesn't seem like much action we can take.

gamerson:
Moving to abeyance.  @kriegfrj  if you are able to help me reproduce on MacOS, I will gladly take another look.

bjhargrave:
Closed by #4789

PR #4789, opened 2021-08-12
fixed build loop in eclipse

juergen-albert:
We ran into the problem, that in certain situations a single project was build over and over again in eclipse when using the bndtools m2e connector.

it turned out, that a few things went wrong here. 
1. The bnd-maven-plugin supports two goals: bnd-process and bnd-process-tests. The `BndConfigurator`  however was only called for `bnd-process` and build the artifact and the test artifact. This is now separated.
2. We call the `maven-jar-plugin` after we updated all the decorators. This is something that m2e normally does not do and wasn't expecting. Thus the resulting jar comes as a surprise to eclipse, which causes the incremental builder to be called again and finally dribbles down to us. We never checked the delta that causes the build and blindly rebuild the jars. Thus the loops starts. In addition to that, we get a Delta in some cases, when e.g. the jar is opened or sometimes simply when the target folder becomes visible for the first time in the Explorer. It now checks if the delta only contains the suspicious content changes on our jars and will only build when other deltas are found. With this we may miss, if someone editors the jars somehow, but I would see this cases as negligible. It seems to work reliably for all the other cases.  

This should also improve the build performance for larger projects, as the build jar was triggered far to often during a full build.

### DIFF
--- a/bndtools.m2e/src/bndtools/m2e/BndConfigurator.java
+++ b/bndtools.m2e/src/bndtools/m2e/BndConfigurator.java
@@ -6,9 +6,9 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import org.apache.maven.plugin.MojoExecution;
@@ -17,7 +17,6 @@ import org.bndtools.api.ILogger;
 import org.bndtools.api.Logger;
 import org.bndtools.build.api.IProjectDecorator;
 import org.bndtools.build.api.IProjectDecorator.BndProjectInfo;
-import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IProject;
@@ -34,6 +33,7 @@ import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.jdt.core.IJavaModelMarker;
 import org.eclipse.m2e.core.MavenPlugin;
+import org.eclipse.m2e.core.embedder.ArtifactKey;
 import org.eclipse.m2e.core.embedder.IMaven;
 import org.eclipse.m2e.core.embedder.IMavenExecutionContext;
 import org.eclipse.m2e.core.internal.MavenPluginActivator;
@@ -56,6 +56,8 @@ import bndtools.launch.JUnitShortcut;
 
 public class BndConfigurator extends AbstractProjectConfigurator {
 
+	private static final String					ARTIFACT_PATTERN				= "%s-%s.%s";
+	private static final String					CLASSIFIER_ARTIFACT_PATTERN	= "%s-%s-%s.%s";
 	protected final static Class<JUnitShortcut> shortcutClazz;
 	static {
 		/*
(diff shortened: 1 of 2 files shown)


kriegfrj:
Is this in any way related to #4087?

rotty3000 (approved):
LGTM

rotty3000:
thanks for the fix @juergen-albert

bndtools.m2e/src/bndtools/m2e/BndConfigurator.java
```diff
@@ -163,50 +174,87 @@ public AbstractBuildParticipant getBuildParticipant(final IMavenProjectFacade pr
...
-							// One solution would be to find the output artifact
-							// jar and refresh ONLY that. However we
-							// have not been able to create the condition we
-							// are worried about so we are deferring any extra
-							// work on this until it's shown to be a
-							// real problem.
-							targetDir.refreshLocal(IResource.DEPTH_INFINITE, progress.newChild(1));
+						try {
+							SubMonitor progress = SubMonitor.convert(monitor, 3);
+							execJarMojo(projectFacade, progress.newChild(1, SubMonitor.SUPPRESS_NONE), isTest);
```
kriegfrj, line 183: I suggest using `split()` rather than `newChild()`. The main functional difference is that `split()` will efficiently do cancelation checks for you whereas `newChild()` will not do any cancelation checks. The only reason to use `newChild()` is if you need compatibility for Eclipse < 4.6 (which we don't).

There are a few other instances of `newChild()` here too. But this is an old issue so it might make sense to fix these in a new, dedicated PR.

timothyjward: If there are several places that this needs tidying up then perhaps a separate bug should be raised and fixed? I would be keen to get rid of this build loop ASAP!

juergen-albert: @kriegfrj Makes sense, but I would agree with Tim, to do this in a separate PR. I will take care of this specific instance after this PR is through.

stbischof: Same to me. This bug makes Eclipse with bnd hard to use in our , slightly bigger project.

juergen-albert:
@bjhargrave, @pkriens or @kriegfrj would any of you be so kind and hit the merge button on this? It would be helfful if this would be available through the sanpshot build.

bndtools.m2e/src/bndtools/m2e/BndConfigurator.java
```diff
@@ -56,6 +56,8 @@
 
 public class BndConfigurator extends AbstractProjectConfigurator {
 
+	private static final String					ARTIFACT_PATERN				= "%s-%s.%s";
```
bjhargrave, line 59: Pattern is misspelled.

bndtools.m2e/src/bndtools/m2e/BndConfigurator.java
```diff
@@ -56,6 +56,8 @@
...
+	private static final String					CLASSIFIER_ARTIFACT_PATERN	= "%s-%s-%s.%s";
```
bjhargrave, line 60: Pattern is misspelled.

juergen-albert: fixed

bndtools.m2e/src/bndtools/m2e/BndConfigurator.java
juergen-albert: fixed

juergen-albert:
Thanks heaps!

PR #4794, opened 2021-08-18
Improved error handler, if maven not setup properly

juergen-albert:
If not resolver or test plugin has been set up properly, we internally ran into an NPE. The new error handling avoid the Exception and give a proper message to the user with a suggestion on how to fix it.

### DIFF
--- a/bndtools.m2e/src/bndtools/m2e/MavenRunListenerHelper.java
+++ b/bndtools.m2e/src/bndtools/m2e/MavenRunListenerHelper.java
@@ -23,15 +23,17 @@ import org.eclipse.core.resources.IWorkspaceRoot;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.m2e.core.MavenPlugin;
 import org.eclipse.m2e.core.embedder.IMaven;
 import org.eclipse.m2e.core.project.IMavenProjectFacade;
 import org.eclipse.m2e.core.project.IMavenProjectRegistry;
 
 import aQute.bnd.build.Run;
-import aQute.bnd.maven.lib.configuration.Bndruns;
 import aQute.bnd.exceptions.Exceptions;
+import aQute.bnd.maven.lib.configuration.Bndruns;
 
 public interface MavenRunListenerHelper {
 
@@ -137,7 +139,13 @@ public interface MavenRunListenerHelper {
 	default MojoExecution getMojoExecution(IMavenProjectFacade projectFacade, String pluginId, String executionPrefix,
 		Predicate<MojoExecution> predicate, IProgressMonitor monitor) throws CoreException {
 		MavenProject mavenProject = getMavenProject(projectFacade);
-		List<String> tasks = getMavenProject(projectFacade).getPlugin(pluginId)
+		Plugin plugin = getMavenProject(projectFacade).getPlugin(pluginId);
+		if (plugin == null) {
+			IStatus status = new Status(IStatus.ERROR, "bndtools.m2e", IStatus.ERROR,
+				createErrorMessage(pluginId), null);
+			throw new CoreException(status);
+		}
+		List<String> tasks = plugin
 			.getExecutionsAsMap()
 			.keySet()
 			.stream()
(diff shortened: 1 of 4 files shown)


bndtools.m2e/src/bndtools/m2e/MavenRunListenerHelper.java
```diff
@@ -23,15 +23,18 @@
...
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.m2e.core.MavenPlugin;
 import org.eclipse.m2e.core.embedder.IMaven;
 import org.eclipse.m2e.core.project.IMavenProjectFacade;
 import org.eclipse.m2e.core.project.IMavenProjectRegistry;
+import org.eclipse.ui.statushandlers.StatusManager;
```
gamerson, line 33: hey @juergen-albert  this dependency on the `org.eclipse.ui` import is what is breaking the build in the PR.  We haven't yet had a dependency from `bndtools.m2e` onto a UI plugin.  We should probably remove this reference for something and instead use something from `org.eclipse.core`  or just do nothing at all (return null silently)

juergen-albert: The Idiot I am, I forgot to commit the bnd file...

The StatusManager is used to show a message to the user, to tell him what the issue is. Without it, resolving would not work and one has no clue what the issue is.

bndtools.m2e/src/bndtools/m2e/MavenRunListenerHelper.java
```diff
@@ -152,6 +163,21 @@ default MojoExecution getMojoExecution(IMavenProjectFacade projectFacade, String
...
 	}
 
+	default String createErrorMessage(String pluginId) {
+		StringBuilder builder = new StringBuilder("Maven Plugin ");
+		builder.append(pluginId);
+		builder.append(" could not be found in your projects maven setup. Please make sure you configured ");
+		builder.append(pluginId);
+		builder.append(" for the bndrun file you want to resolve. See ");
+		if ("bnd-testing-maven-plugin".equals(pluginId)) {
+			builder.append("https://github.com/bndtools/bnd/tree/master/maven/bnd-testing-maven-plugin");
```
bjhargrave, line 173: We must not have such urls in the source code. We don't want to have to change this code should we change things like repo names or branch names.

bjhargrave (changes requested):

bndtools.m2e/src/bndtools/m2e/MavenRunListenerHelper.java
juergen-albert: I see your point, but this is our official documentation for the maven plugins. Without it, the whole message would become rather long and unreadable, as I can't use any formating in the Staus.
I can build a Dynamic link, that uses our current version and points to a tag in the repo. Would this be a possible solution?

bjhargrave: I don't think that is still a good idea. It would be better to link to bnd.bndtools.org since we can control that url namespace. But better still is to put in enough information so the reader knows what to do to fix the issue.

bjhargrave: Even links to bnd.bndtools.org will become wrong as Bnd changes over time and the old code points to the latest docs. That is an issue with links into the git repo too.

juergen-albert: To my knowledge, the only documentation for the maven plugins is availalbe as readme.md on github.

juergen-albert: Okay, I removed the link and changed it to RTFM.

bndtools.m2e/src/bndtools/m2e/MavenRunListenerHelper.java
gamerson: I would like to propose a different way, other than having bndtools.m2e rely on UI apis.  This method `getMojoExecution` already throws CoreException.  `CoreException` can have a Status object where you can put your message.  Then further up the call stack, we can have CoreException be caught by a bndtools plugin that already interacts with the UI, namely, `LaunchUtils.createRun`.  That method could then catch the CoreException and grab the status and call the StatusHandler.handle() method.

juergen-albert: good point!

gamerson (approved):
LGTM!

bjhargrave (approved):

PR #4796, opened 2021-08-19
Library

pkriens:
Allows the extensions of workspace, project,
or bndrun properties by using the -library
instruction.

See library.md for a description

@juergen-albert this is the code I wrote, I guess it was a rewrite. I do not support nesting, I think that has too many potential error cases. I separated the caching from the rest of the code. Let me know when you have time to go over this.


stbischof:
Could These library bundles have requirements? And are they "activated" if not all requirements are resolved in the workspace?

pkriens:
> Could These library bundles have requirements? And are they "activated" if not all requirements are resolved in the workspace?
No, they cannot now. What would be your use case? It should not be too hard to resolve them although this might add significant time to the startup. Which can of course be mitigated with more caching. However, can't think of a useful use case?

juergen-albert:
@pkriens Looks good at first glance. My holiday starts tomorrow and I can only do a more thorough review afterwards. I hope you can live with it, that I can do it in the week of the 6th of September.

stbischof:
since there is no nesting, it might be that a library bundle expects that an other library exists.
EMF codegen-generator needs a repository with all dependencies.

pkriens:
Hmm, in that use case I think you need one library or have several libraries in the same bundle? I think there is a significant increase in complexity. The cardinality between library (many per bundle) and bundle (host of the requirements) seems easy becoming confusing.

Issue #4805, opened 2021-08-25
[bndtools] Use `split()` rather than `newChild()`


bjhargrave:
I suggest using `split()` rather than `newChild()`. The main functional difference is that `split()` will efficiently do cancelation checks for you whereas `newChild()` will not do any cancelation checks. The only reason to use `newChild()` is if you need compatibility for Eclipse < 4.6 (which we don't).

There are a few other instances of `newChild()` here too. But this is an old issue so it might make sense to fix these in a new, dedicated PR.

_Originally posted by @kriegfrj in https://github.com/bndtools/bnd/pull/4789#discussion_r689931971_

kriegfrj:
Updated the title as it is not only the m2e plugin that has this issue.

PR #4818, opened 2021-09-06
replaced the usage of newChild with split as per #4805

juergen-albert:


kriegfrj (approved):
LGTM!

PR #4820, opened 2021-09-06
[reporter] remove Feature-coordinates from Coordinate

### LINKED ISSUE #3923: [reporter] Move Sling Feature to a dedicated subsection in the readme template
The PR #3916 introduces a new Coordinate subsection named Feature-Coordinate. To improve the readability of the readme file I think we should move this subsection into another main section.

The Coordinate section contains the identity of the artifact being documented by the readme file. Actually it contains two identity under two different systems: OSGi and Maven. The Feature-Coordinate is not an identity, it is a snippet to reference an identity in a Sling Feature. So I suggest to  move the newly Feature-Coordinate subsection to another main section named **Reference** and rename the  Feature-Coordinate subsection to **For Sling Feature**. We can also create in the same way other subsections **For Maven** , For Gradle, .... All these subsections should be optional.

The classifier tag should also be moved to the Reference section.

stbischof:
fixes #3923

### DIFF
--- a/biz.aQute.bnd.reporter/src/biz/aQute/bnd/reporter/plugins/transformer/templates/blocks.twig
+++ b/biz.aQute.bnd.reporter/src/biz/aQute/bnd/reporter/plugins/transformer/templates/blocks.twig
@@ -71,12 +71,6 @@
 {{ printer.printChecksum(report.checksum) }}
 {%- endif -%}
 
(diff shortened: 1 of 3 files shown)


bjhargrave (approved):

bjhargrave:
Thanks!

PR #4821, opened 2021-09-07
[maven] allow bnd-maven-plugin to be used for test bundles without requiring a main src/artifact

### LINKED ISSUE #4350: [maven] allow bnd-maven-plugin to be used for test bundles without requiring a main src/artifact
There are cases where you want to isolate test code in a separate module. The issue is that you then have to break all the test patterns such as using `test` scope, `maven-jar-plugin:test-jar`, `bnd-maven-plugin:bnd-process-tests`, etc.

This feels wrong and has unwanted side effects; put all the test code into src/main/java, accidentally publishing of "test" artifact, etc. 

The only thing that should be required is for both `bnd-maven-plugin` and maven-jar-plugin to have their default mojo's configured with `skipIfEmpty`. maven-jar-plugin already supports this but `bnd-maven-plugin` causes this to break because it always generates a manifest even if the bundle is empty (so adding `skipIfEmpty` here would solve that).

Also the `bnd-maven-plugin`'s `bnd-process-tests` mojo has a hard dependency on their being a `target/classes` dir in the classpath, which should be made optional (if it doesn't exist).

rotty3000:
Fixes #4350


maven/bnd-maven-plugin/README.md
```diff
@@ -82,6 +82,7 @@ For further usage information, see the integration test projects under the inclu
 |`warOutputDir`            | The directory where the `bnd-maven-plugin` will extract it's contents when packaging is `war`. _Defaults to `${project.build.directory}/${project.build.finalName}`._|
 |`packagingTypes`                | The list of maven packaging types for which the plugin will execute. *Defaults to `jar,war`*. Override with property `bnd.packagingTypes`. |
 |`skip`                 | Skip the project. _Defaults to `false`._ Override with property `bnd.skip`.|
+|`skipIfEmpty`         | Skip processing if `includeClassesDir` is `true` and the `${project.build.outputDirectory}` is empty. _Defaults to `false`._|
```
bjhargrave, line 85: Should we also mention the property name here in the doc?

maven/bnd-maven-plugin/README.md
```diff
@@ -242,6 +243,7 @@ The `bnd-process-tests` is not executed by default, therefore at least one expli
 | `skip`                  | Skip the project. _Defaults to `false`._ Override with property `bnd-tests.skip`. |
 | `artifactFragment`      | If true, make the tests artifact a fragment using `${project.artifactId}` as the `Fragment-Host` header and setting the `Bundle-SymbolicName` of the tests artifact to `${project.artifactId}-tests`. *Defaults to `false`*. |
 | `testCases`             | Specify the filter that will determine which classes to identify as test cases. *Defaults to `junit5`*. |
+|`skipIfEmpty`         | Skip processing if `includeClassesDir` is `true` and the `${project.build.testOutputDirectory}` is empty. _Defaults to `false`._|
```
bjhargrave, line 246: Should we also mention the property name here in the doc?

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
```diff
@@ -103,6 +107,9 @@
 	@Parameter(property = "bnd.packagingTypes", defaultValue = PACKAGING_JAR + "," + PACKAGING_WAR)
 	List<String>			packagingTypes;
 
+	@Parameter(property = "skipIfEmpty", defaultValue = "false")
```
bjhargrave, line 110: Is this a good property name? Is `skipIfEmpty` not also used by other plugins? Perhaps we should be `bnd.skipIfEmpty` as the property name?

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
```diff
@@ -154,6 +161,14 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 			return;
 		}
 
+		File classesDir = getClassesDir();
+
+		if (includeClassesDir && isEmpty(classesDir) && skipIfEmpty) {
```
bjhargrave, line 166: I would test `skipIfEmpty` first to possibly avoid the work of `isEmpty(classesDir)`.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
```diff
@@ -649,6 +664,25 @@ protected void reportErrorsAndWarnings(Builder builder) throws MojoFailureExcept
 		}
 	}
 
+	protected boolean isEmpty(File directory) {
+		if (directory == null || !directory.exists()) {
```
bjhargrave, line 668: Perhaps this should be `!directory.isDirectory()`?  Otherwise this method will return `false` if `directory` is an existing file.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
```diff
@@ -649,6 +664,25 @@ protected void reportErrorsAndWarnings(Builder builder) throws MojoFailureExcept
...
+			return true;
+		}
+
+		Path path = directory.toPath();
+		if (Files.isDirectory(path)) {
+			try (Stream<Path> entries = Files.list(path)) {
```
bjhargrave, line 674: Should this test care about empty directories? That is, what if `directory` contains a tree of directories but zero files? What is the definition of "empty"?

bjhargrave (commented):
I am not sure the second commit is worthwhile. Better to keep the tests simple by minimizing inheritance from the parent.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
rotty3000: what consists of an empty jar?

bjhargrave: I would say that a non-empty directory must have at least one file in the tree. Just a tree of empty directories is an empty directory. As for a jar, an empty jar is the same as an empty directory with the addition that `META-INF/MANIFEST.MF` does not count as a file in the tree. That is, if the only file in the jar is `META-INF/MANIFEST.MF`, then the jar is empty.

bjhargrave (approved):

bjhargrave:
Please make sure to note this enhancement in the release wiki page!

rotty3000:
added to the wiki!

PR #4823, opened 2021-09-08
Fix oneshot rerun

### LINKED ISSUE #4813: bndtools: Regression in re-running OSGi JUnit tests
Something about test rerunning is wrong in Bndtools now. After running an OSGi JUnit test, you used to be able to click the rerun button to run the test again. Now that does nothing. If you select the test name in the Junit view and click Run or Debug, you get a strange dialog. This is regression from 5.3

https://user-images.githubusercontent.com/277682/131337433-f40941e5-c2a0-4081-bc9c-9940e4e7733a.mov

This is a real productivity drop since I have to go back to the explorer view to find the test project and do Run OSGi JUnit again…

kriegfrj:
Fixes #4813.

### DIFF
--- a/bndtools.test/tester/main-with-runners/src/test/MyTest.java
+++ b/bndtools.test/tester/main-with-runners/src/test/MyTest.java
@@ -8,40 +8,45 @@ import junit.framework.AssertionFailedError;
 
 public class MyTest {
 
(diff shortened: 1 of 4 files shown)


bjhargrave (approved):
Seems OK to me.

PR #4828, opened 2021-09-09
[external plugin] Support external plugins in -plugin

pkriens:
Until now, the -plugin only accepted concrete classes. 
This patch allows you to register an interface with
an optional glob for an external plugin name.

After the plugins are initialized, the workspace will
try to find external plugins for the interfaces in the
-plugin. It will link these as proxies. If the proxy
is used, it will load the plugin. 

Closable and Plugin interfaces are supported. For
Plugin, the attributes are used for setProperties.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
@@ -509,6 +509,7 @@ public class Workspace extends Processor {
 		if (doExtend(this)) {
 			super.propertiesChanged();
 		}
+		postPropertiesChanged(this);
 		forceInitialization();
 	}
 
@@ -1830,11 +1831,6 @@ public class Workspace extends Processor {
 	 * @return true if the properties were changed
 	 */
 	public boolean doExtend(Processor processor) {
-		PluginsContainer plugins = processor.getPlugins();
-		if (!plugins.getInterfaces()
-			.isEmpty()) {
-			getExternalPlugins().setAbstractPlugins(processor, plugins);
-		}
 		String libraries = processor.mergeLocalProperties(Constants.LIBRARY);
 		if (Strings.nonNullOrEmpty(libraries)) {
 			data.libraryHandler.get()
@@ -1842,6 +1838,21 @@ public class Workspace extends Processor {
 			return true;
 		} else
 			return false;
+	}
+
+	/**
+	 * Called by Workspace, Project, and Run after all properties are
+	 * initialized and the plugins container is reset. This must be called so
+	 * that the Plugins Container is not reset after this call, which was the
+	 * problem with doExtend
+	 */
+
+	public void postPropertiesChanged(Processor processor) {
+		PluginsContainer plugins = processor.getPlugins();
+		if (!plugins.getInterfaces()
+			.isEmpty()) {
+			getExternalPlugins().setAbstractPlugins(processor, plugins);
+		}
 	}
 
 	/**
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
@@ -1830,6 +1830,11 @@ public class Workspace extends Processor {
 	 * @return true if the properties were changed
 	 */
 	public boolean doExtend(Processor processor) {
+		PluginsContainer plugins = processor.getPlugins();
+		if (!plugins.getInterfaces()
+			.isEmpty()) {
+			getExternalPlugins().setAbstractPlugins(processor, plugins);
+		}
 		String libraries = processor.mergeLocalProperties(Constants.LIBRARY);
 		if (Strings.nonNullOrEmpty(libraries)) {
 			data.libraryHandler.get()
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
@@ -509,7 +509,6 @@ public class Workspace extends Processor {
 		if (doExtend(this)) {
 			super.propertiesChanged();
 		}
-		postPropertiesChanged(this);
 		forceInitialization();
 	}
 
@@ -1838,21 +1837,6 @@ public class Workspace extends Processor {
 			return true;
 		} else
 			return false;
-	}
-
-	/**
-	 * Called by Workspace, Project, and Run after all properties are
-	 * initialized and the plugins container is reset. This must be called so
-	 * that the Plugins Container is not reset after this call, which was the
-	 * problem with doExtend
-	 */
-
-	public void postPropertiesChanged(Processor processor) {
-		PluginsContainer plugins = processor.getPlugins();
-		if (!plugins.getInterfaces()
-			.isEmpty()) {
-			getExternalPlugins().setAbstractPlugins(processor, plugins);
-		}
 	}
 
 	/**
--- a/biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceExternalPluginHandler.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceExternalPluginHandler.java
@@ -4,24 +4,34 @@ import java.io.File;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PrintWriter;
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import org.osgi.framework.VersionRange;
 import org.osgi.resource.Capability;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import aQute.bnd.exceptions.Exceptions;
 import aQute.bnd.exceptions.FunctionWithException;
 import aQute.bnd.header.Attrs;
 import aQute.bnd.header.Parameters;
+import aQute.bnd.memoize.Memoize;
+import aQute.bnd.osgi.PluginsContainer;
 import aQute.bnd.osgi.Processor;
 import aQute.bnd.osgi.resource.MainClassNamespace;
 import aQute.bnd.result.Result;
+import aQute.bnd.service.Plugin;
 import aQute.bnd.service.externalplugin.ExternalPluginNamespace;
 import aQute.bnd.service.progress.ProgressPlugin.Task;
 import aQute.bnd.service.progress.TaskManager;
@@ -31,8 +41,10 @@ import aQute.lib.strings.Strings;
 import aQute.libg.command.Command;
 
 public class WorkspaceExternalPluginHandler implements AutoCloseable {
+	final static Logger						logger	= LoggerFactory.getLogger("aQute.bnd.build");
 
-	final Workspace workspace;
+	final Workspace							workspace;
+	final Map<Capability, URLClassLoader>	loaders	= new HashMap<>();
 
 	WorkspaceExternalPluginHandler(Workspace workspace) {
 		this.workspace = workspace;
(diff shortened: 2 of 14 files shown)


pkriens:
@juergen-albert I think this is what you were looking for. There are a number of test cases but I've not had time to use it in practice. However, it should be safe since the code only runs if you put an interface in -plugin, which would blow up before.

biz.aQute.bndlib/src/aQute/bnd/osgi/PluginsContainer.java
```diff
@@ -378,6 +382,12 @@ private Object loadPlugin(Processor processor, ClassLoader loader, Attrs attrs,
 		return null;
 	}
 
+	public void addCloseable(Object plugin) {
```
bjhargrave, line 385: This breaks the thread safety. `plugins` is a copy-on-write set, so it is safe to add plugins after the container is initialized.  But `closeablePlugins` is a plain HashSet.

bjhargrave (commented):
The PR fails the build during the rebuild tests which make sure that the changes still allow us to build Bnd.

biz.aQute.bndlib/src/aQute/bnd/osgi/PluginsContainer.java
bjhargrave: https://github.com/bndtools/bnd/blob/ab6abc8987ebc7f33b80c10c6dc05fc476e3ee61/biz.aQute.bndlib/src/aQute/bnd/osgi/PluginsContainer.java#L50-L53

biz.aQute.bndlib/src/aQute/bnd/osgi/PluginsContainer.java
```diff
@@ -356,14 +360,15 @@ private Object loadPlugin(Processor processor, ClassLoader loader, Attrs attrs,
...
 		try {
 			Class<?> c = loader.loadClass(className);
-			Object plugin = publicLookup().findConstructor(c, defaultConstructor)
-				.invoke();
-			processor.customize(plugin, attrs, this);
-			add(plugin);
-			if (plugin instanceof AutoCloseable) {
-				closeablePlugins.add((AutoCloseable) plugin);
+			if (c.isInterface()) {
+				interfaces.put(c, attrs);
```
bjhargrave, line 364: If the type is an interface, this method returns `null`. This causes `loadPlugins` to try again using the pluginpath which will also return `null` since the interface is still an interface. If the plugin declaration has a `command` directive, then it will end up in the missing commands set. Wouldn't it be better to avoid the retry?

biz.aQute.bndlib/src/aQute/bnd/osgi/PluginsContainer.java
```diff
@@ -388,4 +399,9 @@ protected void close() {
 		plugins.clear();
 		missingCommand.clear();
 	}
+
+	public Map<Class<?>, Attrs> getInterfaces() {
+		return interfaces;
```
bjhargrave, line 404: For thread safety, this method should probably return an unmodifiable view of the map.

biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
```diff
@@ -1830,6 +1830,11 @@ public OnWorkspace on(String ownerName) {
 	 * @return true if the properties were changed
 	 */
 	public boolean doExtend(Processor processor) {
+		PluginsContainer plugins = processor.getPlugins();
+		if (!plugins.getInterfaces()
+			.isEmpty()) {
+			getExternalPlugins().setAbstractPlugins(processor, plugins);
+		}
```
bjhargrave, line 1837: In Workspace.propertiesChanged, it does:

https://github.com/bndtools/bnd/blob/879e8cdc025b59d1e62664805f02293cce824199/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java#L508-L511

Processor.propertiesChanged clears the plugsin:

https://github.com/bndtools/bnd/blob/879e8cdc025b59d1e62664805f02293cce824199/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java#L1021-L1027

So doesn't that discard the "abstract" plugins?

biz.aQute.bndlib/src/aQute/bnd/osgi/PluginsContainer.java
```diff
@@ -145,10 +212,6 @@ public boolean remove(Object plugin) {
 		return plugins().spliterator();
 	}
 
-	@Override
-	public Stream<Object> stream() {
```
bjhargrave, line 149: PluginContainer is a Set<Object> so it must have a valid `stream()` implementation. You removed this which means it falls back to the implementation in Collection which in turn depends on the spliterator method which does not handle the changes here. So the set contact is still wrong. You can't view all the plugins in the set that can be see by the `stream(Class)` method. All the set operations need to see all the plugins in the set.

biz.aQute.bndlib/src/aQute/bnd/osgi/PluginsContainer.java
```diff
@@ -47,17 +51,72 @@
...
+
 	// The following are only mutated during init(), so they don't need to
 	// be concurrent-safe
 	private final Set<String>			missingCommand		= new HashSet<>();
-	private final Set<AutoCloseable>	closeablePlugins	= new HashSet<>();
+	private Workspace					ws;
+	private Processor					processor;
+
+	public interface PluginProvider {
+		Stream<?> provide(Class<?> type);
```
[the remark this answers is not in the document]
kriegfrj: How about `<T> Stream <T> provide(Class<? extends T> type)`?

biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceExternalPluginHandler.java
```diff
@@ -196,6 +204,121 @@ public void abort() {
...
+				}, new InvocationHandler() {
+
+					@Override
+					public Object invoke(Object proxy, Method method, Object[] args) throws Throwable {
+
+						if (method.getDeclaringClass() == Object.class) {
+							return method.invoke(this, args);
+						}
+
+						if (method.getDeclaringClass() == AutoCloseable.class) {
```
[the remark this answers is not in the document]
pkriens: Not sure how useful this is? I think there are many possible changes where this will backfire since the new default method expects a value to be returned? I've changed it to invoke the method instead of calling close but this will fail if the delegate is not yet there or does not implement AutoCloseable

biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceExternalPluginHandler.java
```diff
@@ -196,6 +204,121 @@ public void abort() {
...
+			try {
+				Result<File> bundle = workspace.getBundle(cap.getResource());
+				if (bundle.isErr())
+					return bundle.asError();
+
+				URL url = bundle.unwrap()
+					.toURI()
+					.toURL();
+
+				urlClassLoader = new URLClassLoader(new URL[] {
```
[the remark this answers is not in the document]
pkriens: Makes sense. The file busy error on windows is the primary reason I hear people like Intellij

biz.aQute.bndlib/src/aQute/bnd/osgi/PluginsContainer.java
```diff
@@ -47,17 +51,72 @@
...
+					}
+				} else {
+					ws.error("failed to load external plugins for %s, attrs = %s", serviceClass, attrs);
+				}
+			}
+		}
+
+		@SuppressWarnings("unchecked")
+		@Override
+		public Stream<?> provide(Class<?> type) {
```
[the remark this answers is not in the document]
pkriens: Well, you need to give T another name and then cast unchecked. As said, generics suck when you code with dynamically generic ...

biz.aQute.bndlib/src/aQute/bnd/osgi/PluginsContainer.java
```diff
@@ -47,17 +51,72 @@
...
+		boolean			inited;
+
+		AbstractPlugin(Class<T> type, Attrs attrs) {
+			serviceClass = type;
+			this.attrs = attrs;
+		}
+
+		@SuppressWarnings("unchecked")
+		void init() {
+			if (!inited) {
```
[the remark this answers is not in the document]
pkriens: You're right but I do not think AtomicBoolean is correct, it needs to be synchronized or you can get duplicates. I recall I had it synchronized then reasoned it wasn't needed, then made changes and now it can be called on any thread.

biz.aQute.bndlib/src/aQute/bnd/osgi/PluginsContainer.java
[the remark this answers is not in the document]
kriegfrj: Admittedly it is a little untidy, but surely less untidy than the alternative - which is to force that untidiness and unsafe casting onto every caller of the method.

biz.aQute.bndlib/src/aQute/bnd/osgi/PluginsContainer.java
```diff
@@ -40,24 +45,112 @@
 import aQute.libg.cryptography.SHA1;
 
 /**
- * The plugin set for a Processor.
+ * The plugin set for a Processor. Plugins are general service objects and can
+ * be any type. The PluginsContainer treats the @{link {@link PluginProvider}
+ * special. If it is used with a type, it will expand the when it encounters a
+ * {@link ProviderType} plugin, see {@link #getPlugin(Class)} and
+ * {@link #getPlugin(Class)}. These provided plugins are _not_ part of the this
+ * set so they won't be explicitly returned when {@link #plugins()} is called.
```
bjhargrave, line 53: Right. So anyone who looks at the container as a `Set` (iterates, spliterates, contains, size, toArray, ...) will observe the PluginProvider. For example, the bnd `plugins` command iterates over the set and wont display the information on the actual plugins (it will display the toString of AbstractPlugin). The WorkspaceRepository tooltip will do the same.

Perhaps we can do better. The internal set can be `Set<PluginProvider<Object>>` where we wrap simple plugins in a provider. Then at all set method PluginContainer, we unwrap the PluginProvider? If you want to go this way, I can make a PR with this abstaction, and then you could rebase on that?

pkriens:
I am not sure this solves the lazy problem? The reason I do not want to expand the plugin providers at non-type specific operations that this will force a lot of work since the abstract plugins need to search repos. 

Since abstract plugins are new, I do not think we would cause any problems with existing code by treating them as factories? The AbstractPlugin provides a decent toString and we can improve on that if  needed?

Since I expect you're aware of my predicament, I am not sure I understand how you're addressing this? Anybody that gets all the plugins will force an expansion?

bjhargrave:
> I am not sure this solves the lazy problem? 

I think it will since I would only expand the PluginProvider when someone wants an actual plugin object. It also avoids leaking the abstration outside of the PluginContainer. 

> The reason I do not want to expand the plugin providers at non-type specific operations that this will force a lot of work since the abstract plugins need to search repos.
> 
> Since abstract plugins are new, I do not think we would cause any problems with existing code by treating them as factories? The AbstractPlugin provides a decent toString and we can improve on that if needed?
> 
> Since I expect you're aware of my predicament, I am not sure I understand how you're addressing this? Anybody that gets all the plugins will force an expansion?

Well if someone gets the set, that will not force an expansion since we would implement the Set operations. If someone iterates over the set, we would then "expands" to provide all the plugin objects. Otherwise the container does not really implement the set contract. For example, size() does not return the real value. It could be under or over (since the returned list could be empty.)

An alternate solution here is to break the API and change PluginContainer to not implement Set.

pkriens:
The set is as it is now with size and all its contractual obligations is well defined as far as I can see when the set contains the normal plugins and the Plugin Providers.

Removing the set inheritance or turning it into Set<PluginProvider> will have impact in the rest of the code, it is a pretty heavy breaking change.

The current proposal is also breaking but in a way that would be very hard to verify. You would have to somehow correlate the returned results from the typed `getPlugin` methods. This would require to know all possible types ahead of time and you need then some guarantee that nobody else is messing around with the container, since it can always get newer plugins or even lose.

I cannot imagine anyone being broken by this small change of semantics in the `getPlugin` methods?

I might be missing something but the very tiny change in the semantics of the `getPlugin` methods sounds like a less evil solution than removing the set or changing its type?

PR #4833, opened 2021-09-13
Introduced Remote Bundle Install UI Support

### LINKED ISSUE #4826: [Feature] Install/Update bundle directly from IDE to remote OSGi runtimes
We can introduce a feature which would enable us to remotely install or update a bundle from IDE directly. We can introduce a menu item which will only be activated after selecting a jar and we can select the remote IP where we want to install/update the bundle. The IPs of remote runtimes can also be stored in Eclipse preferences for continuous usage. 

I am currently starting to work on it.

amitjoy:
Fixes #4826


kriegfrj:
I haven't looked at this in detail, but one suggestion I have is for the project bsn to be `org.bndtools.remoteinstall` rather than than just `bndtools.remoteinstall`.

amitjoy:
@kriegfrj Thanks for your valuable feedback. I have just updated the namespaces to `org.bndtools.remoteinstall`.

stbischof:
will this work also with maven-workspace.layout?

amitjoy:
@stbischof Yes, it will work with any layout. The requirement is to choose a jar file to install or update.

PR #4834, opened 2021-09-13
fixes Project refresh after mavenBndRepo Update in View

### LINKED ISSUE #4320: [source] MavenBndRepository source download doesn't update projects
I had org.apache.cxf:cxf-core:3.3.1 in a MavenBndRepository.

I had a project with this bundle on its build path.

I attached sources by right-clicking in the repository view.

The sources downloaded and attached, creating "+cxf-core" in the appropriate directory.

However, my project didn't notice and continued linking to "cxf-core" (the bundle without the attached sources). I triggered a re-build of the Bnd dependencies by removing cxf-core from the build path and re-adding it to get it to link to the version of the bundle with attached sources. It would be nice if this were not necessary.

juergen-albert:
- fixes #4320

The issue was caused by the periodic file synchronizer that listens to changes on the index file. Its refresh kicked in during the action that updates the artifact und was not communicating any changes back to the Eclipse projects. 

This is not the most elegant fix. It would be better, if all Projects would register there `BndContainerInitializer`s as `RepositoryListenerPlugin` Services. Unfortunately, Registering wouldn't have been a problem, but I wasn't able to find a point to unregister the service again, if e.g. somebody removes the nature. 

Good thing now is: the `Bnd Bundle Path` will be updated if any of the actions in the maven repo context menu will be triggered and when the index file is edited (which wasn't the case before).

### DIFF
--- a/bndtools.core/src/bndtools/central/Central.java
+++ b/bndtools.core/src/bndtools/central/Central.java
@@ -143,7 +143,6 @@ public class Central implements IStartupParticipant {
 		repoListenerTracker = new RepositoryListenerPluginTracker(bundleContext);
 		repoListenerTracker.open();
 		internalPlugins.open();
-
 	}
 
 	@Override
@@ -588,12 +587,7 @@ public class Central implements IStartupParticipant {
 					refreshFile(file);
 				}
 
-				for (Project p : Central.getWorkspace()
-					.getAllProjects()) {
-					p.setChanged();
-					for (ModelListener l : getInstance().listeners)
-						l.modelChanged(p);
-				}
+				refreshProjects();
 
 				if (repoChanged) {
 					repositoriesViewRefresher.repositoriesRefreshed();
@@ -607,17 +601,29 @@ public class Central implements IStartupParticipant {
 	}
 
 	public static void refreshPlugin(Refreshable plugin) throws Exception {
-		if (plugin.refresh()) {
+		refreshPlugin(plugin, false);
+	}
+
+	public static void refreshPlugin(Refreshable plugin, boolean force) throws Exception {
+		boolean refresh = plugin.refresh();
+		if (refresh || force) {
 			refreshFile(plugin.getRoot());
-			for (Project p : Central.getWorkspace()
-				.getAllProjects()) {
-				p.setChanged();
-				for (ModelListener l : getInstance().listeners)
-					l.modelChanged(p);
-			}
 			if (plugin instanceof RepositoryPlugin) {
 				repositoriesViewRefresher.repositoryRefreshed((RepositoryPlugin) plugin);
+			} else {
+				// If the Plugin was a RepositoryPlugin, the ViewRefresher will
+				// trigger the Project update. If not, we have to do it ourself
+				refreshProjects();
 			}
+		}
+	}
+
+	public static void refreshProjects() throws Exception {
+		for (Project p : Central.getWorkspace()
+			.getAllProjects()) {
+			p.setChanged();
+			for (ModelListener l : getInstance().listeners)
+				l.modelChanged(p);
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


juergen-albert:
@bjhargrave Can we retrigger the build somehow? For some reason the build was canceled,

bjhargrave:
> For some reason the build was canceled,

@juergen-albert, They were cancelled since they hung. After running for 360 minutes, GitHub Actions decided they were hung builds. I restarted to try them again.

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -607,20 +601,31 @@ public static void refreshPlugins() throws Exception {
...
-				p.setChanged();
-				for (ModelListener l : getInstance().listeners)
-					l.modelChanged(p);
-			}
 			if (plugin instanceof RepositoryPlugin) {
 				repositoriesViewRefresher.repositoryRefreshed((RepositoryPlugin) plugin);
+			} else {
+				// If the Plugin was a RepositoryPlugin, the ViewRefresher will
+				// trigger the Project update. If not, we have to do it ourself
+				refreshProjects();
```
bjhargrave, line 615: What if multiple plugins refresh, then we call refreshProjects multiple times. If we are going to refresh the projects, we should do it once after all the plugins are refreshed not once for each plugin.

juergen-albert: The refreshPlugin method is only called by actions triggered in the UI. Thus they are sequential by nature. Who ever uses this method should be aware of what he is doing. 

The part where the Project refresh moved to the `RepositoriesViewRefresher` considers such a case and will only do it in the end, even if it is called multiple times (which might be a possibility).

bndtools.core/src/bndtools/central/RepositoriesViewRefresher.java
```diff
@@ -127,6 +127,12 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
 								busy = false;
 								if (redo) {
 									refreshRepositories(null);
+								} else {
+									try {
+										Central.refreshProjects();
+									} catch (Exception e) {
+										throw new RuntimeException(e);
```
bjhargrave, line 134: This should be `throw Exceptions.duck(e);` to avoid unnecessary wrapping.

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -607,20 +601,31 @@ public static void refreshPlugins() throws Exception {
 	}
 
 	public static void refreshPlugin(Refreshable plugin) throws Exception {
-		if (plugin.refresh()) {
+		refreshPlugin(plugin, false);
+	}
+
+	public static void refreshPlugin(Refreshable plugin, boolean force) throws Exception {
+		if (plugin.refresh() || force) {
```
bjhargrave, line 608: I am wondering if we should use `|` here instead of `||` to be explicit about no short-circuiting.

```java
if (plugin.refresh() | force) {
```

kriegfrj: Unless the JLS permits the arguments to be reordered (I don't think it does?) there doesn't seem to be any harm in short-circuiting.

bjhargrave: It is correct... now. But when someone later looks at this code and reorders the operands, this can subtly break things. `plugin.refresh()` must always be called. To be more explicit about this, the code could be:

```java
boolean refreshed = plugin.refresh();
if (refreshed || force) {
```

kriegfrj: Gotcha - but in that case, wouldn't a warning comment achieve the same thing?

bjhargrave: Sure. I guess I am asking for something...

bndtools.core/src/bndtools/central/RepositoriesViewRefresher.java
```diff
@@ -127,6 +128,12 @@ public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
 								busy = false;
 								if (redo) {
 									refreshRepositories(null);
+								} else {
+									try {
+										Central.refreshProjects();
+									} catch (Exception e) {
+										Exceptions.duck(e);
```
bjhargrave, line 135: This should be `throw Exceptions.duck(e);` for proper escape analysis by tools/compilers.

bjhargrave (changes requested):

PR #4843, opened 2021-09-15
EE inferencing fix

bjhargrave:
When mapping a version string to an EE, we handle when the version
string is just the minor version. That is, passing "8" will return the
EE for JavaSE-1.8. This helps when mapping javac --release arguments to
an EE since the --release arguments do not include the major version 1.

Without this fix, the EE inferencing in the maven and gradle plugins
would return the value for the current JDK  when --release 8 was used
since no EE for 8.0 exists.

Also some bnd file cleanup.


pkriens:
Reminds of the fact Microsoft never had Windows 9 .... :-)

PR #4859, opened 2021-09-23
replace JDK 16 by JDK 17

rotty3000:


bjhargrave:
Gradle 7.2 does not claim Java 17 support. I think we will need to wait for a new Gradle release which supports Java 17 before we can do this.

rotty3000:
sure, just playing with it.

PR #4862, opened 2021-09-27
Workaround fix for JVM Crash Issue while refreshing 'ProblemTreeViewer'

### LINKED ISSUE #4854: [Eclipse Bug] JVM Crash
Even though it is a SWT problem, I thought it might be a good idea to keep you informed about it so that we can track the resolution of this SWT bug.

Here is the link: https://bugs.eclipse.org/bugs/show_bug.cgi?id=576099

amitjoy:
This workaround ensures that the action is performed on the display thread properly by rescheduling it on the display thread. Note that, previously it was also executed on the display thread, but underlying SWT/JFace didn't submit the task properly to the display thread. That's why, this MR ensures that we submit the task properly (or explicitly) to the display thread. The bug still exists in the SWT/JFace which needs to be fixed. Other plugins using `getTreeViewer().refresh()` from `PackageExplorerPart` are prone to crash the JVM as well.

Fixes #4854

### DIFF
--- a/bndtools.core/src/bndtools/explorer/Model.java
+++ b/bndtools.core/src/bndtools/explorer/Model.java
@@ -103,12 +103,8 @@ class Model {
 
 	void update() {
 		dirty.set(true);
-		if (Display.getCurrent() == null) {
-			Display.getDefault()
-				.asyncExec(this::update0);
-		} else {
-			update0();
-		}
+		Display.getDefault()
+			.asyncExec(this::update0);
 	}
 
 	private void update0() {


bndtools.core/src/bndtools/explorer/BndtoolsExplorer.java
```diff
@@ -333,7 +333,8 @@ private void updateTreeViewer() {
 			installFilter();
 		}
 
-		getTreeViewer().refresh();
+		Display.getDefault()
```
bjhargrave, line 336: I don't think this is a proper fix. The issue is here:

https://github.com/bndtools/bnd/blob/1e894467543008cc16696932e0f154c7b052605f/bndtools.core/src/bndtools/explorer/Model.java#L106

`Display.getCurrent() != null` does not means that the current thread is the default Display thread. It just means the current thread is associated with _a_ Display object but not necessarily the default one.

I think the better fix is to remove the `Display.getCurrent() == null` and always call 

```java
Display.getDefault().asyncExec(this::update0);
```
or change the test to

```java
if (Display.getCurrent() != Display.getDefault()) {
```

bjhargrave (changes requested):

bndtools.core/src/bndtools/explorer/BndtoolsExplorer.java
amitjoy: @bjhargrave Yeah, I think, you are absolutely right. The problem seems to be in the model itself as it should always use a display thread to execute the update, whereas, the current thread might not be associated with a display thread at all.

I will quickly verify this and let you know.

amitjoy: I am hereby confirming that the proposed changes do fix the problem 👍

pkriens: I doubt this is true. What I recall is that you need to lookup for each graphic component what its Display object is, the default was just a convenient since it could require some work to figure out the appropriate objects. In this case, confusion cannot be the cause since it happens on macs where this is only a single thread for the UI and since it happens on single screens, only a single Display object. 

> The Eclipse framework itself is an example of an application that can have multiple windows - it still uses a single Display, with a single UI Thread and shared event system. Eclipse documentation on Threading Issues has a basic explanation of this:

bjhargrave: > I doubt this is true.

I looked at the Eclipse code. The default display is a single object in a static field which is returned by getDefault(). The current display is one of multiple Display objects in an array which is searched by looking for a Display whose thread field matches the current thread. So there is clearly potential that the current Display is not the same as the default Display. 

Another possibility is that the actions needed to be serialized with the currently running operation. That is, it is necessary to run them later once the current operation on the default Display thread completes. 

The fix in this PR addresses both possibilities, so it is a good fix. Thanks @amitjoy!

amitjoy: @bjhargrave Thanks to you for pointing it out 👍 It is always my pleasure to contribute to bndtools 😊

pkriens: >  So there is clearly potential that the current Display is not the same as the default Display.

Not in Eclipse ... and definitely not on the Mac with its single UI thread. Code is ok as far as I can see but the current code is a tad slower but current code is fine with me.

Issue #4863, opened 2021-09-27
bnd native Launcher with remote agent on maven project


stbischof:
I tried to start the native-launcher on an maven project to deploy bundles via remote agent.

i directly got 
![remoterunnative](https://user-images.githubusercontent.com/33224746/134936467-991c6a76-b7e4-4ca4-b2b5-26b82725b9a4.png)

when looking on the eclipse run configuration:

![remoteLaunch](https://user-images.githubusercontent.com/33224746/134936572-7fe0c440-abe0-46da-b954-bbf7fa3c2aea.png)

Project:
https://github.com/geckoprojects-org/org.eclipse.sensinact.gateway/tree/gpio
Run:
https://github.com/geckoprojects-org/org.eclipse.sensinact.gateway/blob/gpio/platform/protocol/dio/remote.bndrun

stbischof:
the first `Launch Problems`  Dialog still exists

Example Repo:
https://github.com/stbischof/stbischof-native.test.bundle

PR #4866, opened 2021-09-28
[launch] Made NativeLaunchTabGroup a component

kriegfrj:
In implementing #4703 I missed making the `NativeLaunchTabGroup` a component, which causes errors when you try and configure the launch group. This is directly responsible for the second error in #4863.

I'm not sure if it will fix the first error, but I am having trouble reproducing the workspace that @stbischof linked to in the issue. I've also never used the native launcher before so I don't know what normal behaviour looks like.


stbischof:
I can try to make a much simpler setup that should help.

kriegfrj:
That might be helpful @stbischof, thanks. In the meantime, if you are able to try out this PR and see if it fixes one or both of those problems, that would be much appreciated.

bjhargrave:
I will cherry-pick this fix onto next for the 6.0 release.

kriegfrj:
I'm keen to see if the other part of the problem is fixed as I think that any additional fix needs to go into 6.0.0 too.

Issue #4861, opened 2021-09-24
Bndtools: java.lang.IllegalStateException: Already closed MavenBndRepository 


bjhargrave:
I am now seeing this error a lot in Bndtools. I am wondering if it is caused by this recent change: https://github.com/bndtools/bnd/commit/f0928dab068431590c0a0c1abbbc454588e098f1#diff-866a6f249c64d2b346daf2384e0f74820b2e0dc66f45fad9e637f61601a8975c

```
Exception: java.lang.IllegalStateException: Already closed MavenBndRepository [localRepo=/Users/hargrave/.m2/repository, storage=Baseline, inited=true, redeploy=false]
	at aQute.bnd.repository.maven.provider.MavenBndRepository.init(MavenBndRepository.java:557)
	at aQute.bnd.repository.maven.provider.MavenBndRepository.versions(MavenBndRepository.java:543)
	at aQute.bnd.build.ProjectBuilder.getBaselineJar(ProjectBuilder.java:521)
	at aQute.bnd.build.ProjectBuilder.doBaseline(ProjectBuilder.java:319)
	at aQute.bnd.osgi.Builder.build(Builder.java:176)
	at aQute.bnd.osgi.Builder.builds(Builder.java:1447)
	at aQute.bnd.build.ProjectBuilder.builds(ProjectBuilder.java:841)
	at aQute.bnd.build.Project.buildLocal(Project.java:1895)
	at aQute.bnd.build.Project.build(Project.java:1708)
	at aQute.bnd.build.Project.build(Project.java:2485)
	at org.bndtools.builder.BndtoolsBuilder.lambda$build$5(BndtoolsBuilder.java:284)
	at bndtools.central.Central.lambda$bndCall$13(Central.java:819)
	at aQute.bnd.service.progress.TaskManager.with(TaskManager.java:37)
	at bndtools.central.Central.bndCall(Central.java:819)
	at org.bndtools.builder.BndtoolsBuilder.build(BndtoolsBuilder.java:165)
	at org.eclipse.core.internal.events.BuildManager$2.run(BuildManager.java:853)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:232)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:281)
	at org.eclipse.core.internal.events.BuildManager$1.run(BuildManager.java:334)
[...]

PR #4867, opened 2021-09-28
bndtools: Fix project refreshing causing performance issues

bjhargrave:
f0928dab068431590c0a0c1abbbc454588e098f1 causes performance issues and possibly #4861. The way all projects were refreshed and the refreshing of project from a view refresher caused repetitive project refreshing.

This PR fixes Central.refreshProjects() to mark all projects changed before advising model listeners of the changes. It also removes refreshing projects from the view refresher. Projects should be refreshed only when some action is taken not in a UI view refresher which is itself responding to some taken action. I saw a stack trace where the view refresher refreshed the projects which ended back up in the view refresher refreshing the projects again!

Also, we should not be refreshing projects on the UI thread as that causes responsiveness issues.

So far this PR seems to address recent performance issues I have had using the latest Bndtools.

### DIFF
--- a/bndtools.core/src/bndtools/Activator.java
+++ b/bndtools.core/src/bndtools/Activator.java
@@ -143,11 +143,10 @@ public class Activator extends AbstractUIPlugin {
 	}
 
 	static void async(Runnable run) {
-		if (Display.getCurrent() == null) {
-			Display.getDefault()
-				.asyncExec(run);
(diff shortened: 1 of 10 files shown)


kriegfrj:
It always *felt* like there was something going on here that made refreshing take a lot longer than it should. Here's hoping this fixes it!

PR #4869, opened 2021-09-29
[m2e native] Fix native launch delegate for m2e

### LINKED ISSUE #4868: [m2e] Native launch doesn't work in m2e
the first `Launch Problems`  Dialog still exists

remoterunnative

Example Repo:
https://github.com/stbischof/stbischof-native.test.bundle

This runs (see readme) with plain maven but not when starting inside the ide

_Originally posted by @stbischof in https://github.com/bndtools/bnd/issues/4863#issuecomment-929384704_

kriegfrj:
Changed `NativeBndLaunchDelegate` launch delegate so that it uses `LaunchUtils.createRun()`, which abstracts away the differences
between the Bnd workspace model and M2E model.

Fixes #4868.

### DIFF
--- a/org.bndtools.launch/src/bndtools/launch/bnd/NativeBndLaunchDelegate.java
+++ b/org.bndtools.launch/src/bndtools/launch/bnd/NativeBndLaunchDelegate.java
@@ -2,13 +2,12 @@ package bndtools.launch.bnd;
 
 import static bndtools.launch.LaunchConstants.PLUGIN_ID;
 
-import java.io.File;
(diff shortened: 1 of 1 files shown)


kriegfrj:
I've never really used the native launcher so I'm not sure if this will break anything in a subtle way. But it seemed to work when I used it with @stbischof's minimal example for the remote launcher.

rotty3000 (approved):
I think this is the correct approach because it will allow the m2e listeners to interact with the launch and setup the correct workspace/Run

PR #4881, opened 2021-10-06
makes maven project properties available to the bndrun

juergen-albert:
Currently bndrun files in maven have no access to any properties of the maven project. We have a Project with a couple of of tests that are housed in bndrun files. They all have a couple of duplicate properties (`-tester`, `-runproperties` etc.). We wanted to put this in a centralized file and include them from each file as we already do in the `bnd.bnd ` files. Long story short: it did not work, because we have a mojo that calculates the root folder and we need to access this property.

This is now fixed with this commit. It unfortunately does not work when the bndrun has a `-standalone` but I couldn't say why.

### DIFF
--- a/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
+++ b/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
@@ -194,7 +194,7 @@ public class BndrunContainer {
 		File cnf = new File(temporaryDir, Workspace.CNFDIR);
 		aQute.lib.io.IO.mkdirs(cnf);
 
-		Bndrun run = Bndrun.createBndrun(null, runFile);
+		Bndrun run = Bndrun.createBndrun(runFile, project.getProperties());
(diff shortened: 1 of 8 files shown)


bjhargrave (changes requested):
See https://github.com/bndtools/bnd/blob/47de3f900ad9b9a374a3d2c1ed3bd92f69ce3db8/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java#L204

It is already that case that the maven project properties are made available to the bndrun. So I am not sure why you think they are not. If it is not working, then we need to debug the existing code.

In any case, this code very wrong and cannot be used as is.

rotty3000:
I second @bjhargrave's argument that this already exists since I'm using that in many builds like this one: https://github.com/apache/aries-cdi/blob/master/cdi-executable/base.bndrun#L33-L36

rotty3000:
the part you reference is very important in order to have access to fields of the project and session objects, it lets you reference such values using dot or array notation.

bjhargrave:
Superseded by #4882.

PR #4892, opened 2021-10-14
Testing updates

bjhargrave:
Define new `InjectTemporaryDirectory` testing annotation to make it easy and consistent to use a temp dir in a test method.

### DIFF
--- a/aQute.libg/bnd.bnd
+++ b/aQute.libg/bnd.bnd
@@ -9,6 +9,7 @@
 	osgi.core;version=latest
 
 -testpath: \
+	biz.aQute.bnd.test;version=project,\
 	${junit},\
 	slf4j.simple;version=latest
 
--- a/aQute.libg/test/aQute/lib/index/TestIndex.java
+++ b/aQute.libg/test/aQute/lib/index/TestIndex.java
@@ -5,21 +5,15 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import java.io.File;
 
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.TestInfo;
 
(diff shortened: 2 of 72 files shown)


kriegfrj:
Is this a candidate for osgi-test?

bjhargrave:
> Is this a candidate for osgi-test?

Maybe. A couple of issues. 

1. To simplify the use in the Bnd build (a Bnd workspace model build), I default the base to `generated/tmp/test` where `generated` is the build folder and thus `generated/tmp/test` is the Gradle temp dir for the `test` task. But that folder name is not correct for all users of osgi-test. In a normal gradle build, `build` the build folder and `build/tmp/test` would be the temp dir name for the `test` task and `build/tmp/testOSGi` for the `testOSGi task` In a maven build it would be something more like `target/tmp/test` for surefire testing. 

So to do this nicely in osgi-test, we would need to know things we can't easily know like what is the build directory? `target`, `build`, `generated`? What task as we running in? `test`, `testOSGi`, maven surefire? And where would the build tool put the temp dir for that task so that the user will know the convention and be able to find the temp dir?

2. The extension impl currently does not clean up the temp dir on purpose. The clean task can do that since the temp folders are under the tool's build dir. But for an impl in osgi-test, we would want to have a post-clean option I would think (but this is more easily solved than the above problem). I like to keep the temp dir around to view what the test did especially if it fails since the temp dir contents can help in diagnosing the problem.

JUnit 5 already has a [`TempDir`](https://junit.org/junit5/docs/5.4.2/api/org/junit/jupiter/api/io/TempDir.html) annotation which can be used to make a temp dir which is cleaned up at the end of the test scope. This is probably what most people will be happy with.

PR #4898, opened 2021-10-16
DRAFT: Add DocSearch

danelowe:


danelowe:
There's a little bit of config to set up for the crawler if you're not wanting to add an index step to the build pipeline and not using one of their recommended platforms. I think they may set it up for you, or if they don't I'd be happy to sort it. 

This does mean if you are keen for a refresh, I'd maybe do that first. Not a big deal either way though.

bjhargrave:

PR #4907, opened 2021-10-21
[editors] Add action contributor to editors

### LINKED ISSUE #4433: [bndtools] Undo functionality not working
A few of us had noticed that the undo functionality in Bndtools on Eclipse had gone a git awry. Often (always?) it doesn't work at all.

I think we had assumed that it was an Eclipse bug (on Slack, https://bugs.eclipse.org/bugs/show_bug.cgi?id=564366 was floated as a candidate). But what I've observed doesn't seem to apply to editing Java files - only bnd.bnd files. But I just observed something that might suggest otherwise and seems possibly related to some other weird issues I've seen with bndrun file editing (see #4202 for example). Forgive me for the long-winded description here; I'm not sure if I'll be able to reproduce it so I want to experiment and capture everything I'm doing while I've still got the issue happening.

* I have two bnd.bnd files open in two different editor windows.
* One of them is in a project in my workspace (let us call it "A"). One of them is from a project that is not in my workspace ("B"). * * A and B were both open and dirty (though I'm not sure if they started out dirty). The were in separate windows on separate screens so I could see both at the same time.
* My cursor was in A. I made a change in A and attempted to undo it with Ctrl-Z.
* A did not undo; however I noticed that B changed.
* Further use of Ctrl-Z and Ctrl-Y showed that the undo/redo was being consistently applied to editor B, even though regular keystrokes were being consistently picked up by editor A.
* If I click in editor B and start typing, the regular keystrokes get picked up by editor B rather than editor A. Undo/redo continues to get picked up by editor B.
* If I click back to editor A and start typing, regular keystrokes return to editor A, but undo/redo continues to be applied to editor B.
* I open a third bnd.bnd file from a second workspace project ("C"). Edits now get sent to C; undo/redo continues to get sent to B.
* I save the changes to C; undo/redo still goes to B.
* I close A without saving changes and return editor focus to C; undo/redo still goes to B.
* I save B and return focus to C; undo/redo still goes to B.
* I close B without saving; undo/redo seems to go nowhere (or else is going to an editor that I haven't found yet).
* I re-open A and B and make edits in both. Undos in A again go to B.
* Close B and open another bnd.bnd file from a non-workspace project ("D"). Undo/redo now goes to project D.
* Open B again. Undo/redo goes to B. Now even when editing project D, undo/redo goes to project B.
* Close B. Undo/redo in editor D now seems to go nowhere.
* Open another workspace bnd.bnd file ("E"). All undo/redo now goes to this file. If I close this file, it goes nowhere.

So I think the summary from the above is: whenever you open a new Bnd editor, it seems to steal the "undo focus" from whichever Bnd editor had it. When you close that editor, the focus does not seem to return to any other Bnd editor.

This also explains why some of us had seen it and others not, or why we'd only seen it sometimes - if you freshly open a new file and start editing it, undo will work. But if you often have multiple editors open and switch between them, it won't work.

I just double checked and the same applies to Bndrun editors. There doesn't appear to be any cross-contamination between Bnd and Bndrun editors; it seems to remain "in the family".

It is possible that this is still an Eclipse issue; but it's also possible that we're doing something wrong with our working copy management that has exposed this bug. Not handling working copies properly could also cause other weird issues like #4202. Another possibility is a problem in BndEditModel somewhere.

I recall at some point it used to be the case that when you tried to open 
A couple of possibly-related bug fixes that may have contributed:
* Something that I've not been able to find a reference to - I remember at one point when you tried to open a bnd.bnd file that was already open in an editor, it would open a new editor for the same file. This is in contrast to JDT, which attempts to move focus to the editor that is already opened. Now, it seems opening bnd.bnd files also moves focus to the editor that's already open (if any). Was this an active fix from one of us (if so, the fix is possibly related to this issue)? Or was it a side-effect of some other change?
* #3773 (editor support for Bnd files outside of the workspace)

kriegfrj:
This seems to fix #4433 (among possible other things).

Huge shout out to @Gohla for this tip.

### DIFF
--- a/bndtools.core/_plugin.xml
+++ b/bndtools.core/_plugin.xml
@@ -60,22 +60,27 @@
 	<extension point="org.eclipse.ui.editors">
 		<editor class="bndtools.editor.BndEditor" default="false"
 			extensions="bnd" icon="${icons.bundle}" id="bndtools.bndEditor"
+			contributorClass="org.eclipse.ui.texteditor.BasicTextEditorActionContributor"
 			name="Bnd Bundle Editor">
 		</editor>
 		<editor class="bndtools.editor.BndEditor" default="false"
 			filenames="bnd.bnd" icon="icons/bricks.png"
+			contributorClass="org.eclipse.ui.texteditor.BasicTextEditorActionContributor"
 			id="bndtools.bndProjectEditor" name="Bnd Project Editor">
 		</editor>
 		<editor class="bndtools.editor.BndEditor" default="false"
 			extensions="bndrun" icon="${icons.bndrun}" id="bndtools.bndRunEditor"
+			contributorClass="org.eclipse.ui.texteditor.BasicTextEditorActionContributor"
 			name="Bnd Run File Editor">
 		</editor>
 		<editor class="bndtools.editor.BndEditor" default="false"
 			filenames="build.bnd,ext/*.bnd" icon="icons/bndtools-logo-16x16.png"
 			id="bndtools.bndWorkspaceConfigEditor"
+			contributorClass="org.eclipse.ui.texteditor.BasicTextEditorActionContributor"
 			name="Bnd Workspace Config Editor" />
 		<editor
 			class="org.bndtools.core.editors.pkginfo.PackageInfoEditor"
+			contributorClass="org.eclipse.ui.texteditor.BasicTextEditorActionContributor"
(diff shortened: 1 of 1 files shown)


kriegfrj:
> I am more inclined to just make a 6.1 release sooner rather than wait too long.

:+1:

Issue #4858, opened 2021-09-23
[6.1 DEV] Expanding search results in the Build Path Dialog doesnt work properly


maho7791:
Related to #4857, I further experienced that expanding search results doesn work anymore. 
OS Linux, Ubuntu 20.04 GTK3

To reproduce:
* Open Build Path Dialog
* Search for a string, I took *promise*
* I get the corresponding results, expanded correctly (e.g. org.osgi.util.promise)
* Leave the dialog open
* Replace the search string by e.g. *function*
* Nothing is expanded now, but you can find the search result *org.osgi.util.function* under the *OSGi R7 Reference Implementation* repository, but not expanded
* All other repositories indicate, that they have children, but they don't have

maho7791:
https://user-images.githubusercontent.com/8199528/140495630-b1371149-c01a-49f3-89a0-7e0080134547.mp4

maho7791:
https://user-images.githubusercontent.com/8199528/140499077-ac607f63-8fab-4948-9763-bc500fdcbb4e.mp4

maho7791:
https://user-images.githubusercontent.com/8199528/140499251-776b419f-a520-437d-b3e4-e2c616c99d21.mp4

maho7791:
I also tried it with the latest 6.1.0.202111050242-SNAPSHOT

PR #4929, opened 2021-11-03
#4858 Confusion in bndeditor/build/add dialog

pkriens:
See https://github.com/bndtools/bnd/issues/4858

### DIFF
--- a/aQute.libg/src/aQute/lib/collections/MultiMap.java
+++ b/aQute.libg/src/aQute/lib/collections/MultiMap.java
@@ -362,4 +362,20 @@ public class MultiMap<K, V> implements Map<K, List<V>> {
 		return map.toString();
 	}
 
+	/**
+	 * Index a collection.
+	 *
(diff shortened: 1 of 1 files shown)


kriegfrj:
What about if the ones that are already on the buildpath were greyed out? This is the convention used by Eclipse's "Import Project" dialog, for example - if the project already exists in the workspace, it will be greyed out. (There is also the option in the Eclipse dialog to filter out those that are already in the workspace, which might be a useful addition.) I suggest this because it might be confusing at first as to why the add button gets enabled or disabled. Just a thought.

pkriens:
I am disabling the button, I am still doing some work on this so maybe I find the courage to figure out how to selectively enable/disable entries.

kriegfrj:
If I convert the editor package to a facade, it will probably make your dev experience much more pleasant. Maybe then you'll find the necessary courage. 😄

PR #4931, opened 2021-11-03
gradle: Add support for user enabling use of Configuration Cache

bjhargrave:
We allow the user to configure properties which are available when
evaluating bnd instructions. The default uses the Project object
at task execution time which prevents Gradle from putting the
task in the configuration cache. The user can now configure the
properties value which prevents the default use of the Project object
at task execution time.


bjhargrave:
I still need to update the readme to dicuss this.

PR #4953, opened 2021-11-17
better naming for archives as part of a multi zip

juergen-albert:
The Maven repo can be configured to download an archive containing jars. These jars will then be added to the repos index. Currently they land in the local m2 repo with generically numbered names like: "groupid-artifactid-1.0.0-bin123.jar" 

When we export a bndrun, we add them with exactly that name. This makes it hard to identify what really is in there and also "looks bad".

Thus we now modify the classierfier to reflect the original jar name instead of the bin123.

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
@@ -17,10 +17,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Pattern;
 
 import org.osgi.resource.Resource;
 import org.osgi.util.promise.Deferred;
@@ -49,6 +51,7 @@ import aQute.bnd.version.MavenVersion;
 import aQute.bnd.version.Version;
 import aQute.lib.io.IO;
 import aQute.lib.strings.Strings;
+import aQute.lib.zip.ZipUtil;
 import aQute.maven.api.Archive;
 import aQute.maven.api.IMavenRepo;
 import aQute.maven.api.Program;
@@ -65,6 +68,7 @@ import aQute.service.reporter.Reporter;
  */
 class IndexFile {
 	private final static Logger					logger		= LoggerFactory.getLogger(IndexFile.class);
+	private final static Pattern				mavenNamePattern	= Pattern.compile("[.,;\"'-]+");
 
 	final File									indexFile;
 	final IMavenRepo							repo;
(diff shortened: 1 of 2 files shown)


biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -338,17 +339,25 @@ private boolean isMulti(String name) {
...
+			.getParentFile();
 		Map<Archive, Resource> result = new HashMap<>();
 		try (Jar jar = new Jar(multi)) {
-			int n = 1000;
 			for (Entry<String, aQute.bnd.osgi.Resource> entry : jar.getResources()
 				.entrySet()) {
 				String path = entry.getKey()
 					.toLowerCase();
 				if (path.endsWith(".jar")) {
+					String name = Strings.lastPathSegment(path)[1];
```
bjhargrave, line 352: `lastPathSegment` can return `null`. So you need to guard against NPE here.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -338,17 +339,25 @@ private boolean isMulti(String name) {
...
+					name = Strings.extension(name)[0];
```
bjhargrave, line 353: `extension` can return `null`. So you need to guard against NPE here.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -338,17 +339,25 @@ private boolean isMulti(String name) {
...
+					name = ZipUtil.cleanPath(name);
+					name = name.replaceAll("[.,;\"'-]+", "_");
```
bjhargrave, line 355: You should precompile the regex into a static final Pattern field. There is no point in compiling the regex each iteration of the loop.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -338,17 +339,25 @@ private boolean isMulti(String name) {
...
+					while (!multis.add(name)) {
+						name += "_";
```
bjhargrave, line 357:
```suggestion
						name = name.concat("_");
```

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -338,17 +339,25 @@ private boolean isMulti(String name) {
 	}
 
 	private Map<Archive, Resource> parseMulti(Archive archive, File multi) throws Exception {
+		Set<String> multis = new HashSet<String>();
```
bjhargrave, line 342:
```suggestion
		Set<String> multis = new HashSet<>();
```

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
```diff
@@ -338,17 +339,25 @@ private boolean isMulti(String name) {
...
+		File localArchives = repo.toLocalFile(archive)
```
bjhargrave, line 343: What is this variable for? You don't seem to use it anywhere.

bjhargrave (changes requested):

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
pkriens: But he checks for .jar at the end? So no null check needed as far as I can see?

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
pkriens: Same, there must be an extension

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
juergen-albert: done

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
juergen-albert: In this case we don't because we already checked if the name ends with ".jar" in the if above.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
juergen-albert: done

pkriens: This block is a very rare occurrence and it is less readable. I am not in favor in making code less readable for an immeasurable performance gains? If performance is that important, should a string builder then not even be better?

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
juergen-albert: @pkriens I believe BJ is right. if the jar is at the root of the zip, I can see a path like "somejar.jar", in which case `lastPathSegment` would return null.

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/IndexFile.java
juergen-albert: done

bjhargrave (approved):

PR #4963, opened 2021-11-21
[maven] allow the EE to be implied from the running JDK  rather than …

rotty3000:
…from the maven-compiler-plugin (which should be constant rather than floating)

This enables use cases like running multiple different Java versions in CI affecting the `-runee` used to resolve and run integration tests

### DIFF
--- a/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
+++ b/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java
@@ -293,7 +293,8 @@ public class BndrunContainer {
 		String runee = run.getProperty(Constants.RUNEE);
 
 		if (runee == null) {
-			EE ee = Optional.ofNullable(project.getPlugin("org.apache.maven.plugins:maven-compiler-plugin"))
+			EE ee = isImplicitEE() ? getJavaSpecificationVersionOrMinSupportedVersion()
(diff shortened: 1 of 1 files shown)


bjhargrave (changes requested):
I would rather have a system property (e.g. `bnd.runee`) which is used to configure the inferred runee instead of the computation. So you can set `-Dbnd.runee=JavaSE-${java.specification.version}` from outside bnd.

`-Dbnd.runee=JavaSE-${{ matrix.java }}` for github actions config.

You could also set it from a maven profile (`-Pbnd.runee.from.runtime`)

We would need to support for the gradle side also. And tests!

rotty3000:
After further thought and having considered your feedback the solution really is to just use `-runee: JavaSE-${java.specification.version}` as you suggested. This works perfectly fine today with no change to bnd so I'm just going to close this.

For posterity, the complication I had in some projects was that in order to achieve implicit calculation of the `-runee` such that it would be affected by the version of Java being executed, I made the compilation target float, which was wrong. We typically want to keep the compilation target level pinned to our code's the minimum supported version of Java, but then test resolution on a floating `-runee` in order to test those resolutions out. A side effect of having a floating compilation target was that Eclipse would easily get confused and choose an EE which was based on it's own runtime rather than a value set at the project level as is most typical or accidentally compiling the project with a higher version JDK causing the modules to accidentally become incompatible with a lower EE.

PR #4976, opened 2021-12-03
Adds targetDir and templates to runbundles exporter

juergen-albert:
* targetDir allows to place the run bundles inside a directory in the
resulting zip
* As the runbundles exporter ignores all include resources, a template
allows to fill all the runbundles in a predefined setup


biz.aQute.bndlib/src/aQute/bnd/exporter/runbundles/RunbundlesExporter.java
```diff
@@ -33,23 +36,35 @@ public RunbundlesExporter() {}
...
 
-	private String nonCollidingPath(Jar jar, String fileName) {
+	private String nonCollidingPath(Jar jar, String fileName, String prefix) {
 		String[] parts = Strings.extension(fileName);
 		if (parts == null) {
 			parts = new String[] {
 				fileName, ""
 			};
 		}
+		fileName = prefix.concat(fileName);
```
bjhargrave, line 67: If you enter the while loop, the prefix is lost. You should also add a test for this.

biz.aQute.bndlib/src/aQute/bnd/exporter/runbundles/RunbundlesExporter.java
```diff
@@ -33,23 +36,35 @@ public RunbundlesExporter() {}
...
 		project.prepare();
 		Collection<Container> runbundles = project.getRunbundles();
-		Jar jar = new Jar(project.getName());
+		String prefix = options.getOrDefault(TARGET_DIR, "");
+		if (!prefix.endsWith("/")) {
+			prefix = prefix.concat("/");
+		}
+		String template = options.get(TEMPLATE);
+
+		Jar jar = null;
```
bjhargrave, line 45: You don't need to initialize to null since you set jar in both sides of the if.
```suggestion
		Jar jar;
```

docs/_instructions/export.md
```diff
@@ -30,7 +30,7 @@ The Exporters use a plugin mechanism and therefore the list is not closed. The f
 
 * `bnd.executablejar.pack` – Exports a JAR file using the launcher from the `-runpath` or the default if no launcher is specified.
 * `bnd.executablejar` – Similar to the previous but does not support profiles, has no automatic bsn assigned, and entries are not signed. The reason there are two types for the more or less the same format with subtle differences is for backward compatibility.
-* `bnd.runbundles` – A JAR with the runbundles
+* `bnd.runbundles` – A JAR with the runbundles. As optional Attributes `targetDir` will output all runbundles in the set directory in the resulting jar and `template` will use a given jar/zip as template to which the bundles will be added.
```
bjhargrave, line 33: template will also support a directory value given how you construct the Jar object from the template.

biz.aQute.bndlib/src/aQute/bnd/exporter/runbundles/RunbundlesExporter.java
```diff
@@ -33,23 +36,35 @@ public RunbundlesExporter() {}
...
+		if (template == null) {
+			jar = new Jar(project.getName());
+		} else {
+			jar = new Jar(project.getName(), template);
```
bjhargrave, line 49:
```suggestion
			jar = new Jar(project.getName(), project.getFile(template));
```

This will resolve a relative path against the base of the project.

biz.aQute.bndlib/src/aQute/bnd/exporter/runbundles/RunbundlesExporter.java
```diff
@@ -33,23 +36,35 @@ public RunbundlesExporter() {}
 	public Entry<String, Resource> export(String type, Project project, Map<String, String> options) throws Exception {
 		project.prepare();
 		Collection<Container> runbundles = project.getRunbundles();
-		Jar jar = new Jar(project.getName());
+		String prefix = options.getOrDefault(TARGET_DIR, "");
+		if (!prefix.endsWith("/")) {
```
bjhargrave, line 40: This means the default for `prefix` will be `/` since you default to empty string which does not end with slash and that is then suffixed with `/`. I don't think that is the proper default. Empty string should be the proper default.
```suggestion
		if (!prefix.isEmpty() && !prefix.endsWith("/")) {
```

biz.aQute.bndlib/src/aQute/bnd/exporter/runbundles/RunbundlesExporter.java
```diff
@@ -33,23 +36,35 @@ public RunbundlesExporter() {}
...
 		String[] parts = Strings.extension(fileName);
 		if (parts == null) {
 			parts = new String[] {
 				fileName, ""
 			};
 		}
+		fileName = prefix.concat(fileName);
 		int i = 1;
 		while (jar.exists(fileName)) {
 			fileName = String.format("%s[%d].%s", parts[0], i++, parts[1]);
```
bjhargrave, line 70:
```suggestion
			fileName = String.format("%s%s[%d].%s", prefix, parts[0], i++, parts[1]);
```

bjhargrave (changes requested):

bjhargrave:
> As the runbundles exporter ignores all include resources, a template
allows to fill all the runbundles in a predefined setup

That is the purpose of the runbundles exporter: to JUST export the runbundles. If you want something else that would be a different exporter, no? :-)

biz.aQute.bndlib/src/aQute/bnd/exporter/runbundles/RunbundlesExporter.java
juergen-albert: done

biz.aQute.bndlib/src/aQute/bnd/exporter/runbundles/RunbundlesExporter.java
juergen-albert: done

biz.aQute.bndlib/src/aQute/bnd/exporter/runbundles/RunbundlesExporter.java
juergen-albert: makes sense. done!

docs/_instructions/export.md
juergen-albert: good point! If modified the sentence

biz.aQute.bndlib/src/aQute/bnd/exporter/runbundles/RunbundlesExporter.java
juergen-albert: done

biz.aQute.bndlib/src/aQute/bnd/exporter/runbundles/RunbundlesExporter.java
juergen-albert: Good point. I totally missed that part. done

docs/_instructions/export.md
rotty3000: BTW, while you're here there is an error in this description. bnd.runbundles just drops the bundles into a directory so it should actually read something like: `A directory containing the runbundles. ...`

juergen-albert: @rotty3000 do you refer to the description for the `template`? If so, the new Version is: `template` will use a given jar/zip or directory as template to which the bundles will be added.

rotty3000: No I meant that the bnd.runbundles exporter produces a directory with bundles in it, but the current doc says it creates a jar with bundles in it.

juergen-albert: Well, yes and no. The `RunBundlesExporter` quite specifically creates a `Jar` and adds all the resources and by default bnd will export write it to a jar, especially when using the `-export` instruction. I guess it works different for maven, but I have no clue how I would need to describe it without confusing any user of the workspace model.

rotty3000: odd! you are correct. I was mistaken.

bjhargrave (approved):

PR #4979, opened 2021-12-06
Fixed potential unclosed file Handle

juergen-albert:

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
@@ -1979,6 +1979,7 @@ public class Project extends Processor {
 					File actual = write(f -> {
 						IO.copy(ee.getValue()
 							.openInputStream(), f);
+						IO.close(ee.getValue());
 					}, outputFile);
 
 					if (actual != null) {
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
@@ -1977,9 +1977,9 @@ public class Project extends Processor {
 				for (Map.Entry<File, Resource> ee : exports.entrySet()) {
 					File outputFile = ee.getKey();
 					File actual = write(f -> {
-						IO.copy(ee.getValue()
-							.openInputStream(), f);
-						IO.close(ee.getValue());
+						try (Resource r = ee.getValue()) {
+							IO.copy(r.openInputStream(), f);
+						}
 					}, outputFile);
 
 					if (actual != null) {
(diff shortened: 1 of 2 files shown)


biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -1979,6 +1979,7 @@ boolean isStale(Set<Project> visited) throws Exception {
 					File actual = write(f -> {
 						IO.copy(ee.getValue()
 							.openInputStream(), f);
+						IO.close(ee.getValue());
```
bjhargrave, line 1982:
```suggestion
						try (Resource r = ee.getValue()) {
							IO.copy(r.openInputStream(), f);
						}
```

bjhargrave (changes requested):

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -1977,8 +1977,9 @@ boolean isStale(Set<Project> visited) throws Exception {
 				for (Map.Entry<File, Resource> ee : exports.entrySet()) {
 					File outputFile = ee.getKey();
 					File actual = write(f -> {
-						IO.copy(ee.getValue()
-							.openInputStream(), f);
+						try (Resource r = ee.getValue()) {
+							IO.copy(r.openInputStream(), f);
+						}
```
bjhargrave, line 1982: Actually, we cannot close the Resource in the Consumer as the write method can call the Consumer multiple times. So we need to hoist the try-with-resources for the Resource outside of the Consumer:

```java
					try (Resource r = ee.getValue()) {
						File outputFile = ee.getKey();
						File actual = write(f -> {
							try (OutputStream out = IO.outputStream(f)) {
								r.write(out);
							}
						}, outputFile);

						if (actual != null) {
							buildFilesSet.add(actual);
						} else {
							error("Could not save %s", outputFile);
						}
					}
```

I also changed from `IO.copy` to `Resource.write` since it is way more memory efficient as it does not require writing the Jar to a ByteBuffer and then writing the ByteBuffer to a File.

bjhargrave (changes requested):
A better fix which avoids the mistake of closing the Resource inside of the Consumer.

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
juergen-albert: Sounds reasonable. Is this already in here or do we need to do something else?

bjhargrave: It is not in "here". The PR needs to be fixed. That is why I made the comment with the suggested change.

PR #4984, opened 2021-12-07
DRAFT: initial Version of the template exporter

juergen-albert:
@bjhargrave you last [comment on my PR #4976](https://github.com/bndtools/bnd/issues/4976#issuecomment-985656658_) got me thinking and I saw that you are correct. Thus here is a new Exporter that goes a bit beyond the previous PR.

With such a template based exporter more complex runtimes can be exported. I'm not sure if I got all aspects, but it would cover the cases I currently have to export e.g. something like the custom sensinact server and also what we previously did with our gecko exporter. 

I'm not sure of `bnd.templated` is a name that fits or is even proper English. I'm open to changes and improvements.


bjhargrave (changes requested):
New exporters should be in the biz.aQute.bnd.exporters project.

juergen-albert:
I don't get around finalizing this. I will open a new one, when I have it ready.

Issue #4575, opened 2021-03-03
Eclipse, bnd & Concurrency


pkriens:
There is a PR to refresh the workspace. While working on this PR, it became clear that the current Eclipse interface for bnd is a mess. We've always had the problem that we did not really know Eclipse and struggled to get things to work. Neil wrote the first incarnation of the builder, we rewrote it during a bndtools hackaton, BJ made serious improvements several times. However, as member of this group I think I can say that none of use really knew what we were doing. The result is a mess. Most of the direct problems have found their band-aids but problems are prevalent when the workspace undergoes large changes on disk.

* No rules about locking and what is thread safe to call
* No clear rule when to refresh what resource
* Synchronous inline actions, for example the builder calls things that should work from listeners
* Mostly ignoring locking
* Not taking advantage of some of the (newer?) Eclipse facilities
* We do not take advantage of parallel builds

This is an attempt to setup some rules before we start programming them.

## Concurrency

To my great surprise, parallel gradle tends to work very well. Surprise, because bnd is not designed to run concurrently. It likely works so well because Gradle builds up the Workspace and Project objects ahead of any concurrent actions and this structure is never changed during the run. In Eclipse, the structure does change and this makes it necessary to lock. We've chosen for a single lock in Central to execute all actions. A few years ago, the Workspace got a `readLocked` and `writeLocked` method that takes a lambda. So far, these have only be used by the Remote Workspace.

Looking at the code, I think we have different types of bnd code:

* **Repositories** – The repository handling must be completely thread safe and not require any global bnd locks.
* **Workspace** – Actions that change the workspace (refresh bnd files, adding/removing projects, etc.) need a write lock on the workspace. 
* **Project** – Most actions on a project could use a read lock on the workspace. The exception is refresh, it should hold a write lock on the workspace.
* **Analyzers/Builders/Launchers** – All this code should not be safe to use concurrently. Some of these classes are created via a Workspace or Project and hold a reference via the Processor. We might consider creating copies of the parent since a refresh of the parent invalidates the properties the object is based on. Open for discussion, could we create some copy on write model?

## Workspace Access

[...]

PR #4988, opened 2021-12-09
workspace: Utilize workspace ReentrantReadWriteLock for locking

bjhargrave:
We replace the use of Bndtools' bndLock with the Workspace's
ReentrantReadWriteLock. This allows us to use read locking in most
places and write locking only when necessary.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
@@ -39,9 +39,8 @@ import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReadWriteLock;
 import java.util.concurrent.locks.ReentrantLock;
-import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.BooleanSupplier;
 import java.util.jar.JarEntry;
 import java.util.jar.JarInputStream;
 import java.util.jar.Manifest;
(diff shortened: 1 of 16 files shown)


kriegfrj:
This PR is related to #4575?

pkriens:
Hmm, shouldn't we have an agreement on where we do the locking? I am not sure it should be the responsibility of the bndlib to do its locking itself since in many cases you can guarantee the concurrency issues from outside. 

Not sure this is the right way to go.

chrisrueger:
From 6ef310ec44697cd815697c1b0db51f074e6fe5b9
> We fix the remaining watcher
> to properly refresh the workspace in a job rather than on the main
> thread.

Could that reduce e.g. some UI-Freezes during a (longer) build? At the moment when a build caused by a changed file, typing e.g. in Java-Code editor becomes unresponsive / sluggish. I assume because some thing happening in the main thread. Would this change affect this somehow?

bjhargrave:
> This PR is related to #4575?

Perhaps tangentially. I was trying to fix a nagging issue the quickfix tests would occasionally service and I wanted to use a read/write lock since we don't need exclusive locks in most places.



> Hmm, shouldn't we have an agreement on where we do the locking? I am not sure it should be the responsibility of the bndlib to do its locking itself since in many cases you can guarantee the concurrency issues from outside.

We should certainly discuss this. But by putting the burden on the caller, we make sure new bugs are always introduced as code is changed/added as the caller forgets to do locking or does it wrong. By having it in bndlib, the code is just-thread-safe.

> Could that reduce e.g. some UI-Freezes during a (longer) build?

I don't think so since  build wont be modifying the workspace properties files which is what triggers the refresh action.

kriegfrj:
Did you consider a stamped lock instead of reentrant re-write lock?

bjhargrave:
> Did you consider a stamped lock instead of reentrant re-write lock?

StampedLocks are not reentrant which is I think rather important in this case.

PR #4990, opened 2021-12-11
Spi

rotty3000:
[spi] improve the SPI generated descriptor warning message to be more clear and precise and only appear when there is also no descriptor at all for the service type

/cc @mnlipp


bjhargrave:
Also, Bnd should only generate the warning if both (1) the register directive is not present, and (2) the resource is also not present. If fact, Bnd could simply choose not to generate the resource if the resource is already in the jar (trusting the developer).

rotty3000:
given that we're going to not warn unless both the directive and resource are missing (I had already come to that conclusion myself ;) ) I think it's best to forgo the flag entirely to keep things simple.

bjhargrave (approved):

bjhargrave:
Make sure to note this change in the release notes.

rotty3000:
Noted!

Issue #5004, opened 2021-12-20
[builder] Convert Bndtools builder to a facade pattern


kriegfrj:
In order to streamline future development on the Bndtools builder (eg, #3175), I would like to convert it to a facade pattern so that the builder bundle can be restarted without needing to restart Eclipse. Creating this issue to hang a corresponding commit/PR from.

kriegfrj:
> Care is needed since this is the crux of the Bndtools value.

pkriens:
@kriegfrj I am not sure how this can be done? Creating the facade is easy but BndtoolsBuilder is connected with the project. So if the service goes, you need to save the state and restore it?

I like the idea but I would like to close this issue, it is has been outstanding for a long time. I can move it to abeyance if you want.

pkriens:
Ok, I am moving it to abeyance for now. It looks very interesting though, would be great if you could finish it one day.

PR #5008, opened 2021-12-20
[validators] Fix marker generation for validators

### LINKED ISSUE #5006: [validator] Project/builder validators not generating markers
While working on #5004, I was running a few tests to see if I had broken anything with the standard build behaviour. I tried to misconfigure an Eclipse project to see if would still generate error markers, and found that it didn't. When I dug further, I discovered the following bit of code in `master`:

https://github.com/bndtools/bnd/blob/f848ef99deed1cbd1c4c576f9575781ca4a07a7d/bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java#L207-L213

The problem here is that "markers.validate()" is run on `model` (and more importantly, generates its errors on `model`), and it is run *after* the call to `processor.getInfo(model)` (which is supposed to populate `processor` with the errors in `model`). Then the markers are generated from `processor` (which still has the snapshot of errors from before `validate()` was run).

A potential fix is to run `markers.validate()` before the call to `processor.getInfo()`. This seems to be working in my test environment, but I'm not sure if there was a reason why the previous author attempted to defer the call to `validate()` into the "after" block - for performance - to minimise time running inside the locked section, or to avoid potential deadlock? Maybe the solution would be to put another `getInfo()` call inside the `after` block, after the `validate()` call?

kriegfrj:
Ensure that the validators are run on the model before the processor copies them into itself.

Fixes #5006

### DIFF
--- a/bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
+++ b/bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
@@ -205,9 +205,9 @@ public class BndtoolsBuilder extends IncrementalProjectBuilder {
 						model.prepare();
 
 						Processor processor = new Processor();
+						markers.validate(model);
 						processor.getInfo(model);
 						after.accept("Decorating " + myProject, () -> {
-							markers.validate(model);
 							markers.setMarkers(processor, BndtoolsConstants.MARKER_BND_PATH_PROBLEM);
 						});
 						model.clear();


pkriens:
LGTM. Just see that the validation happens in the MarkerSupport class is a bit strange, it never creates any markers.

But I am ok

kriegfrj:
Sorry, I didn't understand your comment?

pkriens:
I looked at the code and wondered why the validation happened on the MarkerSupport class. Since it does not create markers but only errors/warnings on a processor, it is not the logical class to put this code in. I looked at this because we should not create markers inside the lock.

However, been like that forever and we don't have to fix this now.

kriegfrj:
I agree. In #5004 it was natural to move it into `BndtoolsBuilder`, which is now a DS component and tracks the validators as `@Reference`s.

Issue #5005, opened 2021-12-20
[arch] Deprecate custom extension points for use in Bndtools


kriegfrj:
For discussion:

Since experimenting with the "facade" pattern in the core Bndtools implementation, I have come across a few custom extension point definitions that are part of Bndtools. The most recent one that I discovered was for the `IValidator` interface. (Note: I'm not talking about custom implementations of *existing* extension points, but where we have both defined and used our own extension points.)

I think I have come to the conclusion that there are not any strong use cases for us to use custom extension points within Bndtools. Declarative Services can do the same thing, but they are easier to implement, maintain and deploy (particularly when Bnd is your build tool) and enable dynamic re-deployment (which isn't much help for the end user but it super-helpful for the developer). `ServiceTracker`s (and now even the `ExtensionFacade`) can be used to interface non-DS clients with DS implementations - and at the very least this is no more complicated than using the Eclipse Registry API to interface with extension implementation (and I would argue probably less complicated).

Assuming we agree to the above, then:
1. we agree not to add any new custom extension points to Bndtools;
2. we remove the existing custom extension points and replace them with DS implementations as we have the time/energy/motivation. If we feel it is necessary (as the extension points are technically part of a public API for Bndtools), we do this after formally deprecating them; Perhaps this would require a major version change to Bndtools. However, I doubt there are too many 3rd-party plugins out there that are extending Bndtools using these extension points - they are mainly for internal use (I think), so we might be safe with a minor version increment.

Thoughts?

pkriens:
* No custom extension points – Fully agree
* Deprecating – After we've got a replacement, we can deprecate them, but we should then have at least one major release in between before removal imho. They cost little and even one annoyed customer is one too many.

bjhargrave:
As for deprecating for removal. I am not convinced that anyone but ourselves have even implemented one of the Bndtools extension points. I am not sure we need a drawn out process to remove them. I would suggest we deprecate and cease using them immediately and state that they will be removed in the next major release.

Issue #3175, opened 2019-05-08
[Feature] Smarter builder dependency checking


kriegfrj:
While working on #3145, I noticed that the tester bundle (and the `biz.aQute.junit.bundle.engine` bundle I created, on which it depends) is quite a way down the build dependency hierarchy, and I was spending quite a bit of time waiting for Bndtools to rebuild a large number of bundles in the workspace. It seems that relatively small changes to the `bundle.engine` bundle result in a rebuild of all the dependent bundles. As an extreme example, merely changing a comment in a class file in a bundle causes all of the dependent bundles to rebuilt.

Though I haven't looked at it in detail, I think that the build dependency checking seems to depend on the timestamp of the individual bundles' jar files - if the jar file changes, then any other bundle in the workspace that has that bundle on its build path will get rebuilt.

This is an extremely coarse and conservative approach, which virtually guarantees you won't get any out-of-date bundles higher up the build hierarchy. However, it also means that Bndtools spends quite a bit of time unnecessarily rebuilding bundles that really don't need to be rebuilt. A bundle really only needs to be rebuilt if a dependent bundle's ''API'' changes in a way that is not backward compatible. Changes that fall in this category:

1. Changing a method signature.
2. Removing a method.
3. Removing a class (arguably, only when an exported class is removed, but given that dependent bundles can violate the export provisions at build time at least, the most conservative way is to rebuild when any class is removed).
4. Removing an exported package.

This is one area (perhaps the first, in my experience) where I think PDE still has a bit of an edge over Bndtools - as Eclipse doesn't base its build decisions on fully assembled bundles, but on the classes that they contain, and it uses its regular inter-project dependency mechanisms to make decisions about when to rebuild a class.

I know that Bnd already has some great built-in capabilities for calculating external imports and baselining. It occurred to me that a combination of these two capabilities that in theory could combine to form the basis of smarter dependency/rebuilding decisions. You keep track of which classes you depend on from each bundle (a capability used by the automatic package import calculation), and you keep track of incompatible changes in classes from each bundle (a capability used by the baselining), and you only rebuild if there is an incompatible change in a dependent class.

And while #3032 is strictly speaking a different feature request to this one, I think if this feature were implemented it might also indirectly solve #3032 for my specific use case (as most of the downstream dependencies wouldn't get rebuilt by a builder with smarter dependency checking).

[...]

PR #5036, opened 2022-01-07
[builder] Change to facade

### LINKED ISSUE #5004: [builder] Convert Bndtools builder to a facade pattern

kriegfrj:
This is an early-access PR for those interested in how it's coming together. 

The code is in obvious need of some tidy-ups. But I think I have reached a point where I can start trying to implement some regression tests. Before I ploughed ahead too much further I wanted to give the opportunity for some feedback on the direction/architecture.

Key features:
* I have added whiteboard behaviour for `IResourceChangeListener`s. You don't have to worry about registering/unregistering your listeners as DS will do it for you courtesy of the ResourceWhiteboard component. Just register a service/component that implements `IResourceChangeListener` and it will automatically receive resource change event notifications.
* Validators and ProjectValidators are now exclusively service-driven (part of fix for #5005).
* BndtoolsBuilder is now a dynamically-restartable component, using an adapted facade pattern. Because the base class for builders is an abstract class rather than an interface, this required a slightly different approach, but it seems to be working. (Closes #5004).
* IStartupParticipant is superseded by a pattern of using DS component activation for start/stop (part of fix for #5005).
* Central is now itself a DS component. It also automatically triggers background initialization and registration of the Workspace object as a service (whereas before it would happen when the first client tried to access it). This makes it easy for components that depend on Central being initialized to simply add it as a static reference.

There is still further simplification to be made by converting other pieces to DS components (eg, BndPreferences). But I think this is a good point to stop and add regression tests that will then make further such refactoring safer.


kriegfrj:
Still a little bit of work to do but I think it is mostly working and probably at a good point where it could benefit from a review.

kriegfrj:
Think I've done most of the tidying-up now. Let me have it, @bjhargrave! :smile:

org.bndtools.builder/src/org/bndtools/builder/impl/CnfWatcher.java
```diff
@@ -37,9 +37,11 @@
 	@Reference
 	BndtoolsBuilder					builder;
 
-	// Bnd workspace, not Eclipse workspace.
 	@Reference
-	Workspace						workspace;
+	IWorkspace						eclipseWorkspace;
```
bjhargrave, line 41: I like these var names to clearly distinguish between the 2 workspaces!

bjhargrave (commented):
Wow, what a massive piece of work!

I, of course, have some comments :-)

org.bndtools.builder/src/org/bndtools/builder/classpath/BndContainerCompilationParticipant.java
```diff
@@ -5,7 +5,10 @@
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.compiler.CompilationParticipant;
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.ServiceScope;
 
+@Component(scope = ServiceScope.PROTOTYPE, service = CompilationParticipant.class)
```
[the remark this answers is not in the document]
bjhargrave: We might want to put comments in the places where the scope is set to explain why we need that scope for the component. There is a mix of scope types.

kriegfrj:
> Wow, what a massive piece of work!
>

Well, when you take out moved files and updates from Eclipse automated refactoring, it's probably not as big as it looks at first glance. :smile:
 
> I, of course, have some comments :-)

I was hoping (and expecting) as much! :smile:

org.bndtools.builder/src/org/bndtools/builder/impl/CnfWatcher.java
kriegfrj: :+1: I got confused one too many times...

bndtools.api/src/org/bndtools/api/builder/BuildLoggerConstants.java
```diff
@@ -0,0 +1,9 @@
+package org.bndtools.api.builder;
+
+public interface BuildLoggerConstants {
```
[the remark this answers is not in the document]
kriegfrj: :+1: Will fix.

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -74,7 +76,8 @@
 import bndtools.central.RepositoriesViewRefresher.RefreshModel;
 import bndtools.preferences.BndPreferences;
 
-public class Central implements IStartupParticipant {
```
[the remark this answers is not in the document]
kriegfrj: Good catch.

bndtools.core/bnd.bnd
```diff
@@ -105,7 +105,8 @@ Export-Package: org.osgi.service.metatype.annotations
 	bndtools.utils;version=project;packages='*',\
 	slf4j.api,\
 	org.eclipse.ui.ide.application;version='1.3',\
-	org.eclipse.ui.console
+	org.eclipse.ui.console,\
+	org.eclipse.osgi.services
```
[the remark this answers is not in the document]
kriegfrj: I think I had problems with this but I will try again.

bndtools.core/src/bndtools/central/ResourceWhiteboard.java
```diff
@@ -0,0 +1,66 @@
...
+		listeners.close();
+		listeners = null;
+	}
+
+	void addResourceListener(IResourceChangeListener service, ServiceReference<IResourceChangeListener> reference) {
+		Integer mask = (Integer) reference.getProperty("org.eclipse.core.resources.eventmask");
+		if (mask == null) {
+			workspace.addResourceChangeListener(service);
+		} else {
+			workspace.addResourceChangeListener(service, mask);
```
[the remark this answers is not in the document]
kriegfrj: I thought I tested it and it seemed to work. At least, it seemed that the `ResourceWhiteboard` was doing the right thing. I cannot guarantee that the `IWorkspace` was doing the right thing. I will have a closer look.

bndtools.core/src/bndtools/central/QueryWhiteboard.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * This component adds whiteboard facility to IQueryListeners.
+ */
+@Component(immediate = true)
+public class QueryWhiteboard implements ServiceTrackerCustomizer<IQueryListener, IQueryListener> {
+	@Reference
+	IWorkbench										wb;
+
+	ServiceTracker<IQueryListener, IQueryListener>	listeners;
```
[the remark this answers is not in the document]
kriegfrj: This was one of the things that I hoped you would comment on. :smile:

I originally tried bind/unbind/updated, but I found that I was getting NPEs. I realised that this was because the get called before the reference to IWorkbench was satisifed and before the component was activated. It got complicated to track them as they were removed/added before activation, and then to track them differently after activation (when the workspace is available), and as I thought about how to handle this problem I realised that this was exactly the sort of problem that `ServiceTracker` was designed to solve.

However, I admit to not being a DS expert. If there is an easy way to configure DS so that the dynamic reference updates are queued by DS and only delivered *after* activation, then that would work. (If there is no such feature in DS, maybe we could consider it a feature request?)

If you have a better suggestion, I'd love to hear it.

bndtools.core/src/bndtools/central/Central.java
```diff
@@ -89,85 +92,60 @@
...
 		instance = this;
-
-		repoListenerTracker = new RepositoryListenerPluginTracker(bundleContext);
+		repoListenerTracker = new RepositoryListenerPluginTracker(bc);
 		repoListenerTracker.open();
+		internalPlugins = new InternalPluginTracker(bc);
 		internalPlugins.open();
+		// Trigger building of the Workspace object and registering it a service
+
+		promiseFactory().submit(Central::getWorkspace);
```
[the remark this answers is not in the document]
kriegfrj: I think that the workspace creation was deferred to avoid prematurely trying to access the Eclipse workspace before it ready. Before my changes, the bndtools.core bundle would fail to activate if it activated too early (before the workbench was running).

With regard to increasing apparently delay: Adding this async call to getWorkspace() here should actually be an improvement in this regard. Previously, the workspace was not initialized until the first time a consumer wished to use it, and it was initialised on whatever thread made that first call to `getWorkspace()`. By submitting this promise, we are pre-emptively triggering the background initialisation of the workspace before any consumer actually needs it - it will be that much closer to ready by the time it is needed.

bndtools.builder/src/org/bndtools/builder/facade/IProjectBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+import org.eclipse.core.resources.IBuildConfiguration;
+import org.eclipse.core.resources.IBuildContext;
+import org.eclipse.core.resources.ICommand;
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResourceDelta;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.jobs.ISchedulingRule;
+
+public abstract class IProjectBuilder {
```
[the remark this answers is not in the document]
kriegfrj: :+1: Originally it was an interface and then I realised I needed to store some state and needed the abstract class. Perhaps I will use "IncrementalProjectBuilderDelegate".

But I'm not the first one to use the "I" naming scheme in Bndtools - there are a few pre-existing in org.bndtools.api. I also added `ICentral` in the very first commit. I understand that Eclipse can tell you if it's a type or an interface, but usually this is not quite as convenient as the "I" convention which you can easily see at a glance in the source without having to do any kind of inspection (unless, of course, someone does something silly like I did here which violates the convention... :smile:). I'm not fussed either way - happy to avoid using the "I" convention, but with "Central" I needed something different to distinguish it from the concrete class from which I extracted it originally.

org.bndtools.builder/src/org/bndtools/builder/classpath/BndContainerCompilationParticipant.java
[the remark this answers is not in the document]
kriegfrj: Eclipse registry extension points generally assume that a new instance is returned every time it is instantiated. So unless you know it is safe to do otherwise, you must use PROTOTYPE scope.

When it is safe to use SINGLETON scope: if the base class is an interface, and your implementation doesn't maintain any per-instance state, then you can sometimes get away with SINGLETON. But in this case, the base service type `CompilationParticipant` is a class and so it might maintain some state of its own. So I went with the PROTOTYPE as the least likely to cause problems. To be honest, I haven't looked much further than this and it's *possible* that some of the ones I've set as PROTOTYPE could be SINGLETON.

bndtools.core/src/bndtools/central/Central.java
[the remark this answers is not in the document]
kriegfrj: It's in a promise because my understanding was that `@Activate` methods should run quickly and that `getWorkspace()` might take time to initialize. If this is not a requirement I'm happy to inline the call. Otherwise I'll add the `onFailure()`.

djencks:
Could you describe the problem leading to NPEs when using DS in more detail?  From the description below it doesn’t make any sense to me that this could be happening.

Thanks
David Jencks

kriegfrj:
> Could you describe the problem leading to NPEs when using DS in more detail? From the description below it doesn’t make any sense to me that this could be happening. Thanks David Jencks

Sorry, it wasn't a generic DS problem but a problem with my specific implementation.

The original implementation simply called `workspace.add/RemoveResourceChangeListener()` as was appropriate. The problem was that the component wasn't fully activated (`workspace` was still `null`) when the first `ResourceChangeListener` services started getting implemented, which is why my implementation was causing an NPE. I couldn't see an easy way to fix that without buffering the resource listeners as they were registered and then registering them once the component was fully activated, which is why I employed a `ServiceTracker` to do that for me.

bndtools.core/bnd.bnd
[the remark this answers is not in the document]
kriegfrj: Turns out org.eclipse.osgi.services was surplus to requirements. Removed.

djencks:
I looked at the code, I see what you mean,  thanks!!

David Jencks

bndtools.builder/src/org/bndtools/builder/facade/IProjectBuilder.java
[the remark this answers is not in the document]
kriegfrj: Renamed it ProjectBuilderDelegate

bndtools.builder/bnd.bnd
```diff
@@ -35,7 +35,8 @@
 	org.eclipse.equinox.common,\
 	org.eclipse.equinox.registry,\
 	org.eclipse.equinox.preferences,\
-	slf4j.api
+	slf4j.api,\
+	org.osgi.service.component.annotations
```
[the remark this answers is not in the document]
kriegfrj: There were a lot of dependencies in this bnd file that were no longer used after I moved a lot to the new `org.bndtools.builder` bundle. I did a prune of unused dependencies and this one disappeared.

bndtools.api/src/org/bndtools/facade/ExtensionFacade.java
```diff
@@ -42,9 +42,13 @@
...
-	static final BundleContext	bc			= Optional.ofNullable(FrameworkUtil.getBundle(ExtensionFacade.class))
-		.map(Bundle::getBundleContext)
-		.orElse(null);
+	// Make this an instance variable rather than a static variable on purpose,
+	// because
+	// sometimes (due to lazy bundle activation) the bundle context can come
+	// back null
+	// if you call getBundleContext() during class initialization.
+	final BundleContext			bc			= FrameworkUtil.getBundle(ExtensionFacade.class)
+		.getBundleContext();
```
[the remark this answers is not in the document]
kriegfrj: Thanks for the tip!

org.bndtools.builder/src/org/bndtools/builder/classpath/BndContainerCompilationParticipant.java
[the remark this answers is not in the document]
kriegfrj: Sorry, I just had a closer look at `CompilationParticipant` and realised that it is stateless. So now I understand your confusion as to why it was marked PROTOTYPE.

I think, in my rush to try and get something working, I set everything to PROTOTYPE as the safest option, with the thought that I would at some point return to double-check and change some to SINGLETON if they were required. I'll tidy this up.

kriegfrj:
New push incorporates most of the changes above. Waiting for further discussion on the remainder.

bndtools.core/src/bndtools/central/QueryWhiteboard.java
```diff
@@ -0,0 +1,54 @@
...
+
+	BundleContext									bc;
+
+	@Activate
+	void activate(BundleContext bc) {
```
rotty3000, line 28:
```suggestion
        @Activate
	public QueryWhiteboard(BundleContext bc) {
```
will cause activation to always occur before any bind methods are fired.

rotty3000, line 28:
```suggestion
	@Activate
	public QueryWhiteboard(BundleContext bc) {
```
will cause activation to occur before any bind methods are called.

kriegfrj: 👍 thanks a bunch @rotty3000, I think that will do the trick! I'll try it out tomorrow.

bndtools.core/src/bndtools/central/Central.java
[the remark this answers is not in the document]
bjhargrave: A Promise is generally used to obtain the result of a calculation in the future. Here you do not even keep the returned promise :-) But, as I said, you can use the returned promise to log any error that occurs. You should do that.

bndtools.core/src/bndtools/central/QueryWhiteboard.java
bjhargrave: This is a solution. You will need to update the org.osgi.service.component.annotations dependency to 1.4.0 since that is where constructor injection support is added. The Eclipse version we support (2020-06) uses a version of Felix SCR which supports DS 1.4. So we can update all our Bndtools project dependencies on org.osgi.service.component.annotations to 1.4.0.

rotty3000: to be much more clear, the entire solution which ensures workspace is present at activation before any bind methods are called looks like:
```java
final IWorkbench wb;
@Activate
public QueryWhiteboard(BundleContext bc, @Reference IWorkbench wb) {
 ...
 }
 ```
 as well as ensuring ds version 1.4.0.

Comment (kriegfrj) [reply bndtools.core/src/bndtools/central/QueryWhiteboard.java:28]: Done, thanks! I think I would probably make greater use of constructor injection in the future due to its ability to leverage the extra safety of `final` for the referenced services.

File: bndtools.core/src/bndtools/central/QueryWhiteboard.java (hunk 7 above)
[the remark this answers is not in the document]
Comment (kriegfrj) [reply bndtools.core/src/bndtools/central/QueryWhiteboard.java:23]: Implemented @rotty3000 's suggestion for both the resource and query whiteboards and removed the `ServiceTracker`s.

File: bndtools.core/src/bndtools/central/Central.java (hunk 8 above)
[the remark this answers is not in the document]
Comment (kriegfrj) [reply bndtools.core/src/bndtools/central/Central.java:117]: Done.

File: bndtools.core/src/bndtools/central/ResourceWhiteboard.java (hunk 6 above)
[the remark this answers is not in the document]
Comment (kriegfrj) [reply bndtools.core/src/bndtools/central/ResourceWhiteboard.java:44]: I had a look at the source for `Workspace`'s implementation of the listener callbacks, and it seems like it should do the right thing here.

I double-checked by adding the following to `WorkspaceRepositoryChangeDetector.resourceChanged()`:

```java
		System.err.println("Checking resource type");
		if ((event.getType() & IResourceChangeEvent.POST_CHANGE) == 0) {
			System.err.println("WASN'T A POST_CHANGE!!!: " + event.getType());
		}
```
I ran the test suite and saw plenty of "Checking resource type" messages, but no "WASN'T A POST CHANGE" messages. So I think it is working.

kriegfrj:
Latest push incorporates the last suggestions.

I think this still needs some more testing before we merge though - in particular, I think there are issues with the Java search and seems occasional startup issues where the builder is not ready in time.

cnf/ext/junit.bnd
```diff
@@ -43,6 +43,7 @@ junit-osgi: \
 	org.osgi.service.resolver;version=latest,\
 	org.osgi.util.function;version=latest,\
 	org.osgi.util.promise;version=latest,\
+	org.osgi.test.assertj.framework;version=latest,\
```
bjhargrave, line 46: Shouldn't `org.osgi.test.junit5` also be in this list?

kriegfrj: Yes, this was a bit messy. `bndtools.core.test` was not even using the `junit-osgi` macro. I have tidied this up now and added `assertj.promise` too for completeness.

Originally I added `assertj.framework` because I wanted to get access to the Dates functionality for the assertj-eclipse.resources library I started building (currently embedded in `bndtools.core.test`). However, I realised that `1.1` has not been released yet so I couldn't get it... at the moment, the bit I need is cut-and-pasted into `bndtools.core.test`. So strictly, I still don't need it yet. However, I'd like to refactor this once osgi-test 1.1 is released.

bjhargrave:
I think we should defer merging this until after 6.2.0 is released. We plan 6.2.0-RC1 for Feb 4.

kriegfrj:
> I think we should defer merging this until after 6.2.0 is released. We plan 6.2.0-RC1 for Feb 4.

I agree. Though I was hoping at least a fix for #3175 might be able to be cherŕypicked.

kriegfrj:
> > I think we should defer merging this until after 6.2.0 is released. We plan 6.2.0-RC1 for Feb 4.
> 
> I agree. Though I was hoping at least a fix for #3175 might be able to be cherŕypicked.

On second thoughts, probably better to defer that fix too to give the early adopters more time to test it out.

kriegfrj:
> Latest push incorporates the last suggestions.
> 
> I think this still needs some more testing before we merge though - in particular, I think there are issues with the Java search and seems occasional startup issues where the builder is not ready in time.

For posterity:
* I haven't invested much time in fixing the Java search issues, as I'm actually hoping that #3175 will make the special Java search filter redundant.
* I think I have fixed many startup timing issues by changing calls to `tracker.getService()` with `tracker.waitForService(10000)` where appropriate.

juergen-albert:
I finally gotten around to look into you changes and testing it for a bit. I've realized that the new Facade mechanism at least initially can cause start order problems. In my first Test it was always complaining, that the some Service was not available the moment it was required. Moving the bndtools.core bundle farther back in the start order solved the problem. Not sure if this is something we can ensure when bnd is installed later on.

I might have a solution, but I wanted to see if you already have any thoughts on that.

kriegfrj:
Thanks @juergen-albert for being brave enough to take this for a spin!

I've just pushed my latest changes (which includes the change to use `waitForService()` in some places) - this seems to have fixed the startup timing issues for me. The latest push also includes my regression tests for the Bnd classpath container (which are a work-in-progress).

I also noted that the startup issues seemed to be transitory - some parts of the builder/classpath container (or dependent services) would fail to initialize on the first attempt, but later attempts would succeed. This left exception stack traces in the log output but other than that cosmetic issue it didn't seem to break anything.

kriegfrj:
Hi @juergen-albert , thanks for being brave enough to test this out!

I have just pushed the latest version of this that includes the change to use `waitForService()` instead of `getService()` in some of the facades. This seems to have fixed the startup issues in my testing (so much so that I had forgotten about the startup issues!). Let me know if this doesn't fix your problems.

The latest push also includes my regression tests for the classpath initializer, which are mostly working but incomplete (they do not yet cover the full gamut of possibilities with classpath configuration).

pkriens:
@kriegfrj Fr Krieg, any chance this is going to happen sometimes in the near future? Otherwise I'd like to close it for now. We can always resurrect later.

kriegfrj:
I have dusted this off and rebased. Surprisingly, a lot of it still seems to work. Will need to look at a few failing tests though.

kriegfrj:
It looks like the hiatus for this PR had paid off. The new baseline Eclipse (2022-09) now registers the workspace as a service, making it easy to control startup dependency using DS. In fact, very helpfully, if you try to access it the old way and its not initialized yet, the resultant exception message will even alert you to this fact and inform you of the DS solution (among others). Kudos to the Eclipse core team for whoever had the foresight to add that little nugget!

bndtools.api/src/org/bndtools/api/central/ICentral.java
```diff
@@ -0,0 +1,22 @@
...
+
+	Project getModel(IJavaProject project);
+
+	void changed(Project model);
+
+	void addModelListener(ModelListener m);
+
+	void removeModelListener(ModelListener m);
+
+	void close();
```
laeubi, line 20: Is it really whise to have close on a service interface?

bndtools.api/src/org/bndtools/api/central/ICentral.java
```diff
@@ -0,0 +1,22 @@
...
+import aQute.bnd.build.Project;
+
+@ProviderType
+public interface ICentral {
+
+	Project getModel(IJavaProject project);
+
+	void changed(Project model);
+
+	void addModelListener(ModelListener m);
```
laeubi, line 16: Maybe model listener can better use the whiteboardpattern instead?

bndtools.api/src/org/bndtools/api/central/ICentral.java
```diff
@@ -0,0 +1,22 @@
+package org.bndtools.api.central;
+
+import org.bndtools.api.ModelListener;
+import org.eclipse.jdt.core.IJavaProject;
+import org.osgi.annotation.versioning.ProviderType;
+
+import aQute.bnd.build.Project;
+
+@ProviderType
+public interface ICentral {
```
laeubi, line 10:
```suggestion
public interface IWorkspaceManger {
```

bndtools.api/src/org/bndtools/api/central/ICentral.java
```diff
@@ -0,0 +1,22 @@
...
+
+	Project getModel(IJavaProject project);
```
laeubi, line 12:
```suggestion
	Optional<Project> getModel(IProject project);
```

kriegfrj: Our codebase hasn't made heavy use of the `Optional` pattern. But I suppose we probably should...

bndtools.api/src/org/bndtools/api/central/ICentral.java
kriegfrj: Since the name is something I inherited, and this is a rather central piece of code (pardon the pun) I don't feel right approving a change by myself.

But if we're going to change the name, I think I would prefer to change it to something like `BndWorkspaceManager`. There has been one too many times while working on Bndtools that I have gotten confused between an Eclipse workspace object and a Bnd workspace object, and that consumes precious brain cycles. And that's from a seasoned contributor's perspective - I pity the newbie coming along wanting to make a contribution to the Bndtools codebase for the first time and trying to comprehend the difference between a Workspace and a Workspace... (I was that newbie once, and I remember it being a painful experience...)

@bjhargrave expressed a strong dislike for the `I`-prefix naming convention for interfaces when he reviewed this PR (https://github.com/bndtools/bnd/pull/5036#discussion_r787112643). I personally do not feel quite so strongly about it either way - as I mentioned, although Eclipse/IDE can easily tell you if it's an interface, I see some utility in that you can see it's an interface without having to hover over it or do any kind of code navigation. Perhaps I'll let @pkriens have the last word here on whether we go for `IBndWorkspaceManager` or `BndWorkspaceManager`.

bndtools.api/src/org/bndtools/api/central/ICentral.java
kriegfrj: Yes, you're probably right. In fact, I just did some code searches and found that the original `Central.close()` is not even called anywhere from within Bndtools, so I think we can safely remove it.

bndtools.api/src/org/bndtools/api/central/ICentral.java
kriegfrj: I agree 100%.

This was more about me trying to limit the scope of my PR which had already grown into a behemoth. This part of the code is not particularly well tested in CI (which is to say, not really tested at all!) and so I had a fear that something might break. So the goal was to make the minimal changes I needed in this PR to get the automated testing working reliably, and also so that it could get merged so that it could be subjected to some real-world testing by us on the bleeding edge. Then I could then proceed with additional refactorings (like this) with a little bit more confidence that I haven't broken anything in a subtle or fundamental way. I can foresee this interface evolving significantly as we start to leverage the power of a more component-driven approach.

bndtools.api/src/org/bndtools/api/central/ICentral.java
kriegfrj: Just noticed that this was also a change of the argument type from `IJavaProject` to `IProject`. I'm curious as to the reasoning? Every Bnd project is also a Java project.

kriegfrj:
@laeubi , thank you so much for reviewing and your suggested changes! I have made some counter-comments above.

bndtools.api/src/org/bndtools/api/central/ICentral.java
laeubi: It is easy to get from `IJavaProject` > `IProject` but the other is not that easy (and better handled on a central 😛place), also as it is maybe now the case who knows what the future brings?
So usually  `IProject` is much more comfortable to use from api user point of view as you almost ever have one at hand.

bndtools.api/src/org/bndtools/api/central/ICentral.java
laeubi: I don't mind the I prefix, personally don't use it but if the projects wants it... as you said its redundant and I prefer to follow Java convention that the Interface is the name WHAT it is and the class should mention HOW it is implemented e.g. with `List` (Interface) and `ArrayList` (an implementation using an array as its backing store)...

So for `BndWorkspaceManger` (the thing) `CentralBndWorkspaceManger` (aka `Central`) and implementation that use **one** Workspace ever.

bndtools.api/src/org/bndtools/api/central/ICentral.java
laeubi: It dosen't hurt to have those though...

bndtools.api/src/org/bndtools/api/central/ICentral.java
pkriens: > Our codebase hasn't made heavy use of the Optional pattern. But I suppose we probably should...

I've been using Result more and more. This has all the methods of Optional but

* it allows to carry a message when it fails,
* easy message construction with improved String.format,
* It allows lambdas that throw exceptions.

The idea is derived from Rust.

bndtools.api/src/org/bndtools/api/central/ICentral.java
pkriens: Maybe call it BndManager? Not only is it even shorter, it will probably be mainly used by non-workspace users?

And as always, <=>

pkriens:
First, I've not looked deep into your implementation, so please forgive me if I am not making sense. You've done very good work and I am ok to include this in the code as is. I am pretty sure it works.

My reaction was intuitive, I am hyper sensitive to very similar classes and multiple classes for what seems one implementation 'thing'. My I see it too simple, but in my head the facade problem should be extremely distinct from the domain code. If I have a BndtoolsBuilder, I want he code to be 99% about building. It is ok to implement an interface that helps the facade code, but this should be very minor. When I got started in my career, DeMarco and Yourdan were nagging about reducing coupling and increasing cohesion. I got coupling, but cohesion escaped me. Today, I see maximizing cohesion is actually the most important. I feel that in this case we can create a facade manager that is completely independent of the EP and all our domain code. And use it with a little interface needed for the state transfer between the old and the new service.

So the following is dumping my ideas. Again, it is up to you. I'd much rather have your solution in place than that you get upset with my obnoxious wise guy stuff. Feel free to ignore.

## Current State

An Extension Point (EP) is implicitly associated with a domain type `D`. The EP gets an implementation class `C` that implements/extends `D` or a factory class that can create a `D`. `C` can be used for multiple extension points and it can be instantiated multiple times.

## Problem

The goal is to _dynamically_ provide a `D`for a specific EP via the service registry/DS. 

Therefore, 

1. the EP must use a _facade_ that delegates all methods of D to a corresponding service,
2. Handle absence of a service when a request comes in
3. Handle actual state of `D` when services are replaced

## Analysis 

As far as I can see we can make 1 and 2 generic? Since the EP can only refer to a class name (no parameters from the EP!) the class name is all the information we have to identify the delegate service.

The biggest problem I see is 3, switching service. For example, the BndtoolsBuilder is a stateful object, it is associated with the project for example. We need to be able to transfer state from the prior service to the new service. This is domain specific.

## Solution

At build time, for each EP, we generate a class that implements IExecutableExtensionFactory using `-generate`. This generated factory class F can create a D and it will register a service Facade<D>. The name of F is used in the EP and is known to the delegate service so they can be matched up.

> Dynamic generation in runtime of the class F is not possible because we need a unique F for each EP in plugin.xml that uses these facades.

```
interface Facade<D> {
    Class<F> getType();
    void delegateTo(Delegate<D> serviceOrNull);
}
```
The service that can provide a D and is related to F must be registered as a Delegate<D>
```
interface Memento {}

interface DelegateInstance<D> extends AutoCloseable, Supplier<D> {
       Memento getState();
       D get();
}
interface Delegate<D> {
    Class<F> getType();
    DelegateInstance<D> create(@Nullable Memento state);
}
```
The Facade Manager will track all Facade and Delegate services. When a Facade and a Delegate with the same type F are registered, the Manager will create a delegate D object and inject it the facade, handling the state transfer with the memento. 

If the facade, an instance of F, is garbage collected, the subsystem will sometime later unregister the related Facade<D> service. If the Delegate<D> service unregisters, the facade is injected with null synchronously.

## Conclusion

As said, I am ok to go along with what you've done so far. I fully realize that my 'better' might be the enemy of good. The reason I outlined my solution is that you always seem eager to improve this stuff. I am willing to provide the Facade Subsystem if you think it is useful, it is pretty straightforward and has no Eclipse connections so testing is simple.

Up to you.



-------------------------------------------------

This is some experiments I did to figure out the API
```java
package bndtools;

import java.lang.ref.ReferenceQueue;
import java.lang.ref.WeakReference;
import java.util.HashMap;
import java.util.Iterator;
import java.util.Map;
import java.util.function.Supplier;

import org.eclipse.core.runtime.CoreException;
import org.eclipse.core.runtime.IExecutableExtensionFactory;
import org.eclipse.jdt.annotation.Nullable;
import org.osgi.service.component.annotations.Reference;
import org.osgi.service.component.annotations.ReferenceCardinality;
import org.osgi.service.component.annotations.ReferencePolicy;

import aQute.lib.io.IO;

@Component
@SuppressWarnings({
	"rawtypes", "unchecked"
})
public class FacadeManager {

	/**
	 * An interface signalling the object can provide a memento. In general,
	 * memento's should be used in the constructor so the set method is default.
	 */
	public interface Memento {
		/**
		 * Get an object from which the current state can be restored.
		 */
		Object getState();

		/**
		 * Restore the state as it was when the {@link #getState()} was called.
		 *
		 * @param state the state from {@link #getState()}
		 */
		default void restoreState(Object state) {

		}
	}

	/**
	 * A Facade can provide a constant object that forwards all calls to a
	 * delegate. A delegate can provide multiple instances for when the facade
	 * is used to create multiple instance.
	 *
	 * @param <D> the domain type
	 */
	public interface Facade<D> {
		/**
		 * The unique identity of the facade. In general this is the name of a
		 * factory class registered as an Extension Point.
		 */
		String getType();

		/**
		 * Set the delegate, can be null if there is no delegate
		 *
		 * @param delegate the delegate
		 */
		void delegateTo(@Nullable
		Delegate<D> delegate);
	}

	/**
	 * A Delegate can create managed objects.
	 *
	 * @param <D> the type of the managed objects
	 */
	public interface Delegate<D> {
		/**
		 * The name of the delegate, this should match the
		 * {@link Facade#getType()}
		 */
		String getType();

		/**
		 * Create a new managed instance. The memento can be used to restore the
		 * state
		 *
		 * @param memento the state, can be null
		 */
		DelegateInstance<D> create(@Nullable
		Object memento);
	}

	/**
	 * Provides a managed instance.
	 *
	 * @param <D> the domain type
	 */
	public static class DelegateInstance<D> implements Supplier<D>, AutoCloseable, Memento {
		final D domain;

		public DelegateInstance(D domain) {
			this.domain = domain;
		}

		@Override
		public Object getState() {
			if (domain instanceof Memento m)
				return m.getState();
			else
				return null;
		}

		@Override
		public void close() throws Exception {
			if (domain instanceof AutoCloseable c) {
				IO.close(c);
			}
		}

		@Override
		public D get() {
			return domain;
		}
	}

	/**
	 * Helper class for Factories. In general, factories should be auto
	 * generated, this is convenience to do the right thing
	 *
	 * @param <D> the domain type
	 */
	public static abstract class BaseFactory<D> implements IExecutableExtensionFactory, Facade<D> {
		final HashMap<WeakReference<D>, Binding>	instances		= new HashMap<>();
		final ReferenceQueue<D>						referenceQueue	= new ReferenceQueue<>();
		final Object								lock			= new Object();

		Delegate<D>									delegate;

		/*
		 * Dynamically binds the domain object to the delegate instance.
		 */
		class Binding implements Supplier<D> {
			Object							memento	= null;
			volatile DelegateInstance<D>	current;

			@Override
			public D get() {
				if (current == null)
					return null;
				return current.get();
			}

			// locked
			void setDelegate(Delegate<D> delegate) {
				if (current != null) {
					memento = current.getState();
					IO.close(current);
					current = null;
				}
				if (delegate != null)
					current = delegate.create(memento);
			}

			public void close() {
				memento = null;
				if (current != null) {
					IO.close(current);
				}
			}

		}

		@Override
		public String getType() {
			return getClass().getName();
		}

		@Override
		public void delegateTo(Delegate<D> delegate) {
			synchronized (lock) {
				this.delegate = delegate;
				instances.values()
					.forEach(binding -> binding.setDelegate(delegate));
			}
			purge();

		}

		@Override
		public Object create() throws CoreException {
			synchronized (lock) {
				Binding binding = new Binding();
				binding.setDelegate(delegate);
				D domain = createDelegator(binding);
				WeakReference wr = new WeakReference<>(domain, referenceQueue);
				instances.put(wr, binding);
				return domain;
			}
		}

		void purge() {
			java.lang.ref.Reference<? extends D> v;
			while ((v = referenceQueue.poll()) != null) {
				synchronized (lock) {
					Binding binding = instances.remove(v);
					if (binding != null)
						binding.close();
				}
			}

		}

		protected abstract D createDelegator(Supplier<D> delegatee);

	}

	final Map<Facade, Delegate>	facades		= new HashMap<>();
	final Map<String, Delegate>	delegates	= new HashMap<>();
	final Object				lock		= new Object();

	@Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC)
	void addFacade(Facade facade) {
		synchronized (lock) {
			Delegate delegate = delegates.get(facade.getType());
			facades.put(facade, delegate);
			facade.delegateTo(delegate);
		}

	}

	void removeFacade(Facade facade) {
		Delegate current;
		synchronized (lock) {
			current = facades.remove(facade);
		}
		facade.delegateTo(null);

	}

	@Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC)
	void addDelegate(Delegate delegate) {

		synchronized (lock) {

			String type = delegate.getType();
			Delegate previous = delegates.put(type, delegate);
			assert previous == null;

			Iterator<Map.Entry<Facade, Delegate>> it = facades.entrySet()
				.iterator();
			while (it.hasNext()) {
				var e = it.next();
				Facade facade = e.getKey();
				if (facade.getType()
					.equals(type)) {
					e.setValue(delegate);
					facade.delegateTo(delegate);
				}
			}
		}

	}

	void removeDelegate(Delegate delegate) {
		synchronized (lock) {

			Delegate previous = delegates.remove(delegate.getType());
			assert previous == delegate;

			Iterator<Map.Entry<Facade, Delegate>> it = facades.entrySet()
				.iterator();
			while (it.hasNext()) {
				var e = it.next();
				Facade facade = e.getKey();
				Delegate current = e.getValue();
				if (current == delegate) {
					e.setValue(null);
					facade.delegateTo(null);
				}
			}
		}

	}

}
```
And a very trivial test case:
```java
@SuppressWarnings("rawtypes")
class FacadeManagerTest {
	interface Domain {
		void set(String state);

		String get();
	}


	@Test
	void test() throws CoreException {


		//
		// auto generated source that delegates each method to the
		// delagatee
		//
		class DomainFactory extends BaseFactory<Domain> {

			@Override
			protected Domain createDelegator(Supplier<Domain> delegatee) {
				return new Domain() {

					@Override
					public String get() {
						return delegatee.get()
							.get();
					}

					@Override
					public void set(String state) {
						delegatee.get()
							.set(state);
					}
				};
			}

		}

		/**
		 * The class that provides the real workl
		 */
		class DomainImpl implements Domain {
			String state;

			@Override
			public String get() {
				System.out.println("get : " + state);
				return state;
			}

			@Override
			public void set(String state) {
				System.out.println("set : " + state);
				this.state = state;
			}

		}

		//
		// The registered service that provides the
		// Domain creation. Probably need a helper class.
		//
		class RealDomainService implements Delegate<Domain> {

			@Override
			public String getType() {
				return DomainFactory.class.getName();
			}

			@Override
			public DelegateInstance<Domain> create(Object memento) {
				Domain d = new DomainImpl();
				d.set((String) memento);

				return new DelegateInstance<>(d) {
					@Override
					public Object getState() {
						return get().get();
					}
				};
			}
		}

		FacadeManager fm = new FacadeManager();
		DomainFactory df = new DomainFactory();
		RealDomainService rds = new RealDomainService();

		fm.addFacade(df);
		assertThat(df.delegate).isNull();

		Object object = df.create();
		assertThat(object).isInstanceOf(Domain.class);

		Domain d1 = (Domain) object;

		fm.addDelegate(rds);
		assertThat(df.delegate).isNotNull();

		d1.set("s1");

		fm.removeDelegate(rds);
		assertThat(df.delegate).isNull();

		fm.addDelegate(rds);
		assertThat(df.delegate).isNotNull();

		assertThat(d1.get()).isEqualTo("s1");

		Domain d2 = (Domain) df.create();
		assertThat(d2.get()).isNull();

	}

}

```

pkriens:
If you want to move this forward I am ok. I removed my facade PR since I felt it was more blocking than doing anything.

pkriens:
Feel slightly bad due to my role in the other PR that I closed but it has been too long without activity. Please feel free to reopen it when you want to work on this again.

Issue #5025, opened 2022-01-04
Build ignoring upstream changes


pkriens:
I've got a project that creates a JAR from other projects. When any of the upstream projects change, the dist builder gets called. However, it never really builds since it cannot see any changes.

I can see that the Java builder gets a FULL_BUILD due to the BndContainer changing but when we get to the Bndtools Builder this has been lowered to an AUTO_BUILD. Since there is no Java code changing, the delta is empty.

I am thinking of checking the dates of the classpath files against the last build. However, this feels rather hackish. @bjhargrave  any ideas?

bjhargrave:
The BndContainer holds a lastModified value which is the lastModified value of the newest element in the container. I guess the Bndtools builder could compare that value against the last modified time of the project's `buildfiles` file and build if the container is newer than the `buildfiles` file.

kriegfrj:
Ive got my hands dep in this code at the moment. Some of my changes will make this easier to develop and test. If it is not urgent, I'm happy to tack a fix on for this too (it will be just as easy as handling the merge conflict...)

pkriens:
The #5038 fix seems to solve the problem for me. I just tried it out. Before the log clearly showed it was not build, with #5038 it does.

PR #5038, opened 2022-01-07
builder: Handle updates to classpath container elements

bjhargrave:
If an upstream classpath container element is updated, the builder
must rebuild the project.

Fixes https://github.com/bndtools/bnd/issues/5025

### DIFF
--- a/bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
+++ b/bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
@@ -12,6 +12,7 @@ import java.util.concurrent.TimeoutException;
 import org.bndtools.api.BndtoolsConstants;
 import org.bndtools.api.ILogger;
 import org.bndtools.api.Logger;
+import org.bndtools.builder.classpath.BndContainer;
 import org.bndtools.builder.classpath.BndContainerInitializer;
 import org.bndtools.builder.decorator.ui.PackageDecorator;
 import org.eclipse.core.resources.IMarker;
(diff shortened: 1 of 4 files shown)


kriegfrj:
I haven't looked too closely but this seems OK. But I'm confused as you assigned the issue to me? 😄

bjhargrave:
> But I'm confused as you assigned the issue to me? 😄

Yes. But we are planning to ramp down for a 6.2 release by the then of the month and I was not sure that your facade work for the builder would be ready in time. So I made this fix in the current code for 6.2.

bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java
```diff
@@ -231,8 +233,8 @@
 						force = true;
 					}
 
-					if (!force && delta.hasNoTarget(model)) {
-						buildLog.basic("project has no target files");
+					if (!force && delta.hasOutOfDateTarget(model, lastModifiedClasspathContainer(myProject))) {
+						buildLog.basic("project has no or out-of-date target files");
```
kriegfrj, line 237: Ok, I've had a bit of think about this and I have one minor nit. I prefer log messages to be precise. Rather than test for multiple conditions and issue a generic leg message when any of them fail, I think it's better to check each one individually and log an error message specific to the failure.

Log messages like these are only useful rarely. However, in those rare situations where they do come in useful, usually the more precise they are the more useful they become. Especially if they are the result of an intermittent failure that you can't readily reproduce. If a build is occasionally doing something unexpected and this message appears in the log, the person looking at it may wish that they knew whether a rebuild was triggered because the file was out-of-date, or else if it triggered because it didn't exist.

So my preference would be to check for "no target" and "out-of-date target" as two separate checks, and log each of them separately as appropriate.

bjhargrave: > I think it's better to check each one individually and log an error message specific to the failure.

But this is just duplicating the loop execution. It is the same loop to look for whether the target files are missing or out-of-date. While having a precise log message is nice, it is not worth duplicating code/work to get there. It is just a log message.

In a way, a missing file is a special case of being out-of-date (which is what the code in the PR reflects). So I would  be more inclined to change the log message to just say out-of-date and not mention missing.

bjhargrave:
@pkriens Does this fix the problem you mention in #5025?

pkriens:
Thanks! seems to work like a charm. of course now it is slower :-( Ah well, perfection has its price

PR #5044, opened 2022-01-08
Remove maven-compat dependency

### LINKED ISSUE #5042: Maven Plugins: Remove dependency on deprecated maven-compat
To reduce the download footprint and to get rid of using deprecated classes (also for making this compatible with the upcoming Maven 4) the plugin should not depend on the deprecated maven-compat artifact (compare with https://issues.apache.org/jira/browse/MNG-7266? and https://cwiki.apache.org/confluence/display/MAVEN/Plugin+migration+to+Maven3+dependencies) but use Resolver/Aether API directly.

kwin:
It is not longer in use
This closes #5042

### DIFF
--- a/maven/bnd-maven-plugin/pom.xml
+++ b/maven/bnd-maven-plugin/pom.xml
@@ -35,10 +35,6 @@
 			<artifactId>maven-artifact</artifactId>
 		</dependency>
 		<dependency>
-			<groupId>org.apache.maven</groupId>
-			<artifactId>maven-compat</artifactId>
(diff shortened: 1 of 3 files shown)


kriegfrj:
> It is not longer in use This closes #5042

Your sign-off is missing in the commit message. You need to run `git commit --amend -s` and then force-push.

bjhargrave:

kwin:
@bjhargrave Done

bjhargrave:
Thanks

PR #5048, opened 2022-01-11
Multiple identical plugins in plugins list

pkriens:
In a sub bundle project, the XML files for components
were generated multiple times. After debugging, I found
that the PluginContainer returned multiple instances
of the same plugin.

This seems to have been caused by using the PluginProvider
mechanism to get the plugins from the parent. Although
this could be fixed, the easiest (and seemingly safest)
solution was to just make the stream that returns the
plugins distinct.

@BJ you might want to take a further look to see if
the logic is causing the traversal of the parent
hierarchy multiple times?

Signed-off-by: Peter Kriens Peter.Kriens@aqute.biz


bjhargrave:
The issue with sub bundles was that the ProjectBuilder for a subbundle
has the main project's ProjectBuilder as its parent so the Builder
plugins are added at two levels of the hierarchy. Adding a call to `distinct()`
was necessary. I made a new PR to move the `distinct()` call so that it does not
happen multiple times (once at each level of the processor hierarchy).

PR #5075, opened 2022-01-28
launcher: Refactor to call calculatedProperties later

### LINKED ISSUE #5061: "Clean storage area before launch" has no (or not the expected) effect
With some previous version (before 6.1.0) I could successfully preserve changes made with Configuration Admin between launches by unchecking the "Clean storage area before launch" flag in the Eclipse Run/Debug configuration.

Using 6.1.0, the flag no longer works. Even if unchecked, the information maintained by Configuration Admin is lost (simply use felix with GoGo shell, enter some configuration information and check after restart). A workaround for preserving the information between launches is to add `-runkeep` to `bnd.bnd` (I'm not using a separate `*.bndrun` because this is just a small test project).

bjhargrave:
We no longer call calculatedProperties during updateFromProject since
that is too soon. We now call calculatedProperties from getRunProperties
so it is done just as it is needed.

For ProjectLauncher.calculatedProperties, we no longer set a number of
values since they are already handled in
ProjectLauncherImpl.getConstants as launcher properties which are
then set as framework properties by Launcher.createFramework. So we
then needed to modify RemoteProjectLauncherPlugin to override
calculatedProperties to set the framework properties.

Fixes #5061

### DIFF
--- a/biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
+++ b/biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
@@ -234,7 +234,7 @@ public class ProjectLauncherImpl extends ProjectLauncher {
 		lc.runbundles.addAll(runbundles);
 		lc.trace = getTrace();
 		lc.timeout = getTimeout();
-		lc.services = super.getRunFramework() == SERVICES ? true : false;
+		lc.services = getRunFramework() == SERVICES ? true : false;
 		lc.activators.addAll(getActivators());
 		lc.name = getProject().getName();
 		lc.activationEager = launcherInstrs.runoptions()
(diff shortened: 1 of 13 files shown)


biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
```diff
@@ -205,8 +207,7 @@ public void prepare() throws Exception {
 			return;
 		prepared = true;
 		super.prepare();
-
-		writeProperties();
+		calculatedProperties(getRunProperties());
```
kriegfrj, line 210: `getRunProperties()` returns by-reference, and `calculatedProperties()` modifies it in-place. Is this what we want? I noticed that `RemoteProjectLauncherPlugin` makes a copy of `runproperties` before passing it in to `calculatedProperties()`.

kriegfrj: I also noted that there is no code path in which `calculatedProperties()` is called from within an `update()` call. Is this by design, or an oversight?

bjhargrave: `getRunProperties()` is fine since `updateFromProject()` sets the `runproperties` from `getProject().getRunProperties()` which returns a new map. `RemoteProjectLauncherPlugin` is somewhat of a mystery to me, so I don't know why it does what it does. 🤷‍♂️ 



> I also noted that there is no code path in which `calculatedProperties()` is called from within an `update()` call. Is this by design, or an oversight?

Mistake on my part. I'll push an updated commit. Good catch!

PR #5084, opened 2022-02-04
[m2e] Multiple versions of an artifact not available in repository

### LINKED ISSUE #4903: [m2e] Multiple versions of an artifact not available in repository
I've been trying to do some (meta-)testing work in osgi-test.

As with the tester projects in Bnd, testing is tricky because you have two versions that you need to worry about: the version you compile against and the version that you run the tests against. In a regular testing scenario you normally just make these both the same (usually a recent version). But because osgi-test is comprised of testing utilities and we don't want to force users to be on the bleeding edge of JUnit, we compile against a lower version.

The problem is that the signature of a method I want to use in junit-platform-launcher has changed between 1.6.0 (current compile version) and 1.8.1 (current run version). (Technically this has not been properly semantically versioned but the method *is* marked as internal so I guess I get what I deserve to an extent.) So my code compiles but then fails with a `NoSuchMethodError` at runtime. I tried to fix this by forcing my bndrun resolve to include the compile version (lower version) rather than the run version.

This is where the problem comes in: unfortunately, the resolution fails. The resolver is unable to find 1.6.0 in the Maven Dependencies repository.

Possibly related: When I look under the `junit-platform-launcher` node in the `Repository` view, I see two nodes listed but both with version 1.8.1. There is no 1.6.0 node listed anywhere.

m2e duplicate version

This happens against HEAD.

rotty3000:
fixes #4903

### DIFF
--- a/bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
+++ b/bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
@@ -50,6 +50,7 @@ import aQute.bnd.osgi.repository.AbstractIndexingRepository;
 import aQute.bnd.osgi.repository.BridgeRepository;
 import aQute.bnd.osgi.repository.BridgeRepository.InfoCapability;
 import aQute.bnd.osgi.resource.CapReqBuilder;
+import aQute.bnd.osgi.resource.FilterBuilder;
 import aQute.bnd.osgi.resource.RequirementBuilder;
 import aQute.bnd.osgi.resource.ResourceBuilder;
 import aQute.bnd.osgi.resource.ResourceUtils;
@@ -119,13 +120,21 @@ public class MavenWorkspaceRepository extends AbstractIndexingRepository<IProjec
 		};
 
 		identificationAndVersionExpressionFunction = (bsn, version) -> {
-			RequirementBuilder builder = new RequirementBuilder(BND_INFO);
-			builder.addFilter("name", bsn, version.toString(), null);
 			return combiner.or( //
 				combiner.identity(
-					CapReqBuilder.createSimpleRequirement(IdentityNamespace.IDENTITY_NAMESPACE, bsn, version.toString())
+					new RequirementBuilder(IdentityNamespace.IDENTITY_NAMESPACE) //
+						.addFilter(new FilterBuilder().and()
+							.eq(IdentityNamespace.IDENTITY_NAMESPACE, bsn)
+							.eq(IdentityNamespace.CAPABILITY_VERSION_ATTRIBUTE, version)
+							.end())
 						.buildSyntheticRequirement()), //
-				combiner.identity(builder.buildSyntheticRequirement()));
+				combiner.identity( //
+					new RequirementBuilder(BND_INFO) //
+						.addFilter(new FilterBuilder().and()
+							.eq(Constants.NAME_ATTRIBUTE, bsn)
+							.eq("version", version)
+							.end())
+						.buildSyntheticRequirement()));
 		};
 	}
 
--- a/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/DependencyResolver.java
+++ b/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/DependencyResolver.java
@@ -47,7 +47,7 @@ public class DependencyResolver {
 	private static final Logger														logger					= LoggerFactory
 		.getLogger(DependencyResolver.class);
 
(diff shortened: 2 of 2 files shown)


bjhargrave (changes requested):

bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
```diff
@@ -119,13 +120,21 @@ public MavenWorkspaceRepository() {
...
 		identificationAndVersionExpressionFunction = (bsn, version) -> {
-			RequirementBuilder builder = new RequirementBuilder(BND_INFO);
-			builder.addFilter("name", bsn, version.toString(), null);
 			return combiner.or( //
 				combiner.identity(
-					CapReqBuilder.createSimpleRequirement(IdentityNamespace.IDENTITY_NAMESPACE, bsn, version.toString())
+					new RequirementBuilder(IdentityNamespace.IDENTITY_NAMESPACE) //
+						.addFilter(new FilterBuilder().and()
+						.eq(IdentityNamespace.IDENTITY_NAMESPACE, bsn)
+						.eq("version", version)
```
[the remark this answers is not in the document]
rotty3000: fixed

bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
```diff
@@ -119,13 +120,21 @@ public MavenWorkspaceRepository() {
...
+							.eq(IdentityNamespace.IDENTITY_NAMESPACE, bsn)
+							.eq(IdentityNamespace.CAPABILITY_VERSION_ATTRIBUTE, version)
+							.end())
 						.buildSyntheticRequirement()), //
-				combiner.identity(builder.buildSyntheticRequirement()));
+				combiner.identity( //
+					new RequirementBuilder(BND_INFO) //
+						.addFilter(new FilterBuilder().and()
+							.eq(Constants.NAME_ATTRIBUTE, bsn)
+							.eq(IdentityNamespace.CAPABILITY_VERSION_ATTRIBUTE, version)
```
bjhargrave, line 135: This is not quite correct since the bnd.info namespace is not the osgi.identity namespace. I don't think there is a constant defined for the bnd.info namespace capability version attribute, so we often just use a "version" constant. I think we should just do that here so that searches for `IdentityNamespace.CAPABILITY_VERSION_ATTRIBUTE` do not incorrectly return this place.

rotty3000: fixed ;)

bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
```diff
@@ -119,13 +120,21 @@ public MavenWorkspaceRepository() {
...
 		identificationAndVersionExpressionFunction = (bsn, version) -> {
-			RequirementBuilder builder = new RequirementBuilder(BND_INFO);
-			builder.addFilter("name", bsn, version.toString(), null);
 			return combiner.or( //
 				combiner.identity(
-					CapReqBuilder.createSimpleRequirement(IdentityNamespace.IDENTITY_NAMESPACE, bsn, version.toString())
+					new RequirementBuilder(IdentityNamespace.IDENTITY_NAMESPACE) //
+						.addFilter(new FilterBuilder().and()
+							.eq(IdentityNamespace.IDENTITY_NAMESPACE, bsn)
+							.eq("version", version)
```
bjhargrave, line 128:
```suggestion
							.eq(IdentityNamespace.CAPABILITY_VERSION_ATTRIBUTE, version)
```
This is the identity namespace.

bndtools.m2e/src/bndtools/m2e/MavenWorkspaceRepository.java
```diff
@@ -119,13 +120,21 @@ public MavenWorkspaceRepository() {
...
+							.end())
 						.buildSyntheticRequirement()), //
-				combiner.identity(builder.buildSyntheticRequirement()));
+				combiner.identity( //
+					new RequirementBuilder(BND_INFO) //
+						.addFilter(new FilterBuilder().and()
+							.eq(Constants.NAME_ATTRIBUTE, bsn)
+							.eq(IdentityNamespace.CAPABILITY_VERSION_ATTRIBUTE, version)
```
bjhargrave, line 135:
```suggestion
							.eq("version", version)
```

This is the `bnd.info` namespace.

bjhargrave (changes requested):

bjhargrave (approved):

PR #5085, opened 2022-02-04
[generate] Exception ignored

pkriens:
If a plugin threw an exception it was ignored

BJ, can this still be in 6.2?


bjhargrave:
> BJ, can this still be in 6.2?

Sure, I have not cut the release yet. Once this is merged I will cut it.

PR #5097, opened 2022-02-10
[resolve] Repository validation

pkriens:
Updating an older repository set can be painful. We already had the `bnd mbr` command to do some validation and updating of the mvn files. However, these commands lacked an overview.

This PR updates the `bnd resolve validate` command. It used to require a OBR index, and it can still accept it. However, without the index, it will assume that it is in a workspace. It will then resolve all files in the workspace and all configured repositories stand alone. 

Added the following options:

-x Print a requirement -> set resource for missing resources

-u print unused resources in the repository. That is, no workspace resource requires it.


rotty3000:
Interesting! I may in the future try to add a utility maven/gradle plugins that does this kind diagnostic report.

bjhargrave (commented):
You should also rebase on the latest master branch as your base branch is quite old.

PR #5116, opened 2022-02-17
Deadlock fixes

bjhargrave:


bndtools.core/src/bndtools/central/Central.java
```diff
@@ -224,54 +224,76 @@ public static Workspace getWorkspace() throws Exception {
...
 					// not the default, so switch the workspace to the default
-					cnfWorkspaceDeferred = promiseFactory().deferred();
-					ws.writeLocked(() -> {
-						ws.setFileSystem(Workspace.BND_DEFAULT_WS, Workspace.CNFDIR);
-						ws.forceRefresh();
-						ws.refresh();
-						ws.refreshProjects();
-						return null;
-					});
+					afterLock = Central::adjustWorkspace;
```
kriegfrj, line 250: Any reason not to collapse these into a single `if()` statement now?

PR #5122, opened 2022-02-18
[libg] Restore old package

pkriens:
When we moved to 6.1 (not a major release) we lost
the package aQute.lib.exceptions because we also
made it available via aQute.bnd.util.

This would've been a nice use case for my shading
code but until we've got something like that we cannot
just remove a package in a minor release.

This is causing so much pain in all my builds that
I am restoring it, unfortunately now having a 
duplicate. 

I am ok to deprecate one of them but this is
a giant pita.

This is the content of the package  at 5.3.0

### DIFF
--- a/aQute.libg/src/aQute/lib/exceptions/SupplierWithException.java
+++ b/aQute.libg/src/aQute/lib/exceptions/SupplierWithException.java
@@ -1,0 +1,59 @@
+package aQute.lib.exceptions;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.function.Supplier;
+
+/**
+ * Supplier interface that allows exceptions.
+ *
+ * @param <R> the result type
+ */
+@Deprecated() // move to package aQute.bnd.exceptions
+@FunctionalInterface
+public interface SupplierWithException<R> {
+	R get() throws Exception;
+
+	default Supplier<R> orElseThrow() {
+		return () -> {
+			try {
+				return get();
+			} catch (Exception e) {
+				throw Exceptions.duck(e);
+			}
+		};
+	}
+
+	default Supplier<R> orElse(R orElse) {
+		return () -> {
+			try {
+				return get();
+			} catch (Exception e) {
+				return orElse;
+			}
+		};
+	}
+
+	default Supplier<R> orElseGet(Supplier<? extends R> orElseGet) {
+		requireNonNull(orElseGet);
+		return () -> {
+			try {
+				return get();
+			} catch (Exception e) {
+				return orElseGet.get();
+			}
+		};
+	}
+
+	static <R> Supplier<R> asSupplier(SupplierWithException<R> unchecked) {
+		return unchecked.orElseThrow();
+	}
+
+	static <R> Supplier<R> asSupplierOrElse(SupplierWithException<R> unchecked, R orElse) {
+		return unchecked.orElse(orElse);
+	}
+
+	static <R> Supplier<R> asSupplierOrElseGet(SupplierWithException<R> unchecked, Supplier<? extends R> orElseGet) {
+		return unchecked.orElseGet(orElseGet);
+	}
+}
--- a/aQute.libg/src/aQute/lib/exceptions/BiConsumerWithException.java
+++ b/aQute.libg/src/aQute/lib/exceptions/BiConsumerWithException.java
@@ -1,0 +1,41 @@
+package aQute.lib.exceptions;
+
+import java.util.function.BiConsumer;
+
+/**
+ * BiConsumer interface that allows exceptions.
+ *
+ * @param <T> the type of the first argument
+ * @param <U> the type of the second argument
+ */
+@Deprecated() // move to package aQute.bnd.exceptions
+@FunctionalInterface
+public interface BiConsumerWithException<T, U> {
+	void accept(T t, U u) throws Exception;
+
+	default BiConsumer<T, U> orElseThrow() {
+		return (t, u) -> {
+			try {
+				accept(t, u);
+			} catch (Exception e) {
+				throw Exceptions.duck(e);
+			}
+		};
+	}
+
+	default BiConsumer<T, U> ignoreException() {
+		return (t, u) -> {
+			try {
+				accept(t, u);
+			} catch (Exception e) {}
+		};
+	}
+
+	static <T, U> BiConsumer<T, U> asBiConsumer(BiConsumerWithException<T, U> unchecked) {
+		return unchecked.orElseThrow();
+	}
+
+	static <T, U> BiConsumer<T, U> asBiConsumerIgnoreException(BiConsumerWithException<T, U> unchecked) {
+		return unchecked.ignoreException();
+	}
+}
--- a/aQute.libg/src/aQute/lib/exceptions/BiFunctionWithException.java
+++ b/aQute.libg/src/aQute/lib/exceptions/BiFunctionWithException.java
@@ -1,0 +1,63 @@
+package aQute.lib.exceptions;
+
+import static java.util.Objects.requireNonNull;
+
+import java.util.function.BiFunction;
+import java.util.function.Supplier;
+
+/**
+ * BiFunction interface that allows exceptions.
+ *
+ * @param <T> the type 1 of the argument
+ * @param <U> the type 2 of the argument
+ * @param <R> the result type
+ */
+@Deprecated() // move to package aQute.bnd.exceptions 
+@FunctionalInterface
+public interface BiFunctionWithException<T, U, R> {
+	R apply(T t, U u) throws Exception;
+
+	default BiFunction<T, U, R> orElseThrow() {
+		return (t, u) -> {
+			try {
+				return apply(t, u);
+			} catch (Exception e) {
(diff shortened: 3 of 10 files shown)


bjhargrave:
> When we moved to 6.1 (not a major release) we lost
> the package aQute.lib.exceptions because we also
> made it available via aQute.bnd.util.

This was done at the 6.0 release. A major release. 

https://github.com/bndtools/bnd/commit/0696e7ed184e101aa71c7aa1ea11220fbbe74312

bjhargrave:
> I am ok to deprecate one of them but this is
> a giant pita.

We must definitely deprecate the old lib version as we do not want to mistakenly use the old non-shared one in any Bnd code.

pkriens:
I can deprecate the classes in the old version.

bjhargrave:
> The beauty of OSGi

The majority of usage of Bnd is not in OSGi (Bndtools). It is maven and gradle plugins which are not API and thus have a flat class space where multiple different versions of packages began to cause real problems which is why we fixed this. This change was discussed and agreed to during the 6.0 development.

> However, the Exceptions class is not public API

The exception package is API since you use it in Result which is API. And Result is used on many Bnd API now. So by extension, the exception package is API. So jars/bundles cannot each have their own private/different copies of the package.

> The util bundle you introduced is an idea I always resisted since util bundles become a huge PITA in runtime because they tend to have lots of transitive dependencies and they tend to explode.

util, as it currently stands and is expected to remain, has no external dependencies. That is the point of this util jar. We must ensure this remains so.

> So I actually think the new package should be deprecated and its interfaces moved to a proper API project :-(
Why do you need to deprecate a package to move it to another jar? 

util is a collection of API (and is thus an "API" project) which many parts of Bnd (and other projects) depend on. We cannot just move it to a bnd API jar since I need to be able to use those classes and not all of bndlib which is very large now  (3.2MB). util is only a svelte 423K and is quite useful in many non-Bnd scenarios. I need to be able to use memoize, etc, outside of a Bnd context.

As we discussed and agreed to the creation of util for 6.0, we should also discuss and agree to unwinding that decision before we take action. And any breaking change like that needs also to come at a major version boundary (e.g. 7.0).

kriegfrj:
> util, as it currently stands and is expected to remain, has no external dependencies. That is the point of this util jar. We must ensure this remains so.
>

Perhaps this could be enforced with a regression test that checks to see that the Import-Package header of the build artifact is empty. This would prevent an external dependency from slipping in by accident.

Gave me the idea for osgi/osgi-test#464

kriegfrj:
> Perhaps this could be enforced with a regression test that checks to see that the Import-Package header of the build artifact is empty. This would prevent an external dependency from slipping in by accident.
> 
> Gave me the idea for [osgi/osgi-test#464](https://github.com/osgi/osgi-test/issues/464)

Could also add to this a check to ensure that the bundle remains "svelte" (now there's a work that gets far too little use in the English language... 😆 ) `assertThat(myBundle).isSvelte()`.

timothyjward:
> Perhaps this could be enforced with a regression test that checks to see that the Import-Package header of the build artifact is empty. This would prevent an external dependency from slipping in by accident.

This would need to be a little more sophisticated - once you move to R7’s use of `java.*` it will be impossible to have no dependencies for your class file. Possibly a restriction on JRE modules/api bundles needed to resolve would be the right thing?

kriegfrj:
> > Perhaps this could be enforced with a regression test that checks to see that the Import-Package header of the build artifact is empty. This would prevent an external dependency from slipping in by accident.
> 
> This would need to be a little more sophisticated - once you move to R7’s use of `java.*` it will be impossible to have no dependencies for your class file. Possibly a restriction on JRE modules/api bundles needed to resolve would be the right thing?

Absolutely - a decent implementation of osgi/osgi-test#464 would be flexible enough to allow the test writer to impose such criteria. If you can brainstorm any other use-cases, feel free to document them there and we'll try and come up with an implementation that's flexible enough to cater for them.

pkriens:
> However, the Exceptions class is not public API

I think you added that? 0696e7ed184e101aa71c7aa1ea11220fbbe74312

> As we discussed and agreed to the creation of util for 6.0, we should also discuss and agree to unwinding that decision before we take action. And any breaking change like that needs also to come at a major version boundary (e.g. 7.0).

I agree. So lets first restore the old lib package and then do some more pondering over what to deprecate or move.

I think I am leaning towards have 2 packages for now.

bjhargrave:
> I think I am leaning towards have 2 packages for now.

We would need to make sure, in the Bnd build, they we do not use the old exceptions package in our code by mistake. Such as modifying all `-conditionalpackage` instructions to exclude it. But there is no way to exclude it from the buildpath, so we can still make silly mistakes since Eclipse to always present us with two choices for each exceptions type. This is just setting ourselves up for mistakes.

I guess I am still unclear why we need to restore the old package. aQute.libg is not an API project. We never use it as such and always conditionally import it into our jars/bundles. So it cannot be considered API with any expectations of APIs like backwards compatibility. I have never had any compunction about breaking compatibility of code in aQute.libg to make it better since it is an internal implementation detail for other parts of Bnd. To the extent that you have source code using the old package, you are probably not using it in API signatures (since aQute.libg code is not API) and can easily just modify the source code to use the new package name. Why is this not a possible way forward?

If you really consider aQute.libg to be API, then we should elevate it properly to that status and import the packages rather than conditional package them.

pkriens:
> I guess I am still unclear why we need to restore the old package

Ever heard from breaking compatibility? We removed a package ... The whole purpose of baselining is to prevent this.

> old exceptions package in our code by mistake. 

I am not sure what the problem is? In the old name it will just be included? Who cares?

> (since aQute.libg code is not API)

We've exported the packages forever ...

> Why is this not a possible way forward?

Because it becomes a terrible mess. You need to make sure all your dependencies are updated in a synchronized way. Gee, a breaking change tends to do this.

> If you really consider aQute.libg to be API, then we should elevate it properly to that status and import the packages rather than conditional package them.

Well there are two things. External usage and internal usage. Externally I am perfectly happy to use it as a normal library or as conditional includes. Whatever, don't see why we would care. We published it and exported the packages so we have a commitment to be backward compatible.

Internally. We might consider this since we're now  breaking bndlib apart part by part. The primary purpose of the -conditionalpackage was to create a _leaf_, e.g. an application, that has no external dependencies. Bndlib is growing up and becoming a proper library. Maybe it is time to not have it include aQute.libg. I am ok with this. With the osgi packages gone, it seems to be the sensible thing to do.

It would then be nice if we had also then a JAR that contains all the dependencies that can then be used in bnd, gradle, testing, etc. 

> two choices for each exceptions type

Yeah that sucks.

If you want to make sure there is no aQute.lib included you could make a macro that generates an error if a package contains that name by using that macro in a manifest header? We have macro for the private packages I think?

bjhargrave:
> > I guess I am still unclear why we need to restore the old package
> 
> Ever heard from breaking compatibility? We removed a package ... The whole purpose of baselining is to prevent this.

Yes, but we do not treat the aQute.libg project as an API project in Bnd. Also, baselining does not prevent the removal of packages since we do not baseline Bundle-Version as this is mostly meaningless. Changing the major version of a package is done all the time in Bnd so we are OK with that level of breaking compatibility.

> > old exceptions package in our code by mistake.
> 
> I am not sure what the problem is? In the old name it will just be included? Who cares?

Well, obviously I care :-) And the issue is mistakes and duplication. I assume we all care about not making mistakes and duplicating things. If we are to have the old version of the package in the source tree, then it must be marked deprecated to help us avoid using it.

> > two choices for each exceptions type
> 
> Yeah that sucks.

It totally sucks. :-)

How about this. We leave things as they are now in 6.x (the one new package). At the 7.0 release, we restore the package to its original name and (1) we include it under the original name in the util jar, (2) all Bnd jars do not conditional package the exceptions package but import it since we use the package as API.

This means we have one package, it is under the original name, we treat it as API and share a single copy in the util jar in maven, gradle, Bndtools(OSGi). 

I think this gets us what we both want, but it means we break API at the major version boundary (7.0).

In the interim, you could put the old package in a personal jar for use your other work which needs the old package. Or you can put the old package in the build but it must be marked deprecated.

pkriens:
No, I really need to the old package name back. It is a surprising mess because you get conflicting problems in compile time, dependencies, and the runtime resolver. It is really badly broken, if it had an easy fix I would've done it. We broke compatibility and we should restore it.

For release 6 I guess I can survive having it being deprecated. I guess then we have time to think a bit further.

That said, I really would appreciate it if we did not remove packages **ever** without a proper discussion.

aQute.libg/src/aQute/lib/exceptions/SupplierWithException.java
```diff
@@ -0,0 +1,59 @@
...
+import static java.util.Objects.requireNonNull;
+
+import java.util.function.Supplier;
+
+/**
+ * Supplier interface that allows exceptions.
+ *
+ * @param <R> the result type
+ */
+@Deprecated() // move to package aQute.bnd.exceptions
```
bjhargrave, line 12: You need to mark all the types in the package as Deprecated.

PR #5134, opened 2022-02-22
Exceptions avoid old package

bjhargrave:


pkriens:
Aren't you exaggerating? I see no reason to make an exception for the exceptions package if it is deprecated. This is beyond ugly. We should try to keep some bnd file hygiene. This is a road to more and more messy stuff in the bnd files. if you have to do this on so many places you're doing something really wrong.

Can we please revert these bnd file changes?

bjhargrave:
I think this is helpful to use the tooling to prevent making mistakes in other Bnd projects. Generally no one will be concerned about this in the files since it is already in the files. I am not happy we have two copies of the package in the build, but we compromised and this is just tooling assistance to make sure we don't make mistakes.

Also, if we do move back to the single, old package in 7.0, then we still need the conditional package changes to ensure projects always import the package (from the util bundle) instead of conditional packaging it. So the conditional package changes to exclude the old package will always be necessary in our build. At that point we will remove the packages attribute from the buildpath.

pkriens:
> The buildpath changes make life much nicer (when using the latest snapshot bndtools), since they hide the older exceptions package from code completion to help avoid mistakenly using the old package.

In what way does it help? I can only see that now you have 2 warnings instead of one?

> if we do move back to the single, old package in 7.0

Lets cross that bridge when we get there. Basically we tell people now to get it from util. Removing the new package from aQute.lib in 7.0 is fine with me.

You know how much I dislike the extra, in my unnecessary cruft on the -buildpath .. the maven stuff is still an eye sore to me. And unless I am missing something, this does not provide anything essential.

bjhargrave:
> In what way does it help? I can only see that now you have 2 warnings instead of one?

Did you update to the very latest snapshot bndtools? It will hide the old package's class now during autocomplete, etc.

PR #5136, opened 2022-02-22
[analyzer] Bnd does not carry forward the attributes/directives of pa…

### LINKED ISSUE #5135: [analyzer] Bnd does not carry forward resolution mode of packages contained on the Bundle-ClassPath
Bnd provides a convenient behaviour of calculating the package imports required by classes found in containers on the Bundle-ClassPath. This is automatic behaviour which happens without manual intervention. However, bnd does not retain the resolution mode of the imported packages from containers on the BCP (except for the root container) during this process which means that the result is lossy.

While it is true that one can resolve this by manually intervening using any of the various bnd instructions to enrich the metadata, the default behaviour is not friendly, can be quite painful when the BCP is very large and could easily be lossless by default (where no intervention is required).

rotty3000:
…ckages contained on the Bundle-ClassPath

fixes #5135

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -8,6 +8,8 @@ import static java.util.stream.Collectors.toCollection;
 import static java.util.stream.Collectors.toList;
 import static java.util.stream.Collectors.toMap;
 import static java.util.stream.Collectors.toSet;
+import static org.osgi.framework.Constants.RESOLUTION_MANDATORY;
+import static org.osgi.framework.Constants.RESOLUTION_OPTIONAL;
 
 import java.io.File;
 import java.io.FileNotFoundException;
@@ -2614,6 +2616,11 @@ public class Analyzer extends Processor {
 	private boolean analyzeJar(Jar jar, String prefix, boolean okToIncludeDirs, String bcpEntry) throws Exception {
 		Map<String, Clazz> mismatched = new HashMap<>();
 
+		Parameters importPackage = Optional.ofNullable(jar.getManifest())
+			.map(Domain::domain)
+			.map(Domain::getImportPackage)
+			.orElseGet(() -> new Parameters());
+
 		next: for (String path : jar.getResources()
 			.keySet()) {
 			if (path.startsWith(prefix)) {
@@ -2663,7 +2670,8 @@ public class Analyzer extends Processor {
 						Set<PackageRef> refs = new LinkedHashSet<>(clazz.getReferred());
 						refs.addAll(referencesByAnnotation(clazz));
 						for (PackageRef p : refs) {
-							referred.put(p);
+							referred.compute(p,
+								(pRef, existing) -> mergeImportAttrs(existing, importPackage.get(pRef.getFQN())));
 						}
 						refs.remove(packageRef);
 						uses.addAll(packageRef, refs);
@@ -3764,6 +3772,29 @@ public class Analyzer extends Processor {
 	private void addDefinedContracts() {
 		Parameters definedContracts = getMergedParameters(Constants.DEFINE_CONTRACT);
 		contracts.collectContracts(Constants.DEFINE_CONTRACT, definedContracts);
+	}
+
+	private Attrs mergeImportAttrs(Attrs existing, Attrs imported) {
+		if (existing == null) {
+			if ((imported != null) && imported.containsKey(RESOLUTION_DIRECTIVE)) {
+				return Attrs.create(RESOLUTION_DIRECTIVE, imported.get(RESOLUTION_DIRECTIVE));
+			} else {
+				return new Attrs();
+			}
+		}
+		if (imported != null) {
+			String existingResolution = existing.get(RESOLUTION_DIRECTIVE);
+			String importResolution = imported.get(RESOLUTION_DIRECTIVE);
+
+			if (RESOLUTION_OPTIONAL.equals(existingResolution)) {
+				if (null == importResolution) {
+					existing.remove(RESOLUTION_DIRECTIVE);
+				} else if (RESOLUTION_MANDATORY.equals(importResolution)) {
+					existing.put(RESOLUTION_DIRECTIVE, RESOLUTION_MANDATORY);
+				}
+			}
+		}
+		return existing;
 	}
 
 }
--- a/biz.aQute.bndlib.tests/test/test/BuilderTest.java
+++ b/biz.aQute.bndlib.tests/test/test/BuilderTest.java
@@ -1,5 +1,6 @@
 package test;
 
+import static aQute.bnd.osgi.Constants.RESOLUTION_DIRECTIVE;
 import static aQute.bnd.test.BndTestCase.assertOk;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -23,6 +24,7 @@ import java.util.Collection;
 import java.util.Date;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.TreeSet;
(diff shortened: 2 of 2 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -2663,7 +2668,12 @@ private boolean analyzeJar(Jar jar, String prefix, boolean okToIncludeDirs, Stri
 						Set<PackageRef> refs = new LinkedHashSet<>(clazz.getReferred());
 						refs.addAll(referencesByAnnotation(clazz));
 						for (PackageRef p : refs) {
-							referred.put(p);
+							Attrs attrs = importPackage.get(p.getFQN());
+							if (attrs != null) {
+								referred.put(p, attrs);
```
bjhargrave, line 2673: The `put(PackageRef ref)` method only adds the ref if it is not already there, so it will not overwrite any previous attrs. The `put(PackageRef key, Attrs value)` method blindly overwrites what was already there. Since a given package may be split along the Bundle-Classpath, you end up with only the attrs of the last encountered package with attrs. So if a package exists more than once with attrs along the Bundle-Classpath, last attrs wins. 

So you may want to do

```java
referred.merge(p, false, attrs);
```

instead to merge all encountered attrs. 

This still may not be "right" since if one jar imports without resolution:=optional and another imports with resolution:=optional, the real answer is not resolution:=optional. But merging will end up with resolution:=optional unless the last jar imports with an explicit resolution:=mandatory. So even a blind merge is not correct. You may need to write a method for this which does an explicit merge on a defined set of Import-Package directives and attributes.

bjhargrave (changes requested):

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
rotty3000: :+1:

rotty3000:
@bjhargrave can you take another look at this please?

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -3766,4 +3773,31 @@ private void addDefinedContracts() {
...
+		if (existing == null) {
+			return (imported == null) ? new Attrs() : new Attrs(imported);
+		}
+		if (imported != null) {
+			if (!imported.containsKey(RESOLUTION_DIRECTIVE)) {
+				// Widen to implied "mandatory" since at least one consumer
+				// imports it as such
+				existing.remove(RESOLUTION_DIRECTIVE);
+			} else if (imported.containsKey(RESOLUTION_DIRECTIVE)
+				&& RESOLUTION_MANDATORY.equals(imported.get(RESOLUTION_DIRECTIVE))) {
```
bjhargrave, line 3786:
```suggestion
			} else if (RESOLUTION_MANDATORY.equals(imported.get(RESOLUTION_DIRECTIVE))) {
```

Why retest for `imported.containsKey(RESOLUTION_DIRECTIVE)`? We know the key is present as we are in the else side of the if.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -3766,4 +3773,31 @@ private void addDefinedContracts() {
...
+				// If resolution is "mandatory", override the existing value
+				existing.put(RESOLUTION_DIRECTIVE, "mandatory");
```
bjhargrave, line 3788:
```suggestion
				existing.put(RESOLUTION_DIRECTIVE, RESOLUTION_MANDATORY);
```

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -3766,4 +3773,31 @@ private void addDefinedContracts() {
...
+			}
+			if (imported.containsKey(VERSION_ATTRIBUTE) && !existing.containsKey(VERSION_ATTRIBUTE)) {
+				// Only add a version if absent (take the first version only)
+				existing.put(VERSION_ATTRIBUTE, imported.get(VERSION_ATTRIBUTE));
+			}
+
+			// Merge anything else
+			imported.stream()
```
bjhargrave, line 3796: I am not sure this is wise. Nor is it clear what this means. What if a jar has a mandatory attribute but the main bundle (`.`) does not? It seems wrong to take the union of all attributes and directives of all importers.

This whole merge thing is very messy. It lets the first BCP element set the import version value which may be less restrictive that needed by other BCP elements. If you want to be super precise, you need to keep intersecting encountered version ranges to compute the narrowest version range acceptable to all BCP entries (which could end up being the empty set!)

Resolution is easier since you can "throw away" the optionals. But for version what is the right answer? For other attributes and directives, I don't think we should be collecting them from BCP entries.

bjhargrave (commented):
I still think there are problems here.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
rotty3000: sure, fixed...

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
rotty3000: then I suggest this minimalist merge:
```
	private Attrs mergeImportAttrs(Attrs existing, Attrs imported) {
		if (existing == null) {
			return (imported == null) ? new Attrs() : new Attrs(imported);
		}
		if (imported != null) {
			if (!imported.containsKey(RESOLUTION_DIRECTIVE)) {
				// Widen to implied "mandatory" since at least one consumer
				// imports it as such
				existing.remove(RESOLUTION_DIRECTIVE);
			} else if (RESOLUTION_MANDATORY.equals(imported.get(RESOLUTION_DIRECTIVE))) {
				// If resolution is "mandatory", override the existing value
				existing.put(RESOLUTION_DIRECTIVE, RESOLUTION_MANDATORY);
			}
		}
		return existing;
	}
```
thoughts? the classloader will only use classes from the first package it finds, so that package should win version-wise in my opinion. The only change now is if a later entry on BCP has either implicit/explicit mandatory resolution; again, I think this is perfectly justifiable.

rotty3000: > classes from the first package it finds
even if the package is split across entries the first package really sets the bar and future ones have to be ready to blow up anyway. This is no different than flat classpath behaviour we have today.

bjhargrave: If exising is null, you only want the result to include the resolution directive from imported. So I think the first _if_ becomes:

```java
if (existing == null) {
  existing = new Attrs();
}
```

Then the rest only picks up the resolution directive.

bjhargrave: > the classloader will only use classes from the first package it finds, so that package should win version-wise in my opinion. 

Not true for BCP, you can split a package across the BCP entries. So the bundle class loader will search through all BCP entries for a class/resource. BCP is really a flat space inside the bundle.

rotty3000: that's exactly what I mean! but if there is anything in the first package don't you think that it's requirements need to be the defaults?

rotty3000: doing `new Attrs()`  loses the information from imports (since it returns), At least we need to add resolution directive if it exists.

rotty3000: we need at least:
```
	private Attrs mergeImportAttrs(Attrs existing, Attrs imported) {
		if (existing == null) {
			if ((imported != null) && imported.containsKey(RESOLUTION_DIRECTIVE)) {
				return Attrs.create(RESOLUTION_DIRECTIVE, imported.get(RESOLUTION_DIRECTIVE));
			} else {
				return new Attrs();
			}
		}
		if (imported != null) {
			if (!imported.containsKey(RESOLUTION_DIRECTIVE)) {
				// Widen to implied "mandatory" since at least one consumer
				// imports it as such
				existing.remove(RESOLUTION_DIRECTIVE);
			} else if (RESOLUTION_MANDATORY.equals(imported.get(RESOLUTION_DIRECTIVE))) {
				// If resolution is "mandatory", override the existing value
				existing.put(RESOLUTION_DIRECTIVE, RESOLUTION_MANDATORY);
			}
		}
		return existing;
	}

```

bjhargrave (approved):

PR #5139, opened 2022-02-23
[analyzer] clarify that precedence of manifest metadata happens when …

rotty3000:
…at-least-one user specified attribte is found

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -663,8 +663,8 @@ public class Analyzer extends Processor {
 		// The manifest has priority over the packageinfo or package-info.java
 		//
 
-		Attrs attrs = map.get(packageRef);
-		if (attrs != null && attrs.size() > 1)
+		if (MapStream.ofNullable(map.get(packageRef))
+			.anyMatch((k, __) -> !k.startsWith("-internal")))
 			return;
 
 		//


rotty3000:
cc @kriegfrj

pkriens (commented):
If you do this, at least make it a method with a readable name. I would add it to Parameters? You could make a filter method that removes key names using a glob expr?

The current method was simple but unclear what it meant, making it a method gives it a name and might be useful in other cases?

rotty3000:
sure, I could do that.

rotty3000:
a static method? Parameters isn't used here so seems a little weird to put it there.

pkriens:
map is a Packages object, same story because this is about Attrs. I would add something like:

     Attrs filter( Predicate<String> pred ) { ... }

Then make a predicate in Constants probably to hold the internal keys (and use it in Processor.skipPrint and any other places where need this decision.

I know it is always more work to do it system wide but it does save time in the future. I think ...

kriegfrj:
Perhaps make it `filterKeys()` in case at some point in the future you want a filter method that takes an Attrs or name- value pair?

rotty3000:
hey gents, this is such trivial code and I don't personally feel like there is much to gain from extracting this into a method nor is there an obvious bug here that we know of. Or maybe if we had more than one existing use case or it if was an itch I really needed to scratch then I might pursue it. However feel free to proceed with your ideas :)

bjhargrave (approved):

kriegfrj:
> hey gents, this is such trivial code and I don't personally feel like there is much to gain from extracting this into a method nor is there an obvious bug here that we know of. Or maybe if we had more than one existing use case or it if was an itch I really needed to scratch then I might pursue it. However feel free to proceed with your ideas :)

I'm not fussed if it stays how it is. The only other function that looks like it might sort-of do the same thing is `Processor.skipPrint()`, but that also skips some non-internal attributes so the set is not the same.

pkriens:
Sorry I am a pain in the ass but you even use a prefix string that is not even a constant! The original code might have had some excuse because it was before we had the internal properties but hard coded strings are not good. You fixed a bug because we had not defined what those internal properties really are and now you're making it actually worse imho.

To properly fix this we need a single place where we define internal properties. And as fr Krieg pointed out, there is already another place where we use this information. 

This is one of the most core methods in bnd.

PR #5143, opened 2022-02-24
[analyzer] Internal attributes not well defined

pkriens:
In  #5136 i observed that the fix suggested was not really helping to solve the confusion about the internal attributes. Internal attributes crept in as a quick hack to solve a problem but there were many similar problems. This is a proposal to define this concept properly. 

## Problem
It was observed that Analyzer decided to not learn a package if there was already an attribute set on that package in the manifest. However, over time lots and lots of 'internal' attributes entered the fray. 

## Solution
This patch defines the concept of internal attributes and an `isInternal` method and a toExternal to clean up an Attrs.

It also creates constant in Constants that holds all the internal attributes. However, isInternal will treat any attribute key that starts with an INTERNAL_PREFIX ('-internal-') as internal.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,3 @@ Archive
 /generated/
 /install-jdk.sh
 .to
-docs/.jekyll-metadata
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ Archive
 /generated/
 /install-jdk.sh
 .to
+docs/.jekyll-metadata
--- a/biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
@@ -629,8 +629,7 @@ public class Attrs implements Map<String, String> {
 		Attrs attrs = new Attrs();
 		forEach((k, v) -> {
 			if (predicate.test(k)) {
-				attrs.map.put(k, v);
-				attrs.types.put(k, this.types.get(k));
+				attrs.put(k, this.types.get(k), v);
 			}
 		});
 		return attrs;
--- a/biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
@@ -11,6 +11,7 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
 import java.util.function.BiConsumer;
+import java.util.function.Predicate;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -613,5 +614,25 @@ public class Attrs implements Map<String, String> {
 	public Attrs with(String key, String value) {
 		put(key, value);
 		return this;
+	}
+
+	/**
+	 * Return a new Attrs that has only the attributes that match the predicate.
+	 * The primary use case for this is AttributeClasses.
+	 *
+	 * @param predicate a predicate that returns true if the attribute must be
+	 *            included in the result
+	 * @return a new Attrs that can be used and modified by the caller
+	 */
+
+	public Attrs select(Predicate<String> predicate) {
+		Attrs attrs = new Attrs();
+		forEach((k, v) -> {
+			if (predicate.test(k)) {
+				attrs.map.put(k, v);
+				attrs.types.put(k, this.types.get(k));
+			}
+		});
+		return attrs;
 	}
 }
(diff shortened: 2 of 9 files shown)


pkriens:
I've made a proposal #5143 that should replace this PR

biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
```diff
@@ -534,6 +535,31 @@
 	String		LAUNCH_ACTIVATORS							= "launch.activators";
 	String		LAUNCH_ACTIVATION_EAGER						= "launch.activation.eager";
 
+	/**
+	 * Any attributes that should be removed from the attributes before
+	 * printing.
+	 */
+	String[]	INTERNAL_ATTRIBUTES							= {
```
bjhargrave, line 542:
```suggestion
	Set<String>	INTERNAL_ATTRIBUTES							= Sets.of(
```

Can we make this `Sets.of`? This will catch duplicate entry mistakes and will improve lookup time during execution.

kriegfrj:
I feel that this PR conflates two different types of "internal" attribute. As I see it, they are:

1. Attributes that are not part of Bnd's public API (ie, are not legal to be used in a `bnd.bnd` file). These only exist because they are generated by internal Bnd processes (eg, Builder and Analyzer). Those prefixed with `-internal` fall into this category.
2. Attributes that are not part of OSGi but are part of Bnd's public API. These are intended to be used in `bnd.bnd` files. Examples are `-split-package`, `-noimport`, `-from`.

I'm not sure if these two different classes should always be handled the same way.

pkriens:
That is why we need this abstraction. Clearly for the current goal the bnd attributes like -split-package should be discounted? I.e. if someone specified -split-package in the bnd file, it should not be counted as taking over the version from the package files? 

I theoretically agree with you but I think it is also awfully subtle. That said, I could make it based on an enum that has EXTERNAL, BND_USE, INTERNAL. That would force users to think about it. In our particular case, I am not sure if it is worth it. If we need more attribute classes then we can easily add appropriate API at the time we need it. And I've already added too many things to bnd with a future in mind that never arrived.

Do you have a use case in mind where you want to distinguish between these classes beyond and above what is there now?

biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
pkriens: I am not sure how it catches duplicates? It will silently ignore them? I also seriously doubt it is faster for these short lists and they add to initialization time ... but I've updated them.

.gitignore
```diff
@@ -8,4 +8,5 @@ Archive
 /reports/
 /generated/
 /install-jdk.sh
-.to
\ No newline at end of file
+.to
+docs/.jekyll-metadata
```
bjhargrave, line 12: docs already has its own .gitignore, so this should go there to keep all the docs related gitignore information together.

biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
```diff
@@ -614,4 +615,24 @@ public Attrs with(String key, String value) {
...
+	 *            included in the result
+	 * @return a new Attrs that can be used and modified by the caller
+	 */
+
+	public Attrs select(Predicate<String> predicate) {
+		Attrs attrs = new Attrs();
+		forEach((k, v) -> {
+			if (predicate.test(k)) {
+				attrs.map.put(k, v);
+				attrs.types.put(k, this.types.get(k));
```
bjhargrave, line 633:
```suggestion
				attrs.put(k, getType(k), v);
```

This uses the new put method which take a type.

biz.aQute.bndlib/src/aQute/bnd/osgi/AttributeClasses.java
```diff
@@ -0,0 +1,49 @@
...
+		public boolean test(String key) {
+			return key.startsWith(Constants.INTERNAL_PREFIX);
+		}
+	},
+	/**
+	 * Attributes set by the user but solely with the purpose to control bnd
+	 * processing. For example {@value Constants#SPLIT_PACKAGE_DIRECTIVE}. These
+	 * attributes must not end up in the manifest.
+	 */
+	BND_USE {
```
bjhargrave, line 43: Maybe this could just be `BND` or `TOOL`? INTERNAL is Bnd internal.

kriegfrj:
> That is why we need this abstraction. Clearly for the current goal the bnd attributes like -split-package should be discounted? I.e. if someone specified -split-package in the bnd file, it should not be counted as taking over the version from the package files?
> 

Yes,  I tend to agree. 

> I theoretically agree with you but I think it is also awfully subtle. That said, I could make it based on an enum that has EXTERNAL, BND_USE, 
> Do you have a use case in mind where you want to distinguish between these classes beyond and above what is there now?

No. I sympathise if you think it's overkill. Just wanted to make sure we considered it.

I also think we need a test that fails without this fix and passes with it.

PR #5155, opened 2022-03-02
More decoration

bjhargrave:
Use centralized decoration to thin out the individual bnd files.


pkriens:
Looks nice & clean! Thanks. Maybe write a tip topic on bnd.discourse.group?

Issue #5160, opened 2022-03-04
m2e: BndConfigurator does not decorate packages when using new jar/test-jar goals


bjhargrave:
The BndConfigurator class was decorating packages in Eclipse when invoking the jar and test-jar goals of maven-jar-plugin for use of the bnd-process/bnd-process-tests goals of the bnd-maven-plugin.

But when switching to use the jar/test-jar goals of bnd-maven-plugin, the package decoration is no longer running. So any existing decorations are persistently stuck and not updated.

PR #5161, opened 2022-03-05
m2e: BndConfigurator does not decorate packages when using new jar/test-jar goals #5160 

rotty3000:


bndtools.m2e/src/bndtools/m2e/BndConfigurator.java
```diff
@@ -329,21 +334,40 @@ private void execJarMojo(final IMavenProjectFacade projectFacade, IProgressMonit
...
+	public void scheduleDecorate(final IMavenProjectFacade projectFacade) {
+		final IProject project = projectFacade.getProject();
+
+		Job job = new WorkspaceJob("Executing " + project.getName() + " Bndtools decoration") {
+			@Override
+			public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
+				try {
+					SubMonitor progress = SubMonitor.convert(monitor, 1);
+
+					MavenProject mavenProject = getMavenProject(projectFacade, monitor);
```
bjhargrave, line 352: Should `progress` be used instead of `monitor` as the argument here? I guess an alternate question is whether we need a SubMonitor here?

rotty3000: yeah, you're right. Fixing.

bjhargrave (approved):

PR #5164, opened 2022-03-09
workspace: Add methods for write locking downgrading to read locking

bjhargrave:
These method take a callable and a function. The callable is called
while holding the write lock. The lock is downgraded to the read lock
and then the function is called with the result of the callable. The
result of the function is finally returned after releasing the read
lock.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceLock.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceLock.java
@@ -19,6 +19,7 @@ import java.util.stream.Collectors;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import aQute.bnd.exceptions.FunctionWithException;
 import aQute.lib.strings.Strings;
 
 /**
@@ -101,6 +102,30 @@ final class WorkspaceLock extends ReentrantReadWriteLock {
 				getQueuedThreads()));
 		e.initCause(getOwnerCause());
 		return e;
+	}
+
+	<T, U> T writeReadLocked(final long timeoutInMs, final Callable<U> underWrite,
+		final FunctionWithException<U, T> underRead, final BooleanSupplier canceled) throws Exception {
+		Callable<U> writeLocked = () -> {
+			U writeResult = underWrite.call();
+			// It is safe to let an exception propagate from here.
+			// Downgrade lock by taking read lock before returning to release
+			// write lock.
+			trace("Downgrading", writeLock());
+			readLock().lock();
+			return writeResult;
+		};
+		U writeResult = locked(writeLock(), timeoutInMs, writeLocked, canceled);
+		// If the call to locked completed normally, we hold the read
+		// lock obtained in the writeLocked callable.
+		try {
+			Callable<T> readLocked = () -> underRead.apply(writeResult);
+			return locked(readLock(), timeoutInMs, readLocked, canceled);
+		} finally {
+			// Must unlock the read lock taken above when we held the write
+			// lock.
+			readLock().unlock();
+		}
 	}
 
 	<T> T locked(final Lock lock, final long timeoutInMs, final Callable<T> callable, final BooleanSupplier canceled)
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
@@ -59,6 +59,7 @@ import aQute.bnd.annotation.plugin.BndPlugin;
 import aQute.bnd.build.WorkspaceNotifier.ET;
 import aQute.bnd.build.api.OnWorkspace;
 import aQute.bnd.exceptions.Exceptions;
+import aQute.bnd.exceptions.FunctionWithException;
 import aQute.bnd.exporter.executable.ExecutableJarExporter;
 import aQute.bnd.exporter.runbundles.RunbundlesExporter;
 import aQute.bnd.header.Attrs;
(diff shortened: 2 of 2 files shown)


biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceLock.java
```diff
@@ -103,6 +104,30 @@ private TimeoutException timeout(Lock lock) {
...
 	}
 
+	<T, U> T writeReadLocked(final long timeoutInMs, final Callable<U> underWrite,
+		final FunctionWithException<U, T> underRead, final BooleanSupplier canceled) throws Exception {
+		Callable<U> writeLocked = () -> {
+			U writeResult = underWrite.call();
+			// It is safe to let an exception propagate from here.
+			// Downgrade lock by taking read lock before returning to release
+			// write lock.
+			trace("Downgrading", writeLock());
```
pkriens, line 114: not the right comment is it? It is taking a read lock

biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceLock.java
```diff
@@ -103,6 +104,30 @@ private TimeoutException timeout(Lock lock) {
...
+			readLock().lock();
+			return writeResult;
+		};
+		U writeResult = locked(writeLock(), timeoutInMs, writeLocked, canceled);
+		// If the call to locked completed normally, we hold the read
+		// lock obtained in the writeLocked callable.
+		try {
+			Callable<T> readLocked = () -> underRead.apply(writeResult);
+			return locked(readLock(), timeoutInMs, readLocked, canceled);
```
pkriens, line 123: ? We know we have a read lock, why would we try to lock it again? Unless canceled, we can directly call the underRead lambda? We do not need 2 read locks? 

The timeout is now questionable, would it need to be adjusted? However it should always succeed since we have a read lock.

biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceLock.java
```diff
@@ -103,6 +104,30 @@ private TimeoutException timeout(Lock lock) {
...
+	<T, U> T writeReadLocked(final long timeoutInMs, final Callable<U> underWrite,
+		final FunctionWithException<U, T> underRead, final BooleanSupplier canceled) throws Exception {
+		Callable<U> writeLocked = () -> {
+			U writeResult = underWrite.call();
+			// It is safe to let an exception propagate from here.
+			// Downgrade lock by taking read lock before returning to release
+			// write lock.
+			trace("Downgrading", writeLock());
+			readLock().lock();
+			return writeResult;
```
pkriens, line 116: Although I cannot see it can happen, it bugs me that an exception in the remainder of the return path would lose a read lock. I know it is not necessary looking at the code but I would call the `writeLocked` lambda in the `try/finally` that use for the `readLocked` and check a flag if we got/did not get the lock. I agree it is very defensive and not necessary :-)

biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceLock.java
```diff
@@ -103,6 +104,30 @@ private TimeoutException timeout(Lock lock) {
...
 
+	<T, U> T writeReadLocked(final long timeoutInMs, final Callable<U> underWrite,
+		final FunctionWithException<U, T> underRead, final BooleanSupplier canceled) throws Exception {
+		Callable<U> writeLocked = () -> {
+			U writeResult = underWrite.call();
+			// It is safe to let an exception propagate from here.
+			// Downgrade lock by taking read lock before returning to release
+			// write lock.
+			trace("Downgrading", writeLock());
+			readLock().lock();
```
pkriens, line 115: I see this is the recommended way but is it guaranteed not to block in all lock modes? Looks a bit scary. I think it would be nicer to use `tryLock()` to show (and guarantee) that we do not block here. An assert could be done on the return value to be true.

pkriens (changes requested):
Some comments but they are mostly style issue that I am ok to ignore. However, I think calling the readLocked lambda via the locked method is wrong, we already got a read lock.

biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceLock.java
bjhargrave: We are downgrading the write lock which is the current lock being held to a read lock. So the comment is correct.

biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceLock.java
bjhargrave: This is using the existing locked method which has tracing, interrupt handling etc. already in place. Better to use the one method rather than another code path. Taking 2 counts on the read lock is no big deal. The timeout is meaning less as we already hold the read lock, so we well not be delaying in getting it again.

biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceLock.java
bjhargrave: There is nothing left in the return path which would throw an exception and lose the read lock. I carefully checked this.

biz.aQute.bndlib/src/aQute/bnd/build/WorkspaceLock.java
bjhargrave: If we cannot trust the documented way to downgrade a write lock, we are doomed. What would such an assert message say? "Java lied to us" :-)

bjhargrave:
I merged this. Please do try it in your use cases and if there are issues, I will debug them!

Issue #5081, opened 2022-02-03
bnd-maven-plugin: Allow resolving -includeresource instructions referencing jars built from other Eclipse Workspace projects


kwin:
If you have multiple projects inside Eclipse and project B has a dependency towards project A via the instruction 

```
@<JAR name of project A.jar>!/<some class name in A>
```

this leads to the following m2e error when the JAR file has not yet been built:

```
Can not find JAR file '<JAR name of project A.jar>' (biz.aQute.bnd:bnd-maven-plugin:6.1.0:bnd-process:generate-osgi-metadata:process-classes)

org.apache.maven.plugin.MojoFailureException: Can not find JAR file '<JAR name of project A.jar>'
	at aQute.bnd.maven.plugin.AbstractBndMavenPlugin.reportErrorsAndWarnings(AbstractBndMavenPlugin.java:661)
	at aQute.bnd.maven.plugin.AbstractBndMavenPlugin.execute(AbstractBndMavenPlugin.java:483)
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:137)
	at org.eclipse.m2e.core.internal.embedder.MavenImpl.execute(MavenImpl.java:336)
	at org.eclipse.m2e.core.internal.embedder.MavenImpl.lambda$8(MavenImpl.java:1423)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.executeBare(MavenExecutionContext.java:182)
	at org.eclipse.m2e.core.internal.embedder.MavenExecutionContext.execute(MavenExecutionContext.java:117)
	at org.eclipse.m2e.core.internal.embedder.MavenImpl.execute(MavenImpl.java:1422)
	at org.eclipse.m2e.core.project.configurator.MojoExecutionBuildParticipant.build(MojoExecutionBuildParticipant.java:55)
	at org.eclipse.m2e.core.internal.builder.MavenBuilderImpl.build(MavenBuilderImpl.java:135)
	at org.eclipse.m2e.core.internal.builder.MavenBuilder$1.method(MavenBuilder.java:169)
	at org.eclipse.m2e.core.internal.builder.MavenBuilder$1.method(MavenBuilder.java:1)
	at org.eclipse.m2e.core.internal.builder.MavenBuilder$BuildMethod.lambda$1(MavenBuilder.java:114)
[...]

PR #5175, opened 2022-03-15
bnd-maven-plugin: Set alternate source for classpath directories

bjhargrave:
In command line maven, maven always supplies artifact files which are
the built jar. Either in m2 or in a project's target folder.

However, in m2e, probably since the jar goal is not normally executed,
the artifact file for another project the reactor is for the project's
target/classes folder. This causes jar name matching in
-includeresource to fail since it matched in the file name and all the
file names for these project artifacts is "classes".

So we change the plugin to supply the directory classpath entries with
a file name which is synthesized from the artifact information. This
will then be matched to -includeresource references in m2e as well as
command line maven.

Fixes https://github.com/bndtools/bnd/issues/5081

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
@@ -1061,6 +1061,10 @@ public class Jar implements Closeable {
 		return source;
 	}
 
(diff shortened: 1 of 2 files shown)


rotty3000 (approved):
LGTM

PR #5176, opened 2022-03-16
WIP:[maven] use outputDir and waroutputDir correctly

stbischof:
The new `jar`and `testjar`goals / BndMavenPackagingTestsPlugin and BndMavenPackagingPlugin could be configures according the documentation using the confguration-propertiers `outputDir` and `waroutputDir`. But this properties are not used to save the jars in this position.

TODO: 
- test that shows that a not default `outputDir` and `waroutputDir` work properly.
- since `waroutputDir` and `outputDir` never worked we can decide to use only `outputDirectory` for both like in maven-jar-plugin and `maven-war-plgin`


bjhargrave:
It is unclear to me what problem you are solving. There is no issue describing the problem and there is no text in the pr or commit which describes the purpose of the change. The PR is also lacking tests for the change.

bjhargrave:
A war file never goes in the web app directory. This is only for the exploded wab. What does appear to be missing is the exploding of the wab into the web app directory directory when using a packaging goal. I will make a new PR for that.

bjhargrave:
See #5310.

PR #5181, opened 2022-03-21
[exporter] Add vmargs and program args to start/start.bat

### LINKED ISSUE #5180: [exporter] Jar exporter does not render vmargs
It would be helpful if the Jar exporter would add VM args to the `start`/`start.bat` scripts that are included in the exported Jar. Will give better fidelity between launching a `.bndrun` from within Bndtools and running the standalone program directly.

kriegfrj:
If you have `-runvm` and/or `-runprogramargs` settings that you need to launch your application, this fix will make sure that they are added to the command line in `start`/`start.bat`.

This helps to preserve the launching fidelity when comparing launching under Bndtools vs launching the exported application.

Fixes #5180.


kriegfrj:
Note: proposed PR adds support for `-runvm` *and* `-runprogramargs`.

bjhargrave (approved):

Issue #5178, opened 2022-03-20
[bndlib] ResourceBuilder fails on bundle versions with leading whitespace


kriegfrj:
I was working with a P2 repo that contained artifacts that had a leading space in the Bundle-Version. Ie: 
<pre>Bundle-Version:  1.2.3</pre>

This manifested as the artifacts not being silently omitted from the repository. (More precisely: the reason for their omission was logged in the slf4j log but this is not typically visible to regular users - only developers.)

There are two separate issues here:
1. That the failure was silent from the end-user's perspective. A visual indication that something had gone wrong would have been more appropriate. This issue is (probably) specific to the P2 implementation.
2. That it failed in the first place. Correct me if I am wrong, but there is nothing in this which violates the OSGi spec, so we should handle it more gracefully. This issue does not seem to be specific to the P2 implementation.

For me it manifested here:
https://github.com/bndtools/bnd/blob/eb0cfd9baf04ddc1b848a5957f3d62b4eee461c8/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java#L227-L230

This IAE was caught and logged by the P2 implementation but no further feedback was given (hence the first sub-issue).

And also here:
https://github.com/bndtools/bnd/blob/eb0cfd9baf04ddc1b848a5957f3d62b4eee461c8/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/CapReqBuilder.java#L675-L680

A quick fix which seems to have fixed my problem was to add a `.trim()` to these strings in the two code points listed above before handing them off to `Version::new` or `Version::isVersion`. But I'm not sure if this is the most appropriate approach as there may be other code paths that are affected and a more holistic solution might be required.

* Should `getBundleVersion()` be doing the `trim()` before it returns the value? Or do we need it to return the raw value?
* Should `Version(String)` do a `trim()` on the supplied value? This is likely to fix most code paths, but will it break others that might be relying on the current "strict" behaviour of this constructor?
* Any other considerations that my limited experience/exposure has not anticipated?

kriegfrj:
> 1. That the failure was silent from the end-user's perspective. A visual indication that something had gone wrong would have been more appropriate. This issue is (probably) specific to the P2 implementation.

I have spun this out into its own issue (#5179) so that this one can be focused on fixing `ResourceBuilder`/bndlib.

bjhargrave:
[...]

PR #5182, opened 2022-03-21
Version parsing in resource creation

bjhargrave:
Fixes https://github.com/bndtools/bnd/issues/5178


kriegfrj:
👍 Regression test? 😄

bjhargrave:
> Regression test?

Would be nice if you have time :-)

PR #5193, opened 2022-03-28
Switch to not generate REQ `osgi.extender=osgi.serviceloader.registrar`

stbischof:
in some cases you want use the features of the `@ServiceProvicer`
annotation but not generate the requirement for
`osgi.serviceloader.registrar`.

If there are better solutions for this please give me a hint.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -63,12 +63,14 @@ import aQute.bnd.osgi.Clazz.JAVA;
 import aQute.bnd.osgi.Descriptors.Descriptor;
 import aQute.bnd.osgi.Descriptors.PackageRef;
 import aQute.bnd.osgi.Descriptors.TypeRef;
+import aQute.bnd.osgi.resource.ResourceUtils;
 import aQute.bnd.service.AnalyzerPlugin;
 import aQute.bnd.service.classparser.ClassParser;
 import aQute.bnd.signatures.ClassSignature;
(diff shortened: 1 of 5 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -1118,6 +1120,21 @@ public Manifest calcManifest() throws Exception {
 				requirements.add(ExecutionEnvironmentNamespace.EXECUTION_ENVIRONMENT_NAMESPACE, attrs);
 			}
 
+			if (isTrue(getProperty(NOSERVICELOADERREGISTRAR))) {
```
bjhargrave, line 1123: This would be the wrong place. You want to stop it from being added rather that delete it after the fact. Also, this would remove any user defined caps from the bnd file.

bjhargrave (changes requested):
I would think it would be much nicer to add some option to the annotation so the user of the annotation can control what each use does. Putting another -noXX instruction seems rather coarse.

bjhargrave:
> in some cases you want use the features of the `@ServiceProvicer`
> annotation but not generate the requirement for
> `osgi.serviceloader.registrar`.

Can you be more clear here. Do you mean you just want META-INF/services resource generation?

stbischof:
```The resolution policy of the {@code osgi.extender} requirement.
A mandatory requirement forbids the bundle to resolve when this
requirement is not satisfied; an optional requirement allows a bundle to
resolve even if this requirement is not satisfied.
If not specified, the {@code resolution} directive is omitted from the
requirement clause.```

so i can use `Resolution.OPTIONAL;`

PR #5205, opened 2022-04-08
Callable for main thread will not kill framework on return code 198

juergen-albert:
In some cases we need to run things on the main thread, without shutting down the framework afterwards. This should do the trick and allow for repeated use of this trick.


demo/src/test/TestActivator.java
```diff
@@ -222,6 +225,52 @@ public Integer call() throws Exception {
...
+									@Override
+									public Integer call() throws Exception {
+										System.err.println("In main, return 500");
+										return 500;
+									}
+
+								};
+								Properties props = new Properties();
+								props.setProperty("main.thread", "true");
+								System.err.println("regsitering first main callable round");
```
bjhargrave, line 252: Isn't this the second round?

demo/src/test/TestActivator.java
```diff
@@ -222,6 +225,52 @@ public Integer call() throws Exception {
...
+				}
+
+			};
+
+			Properties props = new Properties();
+			props.setProperty("main.thread", "true");
+			System.err.println("regsitering first main callable round");
+			ServiceRegistration registerService = context.registerService(Callable.class.getName(), r,
+				(Dictionary) props);
+			System.err.println("registered first main callable round");
```
bjhargrave, line 270: This is a duplicate output

bjhargrave (changes requested):

demo/src/test/TestActivator.java
```diff
@@ -222,6 +225,52 @@ public Integer call() throws Exception {
...
+							}
+						});
+					}
+				}
+
+			};
+
+			Properties props = new Properties();
+			props.setProperty("main.thread", "true");
+			System.err.println("regsitering first main callable round");
```
bjhargrave, line 267: Spelling error

biz.aQute.launcher/src/aQute/launcher/Launcher.java
```diff
@@ -464,6 +448,31 @@ private int launch(String args[]) throws Throwable {
...
+			// report(System.err);
+			Integer code = mainThread.call();
+			if (code == null || code != LauncherConstants.RELISTEN_FOR_MAIN_CALLABLE) {
+				trace("main return, code %s", code);
+				return code;
+			}
+			synchronized (this) {
+				mainThread = null;
+			}
+			return handleMainCallable();
```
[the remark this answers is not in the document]
juergen-albert: do you mean a construct like this:
```
	private Integer handleMainCallable() throws Exception {
		boolean repeat = true;
		while (repeat) {
			synchronized (this) {
				while (mainThread == null && !restart) {
					trace("will wait for a registered Runnable or Callable");
					wait();
				}
			}
			if (mainThread != null) {
				trace("will call main");
				// report(System.err);
				Integer code = mainThread.call();
				if (code == null || code != LauncherConstants.RELISTEN_FOR_MAIN_CALLABLE) {
					trace("main return, code %s", code);
					return code;
				}
				synchronized (this) {
					mainThread = null;
				}
			} else {
				repeat = false;
			}
		}
		return null;
	}
```

Out of curiosity: What is the issue with a recursive call?

demo/src/test/TestActivator.java
juergen-albert: true

demo/src/test/TestActivator.java
juergen-albert: done

demo/src/test/TestActivator.java
juergen-albert: Yes and no. It helps figuring out the order of things in this highly async process.

biz.aQute.launcher/src/aQute/launcher/Launcher.java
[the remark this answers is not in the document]
bjhargrave: Using recursion here does not provide any value for the downsides of recursion. The stack frame becomes a memory leak. There is no bound here on how many times you may recurse. The user could want thousands, millions of "relistens" here. Then at the end, all of those stack frames must be unwound. The stack frames also hold no information useful to the algorithm (like intermediate values to be combined with computed values). So there is no value here in using recursion to offset the costs of recursion.

demo/src/test/TestActivator.java
bjhargrave: The difference between the two outputs is hard to see "ing" -> "ed".

juergen-albert: true

biz.aQute.launcher/src/aQute/launcher/Launcher.java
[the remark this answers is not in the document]
juergen-albert: good point. I haven't given the call stack any thought. Thanks for the explanation.

juergen-albert:
Can someone do me a favor and try to run the tests on his machine? I have a classically works on my machine situation. The test runs as expected in the IDE and with gradle on windows. Here however, my new Test failed always fails with exit code 244 and I can't find the source for that.

demo/src/test/TestActivator.java
juergen-albert: I rephrased it a bit, so it is easier to differentiate.

PR #5210, opened 2022-04-11
bndtools container: Allow user to exclude project and have only jar

bjhargrave:
This is useful for version=snapshot or version=latest when the
project's built jar contains transformed classes and we want the
Eclipse compiler to use the jar's classes and not the project's
untransformed classes which lead to compilation errors.

By adding `ide=jar-only` to the -buildpath entry, Bndtools
will include the built jar and not include the project in the
class path container.

### DIFF
--- a/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
+++ b/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
@@ -384,9 +384,14 @@ public class BndContainerInitializer extends ClasspathContainerInitializer imple
 						IPath projectPath = root.getFile(path)
 							.getProject()
 							.getFullPath();
-						addProjectEntry(projectPath, accessRules, extraAttrs);
+						List<String> options = Strings.split(c.getAttributes()
+							.get("ide"));
+						boolean versionProject = isVersionProject(c);
+						if (versionProject || !options.contains("jar-only")) {
+							addProjectEntry(projectPath, accessRules, extraAttrs);
+						}
 						// if not version=project, add entry for generated jar
-						if (!isVersionProject(c)) {
+						if (!versionProject) {
 							/*
 							 * Supply an empty index for the generated JAR of a
 							 * workspace project dependency. This prevents the


kriegfrj (approved):
Not that you need my approval, but having had my head deep in this section of code relatively recently, I can understand the need for it and I think that the way you've done it is probably the best way that it could be done.

pkriens:
Could you not name it 'bndtools'? I think what it signifies is that the source/project information is invalid. So it could be 'ide' or maybe even better, borrow from Java and call it 'synthetic'? Anyway, I'd prefer a functional name that refers to what happened than make it a specific bndtools use case. I.e. I am working on a LSP server where I will have the same situation I guess.

bjhargrave:
> Could you not name it 'bndtools'?

The name needs to be clear that it is not for normal command line build (gradle). It only applies to the Eclipse build. So I picked `bndtools`. I could have picked `eclipse`. I don't know if this issue will surface in other situations like LSP (or even IntelliJ). 

For eclipse, we had to also include the project with the jar to get source access for F3 so you went to the project source code and not the class file in the jar. I have no idea if this sort of issue will be a problem in other IDE-like settings.

I also chose the key name `bndtools` since the value are options for Bndtools handling of the class path entry. The suggested alternate key name of `ide` is probably ok if that is agreeable. I don't think a key name of `synthetic` makes sense.

pkriens:
> I also chose the key name bndtools since the value are options for Bndtools handling of the class path entry. The suggested alternate key name of ide is probably ok if that is agreeable. I don't think a key name of synthetic makes sense.

That is _not_ true. This will be necessary for _any_ driver that has multiple sources for the class code. This is typical for an 'ide' but I could imagine other cases. Imho it is much better to use a name that describes the mechanism than relying on another tool's semantics. It creates a semantic coupling in the wrong direction. bnd should be self contained. I am not married to synthetic but it reflects that the JAR contents come from a transformation, like the bridge methods are generated by Java. 

I could live with ide but I'd really prefer a name that maps to what is going, a name that allows the user to interpret what to do, instead of relying what tool X does.

kriegfrj:
> > I also chose the key name bndtools since the value are options for Bndtools handling of the class path entry. The suggested alternate key name of ide is probably ok if that is agreeable. I don't think a key name of synthetic makes sense.
> 
> That is _not_ true. This will be necessary for _any_ driver that has multiple sources for the class code. This is typical for an 'ide' but I could imagine other cases. Imho it is much better to use a name that describes the mechanism than relying on another tool's semantics. It creates a semantic coupling in the wrong direction. bnd should be self contained. I am not married to synthetic but it reflects that the JAR contents come from a transformation, like the bridge methods are generated by Java.
> 
> I could live with ide but I'd really prefer a name that maps to what is going, a name that allows the user to interpret what to do, instead of relying what tool X does.
I see your point here. 

Taking this line of thinking further... it actually shouldn't be a user-supplied attribute. Bnd should  provide utility methods that enable the driver to tell if the classes have been transformed as they were jarred (possibly with finer granularity, eg package or class level). The driver can then use this information as it sees fit. 

It does not make sense to allow the user the freedom to configure this attribute. This is only giving them the opportunity to misconfigure it.

bjhargrave:
> it actually shouldn't be a user-supplied attribute

I don't think we can go that far. Bnd does not even know that the jar does not match the source. That was done by a user configured plugin. So the user configures the plugins to build the jar and then has to tell downstream projects that the built jar does not match the source.

After discussion with Peter, I changed the attribute name to be `source` where `source=project` is the default. You can set to `source=none` to tell Bndtools that the project source does not match the jar. This allows for supporting other `source` values in the future.

kriegfrj:
> > it actually shouldn't be a user-supplied attribute
> 
> I don't think we can go that far. Bnd does not even know that the jar does not match the source. That was done by a user configured plugin. So the user configures the plugins to build the jar and then has to tell downstream projects that the built jar does not match the source.
> 

Couldn't the plugins signal to Bnd that they have transformed the classes? Or as a more coarse measure, couldn't the mere presence of a "transformer" plugin be the signal?

bjhargrave:
> Couldn't the plugins signal to Bnd that they have transformed the classes?

Yes, but that would mean we would have to design a means for this signalling.



> Or as a more coarse measure, couldn't the mere presence of a "transformer" plugin be the signal?

How would Bnd know that a plugin was a "transformer"? Bnd just knows it is a plugin.

In any case, this is now moot since I found out we don't need to tell Bndtools anything. See https://github.com/bndtools/bnd/pull/5213 where I figured out how to tell Eclipse to ignore the project's bin folder which means it will then use the jar file for the class files.

PR #5225, opened 2022-04-21
Support a "project" attribute to name the bsn source project

bjhargrave:
Normally, Bnd expects the bsn of a bundle built in the workspace to be
prefixed by the project name. This convention allows a quick lookup
from a bsn to find the project which builds the bundle with the bsn.

However, sometimes you need to refactor code, or when transforming,
use a completely different bsn for one of the sub bundles. The bundle
building works fine. But if another project refers to the bsn which
is not prefixed by the project name, Bnd cannot find the bundle and
fails.

Bnd could just exhaustively search the workspace but this would be cost
prohibitive in very large workspaces such as Open Liberty.

So this change introduces a "project" attribute which can be used on
a buildpath/testpath entry to name the project in which to look for the
bsn. So this does not affect the normal project-name=bsn case but allows
for bsn refactoring in a build at the cost of specifying the "project"
attribute. Fortunately, decoration can be used to centralize the
mappings of bsn to project name.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Container.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Container.java
@@ -68,7 +68,7 @@ public class Container {
 	}
 
 	public Container(Project project, String bsn, File file, Map<String, String> attributes) {
-		this(project, bsn, "project", TYPE.PROJECT, file, null, attributes, null);
+		this(project, bsn, Constants.VERSION_ATTR_PROJECT, TYPE.PROJECT, file, null, attributes, null);
 	}
 
 	public Container(Project project, File file, Map<String, String> attributes) {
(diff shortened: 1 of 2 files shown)


pkriens:
I really, really, really don't like this :-(

This is going in the same direction as so many systems I know that started simple and became unwieldy over time. A single, relatively small inconvenience for a few is used to kill an important invariant of the overall system. For me, the workspace power is the simple link between bsn and project. Making this indirection means that you need to change every script and code that works with the workspace and it introduces a whole new space of potential errors.

I think we need to discuss this at the next meeting.

bjhargrave:
Sure. But no one has to use this and it is a very small change. 

However, people need to use different bsns as things change. The jakarta rename has caused all sorts of naming issues. Using decoration can hide all of this in a central place. See <https://github.com/bjhargrave/transformer-bnd-analyzer-demo/blob/a7b7d1930f0a08eac3395c4a2c2185f9fa235699/cnf/build.bnd#L19-L20> in my demo project for example.



> an important invariant of the overall system

I don't think it is an invariant. It is a convention. A project can make bundles with any BSNs they choose. But you cannot reference them from other projects with outside advice on where to look.

PR #5238, opened 2022-05-05
Maven lifecycle participation fix

bjhargrave:
Fixes to the bnd-maven-plugin life cycle participant to more carefully replace goals.

Only alter jar plugin executions for jar packaging. Only alter war
executions for war packaging.

Don't remove execution if one of the executions goals matches.
First we remove the goals and if no goals are left, we remove the
execution.


rotty3000 (approved):
LGTM

PR #5252, opened 2022-05-13
Maven integration test refactoring

bjhargrave:
The integration tests for the Bnd maven plugins are refactored so that there are multiple integration tests per plugin instead of a single integration test having a reactor with multiple projects. The integration tests now use the normal maven conventions for directory and file names (e.g. `src/it`, `postbuild`).

This refactoring allows running individual tests rather than all the tests for a plugin. Using the `invoker.properties` files allows for tests to having individual goals and debug log level. It also allows for an individual test to be started in debug mode and wait for a debugger to attach. It allows for all tests to execute even if one fails.

The normal output from integration test execution is now minimal unless an error occurs in the test so the build output is much shorter.


rotty3000:
wow, nice effort here...

bjhargrave:
> wow, nice effort here...

I decided to learn how to properly do maven integration testing using `maven-invoker-plugin` for the new maven plugin I wrote for Eclipse Transformer. So I wanted to apply that knowledge here to improve Bnd's rather awkward integration testing setup.

PR #5258, opened 2022-05-23
[export] JPMS module info calculation fails

### LINKED ISSUE #5257: Gradle Export Task fails
Currently, with the latest version (`6.4.0-SNAPSHOT`), the bndruns cannot be exported anymore.

It reports several warnings regarding the component property types denoting that the annotation should be in the project build path whereas all our custom property types are in a single base API project which is added to the builpath of all the workspace projects. It also reports the similar problems for `org.osgi.service.event.propertytypes.EventTopics` component property type which is also in the aforementioned base API and by default added to the buildpath of all projects.

After reporting the above-mentioned warnings, it fails with the following **ERROR** message:

```
The Bundle-Activator header only supports a single type. The following types were found: com.qivicon.gateway.dmt.simple.provider.Activator,com.qivicon.gateway.dmt.systemproperty.synchronizer.Activator,com.qivicon.gateway.eventnotification.internal.Activator,com.qivicon.gateway.logger.forwarder.LoggerFacadeController,com.qivicon.gateway.slm.agent.internal.Activator,com.qivicon.runtime.healthcheck.core.osgi.OSGiCoreServicesCheckActivator,com.qivicon.runtime.inithooks.adapter.Activator,com.qivicon.runtime.osgiagent.Activator,com.qivicon.runtime.redirect.internal.Activator,com.qivicon.runtime.systemstate.starter.Activator,com.qivicon.tools.console.commands.RandomBundleRestartCommand,org.apache.felix.gogo.command.Activator,org.apache.felix.gogo.runtime.activator.Activator. This usually happens when a macro resolves to multiple types
```

Surprisingly, the export task thinks that all these different bundle activators are inside a single bundle whereas these activators are in separate bundles. Since a bundle cannot have more than one activator, it is also prevented by bndtools by default during build phase as a precautionary step. But, somehow during export the gradle task considers them to be part of a single bundle and that's why the build fails.

For your reference, all our activators in the workspace use `org.osgi.annotation.bundle.Header` annotation as follows:
`@Header(name = org.osgi.framework.Constants.BUNDLE_ACTIVATOR, value = "${@class}")`. We don't specify the `Bundle-Activator` manifest entry manually in the respective project's `bnd.bnd` file and surprisingly, the abovementioned **ERROR** log only comprises those projects that have activators specified using `org.osgi.annotation.bundle.Header` annotation. We also have a few projects where `Bundle-Activator` manifest entry has been added manually to their respective `bnd.bnd` files but they are exempted from the aforementioned **ERROR** log.

Please also find the gradle execution log with `--full-stacktrace` enabled here.

[log.txt](https://github.com/bndtools/bnd/files/8753140/log.txt)

pkriens:
There was a recent change https://github.com/bndtools/bnd/commit/dd2f0c56f9c102b9b7aa969c389787f541e76528 that took a shortcut to calculate the embedded packages unconditionally. It used the bundle class path analysis to analyze all the jars in the executable. 

However, this significantly slows down Eclipse but worse, it then tries to reinterpret the annotation headers. In my case, I got 3 types on the Bundle-Activator.

The analysis should only be done when it is needed and I'd suggest to do it either in a separate Builder or try to analyze the manifests for speedup.

This patch just makes the analysis conditional but that is a temporary fix to make it compatible with the the previous code.

This needs to be cherry picked for 6.3.0 since it is a very serious regression.

Fixes  #5257

### DIFF
--- a/biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
+++ b/biz.aQute.launcher/src/aQute/launcher/plugin/ProjectLauncherImpl.java
@@ -357,19 +357,24 @@ public class ProjectLauncherImpl extends ProjectLauncher {
 				}
 			}
 
-			List<String> bcpList = new ArrayList<>();
-			bcpList.add(".");
-			bcpList.addAll(classpath);
-			bcpList.addAll(actualPaths);
+			String moduleProperty = builder.getProperty(JPMS_MODULE_INFO);
+			if (moduleProperty != null) {
 
-			builder.setProperty(BUNDLE_CLASSPATH, bcpList.stream()
-				.collect(joining(",")));
-			builder.setProperty(EXPORT_CONTENTS, "aQute.launcher.pre");
+				List<String> bcpList = new ArrayList<>();
+				bcpList.add(".");
+				bcpList.addAll(classpath);
+				bcpList.addAll(actualPaths);
 
-			// Ignore common warnings resulting from the Bundle-ClassPath which
-			// are irrelevant in this use case
-			builder.setProperty(FIXUPMESSAGES,
-				"Classes found in the wrong directory, private references, Export-Package duplicate package name, Invalid package name: * in Export-Package");
+				builder.setProperty(BUNDLE_CLASSPATH, bcpList.stream()
+					.collect(joining(",")));
+				builder.setProperty(EXPORT_CONTENTS, "aQute.launcher.pre");
+
+				// Ignore common warnings resulting from the Bundle-ClassPath
+				// which
+				// are irrelevant in this use case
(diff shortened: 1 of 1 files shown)


rotty3000 (commented):
For the record this change is perfectly fine with me. Also note that I do see that the current approach has limitations that should be improved and have some ideas on how to accomplish that. Stay tuned.

pkriens:
> has limitations

It will generate errors.  Bundle-Activator defined by  Header annotations is one since you get the aggregate of all bundlles & jars on the runpath. I also got an error because a bundle had some stuff in the default package.

Best approach imho is to use a new builder.

rotty3000:
:+1:

> a new builder

is what @bjhargrave and I thought as well.

PR #5261, opened 2022-05-24
[decoration] Backward compatibility

pkriens:
The current decoration rewrite was not backward
compatible since it added literals for the `-runbundles`.

After some discussion we decided to add `++` for decorations
that add literals and a single `+` for the 6.2 behavior.

Since we use the decoration in our own build, this
patch first adds `++` syntax but both the single and double
plus will add literals.

After this works, we need to add test for the single `+` 
(not adding literals) and replace our own use of decoration
to use double plusses where applicable.


pkriens:
> @bjhargrave feel free to take over this PR and adjust, I can finish stuff then tomorrow

Reminds me of 1998! :-)

bjhargrave:
Closing this PR since you used your master branch and I did not want to push to it. So I opened a new PR #5262.

PR #5304, opened 2022-06-30
[resolve] Add cache mode

pkriens:
cache – Will use a cache file in the workspace cache. If that file is stale relative to the workspace or project or it does not exist, then the bnd(run) file will be resolved and the result is stored in the cache file.

### DIFF
--- a/biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java
+++ b/biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java
@@ -15,6 +15,7 @@ import org.slf4j.LoggerFactory;
 import aQute.bnd.build.Container;
 import aQute.bnd.build.Run;
 import aQute.bnd.build.Workspace;
+import aQute.bnd.build.WorkspaceLayout;
 import aQute.bnd.build.model.BndEditModel;
 import aQute.bnd.build.model.clauses.HeaderClause;
 import aQute.bnd.build.model.clauses.VersionedClause;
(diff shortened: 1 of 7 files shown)


biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java
```diff
@@ -185,4 +194,100 @@ public Collection<Container> getRunbundles() throws Exception {
 		return super.getRunbundles();
 	}
 
+	private Collection<Container> cache() throws Exception {
+		File ours = getPropertiesFile();
```
bjhargrave, line 198: What about included files in the bndrun file and in the workspace?

In the gradle plugin all of the files factor in to out-of-date checks:

https://github.com/bndtools/bnd/blob/16607b11580986dadf96d162c7cf8555dcbfd9a3/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java#L981-L986

biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java
```diff
@@ -185,4 +194,100 @@ public Collection<Container> getRunbundles() throws Exception {
...
+	}
+
+	boolean testIgnoreDownloadErrors = false;
+
+	private boolean isAllOk(Collection<Container> containers) {
+		for (Container c : containers) {
+			if (c.getError() != null) {
+				if (testIgnoreDownloadErrors) {
+					if (c.getError()
+						.contains("FileNotFoundException"))
```
bjhargrave, line 284: This seems very fragile (looking for a string in the error message). We could change the Container code and break this without even knowing.

biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java
```diff
@@ -185,4 +194,100 @@ public Collection<Container> getRunbundles() throws Exception {
...
+				.asHex();
+			return getWorkspace().getCache("resolved-cache/".concat(digest)
+				.concat(".resolved"));
+		} catch (Exception e) {
+			// not gonna happen
+			throw Exceptions.duck(e);
+		}
+	}
+
+	boolean testIgnoreDownloadErrors = false;
```
bjhargrave, line 277: Why does testing need this? Seems unusual that the test does not test the normal code path.

bjhargrave (commented):
Some comments/questions.

biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java
pkriens: this is for testing only. If it breaks it should be easy to diagnose. But it would be nice if I had doing error numbers when I started 22 years ago :-( I remember thinking about it but was too lazy.

biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java
pkriens: Because we then need all dependencies ... that is either potential communication errors. That said, I changed the test from the enroute set to a much smaller set so I can see if I can have them locally.

biz.aQute.resolve/src/biz/aQute/resolve/Bndrun.java
pkriens: makes sense

bjhargrave:
I guess we can close this PR since you merged #5305.

PR #5348, opened 2022-08-19
[jvm crash in swt] It is still happening

pkriens:
As a try, I've increased the coalesce time
and then also tried to rearrange some of the overridden
methods. Instead of calling the super last, moved it
to first.

Bit desperate ... it is pretty ok now, it happens
to me once every month. But very frustrating that
we cannot pin this down.

If anybody has a full crash report, appreciated.

### DIFF
--- a/bndtools.core/src/bndtools/explorer/Model.java
+++ b/bndtools.core/src/bndtools/explorer/Model.java
@@ -116,7 +116,7 @@ class Model {
 		try {
 			// coalesce some more updates on
 			// the worker thread(s).
-			Thread.sleep(10);
+			Thread.sleep(50);
 		} catch (InterruptedException e) {
 			Thread.currentThread()
 				.interrupt();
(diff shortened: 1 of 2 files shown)


bndtools.core/src/bndtools/explorer/Model.java
```diff
@@ -116,7 +116,7 @@ private void update0() {
 		try {
 			// coalesce some more updates on
 			// the worker thread(s).
-			Thread.sleep(10);
+			Thread.sleep(50);
```
laeubi, line 119: If you want to wait, maybe `Display#timerExec` would be more appropriate (e.g. Line 108), blocking in the UI thread is bad...

pkriens: I know but this is intentional to give whoever is messing with SWT to finish. 50 ms is not yet noticeable. But I agree it is horrible, just lost for other options .... It is a an SWT problem but they do not want to look into it.

laeubi: > It is a an SWT problem but they do not want to look into it.

Can you link the [SWT issue](https://github.com/eclipse-platform/eclipse.platform.swt/issues)? Is there a reproducer?

PR #5357, opened 2022-08-24
Add support for reading Multi-Release-Jar content

laeubi:
This adds methods to the Jar class to read versioned content of a
multi-release jar for the following cases:

- get a resource or a map of resources for a given release version
- get a merged manifest according to the OSGi specification
- get module name/version for a given given release version
- list all contained release alternative versions

Beside this a unit-test is added to cover these new functions.

Signed-off-by: Christoph Läubrich <laeubi@laeubi-soft.de>

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
@@ -24,16 +24,21 @@ import java.security.MessageDigest;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
+import java.util.AbstractMap.SimpleEntry;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.EnumSet;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.NavigableMap;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.SortedMap;
 import java.util.Spliterator;
 import java.util.Spliterators.AbstractSpliterator;
 import java.util.TreeMap;
@@ -46,6 +51,8 @@ import java.util.jar.JarOutputStream;
 import java.util.jar.Manifest;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 import java.util.zip.CRC32;
@@ -74,6 +81,7 @@ import aQute.libg.cryptography.SHA256;
 import aQute.libg.glob.PathSet;
 
 public class Jar implements Closeable {
+
 	private static final int	BUFFER_SIZE				= IOConstants.PAGE_SIZE * 16;
 	/**
 	 * Note that setting the January 1st 1980 (or even worse, "0", as time)
@@ -104,7 +112,7 @@ public class Jar implements Closeable {
 	private final NavigableMap<String, Resource>				resources				= new TreeMap<>();
 	private final NavigableMap<String, Map<String, Resource>>	directories				= new TreeMap<>();
 	private Optional<Manifest>									manifest;
-	private Optional<ModuleAttribute>							moduleAttribute;
+	private Map<Integer, Optional<ModuleAttribute>>				moduleAttributes				= new HashMap<>();
 	private boolean												manifestFirst;
 	private String												manifestName			= JarFile.MANIFEST_NAME;
 	private String												name;
@@ -124,6 +132,14 @@ public class Jar implements Closeable {
 	private long												zipEntryConstantTime	= ZIP_ENTRY_CONSTANT_TIME;
 	public static final Pattern									METAINF_SIGNING_P		= Pattern
 		.compile("META-INF/([^/]+\\.(?:DSA|RSA|EC|SF)|SIG-[^/]+)", Pattern.CASE_INSENSITIVE);
+	static final String											MULTI_RELEASE_HEADER			= "Multi-Release";
+	static final String											SUPPLEMENTAL_MANIFEST_PATH		= "OSGI-INF/MANIFEST.MF";
+	static final int											MULTI_RELEASE_MIN_VERSION		= 9;
+	static final int											MULTI_RELEASE_DEFAULT_VERSION	= 0;
+	static final int											MULTI_RELEASE_VERSION_GROUP		= 1;
+	static final int											MULTI_RELEASE_PATH_GROUP		= 2;
+	static final Pattern										MULTI_RELEASE_PATTERN			= Pattern
+		.compile("^META-INF/versions/(\\d+)/(.*)$", Pattern.CASE_INSENSITIVE);
 
 	public Jar(String name) {
 		this.name = name;
@@ -262,6 +278,24 @@ public class Jar implements Closeable {
 		this(string, file, DEFAULT_DO_NOT_COPY);
 	}
 
+	public boolean isMultiRelease() {
+		return manifest().map(Manifest::getMainAttributes)
+			.map(attr -> attr.getValue(MULTI_RELEASE_HEADER))
+			.map(Boolean::parseBoolean)
+			.orElse(Boolean.FALSE);
+	}
+
+	public void setMultiRelease(boolean multiRelease) {
+		try {
+			ensureManifest();
+			manifest().get()
+				.getMainAttributes()
+				.putValue(MULTI_RELEASE_HEADER, String.valueOf(multiRelease));
+		} catch (Exception e) {
+			Exceptions.duck(e);
+		}
+	}
+
 	private Jar buildFromDirectory(final Path baseDir, final Pattern doNotCopy) throws IOException {
 		Files.walkFileTree(baseDir, EnumSet.of(FileVisitOption.FOLLOW_LINKS), Integer.MAX_VALUE,
 			new SimpleFileVisitor<Path>() {
@@ -361,13 +395,22 @@ public class Jar implements Closeable {
 	public boolean putResource(String path, Resource resource, boolean overwrite) {
 		check();
 		path = ZipUtil.cleanPath(path);
+		versionedResources = null;
 
 		if (path.equals(manifestName)) {
 			manifest = null;
 			if (resources.isEmpty())
 				manifestFirst = true;
-		} else if (path.equals(Constants.MODULE_INFO_CLASS)) {
-			moduleAttribute = null;
+		} else {
+			if (path.equals(Constants.MODULE_INFO_CLASS)) {
+				moduleAttributes.remove(MULTI_RELEASE_DEFAULT_VERSION);
+			} else {
+				Matcher matcher = MULTI_RELEASE_PATTERN.matcher(path);
+				if (matcher.matches() && matcher.group(MULTI_RELEASE_PATH_GROUP)
+					.equals(Constants.MODULE_INFO_CLASS)) {
+					moduleAttributes.remove(Integer.parseInt(matcher.group(MULTI_RELEASE_VERSION_GROUP)));
+				}
+			}
 		}
 		String dir = getParent(path);
 		Map<String, Resource> s = directories.get(dir);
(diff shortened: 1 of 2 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -397,6 +439,29 @@ public Resource getResource(String path) {
...
 
+	/**
+	 * Returns a resource taking the release version into account.
+	 *
+	 * @param path the path of the resource to read
+	 * @param release the release to use
+	 * @return an optional representing the resource or an empty optional if the
+	 *         resource do not exits
+	 */
+	public Optional<Resource> getVersionedResource(String path, int release) {
```
timothyjward, line 450: I think that this method isn't quite right or needs a different name and more documentation. At the moment it only checks one version of the release, not every version below, which isn't my understanding of how Multi-Release jars work.

If, for example, my jar has:

    /Foo.class
    /Bar.class
    /META-INF/versions/9/Foo.class
    /META-INF/versions/11/Bar.class

And I call:
*  `getVersionedResource(Foo.class, 11)` - I would expect to get returned the resource at `/META-INF/versions/9/Foo.class`, but this method will return an empty `Optional`
* `getVersionedResource(Bar.class, 9)` - I would expect to get returned the resource at `/Bar.class` but this method will return an empty `Optional`

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -432,6 +497,72 @@ public Map<String, Resource> getResources() {
...
+	public Map<String, Resource> getVersionedResources(int release) {
+		if (isMultiRelease()) {
+			check();
+			Map<String, NavigableMap<Integer, Resource>> versionedResources = getAllVersionMap();
+			return versionedResources.entrySet()
+				.stream()
+				.map(versions -> {
+					Entry<Integer, Resource> releaseEntry = versions.getValue()
+						.headMap(release, true)
+						.lastEntry();
```
timothyjward, line 515: I think that this method has the same issue as `getVersionedResource`, in that it fails to take account of all version folders lower than `release`, with higher versions taking precedence over lower versions.

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -448,6 +579,48 @@ public Manifest getManifest() throws Exception {
...
+	 * @return a <b>copy</b> that is <b>not</b> backed by the original manifest
+	 *         but copied and enhanced with the supplemental manifest data (if
+	 *         any) for the given release
+	 */
+	public Optional<Manifest> getManifest(int release) {
+		if (isMultiRelease()) {
+			return manifest().map(original -> {
+				Manifest copy = new Manifest(original);
+				if (release >= MULTI_RELEASE_MIN_VERSION) {
+					Optional<Resource> releaseEntry = getVersionedResource(JarFile.MANIFEST_NAME, release);
```
timothyjward, line 596: Having checked the specification the location of supplemental manifests is `/META-INF/versions/X/OSGI-INF/MANIFEST.MF`. See https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module-multireleasejar

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -486,25 +669,34 @@ Optional<ModuleAttribute> moduleAttribute() throws Exception {
...
 			.findFirst();
 	}
 
 	public String getModuleName() throws Exception {
-		return moduleAttribute().map(a -> a.module_name)
-			.orElseGet(this::automaticModuleName);
+		return getModuleName(MULTI_RELEASE_DEFAULT_VERSION);
 	}
 
-	String automaticModuleName() {
```
timothyjward, line 500: It looks like we lost a method here, which would break version compatibility.It could just delegate to the new method with version `0`.

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -467,16 +640,26 @@ Optional<Manifest> manifest() {
 		}
 	}
 
-	Optional<ModuleAttribute> moduleAttribute() throws Exception {
```
timothyjward, line 470: It looks like we lost a method here which would break version compatibility. It could just delegate to the new method with version `0`.

biz.aQute.bndlib/test/aQute/bnd/build/JarTest.java
```diff
@@ -0,0 +1,160 @@
...
+import aQute.lib.io.ByteBufferDataOutput;
+
+public class JarTest {
+
+	private static final String	MODULE_INFO_PATH			= "META-INF/versions/9/" + Constants.MODULE_INFO_CLASS;
+
+	private static final String	MAIN_MANIFEST_PATH			= JarFile.MANIFEST_NAME;
+
+	private static final String	SUPPLEMENTAL_MANIFEST_PATH	= "META-INF/versions/9/" + MAIN_MANIFEST_PATH;
+
```
timothyjward, line 37: I think this is the wrong path as per the OSGi core spec (link in earlier comments).

timothyjward (changes requested):
Overall I think the direction is good, with just a few small fixes required to match spec behaviour.

I would be interested in seeing the `aQute.bnd.build.model.EE` updated to provide access to the `release` int for each EE.

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
laeubi: I can enhance the test-case but actually the method should work that way (but... if its not testes it most likely fails ..).

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
laeubi: Do you "think" this or did you tried it out? Just curious, because `headMap` returns a map of all keys lower or equal to release and `lastEntry` returns the highest value from it, sounds exactly as what you explained here...

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
laeubi: the method is package private an nowhere used in the package, but if you think it helps, I can add it again...

timothyjward: I know that removing methods is likely to trigger a baselining error (though default visibility may be safe), and if we see a baselining error it will incur the wrath of @bjhargrave :smile:

laeubi: Maybe you can enable the workflow runs so we can get a normative result while I address the mentioned issues? :-)

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
timothyjward: I don't have a slot to check the code out and run it today but rather than letting the PR sit I did these checks by eyeballing the code. 
It's entirely possible that I've misread the data structure you're using and everything here is actually fine. If that is the case I would recommend some JavaDoc on `getAllVersionMap` so that others don't make the same mistake when maintaining the class in future.

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
timothyjward: That would be great - I'm definitely not infallible, and may simply be misunderstanding the role of the internal map structures.

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
laeubi: I'll enhance the javadoc and test-case so we can see at least that passing in the CI (it just requires an approval from a bnd committer as I'm a first-time contributor to this repository...).

laeubi: Done.

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
laeubi: Done

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
laeubi: Done.

biz.aQute.bndlib/test/aQute/bnd/build/JarTest.java
laeubi: Fixed.

laeubi (commented):
I have applied the suggested fixes and enhanced the test-case.

laeubi:
> I would be interested in seeing the `aQute.bnd.build.model.EE` updated to provide access to the `release` int for each EE.

If I understand the code correctly, this is simply `EE.getCapabilityVersion().getMajor()`

timothyjward:
At this point I'm happy and think this is worth merging - @bjhargrave do you have anything to add?

timothyjward:
> If I understand the code correctly, this is simply `EE.getCapabilityVersion().getMajor()`

Yes, although the following isn't exactly readable...

    EE myEE = getResolveTargetEE();
    manifest = jar.getManifest(myEE.getCapabilityVersion().getMajor()).get();

timothyjward (approved):

laeubi:
I think it won't be a problem to add a new "getRelease()" method here, but lets see first if there is other to do.

laeubi:
thanks @timothyjward for the review, the Ci is also happy!

laeubi:
> Yes, although the following isn't exactly readable...
> 
> ```
> EE myEE = getResolveTargetEE();
> manifest = jar.getManifest(myEE.getCapabilityVersion().getMajor()).get();
> ```

I have opened 
- https://github.com/bndtools/bnd/pull/5358

for this as it seems to be a valid enhancement regardless of MR support...

laeubi:
@bjhargrave I have now created two PRs based on this to show how one can uses the Multi-Release-Jar support to:

- Correctly handle JPMS module info: https://github.com/bndtools/bnd/pull/5360
- Create a Multi-Release-Bundle by specify `Multi-Release: true` in the BND instructions: https://github.com/bndtools/bnd/pull/5359

I hope this i sufficient to review this PR and decide if changes are required before this can be merged.

FYI @stbischof @rotty3000

timothyjward:
The following comment is wrong. The version should be 17 for the third check in the test. 

https://github.com/bndtools/bnd/compare/2d1a4c6ba66197105067b44814b5c4033f6f7a0d..9856c6cbda183da1a08ee1a8fd1dbaa77a76e090#diff-5ba9b2d89cbd966cfd1655fb0c9488599e851bc6ae563e20bfd63c9df4e0d93aR152

laeubi:
> The version should be 17 for the third check in the test.

Fixed!

timothyjward:
@bjhargrave - I think that this should be merged, can we go ahead?

laeubi:
@rotty3000 can you maybe help here? Its about a full month now with no feedback at all ... :-\

bjhargrave:
I am still working on ideas about mrjar support in a side branch based upon the narrative in https://github.com/bndtools/bnd/wiki/Multi-release-JAR-support-design-discussion. Sorry it is taking awhile but have other items to also work on.

Until I have a better idea on how we should handle mrjars, I don't want to merge this partial proposal.

laeubi:
So now after about half a year later is there any decision / progress on this?

pkriens:
@laeubi can you contact me via Peter.Kriens@aQute.biz? I'd like to have a zoom call to understand all this. I've kept myself out of this discussion (I am not convinced multi release jars are even remotely a good idea, they seem to suffer a lot of accidental complexity) but now with BJ out of the loop I need to get this of the PR list. So if you can contact me, we setup a call.

laeubi:
@pkriens I send you a mail, I think the main point is not if one likes MR or not, that's not a desiccation made by the tool, I just wanted to support it because more and more libs are start using this and it is part of the OSGi Spec for many years, but yes it might not be used much. That's also why I aim for a quite basic solution as a first step and from my testing with this it works quite good for the rare cases where I needed a MR jar (mostly to add OSGi headers to a 3rd party library).

pkriens:
I will take over this PR

PR #5359, opened 2022-08-25
Add support for building multi-release jars (automatic!)

### LINKED ISSUE #5346: [bnd] Support processing of Multi-Release-Jars

laeubi:
This is based on:

so to see the actual changes from this PR simply select the last commit on the review tab.

an allows to specify 
`Multi-Release: true`
in a BND file to instruct BND building an [OSGi Multi-Release Bundle](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module-multireleasejar) based on the detected versions.

@timothyjward this also includes fixing general processing of BND in cases where multi-release is off, and methods for `Jar` to put versioned resources / supplemental manifests. If you think it helps, both aspects could be split into separate PRs as well.

This will fix #5346


laeubi:
I have adjusted [the example here](https://github.com/ClickHouse/clickhouse-jdbc/blob/ce23675513519a6dc8efdfe9adbb91bba19c8f01/pom.xml#L599) to reflect the changes, the
[clickhouse-http-client-0.3.3-SNAPSHOT.jar.zip](https://github.com/bndtools/bnd/files/9421835/clickhouse-http-client-0.3.3-SNAPSHOT.jar.zip) shows the result for such a bundle build by bnd with supplemental manifests for version 9 and 11.

timothyjward:
I think the direction here is good. We probably need some more multi-release bundle generation tests before it can be merged though. 
One feature that seems likely to have issues is the generation of `java.*` imports for multi release manifests at versions greater than 11.

laeubi:
> We probably need some more multi-release bundle generation tests before it can be merged though.

Can you suggest a test-case that might be adapted for this purpose?

> One feature that seems likely to have issues is the generation of `java.*` imports for multi release manifests at versions greater than 11.

According to my tests (see attached generated bundle) it simply "works", e.g. I see `java.*` imports for the java 11 variant.

laeubi:
> We probably need some more multi-release bundle generation tests before it can be merged though.

I have added now a test here:
https://github.com/bndtools/bnd/blob/51a4947c7473e6dbfa56d2e37e1b08575b652d6b/biz.aQute.bndlib.tests/test/aQute/bnd/osgi/AnalyzerTest.java#L24-L53

PR #5360, opened 2022-08-26
[jpms] Fix calculation of module name with mult-release-jar dependencies

### LINKED ISSUE #5327: Calculate moduleName for multi-release-jar-dependencies

laeubi:
As described here:

currently the JPMS Plugin incorrectly calculates the module name of dependencies that are multi-release jars if they are java 8 and java9+ as in this case no module-info.class is present at the default location.

This is based on

so if one want to see the changes from this PR only, simply select the last commit.

This will fix #5327

FYI @rotty3000 @stbischof


pkriens:
I will take over this PR

PR #5369, opened 2022-09-12
Adds clear option to ProjectGenerate instruction

juergen-albert:
Right now the output directory is cleared before the code generator runs again. Our specific generator is capable of handling existing code and even to keep parts of the generated code that has been modified manually.

This PR adds an option to avoid the automatic clean of the output directory.


juergen-albert:
@pkriens Would be so kind and have a look at this?

juergen-albert:
Thanks!

PR #5374, opened 2022-09-21
Provide a HasConfigurableValue view of BundleTaskExtension.instructions.

chrisr3:
Not intended as an "end-run" around the extension's `bnd` methods, but as a means of telling Gradle when we are finished adding new instructions.

### DIFF
--- a/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
+++ b/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
@@ -50,6 +50,7 @@ import org.gradle.api.file.ProjectLayout;
 import org.gradle.api.file.RegularFileProperty;
 import org.gradle.api.java.archives.internal.DefaultManifest;
 import org.gradle.api.model.ObjectFactory;
+import org.gradle.api.provider.HasConfigurableValue;
 import org.gradle.api.provider.ListProperty;
 import org.gradle.api.provider.MapProperty;
 import org.gradle.api.provider.Provider;
@@ -58,6 +59,7 @@ import org.gradle.api.tasks.ClasspathNormalizer;
 import org.gradle.api.tasks.Input;
 import org.gradle.api.tasks.InputFile;
 import org.gradle.api.tasks.InputFiles;
+import org.gradle.api.tasks.Internal;
 import org.gradle.api.tasks.PathSensitive;
 import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.TaskInputFilePropertyBuilder;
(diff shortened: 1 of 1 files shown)


gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
```diff
@@ -181,6 +183,16 @@ public MapProperty<String, Object> getProperties() {
 		return properties;
 	}
 
+	/**
+	 * Allow limited access to the underlying <code>instructions</code>
+	 * property so that users can control when its value is finalized.
+	 * @return View of the underlying <code>instructions</code> property.
+	 */
+	@Internal
+	public HasConfigurableValue getInstructions() {
```
bjhargrave, line 192: The method name leaks an implementation detail (internal field name). It also exposes 4 methods to the caller. Do we need anything more than a method which can do `instructions.disallowChanges()`?

bjhargrave (changes requested):
I think this is over zealousness on your part. The user cannot be protected from themself. Also using this would mean one cannot mix in another plugin which manipulates bnd instructions in the BundleTaskExtension as there is now a timing issue about your plugin locking the gate before the other plugin can add its changes.

So I don't think this is a smart change to the API.

chrisr3:
> I think this is over zealousness on your part. The user cannot be protected from themself. Also using this would mean one cannot mix in another plugin which manipulates bnd instructions in the BundleTaskExtension as there is now a timing issue about your plugin locking the gate before the other plugin can add its changes.

To be blunt, I don't really want anyone else manipulating the Bnd instructions. I can tolerate extra merged instructions, but as you've already said, not every instruction is a merged instruction.

I've actually back-tracked on this slightly because I probably cannot wait for a Bnd 6.4.0 release. My current approach is to execute a `jar.doFirst()` action that parses `BundleTaskExtension.getBnd()` into a `LinkedHashMap`, and then checks that my chosen keys still have the values I expect. This is slightly more work than I'd like since each `bnd()` invocation can add multiple lines of text, but it does seem to work. (Bnd _must_ already have a function to convert a `bndrun` file into a map of individual instructions, but I doubt it's a public API!)

bjhargrave:
> I don't really want anyone else manipulating the Bnd instructions.

Yes, but you are only one party in the ecosystem.


> My current approach is to execute a `jar.doFirst()` action that parses `BundleTaskExtension.getBnd()` into a `LinkedHashMap`, and then checks that my chosen keys still have the values I expect.

This is a better choice. Confirm your values are present and allow other values to be added. Basically data validation.

> Bnd _must_ already have a function to convert a `bndrun` file into a map of individual instructions, but I doubt it's a public API!

You can use UTF8Properties to parse the instructions into a Properties object and then use Parameters to parse an individual header into clauses. You can then inspect the parsed Import-Package header to make sure the values you expect are still present.

chrisr3:
> You can use UTF8Properties to parse the instructions into a Properties object and then use Parameters to parse an individual header into clauses. You can then inspect the parsed Import-Package header to make sure the values you expect are still present.

I don't need any "deep" parsing into directives etc. Only that `key1` still maps to `value1` despite any extra instructions:
```
key1=value1
key1: value2
```

bjhargrave:
> I don't need any "deep" parsing into directives etc.

In any case, I think we can close this PR.

Issue #5375, opened 2022-09-21
bnd-maven-plugin: Clarify property handling


kwin:
According to https://github.com/bndtools/bnd/blob/7989c5522761b763db41c5c4ec96c763cbe6dfae/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java#L225-L235 all Maven properties and apart from that the settings fields and project fields (with prefix `settings`/`project`) can be used in bnd files processed by the bnd-maven-plugin. This should be clarified in the [README.MD](https://github.com/bndtools/bnd/blob/master/maven/bnd-maven-plugin/README.md)

PR #5379, opened 2022-09-26
maven: Document access to Maven properties in Bnd instructions

bjhargrave:
Fixes https://github.com/bndtools/bnd/issues/5375

### DIFF
--- a/maven/bnd-maven-plugin/README.md
+++ b/maven/bnd-maven-plugin/README.md
@@ -180,6 +180,28 @@ The plugin will by default set some OSGi bundle headers derived from [pom elemen
 | `Bundle-Developers`   | `developers` (child element `id` must be set on each developer) |
 | `Bundle-DocURL`       | `url`                                                           |
 
+### POM Properties
+
+The Bnd instructions can reference properties defined for the Maven project and Maven settings using the `project.` and `settings.` stems, respectively.
+Other properties defined in the POM can also be referenced.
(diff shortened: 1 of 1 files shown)


maven/bnd-maven-plugin/README.md
```diff
@@ -180,6 +180,28 @@ The plugin will by default set some OSGi bundle headers derived from [pom elemen
...
+
+```xml
+<properties>
+	<implementation.vendor>Acme Inc.</implementation.vendor>
+</properties>
+<plugin>
+  <groupId>biz.aQute.bnd</groupId>
+  <artifactId>bnd-maven-plugin</artifactId>
+  <configuration>
+    <bnd><![CDATA[
```
kwin, line 196: @bjhargrave maybe this is not the best example, as with inline bnd instructions the replacement is actually done by Maven. Only for external bnd files it is done by the bnd-maven-plugin IIUC.

bjhargrave: Perhaps you can make a PR to improve the example?

kwin: Either split up the example to show both `pom.xml` and `bnd.bnd` or get rid of it alltogether I would say. But referencing https://maven.apache.org/ref/current/maven-model-builder/#model-interpolation in this context would be beneficial, as it works similarly.

PR #5382, opened 2022-09-28
Use gradle-build-action to speedup builds

Goooler:
https://github.com/gradle/gradle-build-action


bjhargrave (changes requested):
This simply adds additional steps rather than replace or modify the existing build steps.

Goooler:
No need to replace arguments, `gradle-build-action` will automatically cache your Gradle builds.

Goooler:
See https://blog.gradle.org/gh-actions

bjhargrave (approved):

PR #5319, opened 2022-07-15
Maven codegen Mojo

juergen-albert:
This PR is meant to get some early feedback on this. 

The Mojo provides support for the generate mechanism via external plugins. The general bnd generator mechanism reads the instruction and looks in the workspace if it finds a bundle that provides a properly named external plugin capability. Thus I need to provide a Repository containing the dependency with the generator. 
At the moment I use the build dependencies as source for that similar to our Mojos that work on a bndrun. This seems a bit wrong, as this is actually a Plugin dependency and no direct dependency of my maven module. I tried another approach doing the same based on the plugin dependencies. There is a problem though. Plugin dependencies are part of the Mojos Classpath when bnd starts it. If I didn't read the code wrong, generate requires a repository and will load the resulting jar in a completly isolated classloader. This in theory should not cause any issues with the Mojos classloader, that already has this jar.
So much for the theory. Reality poses other issues however. The isolated nature of the external plugin requires an all in one jar, that packs every class it will ever need. In the case of my generator I use and thus package e.g. Promises in an older Version. Because my generator is now on the classpath, it can break the Mojo and causes:
```
java.lang.NoSuchMethodError: org.osgi.util.promise.PromiseFactory.toPromise()Ljava/util/stream/Collector;
    at aQute.bnd.repository.fileset.FileSetRepository.readFiles (FileSetRepository.java:81)
    at aQute.bnd.repository.fileset.FileSetRepository.getBridge (FileSetRepository.java:73)
    at aQute.bnd.repository.fileset.FileSetRepository.list (FileSetRepository.java:189)
    at aQute.bnd.maven.export.plugin.BndContainer.injectImplicitRepository (BndContainer.java:170)
    at aQute.bnd.maven.export.plugin.BndContainer.generate (BndContainer.java:146)
    at aQute.bnd.maven.export.plugin.GenerateMojo.execute (GenerateMojo.java:69)
```

The classloader hirachy looks as follows (my generator is `url[1]` and the correct promise jar is `url{37]`):
```
realm =    plugin>biz.aQute.bnd:bnd-generate-maven-plugin:6.4.0-SNAPSHOT
strategy = org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy
[...]

PR #5385, opened 2022-09-29
Maven codegen

bjhargrave:
This replaces #5319 since @juergen-albert never allows edits by maintainers.

image

I tidied things up, removed some files that should not have been in source control and restored some fixes that were reverted.


juergen-albert (approved):
LGTM! Sorry for not making my PR editable from the start. I definitelly need to make this my default. Right now I simply overlooked the checkbox.

Issue #5363, opened 2022-09-01
Parameter headers and merging


pkriens:
Working on a feature I noticed that we skip duplicate Require-Capability and Provide-Capability clauses. Looking into the code I could not but observe that we handle the merging purely string based. We use a Parameters in CDIAnnotations & DSAnnotations but then turn it into a string representing a clause in  a sorted. Then we join all the requirements and capabilities with a ",", and write it back to the property. This is done for each component & CDI thingy. The code between the two looks copied.

Later in AnnotationHeaders.getHeader(), we get the property. This is quite suspect code. AnnotationHeaders keeps the headers in a MultiMap. The values (the clauses) for the key are then sorted and it is checked if the header already contains the existing value. However, the annotation headers add per clause but the existing property is highly likely a string containing multiple clauses. This means that the resulting set is not properly sorted per clause.

A bit like:

    bnd:          Require-Capability   bar,foo
    java:          @Header(name="Require-Capability", "cat")

    manifest: RC bar,foo,cat

I'd like to propose an addClause(key,attrs)  to Processor. Processor will keep a `Map<String,Parameters>` for any clause that has been used by addClause. The versions of getProperty(key) for a key that has been used this way will the string version
of this Parameters. CDIAnnotations and DSAnnotations must then use this. They should not merge the requirements and capabilities since this should be a global policy.

AnnotationHeaders will then properly use these Parameters. Before the Parameters is printed, we should apply a header specific merge & order function. I.e. requirements will be merged differently then capabilities. Currently we duplicate this policy in CDIAnnotations and DSAnnotations and likely ignore it in other places. This will probably make MergedRequirement superfluous, at least it would now work on the final Parameters instead of a sequence of small subsets.

I think an `addClause(key,attrs)` & `getParameters(key)` will clean up the handling of requirements and capabilities and will likely be useful in other places. As far as I can see, it would significantly clean up DS- and CDIAnnotations.

The reason for this is that I need to be able to know how many service capabilities are becoming merged, I'd like to represent this as an attribute, or at least do not deduplicate the capabilities in the service namespace. And I do not want to hack this functionality into the current code due to the aforementioned issues.

Looking for feedback before I clean up this code.

rotty3000:
for me the proposal sounds reasonable. I was always worried that we may not be handling merge operations correctly. thx for looking into this.

bjhargrave:
I don't understand the root problem here. Is it that you don't want deduplication?

[...]

PR #5389, opened 2022-09-30
[clause manager] A clause manager for Processor

pkriens:
See https://github.com/bndtools/bnd/issues/5363

This is the first part of #5363

This part adds a clause manager that allows
code to set arbitrary clauses instead of doing
lots of string processing. If a property is gotten
that was managed with addClauses, this property
will be consolidated. This goes through the
plugins: see ConsolidatedParameters.

The code intents to be completely backward 
compatible. 


chrisrueger:
This looks good. I wonder why this PR was closed?  @pkriens 

I think it might be a base for an alternative solution to this https://github.com/bndtools/bnd/pull/6868 in https://github.com/bndtools/bnd/issues/6867

PR #5418, opened 2022-11-08
[m2e] Expand bundle version ranges for M2E imports

### LINKED ISSUE #5417: Bndtools M2E integration broken on latest Eclipse Releases
Recent Eclipse releases (e.g. 2022-09) have updated the version of `org.eclipse.m2e.maven.runtime` to `3.8.601.xxx` which doesn't meet the version range syntax used in bndtools m2e. We need to work out how to deal with this as it prevents bndtools m2e from starting.

It looks like the version of the bundle is intended to match the version of maven now, although actually it's a few hundred micro versions higher.

timothyjward:
In recent versions (2022-06 and later) of Eclipse the version of the M2E maven runtime bundle has stopped matching the M2E version and now uses a higher version related to the maven it packages. We need to be able to use this if M2E support is to keep working.

Fixes #5417

### DIFF
--- a/biz.aQute.bnd.maven/bnd.bnd
+++ b/biz.aQute.bnd.maven/bnd.bnd
@@ -6,9 +6,9 @@ Bundle-Description:	    Shared logic used by bnd maven plugins
 -includepackage: aQute.bnd.maven.lib.*
 
 Import-Package: \
-	org.apache.maven.*;bundle-symbolic-name="${@bundlesymbolicname}";bundle-version="${range;[==,+);${@bundleversion}}";version=!,\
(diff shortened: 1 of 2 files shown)


bjhargrave (approved):
Sigh. This problem is caused by https://github.com/eclipse-m2e/m2e-core/commit/bdf007bfa0feb106f58f30d8a5e1959bf3ffc0e4 which made a significant version change in the m2e bundle version to match the embedded maven runtime version (1.18 => 3.8). 

So this change should see us through to Maven 4. 🤞

timothyjward:
Possibly one to make @laeubi aware of as well.

PR #5427, opened 2022-11-13
Resolve artifact's finalName immediately before BndMaven*Plugin execution

HannesWell:
If one wants to customize the name of the artifact build for a project in a Maven build, the property `project.build.finalName` can be set respectively has to be used the following approach:
```
<project>
	...
	<build>
	<finalName>${project.artifactId}-${qualifiedVersion}</finalName>
	...
	<Execute a mojo that creates the 'qualifiedVersion' property>
	...
	<Execute the `bnd-maven-plugin:jar` goal>
	</build>
</project>
```
Unfortunately the `bnd-maven-plugin:jar` goal does not consider properties that are only created by other previous mojo executions for the same project.
The reason is that the `finalName` is read directly from the `MavenProject`'s `Build` object, which is created at the beginning of the (project) build and whose field values are only resolved against the properties initially available. Later no further attempt is done to resolve unresolved references in the fields of that object.
But when the value of 'project.build.finalName' is injected as Mojo parameter the value is resolved against the then available properties and therefore considers properties that were created in a previous mojo execution for the same project.

In order to consider properties created during previous mojo executions the value has to be injected as parameter. This is the way the `maven-jar-plugin` reads the value from the model:
https://github.com/apache/maven-jar-plugin/blob/6eda00939b31aeddcb28cd49763a28ba86fdce5b/src/main/java/org/apache/maven/plugins/jar/AbstractJarMojo.java#L76-L80

Could this also be considered for BND 6.4?

### DIFF
--- a/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
+++ b/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
@@ -99,7 +99,13 @@ public abstract class AbstractBndMavenPlugin extends AbstractMojo {
 	 * The directory where the webapp is built when packaging is {@code war}.
 	 */
 	@Parameter(alias = "warOutputDir", defaultValue = "${project.build.directory}/${project.build.finalName}")
-	File                    webappDirectory;
+	File					webappDirectory;
+
+	/**
+	 * The name of the created artifact (excluding extension and classifier).
+	 */
+	@Parameter(defaultValue = "${project.build.finalName}", readonly = true)
+	String					finalName;
 
 	@Parameter(defaultValue = "${project}", required = true, readonly = true)
 	MavenProject			project;
@@ -656,11 +662,8 @@ public abstract class AbstractBndMavenPlugin extends AbstractMojo {
 	}
 
 	private File getArtifactFile() {
-		return new File(getOutputDir(), project.getBuild()
-			.getFinalName()
-			+ getClassifier().map("-"::concat)
-				.orElse("")
-			+ "." + getExtension(project.getPackaging()));
+		return new File(getOutputDir(), finalName + getClassifier().map("-"::concat)
+			.orElse("") + "." + getExtension(project.getPackaging()));
 	}
 
 	private String getExtension(String type) {
--- a/maven/bnd-indexer-maven-plugin/src/main/java/aQute/bnd/maven/indexer/plugin/IndexerMojo.java
+++ b/maven/bnd-indexer-maven-plugin/src/main/java/aQute/bnd/maven/indexer/plugin/IndexerMojo.java
@@ -59,6 +59,12 @@ public class IndexerMojo extends AbstractMojo {
 	@Parameter(defaultValue = "${project}", readonly = true, required = true)
 	private MavenProject				project;
 
+	/**
+	 * The name of the created artifact (excluding extension and classifier).
+	 */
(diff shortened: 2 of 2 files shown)


bjhargrave:
> Could this also be considered for BND 6.4?

Yes.

maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
```diff
@@ -99,7 +99,13 @@ public abstract class AbstractBndMavenPlugin extends AbstractMojo {
...
 	 */
 	@Parameter(alias = "warOutputDir", defaultValue = "${project.build.directory}/${project.build.finalName}")
-	File                    webappDirectory;
+	File					webappDirectory;
+
+	/**
+	 * The name of the created artifact (excluding extension and classifier).
+	 */
+	@Parameter(defaultValue = "${project.build.finalName}", readonly = true)
+	File					finalName;
```
bjhargrave, line 108:
```suggestion
	String					finalName;
```

bjhargrave (changes requested):

HannesWell:
> > Could this also be considered for BND 6.4?
> 
> Yes.

Great. Should I then create another PR once this is merged against the 6_4 branch or do you prefer to do it yourself?

And thank you for enhancing/extending this PR. Do you have an idea why this has caused build failures/time-outs?

bjhargrave:
> Should I then create another PR once this is merged against the 6_4 branch or do you prefer to do it yourself?

I will cherry pick the commit.

> Do you have an idea why this has caused build failures/time-outs?

The internet is flaky :-) The test make a lot of network connections and sometimes they don't work. I restarted the failed jobs.

HannesWell:
> > Should I then create another PR once this is merged against the 6_4 branch or do you prefer to do it yourself?
> 
> I will cherry pick the commit.

Great, thank you.

> 
> > Do you have an idea why this has caused build failures/time-outs?
> 
> The internet is flaky :-) The test make a lot of network connections and sometimes they don't work. I restarted the failed jobs.

Understand. Looks like this let the Tests succeed. :)

Can you already estimate when BND 6.4 will be available?

bjhargrave:
> Can you already estimate when BND 6.4 will be available?

Our plan is to publish by end of November.

bjhargrave (approved):

HannesWell:
> > Can you already estimate when BND 6.4 will be available?
> 
> Our plan is to publish by end of November.

Thanks. 👍🏽

HannesWell:
@bjhargrave would you mind to make the `finalName` writeable for user? I.e. to remove the `readOnly=true` annotation field.

This would allow to assign another name, e.g. for a sources artifact build with the `jar` goal.

bjhargrave:
> make the `finalName` writeable for user?

It is the same as maven-jar-plugin and the jar goal of bnd-maven-plugin is to replace the jar goal of maven-jar-plugin. So I think it is correct as it currently stands.

You can set the final name using 

```xml
<project>
  <build>
    <finalName>foo</finalName>
  </build>
</project>
```

which is the same way one would set it for maven-jar-plugin.

HannesWell:
> > make the `finalName` writeable for user?
> 
> It is the same as maven-jar-plugin and the jar goal of bnd-maven-plugin is to replace the jar goal of maven-jar-plugin. So I think it is correct as it currently stands.
> 

Yes. I forgot that the classifier is appended when the name of the jar is build. So there will be a distinction and it is not necessary to be able to set name.
Please just disregard my previous comment.

Issue #5409, opened 2022-10-25
bndrun window in maven-layout does not show bundles (e 2022-9)


stbischof:
When opening the Resolve/Run (bndrun) window in eclipse 2022-9 in a project that uses mvn layout you could not longer see the MavenDependencies in the left site of the window.

Left: Eclipse 2022-9        Right: Eclipse 2022-3
![Screenshot from 2022-10-25 14-33-43](https://user-images.githubusercontent.com/33224746/197774097-2e438532-0a9d-4e56-9b80-bf0115362b3d.png)

bjhargrave:
Isn't this addressed by the recent PRs from @timothyjward?

timothyjward:
> Isn't this addressed by the recent PRs from @timothyjward?

Yes and no. They appear, but then disappear again after < 1 second. The recent PRs make the resolve work though, which was completely broken before.

stbischof:
on newest snapshot this came up;
```
An internal error occurred during: "Updating Maven Project".
Receiver class bndtools.m2e.MavenDependenciesRepository does not define or inherit an implementation of the resolved method 'abstract void mavenProjectChanged(java.util.List, org.eclipse.core.runtime.IProgressMonitor)' of interface org.eclipse.m2e.core.project.IMavenProjectChangedListener.
```

timothyjward:
This is probably why the panel paints then disappears on Eclipse 22-06+

timothyjward:
> fixed: mavenProjectChanged method by #5429
> 
> now i found this in error log
> 
> ```
> org.osgi.framework.BundleException: Could not resolve module: bndtools.m2e [439]
>   Unresolved requirement: Import-Package: org.apache.maven.artifact; bundle-symbolic-name="org.eclipse.m2e.maven.runtime"; bundle-version="[1.16.0,2.0.0)"; provider="m2e"
> 
> 	at org.eclipse.osgi.container.Module.start(Module.java:463)
> 	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel$2.run(ModuleContainer.java:1852)
> 	at org.eclipse.osgi.internal.framework.EquinoxContainerAdaptor$1$1.execute(EquinoxContainerAdaptor.java:136)
> 	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1845)
[...]

PR #5429, opened 2022-11-14
[m2e] Avoid runtime errors due to API changes between 1.x and 2.x

timothyjward:
Several types in M2E have been updated to use Lists rather than arrays. We have to implement the extra interface methods (where needed) and use reflective calls to avoid linking to a method which may not exist. All of this code is ugly, and should be removed once we can use a base of M2E 2.x.

### DIFF
--- a/bndtools.m2e/src/bndtools/m2e/AbstractMavenRepository.java
+++ b/bndtools.m2e/src/bndtools/m2e/AbstractMavenRepository.java
@@ -2,15 +2,18 @@ package bndtools.m2e;
 
 import java.io.File;
 import java.io.InputStream;
+import java.util.List;
 
 import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.m2e.core.embedder.ArtifactKey;
 import org.eclipse.m2e.core.project.IMavenProjectChangedListener;
 import org.eclipse.m2e.core.project.IMavenProjectFacade;
 import org.eclipse.m2e.core.project.IMavenProjectRegistry;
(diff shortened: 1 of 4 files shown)


timothyjward:
I believe that this fixes #5409

timothyjward:
I believe that the two commits above make *everything* work again. It's pretty horrible, and I look forward to being able to remove these hacks once Bndtools 7.0.0 uses M2E v2.x as a base.

stbischof:
thx @timothyjward

bjhargrave (approved):
Once this is verified and merged, I will cherry-pick the commits to next for 6.4.0.

timothyjward:
> Screenshot from 2022-11-14 17-20-43
> 
> with 2022-9 I do not see bundles in the bndrun view.
> 
> but the error is not longer there
> 
> ```
> An internal error occurred during: "Updating Maven Project".
> Receiver class bndtools.m2e.MavenDependenciesRepository does not define or inherit an implementation 
> of the resolved method 'abstract void mavenProjectChanged(java.util.List, org.eclipse.core.runtime.IProgressMonitor)' 
> of interface org.eclipse.m2e.core.project.IMavenProjectChangedListener.
> ```

It sounds like an improvement then. I found another place that needed a repeat of an abstract method to satisfy the linker in [c8ea019](https://github.com/bndtools/bnd/pull/5429/commits/c8ea0190db2ffb9d4e4d5781341b7b13aebf9c2b) - if that wasn't the commit you build then it may be worth trying. Otherwise I'm not sure how else to fix this.

stbischof:
i build after https://github.com/bndtools/bnd/pull/5429/commits/c8ea0190db2ffb9d4e4d5781341b7b13aebf9c2b

timothyjward:
I think that this is ready to merge. I can't guarantee that all the M2E integration is working, but it does appear substantially improved. I am able to build/resolve/run my test workspace in both old and new versions of Eclipse.

stbischof:
jep this should be merged!
@timothyjward 
old: 2022-03
new: 2022-06?

PR #5437, opened 2022-11-18
temp workspace was created in the wrong location

juergen-albert:
Due to the use of the wrong workspace folder some remnant directories have been created in the projects root directory instead of the target directory

### DIFF
--- a/maven/bnd-generate-maven-plugin/src/main/java/aQute/bnd/maven/generate/plugin/BndContainer.java
+++ b/maven/bnd-generate-maven-plugin/src/main/java/aQute/bnd/maven/generate/plugin/BndContainer.java
@@ -153,6 +153,7 @@ public class BndContainer {
 			bnd.setProperties(properties.replaceHere(project.getBasedir()));
 
 			bnd.setProperty("project.output", workingDir.getCanonicalPath());
+			bnd.setProperty(aQute.bnd.osgi.Constants.DEFAULT_PROP_TARGET_DIR, workingDir.getCanonicalPath());
 
 			if (logger.isDebugEnabled()) {
 				logger.debug("Generate Project Properties");
@@ -170,7 +171,7 @@ public class BndContainer {
 	}
 
 	public Project init(String task, File wsDir, File workingDir, Properties mavenProperties) throws Exception {
-		File temporaryDir = workingDir.toPath()
+		File temporaryDir = wsDir.toPath()
 			.resolve("tmp")
 			.resolve(task)
 			.toFile();


maven/bnd-generate-maven-plugin/src/main/java/aQute/bnd/maven/generate/plugin/GenerateMojo.java
```diff
@@ -166,6 +167,10 @@ private GenerateOperation getOperation() {
 				}
 			} finally {
 				errors += BndContainer.report(project);
+				if (project.getTargetDir()
+					.exists()) {
+					IO.delete(project.getTargetDir());
+				}
```
bjhargrave, line 173: This seems wrong, like you are writing into a source folder and then need to tidy up after. It would be much better to write into an output folder originally. Can this be done?

juergen-albert: Yes and no. In Maven code is usually generated somewhere in the target directory and many people follow the mantra of not checking in generated code. With my main usecase being EMF, we have to abilitly to overwrite e.g. specific methods in the generated code and prevent them from being overwritten in the next run. Thus my output directory is my src/main/java folder.
Long story short, I achieve this by setting the directory of the bnd project to the directory of the maven Module, so everything is relative to this. Do you have any other idea on how to make something like this work without setting the workingDir to the module dir?
```
-generate: \
    src/main/resource/**.template;\
        generate="somegenerator";\
	output=src/main/java
```

The working dir for the Project is currently set here: https://github.com/bndtools/bnd/blob/355f1985af3ecf5c60de2458657a12298b3a9886/maven/bnd-generate-maven-plugin/src/main/java/aQute/bnd/maven/generate/plugin/BndContainer.java#L183

bjhargrave: If the user configures the generator to output in a source folder, that is their prerogative. But then you should not delete the folder since they may want to commit the changes.

juergen-albert: No, that is not the folder I delete. When the project is initialized, it automatically creates the `generate` dir, we usually have in the bnd workspace layout. I just realized that I can configure the projects target dir as well, so I pointed it to mavens target dir.

bjhargrave (approved):

bjhargrave:
Is this something that needs to be cherry-picked onto next for 6.4.0?

bjhargrave:
I failed to see that this PR was against next. So I have forced pushed next to remove this. Please open PRs against master and then we cherry-pick them to next when needed. Otherwise we end up where we forget to fix on master.

juergen-albert:
Okay, sorry. I thought it might be better mark this as a fix for the RC. Will do next time.

bjhargrave:
So please make a new PR for master asap.

PR #5441, opened 2022-11-30
Extract an abstract mojo so it is possible to have different lockups

laeubi:
Currently BaselineMojo uses a very specific way to lookup the base artifact. If one wants to enhance this, it is currently not possible.

This extracts a common base as AbstractBaselineMojo that could be used by other Mojos to implement a baseline check with other mechanisms.

### DIFF
--- a/maven/bnd-baseline-maven-plugin/src/main/java/aQute/bnd/maven/baseline/plugin/AbstractBaselineMojo.java
+++ b/maven/bnd-baseline-maven-plugin/src/main/java/aQute/bnd/maven/baseline/plugin/AbstractBaselineMojo.java
@@ -1,0 +1,122 @@
+package aQute.bnd.maven.baseline.plugin;
+
+import static java.util.stream.Collectors.toList;
+
+import java.io.File;
+import java.io.IOException;
+import java.util.Comparator;
+import java.util.Formatter;
+import java.util.List;
+import java.util.Locale;
+import java.util.Objects;
+
+import aQute.bnd.differ.Baseline;
+import aQute.bnd.differ.Baseline.BundleInfo;
+import aQute.bnd.differ.Baseline.Info;
+import aQute.bnd.differ.DiffPluginImpl;
+import aQute.bnd.header.Parameters;
+import aQute.bnd.osgi.Instructions;
+import aQute.bnd.osgi.Jar;
+import aQute.bnd.osgi.Processor;
+import aQute.bnd.service.diff.Diff;
+import aQute.lib.io.IO;
+import aQute.lib.strings.Strings;
+import org.apache.maven.plugin.AbstractMojo;
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public abstract class AbstractBaselineMojo extends AbstractMojo {
+
+	protected static final Logger	logger	= LoggerFactory.getLogger(BaselineMojo.class);
+
+	@Parameter(property = "bnd.baseline.full.report", defaultValue = "false")
+	protected boolean				fullReport;
+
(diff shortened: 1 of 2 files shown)


laeubi:
@bjhargrave can you review this?

bjhargrave (commented):
This implies that AbstractBaselineMojo becomes API. Since we do not need to use it ourself, it only exists for other to subclass (it is public).

I am not sure this is wise. Is this sort of behavior (abstract base mojo classes for other plugins to extend) common in the maven plugin ecosystem?

None of the other bnd maven plugins purport to be API.

laeubi:
> This implies that AbstractBaselineMojo becomes API. Since we do not need to use it ourself, it only exists for others to subclass (it is public).

I think it depends a bit how one defines "API" its more an SPI I think and people should be aware that it can change in whatever way between releases (as maven plugins are bound to a specific version that won't be much of an issue(. If you want to be more specific on this I found https://github.com/apiguardian-team/apiguardian usefull and one might add
`@API(status=EXPERIMENTAL)`

> I am not sure this is wise.

At least it would be more wise than people implementing the same things or copy the code of BND to their plugins :-)
As said, no maven-plugin developer will expect any special API policy (at least I'm not aware of it) as such thing would always be rely on internal behavior, but that's actually the point, to inherit as much of the internals as possible.

> Is this sort of behavior (abstract base mojo classes for other plugins to extend) common in the maven plugin ecosystem?

Just a small excerpt, but I think in the end there is no normative answer:
- [AbstractCompilerMojo](https://github.com/apache/maven-compiler-plugin/blob/master/src/main/java/org/apache/maven/plugin/compiler/AbstractCompilerMojo.java)
- [AbstractSurefireMojo](https://github.com/apache/maven-surefire/blob/f6bc398c4d5bc352967434ec6d0e7facccec7afe/maven-surefire-common/src/main/java/org/apache/maven/plugin/surefire/AbstractSurefireMojo.java)
- [AbstractGpgMojo](https://github.com/apache/maven/blob/master/maven-plugin-api/src/main/java/org/apache/maven/plugin/AbstractMojo.java)
- ... probably more ...

> None of the other bnd maven plugins purport to be API.

There is [AbstractBndMavenPlugin](https://github.com/bndtools/bnd/blob/master/maven/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java) I think in the end it does not matter much if there is one or multiple extenders.

bjhargrave:
> Currently BaselineMojo uses a very specific way to lookup the base artifact. If one wants to enhance this, it is currently not possible.

Wouldn't it be better to refactor this logic into a method which a subclass could override instead of splitting the mojo class into 2 classes?

laeubi:
The problem with subclassing would be that then the mojo has probably some parameters that are not useful e.g `[includeDistributionManagement](https://github.com/bndtools/bnd/blob/master/maven/bnd-baseline-maven-plugin/README.md#include-distribution-management)` that only makes sense if the pom repositories are used.

bjhargrave:
In any case, we (bnd) do not need to have an abstract base class. So we should not burden ourselves or pretend to make API/SPI. Deciding which configuration belongs in a base class or subclasses is an API/SPI design decision.

I don't have a problem with refactoring the baseline finding logic into a method which someone could override if they want. But I don't see that having an abstract base class helps the bnd plugin so I don't think it is a good change.

bjhargrave:
> A method override can't serve the same purpose well because then e.g. one always has to make sure that the method is completely replaceable (e.g. it should not change internal state or depend on maybe then unitized fields) and it always makes one inherit parameters that might be of no use in such a context where the base is derived from something else, e.g. an OBR repository...

Yes. But this is true of any methods in the base class since the whole point of making a base class here is to allow external subclasses which can override any method in the base class. Having an abstract base class does not solve the design issues involved in designing a class for subclassing by others.

Issue #5446, opened 2022-12-01
[M2E] bnd 6.4.0 "Error occurred during initialization of boot layer" when launching bndrun in maven shape


maho7791:
I get the following message on the console the try to launch a bndrun.

Error occurred during initialization of boot layer
java.lang.LayerInstantiationException: Package jdk.internal.jimage in both module jrt.fs and module java.base

This problem also occurs at our customers when installing a fresh eclipse 2022-06 or 2022-09 and bndtools 6.4.0.

The problem happens with various Java 11 JDK's. 

I am not sure, if this is really a bndtools bug or JDT. But it first appeared today. I will investigate further and just wanted to inform you as early as possible.

maho7791:
This is no bnd issue!

maho7791:
In my case the Maven project that contained the bndruns is a project with packaging pom. 
So, for the JDT launcher it is not a java project. Changing the packaging type to default, the problem is solved.

I am not sure, if this is really a JDT launcher problem. But for bnd the JDT launcher added the jrt-fs.jar to the module path for a non-Java project. In the end the command ended up with additioanlly having the jrt-fs.jar on the module path, which correctly leads to:
Error occurred during initialization of boot layer
java.lang.LayerInstantiationException: Package jdk.internal.jimage in both module jrt.fs and module java.base

PR #5455, opened 2022-12-05
Avoid adding JDK jrt-fs.jar to the module-path (-p) of the JDT launcher

maho7791:
Avoid adding JDK jrt-fs.jar to the module-path (-p) of the JDT launcher

- filter jrt-js.jar from the module path

This PR goes back to: 
https://github.com/bndtools/bnd/issues/5446

It solves the issue #5446, when the jrt-fs.jar is added to the module path of the JDT launcher execution java command. 
This usually happens in cases when the Eclipse project, that contains the bndrun is no JavaProject.

This is in an solution, independent from the JDT handling of creating default module-paths for non-Java projects.


org.bndtools.launch/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
```diff
@@ -294,6 +296,13 @@ public String[][] getClasspathAndModulepath(ILaunchConfiguration config) throws
 			classpathAndModulepath = new String[2][];
 			classpathAndModulepath[1] = new String[0];
 		}
+		if (classpathAndModulepath[1].length > 0) {
+			classpathAndModulepath[1] = Arrays.asList(classpathAndModulepath[1])
+				.stream()
+				.filter(mp -> !mp.endsWith("jrt-fs.jar"))
+				.collect(Collectors.toList())
+				.toArray(new String[0]);
```
bjhargrave, line 304:
```suggestion
				.toArray(String[]::new);
```

No need to collect to a list and then an array. Just go straight to an array.

org.bndtools.launch/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
```diff
@@ -294,6 +296,13 @@ public String[][] getClasspathAndModulepath(ILaunchConfiguration config) throws
 			classpathAndModulepath = new String[2][];
 			classpathAndModulepath[1] = new String[0];
 		}
+		if (classpathAndModulepath[1].length > 0) {
```
bjhargrave, line 299:
```suggestion
		} else if (classpathAndModulepath[1].length > 0) {
```

This test is only interesting on the else side of the if.

org.bndtools.launch/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
```diff
@@ -294,6 +296,13 @@ public String[][] getClasspathAndModulepath(ILaunchConfiguration config) throws
...
+			classpathAndModulepath[1] = Arrays.asList(classpathAndModulepath[1])
+				.stream()
```
bjhargrave, line 301:
```suggestion
			classpathAndModulepath[1] = Arrays.stream(classpathAndModulepath[1])
```

You can stream an array directly without needing to wrap in a list.

bjhargrave (changes requested):

org.bndtools.launch/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
```diff
@@ -7,12 +7,14 @@
...
 import java.text.MessageFormat;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
+import java.util.stream.Collectors;
```
bjhargrave, line 17:
```suggestion
```

You don't need this import now (it will make a warning in Eclipse).

bjhargrave (approved):

bjhargrave:
Could you squash the 5 commits into 1?

maho7791:
I squashed the commits.

Issue #5462, opened 2022-12-07
Introduction of option to disable setting 'launch.startlevel.default'


amitjoy:
Currently we set the start levels of all the bundles in our runtime which has been specified in `-runbundles+` and `-runstartlevel`. Currently, this is set as a framework property which bnd uses to understand if the runtime uses start level functionality. If bnd finds out that the start level is used, it takes control of the runtime. Underneath, bnd tries to find out the maximum start level defined in the aforementioned instructions. After finding out the currently configured maximum start level amongst the list of bundles, it uses the `number + 1` as the default start level which bnd stores in `launch.startlevel.default`. When we use a firmware with only firmware bundles installed, it sets the value to `11` as for the firmware bundles, the maximum start level is set to `10`. When runtime management agent is at play and downloads a target configuration from the management server, the management agent takes control for the time being and sets the start levels to the ones specified in our metadata bundle that comprises the start levels of all the bundles to be installed in the runtime. Now, if we restart the VM, bnd takes control once again and uses `launch.startlevel.default` and sets this number as the start levels of all the bundles in the runtime which is unexpected since the start ordering doesn't happen according to our expectation at all. That's is why, it would be better to provide an option for `-runstartlevel` to disable setting `launch.startlevel.default` property so that bnd doesn't set the start levels of all the bundles at all.

For example,
```
-runstartlevel =     \
                               begin =    0,\
                               step =   -1,\
                               default = -1,\
                               order =    sortbynameversion
```

The `default` option if set to `-1` can be used to disable setting the property.

PR #5474, opened 2022-12-12
#5462 Limit the Launcher to set startlevels to only the bundles in scope

pkriens:
During bnd 6.2.0 the behavior of the launcher was changed to set start levels on all bundles installed, even if they were not listed in the run bundles scope. This was recently discovered to wreak havoc in systems that use an agent to install additional bundles. I.e. the core set of run bundles were used to create an executable JAR that provided an environment that allowed the agent to install the bundles needed to satisfy the constraints of a remote management system. This management system also provided start levels. The change caused the launcher to set all bundles the agent introduced to the default level. Since the agent used higher numbers, any required ordering was lost.

This change introduces the idea of _narrow_ and _all_ scope. It provides a `-launcher` instruction that can set the `manage` option. This options is either `all` (default), `narrow`, and `none`. This `manage` scope is not limited to start levels in the future. It indicates that the launcher must be aware of additional bundles that are not in its run bundles.

The Project Launcher uses this instruction to set `launch.startlevel.default` to either 0 (none), a negative value (narrow), or a positive value for all. This is then processed appropriately in the Launcher.

The default for the manage instruction is set in the current 7.0.0 defaults.bnd

---

### DIFF
--- a/aQute.libg/src/aQute/lib/startlevel/StartLevelRuntimeHandler.java
+++ b/aQute.libg/src/aQute/lib/startlevel/StartLevelRuntimeHandler.java
@@ -157,16 +157,28 @@ public class StartLevelRuntimeHandler implements Closeable {
 	static public StartLevelRuntimeHandler create(Trace logger, Map<String, String> outerConfiguration) {
 
 		String defaultStartlevelString = outerConfiguration.get(LAUNCH_STARTLEVEL_DEFAULT);
-		if (defaultStartlevelString == null) {
+		if (defaultStartlevelString == null || defaultStartlevelString.equals("0")) {
 			logger.trace("startlevel: not handled");
 			return absent();
 		}
 
-		int defaultStartlevel = toInt(defaultStartlevelString, 1);
+		int defaultStartlevel;
+		boolean manageAll;
+
+		int tmp = toInt(defaultStartlevelString, 1);
+		if (tmp > 0) {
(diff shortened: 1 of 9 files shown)


aQute.libg/src/aQute/lib/startlevel/StartLevelRuntimeHandler.java
```diff
@@ -157,16 +157,28 @@ public void close() {}
 	static public StartLevelRuntimeHandler create(Trace logger, Map<String, String> outerConfiguration) {
 
 		String defaultStartlevelString = outerConfiguration.get(LAUNCH_STARTLEVEL_DEFAULT);
-		if (defaultStartlevelString == null) {
+		if (defaultStartlevelString == null || defaultStartlevelString.equals("0")) {
```
bjhargrave, line 160: After the null check, you should `toInt` the value to handle trimming and equivalence ("+0"). A simple "0" equals test is not robust enough.

bjhargrave (commented):
Looks good. One comment on the check for a zero default start level value.

aQute.libg/src/aQute/lib/startlevel/StartLevelRuntimeHandler.java
pkriens: the trimming is already done in bnd, who sets this property. But I can change it.

PR #5477, opened 2022-12-14
Make BND compatible with SLF4J-2

HannesWell:
SLF4J-2 is a binary compatible drop-in replacement for SLF4J-1 and therefore artifacts compiled against the SLF4J-1 API can use SLF4J-2 as well:
- https://www.slf4j.org/faq.html#changesInVersion200
- https://www.slf4j.org/faq.html#compatibility

To make BND compatible with SLF4J-2 it is sufficient to widen the 'Import-Package' version-range for org.slf4j packages to "[1.x,3)". This change allows to use BND in runtimes where either SLF4J-1 or SLF4J-2 is present.

I hope this simple change covers all all relevant cases.

### DIFF
--- a/biz.aQute.bnd/bnd.bnd
+++ b/biz.aQute.bnd/bnd.bnd
@@ -36,7 +36,7 @@
 
 
 Import-Package: \
- org.slf4j,\
+ org.slf4j;version="${range;[==,3)}",\
  !com.google.appengine.*,\
  !com.google.apphosting.*,\
  !com.cloudius.util.*,\


pkriens:
I feel this is a bad idea and quite contrary to what bnd is about. 

A bundle should describe what it _needs_ and not try to assume it knows more about the world. If the SLF4J does not  support semantic versioning it should change its export package to support multiple versions. Although this is also cheating, at least they are the authority about this backward compatibility and they will look silly not supporting semantic versioning, not us. They should not use a breaking version when in reality it is backward compatible. Of course the best solution would be if they used package versions. They could then still use the marketing version to whatever but keep the API packages representing reality. 

This looks like an easy fix but imho these small hacks will metastasize in your builds over time. OSGi works like a charm if you do not cheat but once you start cheating it quickly goes feral. 

The fact that we _can_ do this non-semantic imports does not imply we _should_. Fix the root problem, not the symptom or the developers that will come after you will regret it.

jcompagner:
@pkriens  this is interesting discussion,. how would you solve that exactly in SLF4J?

they can't say to there org.slf4j package in a 2.0 release that that is a 1.7 package..
because that would not work for people actually starting to use 2.0 features

Because the Logger api is backward compartible, so a 2.0 lib works fine against a 1.7.x compiled version because they didn't break any api for that
But the Logger api has a few new funtions like:

org.slf4j.Logger.atError() which returns a Builder object to construct the messages..

So how would you see that? How can that be any other then libaries opening up what they support
because BND uses 1.7 api, but will work fine in 2.0 api..
But if BND manifest says i only accept 1.x api then it will hold back all other stuff that uses bnd but does want to use 2.0

i guess what we really want is just a minimum version, like what most applications or libraries i know work with if you compare which java version they run on, its a minimum of java8 or java11 but never really see an upper bound in that.

HannesWell:
> I feel this is a bad idea and quite contrary to what bnd is about.
> 
> A bundle should describe what it _needs_ and not try to assume it knows more about the world. If the SLF4J does not support semantic versioning it should change its export package to support multiple versions. 

Yes, that was a consideration of myself as well, when I started the migration of Eclipse to SLF4J-2, but wasn't sure if this is a good idea. I'll propose that to SLF4J.

> @pkriens this is interesting discussion,. how would you solve that exactly in SLF4J?
> 

SLF4J can simply export its packages in multiple versions, just like:
```
Export-Package: org.slf4j.api;version="2.0.5",org.slf4j.api;version="1.7.36"
```
For example `org.apache.felix:org.apache.felix.http.servlet-api:1.2.0` is exporting multiple versions of the `javax.servlet` packages.

timothyjward:
> SLF4J-2 is a binary compatible drop-in replacement for SLF4J-1 and therefore artifacts compiled against the SLF4J-1 API can use SLF4J-2 as well:

Note that due to the lack of baselining this statement isn't totally true. For example there is a break in binary compatibility between: [the current main](https://github.com/qos-ch/slf4j/blob/f85104040ce44ac545e15e4f41ef771a7a7f7add/slf4j-api/src/main/java/org/slf4j/event/LoggingEvent.java) and [the 1.7 maintenance branch](https://github.com/qos-ch/slf4j/blob/e9ee55cca93c2bf26f14482a9bdf961c750d2a56/slf4j-api/src/main/java/org/slf4j/event/LoggingEvent.java)

Specifically the `getMarker` method was deleted and replaced by `getMarkers` in `org.slf4j.event.LoggingEvent`.

bjhargrave:
This is a bad idea at this time. (Also the PR only affects one jar in the whole Bnd build.)

Bnd jars are used in Maven and Gradle, both of which are still based upon slf4j 1.7. We would need the rest of the ecosystem to step up to slf4j 2.x when we do.

bjhargrave:
Also, slf4j 2.x is not semantically backwards compatible with slf4j 1.x. While it is mostly backwards compatible at the syntactic API signature level, it is definitely not backwards compatible at the behavioral level in regards to how bindings are located. While an slf4j API client may not care about this, it is incorrect to state that slf4j 2.x is backwards compatible with slf4j 1.x.

jcompagner:
that is easy to say again,
but from a user point of view (and i tested our product fully)
it works like a charm. i don't have any problems everything is logged as it should

But our product was hold back (still on 1.8beta by the way not 1.7) because it needs to work full in an OSGI environment..
(so we needed more manifest entries to have the service loader mechanism working)

But we couldn't move because the Apache http client component libs where forcing 1.7
and because of the dependency (one plugin of us needs to include http client and also other stuff that has then 2.0) that completely fails in OSGi.

So now we are hold back because of 3rd party libs that work fine with 2.0 (in the end all those libs really use the pure logging api..)


I guess that binary break in the LoggingEvent should kind of be reported so they add it back in (but maybe as deprecated..)

HannesWell:
> > SLF4J-2 is a binary compatible drop-in replacement for SLF4J-1 and therefore artifacts compiled against the SLF4J-1 API can use SLF4J-2 as well:
> 
> Note that due to the lack of baselining this statement isn't totally true. For example there is a break in binary compatibility between: [the current main](https://github.com/qos-ch/slf4j/blob/f85104040ce44ac545e15e4f41ef771a7a7f7add/slf4j-api/src/main/java/org/slf4j/event/LoggingEvent.java) and [the 1.7 maintenance branch](https://github.com/qos-ch/slf4j/blob/e9ee55cca93c2bf26f14482a9bdf961c750d2a56/slf4j-api/src/main/java/org/slf4j/event/LoggingEvent.java)
> 
> Specifically the `getMarker` method was deleted and replaced by `getMarkers` in `org.slf4j.event.LoggingEvent`.

Thanks for that remark. I didn't knew that and just trusted to linked documentation. As suggested by Johan, I created https://jira.qos.ch/browse/SLF4J-577.


> Bnd jars are used in Maven and Gradle, both of which are still based upon slf4j 1.7. We would need the rest of the ecosystem to step up to slf4j 2.x when we do.

This change is not about requiring slf4j-2, it is about being compatible with slf4j. So it is totally fine, if bnd only requires slf4j, but it should also work with slf4j-2 (given that it is not affected by the mentioned binary incompatibilities or the binary compatibility is restored).

> Also, slf4j 2.x is not semantically backwards compatible with slf4j 1.x. While it is mostly backwards compatible at the syntactic API signature level, it is definitely not backwards compatible at the behavioral level in regards to how bindings are located. While an slf4j API client may not care about this, it is incorrect to state that slf4j 2.x is backwards compatible with slf4j 1.x.

That's right. But I think that is an issue for the person that assembles the final applications. At least for Eclipse I spent some amount of work to restrict application/product builders in that decision, mainly be using Import-Package only and removing the bundle from features to let P2 handle its inclusion dynamically. 

> A bundle should describe what it _needs_ and not try to assume it knows more about the world. If the SLF4J does not support semantic versioning it should change its export package to support multiple versions.

Just created https://jira.qos.ch/browse/SLF4J-576 in order to discuss the export of multiple versions.

pkriens:
>  it works like a charm. i don't have any problems everything is logged as it should

Some people  are just lucky ... :-) 

One of the key things that bnd aspires to is to prevent runtime issues during the build process but this is predicated on having the correct meta data. We went out of our way to support this model with automatic generation and extensive verification of  metadata and even provide baselining to detect API compatibility errors while you make them. Once you become sloppy with your metadata, bnd and OSGi just become very expensive dead weights. Something I sadly see too often.

I admit I am a bit defeated in this quest because I've see too many developers sell their OSGi soul for a short term relief and then later blame OSGi for their predicament. A sense of feeling Cassandra has become not foreign to me.

Anyway, I already was highly suspicious that SLF4J 2.0 was not really backward compatible and Tim confirmed my suspicion. If civil engineers worked this way we would see a lot more Genua bridges. The fact that it works for you in your particular situation does not mean it works for everybody. It is uncanny how all the dark holes will be visited when you have hundreds or thousands of users. 

There is no simple solution in this situation ... Multiple exports do not work reliably because, as Timothy Ward shows, there are breaking changes; the major semantic version bump seems correct. The only thing that _might_ work is make a facade bundle with the old API that uses reflection to map the old API to the new API. Since the SLF4J API normally does not show up in the public API of a bundle, the support for multiple versions will allow you to install the SLF4J 2.0 and 1.x facade at the same time. Sucks, but it is the consequence of creating non-backward releases. If you really want to delete stuff, use another package name. Te go practice to use major versions (`org.slf4j.v2`) is really smart, it makes it easy to handle breaking changes and provide compatibility at minimum cost.

Admittedly, as a long time subjugate of bnd baselining, this is surprisingly hard. BJ Hargrave and I once tried to create a 2.0 OSGi APIs but despite our heroic efforts and sleepless nights we had to surrender in the end. This is the reason you still see Dictionary and some null returns in the OSGi API :-( For API packages it is a bit like the hotel California ... you can add something but you can never remove anything. The only strategy that seems to work well imho is baselining. Where you get an error at the moment you violate backward compatibility. Stopping and thinking then often allows you to find another solution that is backward compatible and satisfies your needs.

Anyway, people with OSGi problems invariably have fake metadata _somewhere_, it is prevalent in many open source projects. Sad, because it is so simple to reach magical bless with OSGi, just tell the truth and she rewards you handsomely. Start cheating and she becomes a mean vicious ....

jcompagner:
for SLF4J (which is really simple api if you ask me) to stay backward compatible should be relatively easy
We have a product that is backward compatible for +/- 20 years... Only stuff that we say yeah this really had to break, we did.
But for the most part people building (using our product) that they build 10 years ago will still run fine on top of our product today...

But we are already for years on SLF4J1.8 beta .. because that had this:

Require-Capability: osgi.extender;filter:="(osgi.extender=osgi.service
 loader.processor)",osgi.serviceloader;filter:="(osgi.serviceloader=or
 g.slf4j.spi.SLF4JServiceProvider)"

in our manifest.. that the 1.8 beta had for years. 

But there is another big problem if I can't upgrade now to 2.0.... our backend is log4j..
Log4j stopped supporting SLF4J 1.8, they now only support 2.0
So i can't at the same time also not update to the latest log4j because of that because i don't have the correct binding..

So if there is now suddenly a security problem found in log4j i have a problem.. 

But if I would have known about that multi package version export that you can have in the manifest..
I would have created a case like this: https://jira.qos.ch/browse/SLF4J-576 +3 months ago....

bjhargrave:
I think we can close this PR as there is no agreement to make such a change (and also this PR is insufficient to make the change to all artifacts in the build.)

HannesWell:
> I think we can close this PR as there is no agreement to make such a change (and also this PR is insufficient to make the change to all artifacts in the build.)

Alright. When https://jira.qos.ch/browse/SLF4J-577 lands that should work as well.

Out of curisosity, what would I have need to do in order make the change for all artifacts of BND?

For your information, https://jira.qos.ch/browse/SLF4J-577 was closed as invalid, because the `LoggingEvent` used to be internal in SLF4J-1 and is only used by implementors/providers in SLF4J-2. I will have a more detailed look later and will also check if there are other binary incompatible changes, but so for it looks like SLF4J-1 and 2 are really binary compatible from a client perspective.
Nevertheless I think the fact that some packages are internal in SLF4J-1 should be reflected in https://jira.qos.ch/browse/SLF4J-577 and only those packages intended for clients should be exported in a 1 version:
https://jira.qos.ch/browse/SLF4J-576?focusedCommentId=21553&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21553
But I suggest to discuss that in the SLF4J issue tracker, to have the information there.

PR #5482, opened 2022-12-16
fixed missing build and testpath for bnd maven generator

juergen-albert:

### DIFF
--- a/maven-plugins/bnd-generate-maven-plugin/src/main/java/aQute/bnd/maven/generate/plugin/GenerateMojo.java
+++ b/maven-plugins/bnd-generate-maven-plugin/src/main/java/aQute/bnd/maven/generate/plugin/GenerateMojo.java
@@ -10,6 +10,7 @@ import java.util.StringJoiner;
 
 import aQute.bnd.build.Project;
 import aQute.bnd.result.Result;
+import org.apache.maven.artifact.handler.manager.ArtifactHandlerManager;
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.plugin.AbstractMojo;
@@ -28,7 +29,7 @@ import org.eclipse.aether.RepositorySystemSession;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@Mojo(name = "bnd-generate", defaultPhase = LifecyclePhase.GENERATE_SOURCES, requiresDependencyResolution = ResolutionScope.NONE, threadSafe = true)
+@Mojo(name = "bnd-generate", defaultPhase = LifecyclePhase.GENERATE_SOURCES, requiresDependencyResolution = ResolutionScope.TEST, threadSafe = true)
 public class GenerateMojo extends AbstractMojo {
 
 	protected final Logger			logger	= LoggerFactory.getLogger(getClass());
@@ -72,6 +73,9 @@ public class GenerateMojo extends AbstractMojo {
 
 	@Component
 	private RepositorySystem									system;
+
+	@Component
+	private ArtifactHandlerManager	artifactHandlerManager;
 
 	/**
 	 * File path to a bnd file containing bnd instructions for this project.
@@ -128,6 +132,7 @@ public class GenerateMojo extends AbstractMojo {
 		try {
 			BndContainer container = new BndContainer.Builder(project, session, repositorySession, system)
 				.setDependencies(normalizedDependencies)
+				.setArtifactHandlerManager(artifactHandlerManager)
 				.setAdditionalProperiets(additionalProperties)
 					.build();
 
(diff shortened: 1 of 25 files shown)


maven-plugins/bnd-generate-maven-plugin/src/main/java/aQute/bnd/maven/generate/plugin/GenerateMojo.java
```diff
@@ -28,7 +29,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@Mojo(name = "bnd-generate", defaultPhase = LifecyclePhase.GENERATE_SOURCES, requiresDependencyResolution = ResolutionScope.NONE, threadSafe = true)
+@Mojo(name = "bnd-generate", defaultPhase = LifecyclePhase.GENERATE_SOURCES, requiresDependencyResolution = ResolutionScope.TEST, threadSafe = true)
```
bjhargrave, line 32: I think there is a problem with using the [`generate-sources`](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#default-lifecycle) phase to generate test source code. That should be done at the `generate-test-sources` phase. So I think you will want two mojos here. One for `generate-sources` phase and another for `generate-test-sources` phase.

Also, while we are at it, can we rename the mojo for the `generate-sources` phase to `name="generate"`? (None of the other mojo names start with `bnd`.) Then the name for the mojo for the `generate-test-sources` phase can be `name="generate-test"`. Then on the command line the goals are `bnd-generate:generate` and `bnd-generate:generate-test` (with `bnd-generate` being the plugin prefix).

bjhargrave (changes requested):

maven-plugins/bnd-generate-maven-plugin/src/main/java/aQute/bnd/maven/generate/plugin/GenerateMojo.java
juergen-albert: I can do the Renaming.

> I think there is a problem with using the [generate-sources](https://maven.apache.org/guides/introduction/introduction-to-the-lifecycle.html#default-lifecycle) phase to generate test source code. That should be done at the generate-test-sources phase. So I think you will want two mojos here. One for generate-sources phase and another for generate-test-sources phase.

Right now, there is no distinction between generating test or normal sources, because the underlying codegenerator can't make such a distinction either. The goal is to wirte a generator once and it will run in the bnd, gradle and maven workspace model equally. I added the resolution Scope, so I can have the required dependencies at hand and set it is build and testpath, so it would work similar to bnd. 
However, I can see the merit in providing a second version that can run in the `generate-test-sources` phase. It would be the exact same thing and both would resolve dependencies up to the test scope.

bjhargrave: > It would be the exact same thing and both would resolve dependencies up to the test scope.

That is wrong. generate-sources phase should not need or use test scope dependencies. That can lead to leaking of test scoped dependencies into non-test code. Code generated by the generate-sources phase is compiled by the compile phase which only uses compile scoped dependencies.

If you want to generate source and test source in a maven build, you need 2 executions of the generate plugin. One for generate (generate-sources phase) and one for generate-test (generate-test-sources phase).

juergen-albert: Acknowledged!

The generator can basically generate anything. Thus it can also generate any kind of resources. Do we need two more versions that have the `generate-resources` and `generate-test-resources` as default Lifecyclephase as well?

bjhargrave: I think we just need two mojos where one (`name="generate"`) has the default phase `generate-sources` and has `ResolutionScope.COMPILE` and the other (`name="generate-test"`) has the default phase `generate-test-sources` and has `ResolutionScope.TEST`. 

If someone wants to generate for the `generate-resources` phase, they can configure an execution of the first mojo (`name="generate"`) for that phase. If someone wants to generate for the `generate-test-resources` phases, they can configure an execution of the second mojo (`name="generate-test"`) for that phase.

maven-plugins/bnd-generate-maven-plugin/src/main/java/aQute/bnd/maven/generate/plugin/GenerateMojo.java
```diff
@@ -1,222 +1,17 @@
...
-	 * <p>
-	 * The bnd instructions for this project are merged with the bnd
-	 * instructions, if any, for the parent project.
-	 */
-	@Parameter
-	// This is not used and is for doc only; See {@link
-	// BndConfiguration#loadProperties(Processor)}
-	@SuppressWarnings("unused")
-	String							bnd;
+@Mojo(name = "generate", defaultPhase = LifecyclePhase.GENERATE_SOURCES, requiresDependencyResolution = ResolutionScope.RUNTIME_PLUS_SYSTEM, threadSafe = true)
```
bjhargrave, line 9: This should be `ResolutionScope.COMPILE` (compile + system + provided dependencies). Runtime dependencies should not be included since the generated code will be compiled with compile scope dependencies.

juergen-albert: Done

bjhargrave (approved):

juergen-albert:
Thanks and also for your cleanup commit.

PR #5499, opened 2023-01-10
Drop back to Java 11 as a base for the bnd launcher

timothyjward:
The bnd launcher code is deliberately minimal as it is used in any exported runnable application. The move of bnd to Java 17 has suddenly required that everyone using the bnd launcher uses Java 17, even if there is no other part of their runtime that needs Java 17. This commit rolls the launcher back to a Java 11 base, while leaving the rest of bnd at Java 17. This allows applications to build and export using bnd 7.x and Java 17, but to still target Java 11 as a base runtime.


bjhargrave (changes requested):
This is only a partial solution. The `aQute.libg` project is still compiled with Java 17 and some of those packages are included in `biz.aQute.launcher` and used by `aQute.launcher.Launcher` for example. If we want to change the Java level of a project, we would also need to similarly changes the Java level of its dependencies.

Also, why just back down to Java 11 and not Java 8 if your goal is maximum support?

bjhargrave:
If we do such a thing (mixed Java versions across the Bnd projects), we would need to add some testing to confirm the launcher properly launches in the desired level of VM. Since the PR is only a partial solution, such a test would have caught this issue.

I think we end up with too many corner cases if we try to mix the Java versions across projects especially with our common use of `-conditionalpackage`. And testing to confirm desired behavior of the corner cases will be complex.

timothyjward:
> Also, why just back down to Java 11 and not Java 8 if your goal is maximum support?

The decision was made to move bnd away from Java 8 in version 7. We were both part of that decision process along with other members of the team and I’m not looking to reverse it. I agree that it’s not a good thing for libraries to be stuck with ancient versions of things “just in case”. The reason for selecting Java 11 is that Java 11 is (unlike Java 8) still in active support and not such a huge jump as 8 to 17.

bjhargrave:
> The reason for selecting Java 11 is that Java 11 is (unlike Java 8) still in active support

Java 8 is still in active support from the community. Per https://adoptium.net/support/, Java 8 has LTS support through Nov 2026. Java 11 LTS support is only through Oct 2024. So from that POV, Java 8 is a better choice :-)

pkriens:
I am pretty against this, we discussed this widely. However, I'd be ok to have a 6.4.x branch that we build against Java 8 and only make serious bug fixes. Basically the branch should be left to die over time but people really constrained could you use parts of it.

I do not want a workspace with different versions. These kind of compromises create a lot of unnecessary complexity and problems downstream.

Issue #5498, opened 2023-01-10
Print version of contained items


laeubi:
I'm getting the following error from bnd-testing-maven-plugin (**highlights** by me):

> Error   : tck **org.osgi.test.cases.jdbc**;version=[8.1.0,9) Not found in [bnd-cache [...], tck [org.eclipse.osgi,org.osgi.service.component,org.objenesis,assertj-core,org.opentest4j,net.bytebuddy.byte-buddy,junit-platform-launcher,junit-vintage-engine,org.osgi.util.function,org.mockito.mockito-core,osgi.annotation,net.bytebuddy.byte-buddy-agent,org.osgi.test.common,junit-jupiter-params,junit,org.apiguardian.api,org.apache.servicemix.bundles.junit,org.osgi.test.assertj.framework,junit-jupiter-engine,org.osgi.namespace.extender,org.apache.felix.scr,junit-platform-commons,org.osgi.util.promise,junit-platform-engine,org.osgi.test.junit4,org.awaitility,org.hamcrest,org.osgi.test.junit5,**org.osgi.test.cases.jdbc**,junit-jupiter-api,org.codehaus.mojo:animal-sniffer-annotations r/w=false]]

As one sees, the complained item **org.osgi.test.cases.jdbc** is actually there (but it seems the version do not match), to better understand the problem it might be good to also print the version of the contained items in the error message...

bjhargrave:
Try running maven with debug and see if more information to diagnose is available.

But my first guess is that Bundle-SymbolicName `tck org.osgi.test.cases.jdbc` is not equal to `org.osgi.test.cases.jdbc`. I am not sure what the `tck` is prefixing the BSN.

laeubi:
> Try running maven with debug and see if more information to diagnose is available.

No additional information

> But my first guess is that Bundle-SymbolicName tck org.osgi.test.cases.jdbc is not equal to org.osgi.test.cases.jdbc. I am not sure what the tck is prefixing the BSN.

`tck` is the project name of the maven project, so has nothing to do with the `Bundle-SymbolicName`, the specification in the runbundles looks like this: `org.osgi.test.cases.jdbc;version='[8.1.0,9)'`, but I think you can reproduce it with any setup that has a false version range (even though I can't see here why it should be wrong as the `Bundle-Version: 8.1.0.202212101352` should match, but knowing what BND has in his internal repo would probably reveal an issue, therefore the request to print some more infos.

[...]

PR #5502, opened 2023-01-11
repo: Provide bsn+version in FileSetRepository.toString

bjhargrave:
This will provide more detail in Bnd errors which may be reported.

Fixes https://github.com/bndtools/bnd/issues/5498


laeubi:
Thanks!

PR #5519, opened 2023-01-17
Launcher check

### LINKED ISSUE #5447: "Launch if project contains errors" not honoured
1. Navigate to Preferences -> Run/Debug -> Launching 
2. Set "Continue launch if project contains errors" to "Prompt"
3. Build in Eclipse.
4. Introduce a compile error in your project.
5. Build in Eclipse.
6. Launch. No prompt appears. The last error-free build is launched.

pkriens:
"Launch if project contains errors" not honoured #5447

The code was bypassing the inherited checks from Java. I felt it was too dangerous to restore this so I used the code to do this check inline in the final check.

Refactored the code a bit to make it more readable.

This fixes #5447 
---

### DIFF
--- a/org.bndtools.launch/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
+++ b/org.bndtools.launch/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
@@ -11,6 +11,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.logging.Level;
@@ -25,6 +26,7 @@ import org.eclipse.core.resources.IResourceChangeListener;
 import org.eclipse.core.resources.IResourceDelta;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IAdaptable;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
@@ -33,12 +35,15 @@ import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.jobs.IJobFunction;
 import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.debug.core.DebugPlugin;
 import org.eclipse.debug.core.IDebugEventSetListener;
 import org.eclipse.debug.core.ILaunch;
 import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.debug.core.IStatusHandler;
 import org.eclipse.debug.core.model.IProcess;
+import org.eclipse.debug.internal.core.DebugCoreMessages;
 import org.eclipse.jdt.launching.IVMInstall;
 import org.eclipse.jdt.launching.JavaLaunchDelegate;
 import org.eclipse.jdt.launching.JavaRuntime;
@@ -51,6 +56,7 @@ import org.osgi.service.component.annotations.Reference;
 import aQute.bnd.build.Project;
 import aQute.bnd.build.ProjectLauncher;
 import aQute.bnd.build.Run;
+import aQute.bnd.build.Workspace;
 import aQute.bnd.osgi.Jar;
 import bndtools.Plugin;
 import bndtools.StatusCode;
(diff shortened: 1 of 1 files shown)


pkriens:
Ah well, lets try it out ...

kriegfrj:
> Ah well, lets try it out ...

Sorry Peter, I'm on leave... I'll try and have a look next week.

pkriens:
I decided to just commit and see ... it is a small change and we have some time to figure it out. 

Enjoy your leave :-)

org.bndtools.launch/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
```diff
@@ -171,70 +182,153 @@ public boolean buildForLaunch(ILaunchConfiguration configuration, String mode, I
...
+			.filter(Objects::nonNull)
+			.toArray(IProject[]::new);
 	}
 
 	@Override
 	public void launch(ILaunchConfiguration configuration, String mode, final ILaunch launch, IProgressMonitor monitor)
 		throws CoreException {
 		try {
+
+			IEclipsePreferences prefs = InstanceScope.INSTANCE.getNode("org.eclipse.debug.core");
```
kriegfrj, line 330: I can't see where this is used?

org.bndtools.launch/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
```diff
@@ -171,70 +182,153 @@ public boolean buildForLaunch(ILaunchConfiguration configuration, String mode, I
...
+			return getAllIProjects(p.getWorkspace());
+		}
+		return Central.getProject(p)
+			.map(x -> new IProject[] {
+				x
+			})
+			.orElse(new IProject[0]);
+	}
+
+	private IProject[] getAllIProjects(Workspace ws) {
```
kriegfrj, line 316: Any reason to not make this `static`? Seems to be a helper method that doesn't access any non-static fields.

org.bndtools.launch/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
```diff
@@ -171,70 +182,153 @@ public boolean buildForLaunch(ILaunchConfiguration configuration, String mode, I
...
+					List<IAdaptable> errors = new ArrayList<>();
+					for (IProject project : projects) {
+						monitor.subTask(
+							MessageFormat.format(DebugCoreMessages.LaunchConfigurationDelegate_7, new Object[] {
+								project.getName()
+						}));
+						if (existsProblems(project)) {
+							errors.add(project);
+						}
+					}
```
kriegfrj, line 276: The preferred way to use a progress monitor is not to call `beginTask()`/`done()` directly, but to use a `SubMonitor`. From [here](https://wiki.eclipse.org/Progress_Reporting): 

> Whenever you write a method that accepts an IProgressMonitor, the first thing you should do is convert it to a SubMonitor using SubMonitor.convert(...)

We could also do a better job of reporting progress by setting the number of ticks equal to the number of `IProject`s and calling  `worked(1)` in the loop, also setting the status message to the name of the project. 99 times out of 100 this will probably not make much of a difference to the end user, but in that one time out of 100 it could be a useful debugging tool.

org.bndtools.launch/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
```diff
@@ -171,70 +182,153 @@ public boolean buildForLaunch(ILaunchConfiguration configuration, String mode, I
...
+								.booleanValue();
+						}
+					}
+
+					return true;
+				});
+		} catch (Exception e) {
+			throw new CoreException(Status.error("Unexpected while checking compile errors", e));
+		} finally {
+			monitor.done();
```
kriegfrj, line 292: This will not be necessary if you use a `SubMonitor`.

org.bndtools.launch/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
```diff
@@ -171,70 +182,153 @@ public boolean buildForLaunch(ILaunchConfiguration configuration, String mode, I
...
+						if (prompter != null) {
+							return ((Boolean) prompter.handleStatus(complileErrorProjectPromptStatus, errors))
+								.booleanValue();
+						}
+					}
+
+					return true;
+				});
+		} catch (Exception e) {
+			throw new CoreException(Status.error("Unexpected while checking compile errors", e));
```
kriegfrj, line 290: Would it be worth while to have two different catches here - one that encapsulates the bnd error checking code, and one that encapsulates the Java error checking code? That way we can report a more specific error message upon failure.

org.bndtools.launch/src/bndtools/launch/AbstractOSGiLaunchDelegate.java
```diff
@@ -171,70 +182,153 @@ public boolean buildForLaunch(ILaunchConfiguration configuration, String mode, I
...
 			run = LaunchUtils.createRun(configuration, getRunMode());
+			return Central.getWorkspace()
+				.readLocked(() -> {
+
+					initialiseBndLauncher(configuration, run);
+
+					if (!checkProjectErrors(configuration, mode, monitor))
+						return false;
+
+					if (prefs.getWarnExistingLaunches() && !checkExistingLaunches(configuration))
```
kriegfrj, line 195: There has been a subtle change here: The call here to `prefs.getWarnExistingLaunches()` is referencing the field called `prefs`. Before your changes, `finalLaunchCheck()` would instantiate a new local `BndPreferences` called `prefs` and use that. If I read the code correctly, this should not change the behaviour as `BndPreferences` doesn't cache the results, but reads them on-demand from the preference store. So this is probably an improvement.

kriegfrj (commented):
A few random thoughts... nothing really critical or fundamental.

Issue #5407, opened 2022-10-21
remove `-includeresource` as one of the Markers to ignore Annotations


juergen-albert:
`-includeresource` is together with `Export-Package` and `-privatepackage` a trigger to not look for `@Export` annotations. If you want to include resources with your bundle and still use the comfortable Annotations you need to give the `-includeresource` a namespace and then it works. This however is so much magic, that most of our customers are confused regularly and I need to explain to basically everybody at one point why suddenly all there Packages are marked as excluded. 

I would opt for changing this once an for all. I know that changing default behavior is bad and can break things. In this case however I believe it will not. 
I see the following scenarios:
1. use of `-includeresource` in a reosurce only bundle or one with no code: nothing changes
2. use of `-includeresource` with Annotations via a namespace: nothing changes
3. `-includeresource` with manual Package declaration: nothing changes
4. `-includeresource` with abandoned code that is not intended to be packaged: Well, in this case it will have an impact. But why should this be a pattern?

Anything I may have missed?

stbischof:
I am absolutely with you. Run to often in this.

Maybe just deprecate includeresources and create a Importresource instruction that does it .. right..

bjhargrave:
@pkriens suggests we instead change the defaults to provide the everything-in-project behavior rather than the current way we do this.

PR #5522, opened 2023-01-18
#5407 Decision to include project bin directory

pkriens:
Bnd was changed a couple of years ago to have a default expand when there were no instructions like -includeresource, Private-Package, etc. The implementation decided to put the whole output directory of the project directly in the to be build JAR. This meant that the normal processing did not happen. You had to use -exportcontents to export packages.

This PR changes the following

- It introduces a magic keyword for the `from:` directive. If this is `project`, it will search the class path entry that is the project's output directory and will include only from there. Therefore, doing a `-includepackage *;from:=project` will include the _whole_ project output directory.
- Created a set of EXPAND_HEADERS in Constants. The -includresource was not include.
- Updated the Project builder to set the `-includepackage *;from:=project` instruction in the builder when none of the EXPAND_HEADERS was set

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
@@ -76,6 +76,7 @@ import aQute.libg.generics.Create;
  * Import-Package: package-decl ( ',' package-decl )* @version $Revision: 1.27 $
  */
 public class Builder extends Analyzer {
+
 	private final static Logger			logger						= LoggerFactory.getLogger(Builder.class);
 	private final static Pattern		IR_PATTERN					= Pattern
 		.compile("[{]?-?@?(?:[^=]+=)?\\s*([^}!]+).*");
@@ -695,11 +696,9 @@ public class Builder extends Analyzer {
 
 			Attrs directives = e.getValue();
 
-			// We can optionally filter on the
-			// source of the package. We assume
-			// they all match but this can be overridden
-			// on the instruction
-			Instruction from = new Instruction(directives.get(FROM_DIRECTIVE, "*"));
+			String fromDirective = directives.get(FROM_DIRECTIVE, "*");
+			Instruction from = new Instruction(fromDirective);
+			boolean project = fromDirective.equals(FROM_DIRECTIVE_PROJECT);
 
 			boolean used = false;
 
@@ -728,7 +727,7 @@ public class Builder extends Analyzer {
 				}
 
 				// Do the from: directive, filters on the JAR type
-				List<Jar> providers = filterFrom(from, p.getValue());
+				List<Jar> providers = filterFrom(from, p.getValue(), project);
 				if (providers.isEmpty())
 					continue;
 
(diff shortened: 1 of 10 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
```diff
@@ -695,11 +696,9 @@ private Set<Instruction> doExpand(Jar jar, MultiMap<String, Jar> index, Instruct
...
 			Attrs directives = e.getValue();
 
-			// We can optionally filter on the
-			// source of the package. We assume
-			// they all match but this can be overridden
-			// on the instruction
-			Instruction from = new Instruction(directives.get(FROM_DIRECTIVE, "*"));
+			String fromDirective = directives.get(FROM_DIRECTIVE, "*");
+			Instruction from = new Instruction(fromDirective);
+			boolean project = fromDirective.equals(FROM_DIRECTIVE_PROJECT);
```
bjhargrave, line 701: This is too broad since it covers the `from` directive from more than just `-includepackage`.

biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
```diff
@@ -491,7 +492,29 @@ public interface Constants {
...
 
-	/*
+	/**
+	 * Headers that if **all** are absent will trigger the -includepackage
+	 * `*;from:=project` (include all packages from the project's output.
+	 */
+	String[]	EXPAND_HEADERS							= {
+		Constants.RESOURCEONLY, Constants.INCLUDEPACKAGE, Constants.PRIVATE_PACKAGE, Constants.PRIVATEPACKAGE,
+		Constants.EXPORT_PACKAGE,
+	};
```
bjhargrave, line 502:
```suggestion
	Set<String>	EXPAND_HEADERS							= Sets.of(
		Constants.RESOURCEONLY, Constants.INCLUDEPACKAGE, Constants.PRIVATE_PACKAGE, Constants.PRIVATEPACKAGE,
		Constants.EXPORT_PACKAGE,
	);
```

We have moved to use unmodifiable sets for such groups in Constants. This prevents mistakes like duplicates and mutation.

biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
```diff
@@ -2633,4 +2633,18 @@ private List<File> getSelfAndAncestors(List<File> l) {
...
 		this.propertiesFile = source;
 	}
+
+	/**
+	 * Answer true if any of the headers is set as a property
+	 *
+	 * @param headers list of headers
+	 * @return true if any of the headers is set
+	 */
+	public boolean isHeaderSet(String... headers) {
```
bjhargrave, line 2643: These are not all Headers. Some are instructions? Perhaps `isPropertySet` a better name for Processor?

bjhargrave (changes requested):

biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
```diff
@@ -749,10 +748,17 @@ private Set<Instruction> doExpand(Jar jar, MultiMap<String, Jar> index, Instruct
...
 	 */
-	private List<Jar> filterFrom(Instruction from, List<Jar> providers) {
+	private List<Jar> filterFrom(Instruction from, List<Jar> providers, boolean project) {
 		if (from.isAny())
 			return providers;
 
+		if (project) {
+			for (Jar j : providers) {
+				if (j.getResource(PROJECT_MARKER) != null) {
+					return Collections.singletonList(j);
```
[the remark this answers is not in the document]
pkriens: Yes, and that is the purpose? This is a nice feature in itself. Not sure what the problem is?

biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
pkriens: Yes, that is nice isn't it?

Issue #5060, opened 2022-01-14
[builder] No automatic build when build.bnd changes


kriegfrj:
As part of #5011, I took as my starting point these comments at the top of `BndtoolsBuilder`:

https://github.com/bndtools/bnd/blob/3b3c8a9c2d8e12f907ef52fc0753e95ed116a329/bndtools.builder/src/org/bndtools/builder/BndtoolsBuilder.java#L48-L63

However, I've now implemented most of these tests and found that many of them don't work exactly as expected. I suspect this is mostly  a case of improvements that haven't been reflect in the comments. However, there is is one that I think is a definitely a bug.

It seems changes to build.bnd do not trigger an automatic rebuild.

I have attempted to see if this behaviour is replicated in HEAD, and as far as I can tell it is. I tested on the Bnd workspace by creating a small test project with a single class and an empty bnd.bnd. Then I changed base.version to 6.3.0, and I can see that the jar still has 6.2.0 in it as the version. In fact, even if I remove test.jar the automatic rebuild still produces 6.2.0. It is not unless you do a clean that the new base.version is reflected in the output jar.

This is a bit finnicky and I may have done something silly to cause this behaviour. So if anyone could verify this behaviour, it would be appreciated. I'm a bit surprised if it is the case because it seems like a core behaviour; however perhaps because build.bnd is updated infrequently we haven't noticed much.

If the above is correct, it probably means that BndtoolsBuilder is caching some state somewhere.

pkriens:
I can confirm that changes to build.bnd are not propagated. Will see if I can make a quick fix

PR #5523, opened 2023-01-18
#5060 No automatic build when build.bnd changes

pkriens:
I could not get Eclipse to recognize the dependency between the cnf project changes and the other projects. So I used the new notifier mechanism to get changes in the workspace. When the repos are reported to be ready (and it is not the first time) a new full build is started.

Notice that whenever the registration is done, the notifier will send init/repos. That is why the first time it is ignored.

### DIFF
--- a/bndtools.core/src/bndtools/central/sync/WorkspaceChange.java
+++ b/bndtools.core/src/bndtools/central/sync/WorkspaceChange.java
@@ -1,0 +1,63 @@
+package bndtools.central.sync;
+
+import java.util.Collection;
+
+import org.eclipse.core.resources.IWorkspace;
+import org.eclipse.core.resources.IncrementalProjectBuilder;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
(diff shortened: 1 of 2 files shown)


bndtools.core/src/bndtools/central/sync/WorkspaceChange.java
```diff
@@ -0,0 +1,63 @@
...
+	Workspace				ws;
+
+	boolean					first		= true;
+	volatile boolean		cancel		= false;
+
+	@Activate
+	void activate() {
+		ws.on("workspace-changes")
+			.initial(this::init)
+			.repositoriesReady(this::done);
```
bjhargrave, line 34: I am not sure I see the distinction here. init and repos are only called together (in Workspace.forceInitialization). So init is called, sets cancel to true, repos is called and sets cancel to false before scheduling the job. I am not sure how done0 ever sees cancel==true?

pkriens: It handles the case when there is init -> repo-ready -> start job after 2 s -> init! -> enter job

It aborts a job if we're bound to receive another one. Maybe not such a likely case but it can happen.

PR #5565, opened 2023-02-24
Split the Jar class into its base (zip) and extensions

laeubi:
A Jar is actually a Zip file with a manifest, but currently is a bit of a mixture of both not clearly separated.

This splits the part that are more Zip-Specific into a base class 'Zip'... 

FYI @pkriens just an idea that came into my mind due to the recent discussions ...


pkriens:
I am in the middle of some changes in this area and I fail to see the purpose of this change.

pkriens:
I do not think this relevant anymore? I am closing it because I also do not like to split the JAR and ZIP. 

The JPMS specific parts are in JPMSModule from now on

PR #5566, opened 2023-02-24
Issue/5346 build multi release

pkriens:
This is a proposal to generate the manifest of JPMS versioned jars. After consideration, nothing is done to support multi release JARs in other places since they should work out of the box.

Due to Maven and plain gradle, the process to provide the OSGi data to a JAR is in the Builder (I think), where the jar of the builder is already filled with content. The builder then runs the plugins, creates xml, etc, calculates the manifest, sets it in the jar, and then runs the verifier. The caller can then write out the added resources so they can be included in the JAR. In workspace mode, we can directly use the Jar.

We already have support for creating the JPMS module-info.class. Unfortunately, this plugin ran in the Verifier! I guess this was a **hack** to get the completed manifest, when running in an AnalyzerPlugin, the manifest is not set yet. However since this plugin added the module-info.class, it changed the set of resources and thus could never be signed since it was not in the name section. This made it also awkward to create the module-info for the versioned files since it required running a Verifier. 

The approach taken is to add generic functions to the core parts and then concentrate all functions in a single plugin that runs in the Builder.

## Support Manifest Plugins

I added  support in the Analyzer to run a new ManifestPlugin just before and after the name section. The before plugin is allowed to modify the resource. 

## Incremental analysis

The simplest solution is to collapse a version in a multi-release JAR to a new Jar and restart the analysis. However, this is rather inefficient and might become problematic when the JARs are very big. For the conditional package support, we can incrementally add new classes/resources. This significantly speeds up the analysis. At least theoretically.

## Copying Analyzers

Copy support for an Analyzer, Jar, Descriptors, Parameters, and Contracts are created.

## JPMS Module

To collect all knowledge about JPMS I created a JPMSModule class. It can wrap a JAR and then provides convenience functions to access versioned JARs. It should also be used to store constants and any specific methods we need to work with JPMS. I'd like to migrate much of the JPMS plugins tooling to  this class. I did move the module name and version methods to here and the Jar file now refers to them. I experimented with deprecating them on Jar but that felt too pedantic. Clearly they do not belong there.

## JPMS Release Plugin
For the JPMS Release Plugin, it is necessary to use the existing analysis and only add the versioned jars. This should be done incrementally for each progressive release. At each stage, the manifest should be compared to the previous release version. If it differs, it should be set in the versioned directory.

This work cannot be done in the existing builder since we get called before the manifest is finished. Therefore, a copy of the analyzer that is used to make the main jar is made. This copy is then incrementally updated with the first release in the multi -release jar. 

Since this is an Analyzer, it will not create a module-info.class, that is only done in Builder. However, I think it makes sense to allow the versioned releases in the multi-release jar the pleasure of a module-info.class. Since this class is heavily based on the main jar and project settings since it is required to use the same module name, version, etc, it is almost impossible to build this in another way. I therefore add the JPMSModuleInfoPlugin by hand to the copied analyzer.

Now after the manifest is calculated, the copied JAR will also contain a new module-info.class. The Manifest and module-info are then copied to the versioned area of the main jar.

## Conclusion

I've broken up the changes in different commits to separate the generic parts from the specific parts. I tried to be a nice boy! The advantage is that the actual specific changes for the JPMS are really small and imho :-) very easy to read. The generic mechanism are easy to test standalone

For now, don't go ballistic on the code review. details Let's first get the big picture. If this works then we can do a more narrow code review. So please provide some feedback in the comments on this approach.

That said, if someone could take a look at the Maven code and verify that I've not broken anything that would be great. There was quite a lot of testing around this so I am reasonably confident we're ok. But trust and verify!


biz.aQute.bndlib/src/aQute/bnd/build/model/EE.java
```diff
@@ -240,4 +254,66 @@ private void init() {
 			throw Exceptions.duck(ioe);
 		}
 	}
+
+	public int getRelease() {
```
laeubi, line 258: There is already a method that gets the release from an EE automatically so no need to maintain this manually...

biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
```diff
@@ -544,6 +544,7 @@ public interface Constants {
 	String		STATIC_ATTRIBUTE							= "static";
 	String		SUBSTITUTE_ATTRIBUTE						= "substitute";
 	String		MODULE_INFO_CLASS							= "module-info.class";
+	String		JPMS_MULTI_RELEASE							= "-jpms-multi-release";
```
laeubi, line 547: There is really no such thing as "jpms-multi-release"  see:

https://docs.oracle.com/en/java/javase/16/docs/specs/jar/jar.html#modular-multi-release-jar-files

> A modular multi-release JAR file is a multi-release JAR file that has a module descriptor, module-info.class, in the top-level directory (as for a [modular](https://docs.oracle.com/en/java/javase/16/docs/specs/jar/jar.html#modular-jar-files) JAR file), or directly in a versioned directory.
> 
> ...
> 
> **A module descriptor is generally treated no differently to any other class or resource file**. A module descriptor may be present under a versioned area but not present under the top-level directory.

laeubi:
General remark: JPMS is just a *symptom* it is not really the *issue* with MR-Jars. Also having always a set of "MR-Plugin" and "No-MR-Plugin" variants seem not to scale very well. From a user POV I just want it to always work without enabling just another option (most of the timer uses won't know if they need MR jar or not...).

> After consideration, nothing is done to support multi release JARs in other places since they should work out of the box.

Just repeating question from previous PR: If the classpath do not matter, why do we have a classpath on the analyzer?

pkriens:
> General remark: JPMS is just a symptom it is not really the issue with MR-Jars. Also having always a set of "MR-Plugin" and "No-MR-Plugin" variants seem not to scale very well. From a user POV I just want it to always work without enabling just another option (most of the timer uses won't know if they need MR jar or not...).

That is why it is by default on, this will check the META-INF/versions and if it sees versioned directories it goes ahead. However, there are ALWAYS cases where some users (or test) cases do not like this default. So any feature must always have a flag that can disable this. This resulted in a lot of -noxxxxx options. One day I proposed to have version specific defaults so we can set a default for a version and I am using that here. See https://github.com/bndtools/bnd/pull/5566/files#diff-f2d5528348c0c8b873326dd479f712b890decb30d1fab8a9a1d9bc918184b1c4

> Just repeating question from previous PR: If the classpath do not matter, why do we have a classpath on the analyzer?

The classpath is crucial, we need to pickup versions of the packages to create the proper imports. However, a MRJ == JAR for this part. By definition, a MRJ must never ever introduce a new API. So the compiler will ignore the META-INF/versions directories and so does the Analyzer, it can by the spec not contain relevant information to depend on, it is invisible. So treating it as a normal JAR is fine.

Only when you _build_ a JAR, do you need to analyze the contained resources in the versions directory for additional requirements. And if they differ, create a new manifest + module-info.

laeubi:
> The classpath is crucial, we need to pickup versions of the packages to create the proper imports. However, a MRJ == JAR for this part. By definition, a MRJ must never ever introduce a new API. So the compiler will ignore the META-INF/versions directories and so does the Analyzer, 

What compiler we are talking about? Neither javac nor ecj is "ignoring" them, because then it won't be possible to compile code for different release targets, as Types can change between releases (and their "API" what simply is all visible public types and all package-private types of the same package and probably abstract of extended classes ... maybe more ...).

About "JPMS-API" [the spec that says ](https://docs.oracle.com/en/java/javase/16/docs/specs/jar/jar.html#modular-multi-release-jar-files)

```
Any versioned module descriptor that presides over a lesser versioned module descriptor or that at the top-level, M say, must be identical to M, with two exceptions:

    the presiding versioned descriptor can have different non-transitive requires clauses of java.* and jdk.* modules; and
    the presiding versioned descriptor can have different uses clauses, even of service types defined outside of java.* and jdk.* modules.
```

[Also the OSGi spec says that](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module-multireleasejar):

> the Framework must replace the values of the following manifest headers in the manifest with the values of these headers, if present, in the supplemental manifest file.
> 
> - Import-Package
> - Require-Capability

So certainly the _visible_ API (and therefore classes) can diverge between versions .... what results in different Imports/Requires (especially java.* packages and EEs)

pkriens:
Lets say we have a source code C that is compiled against a class V in an MRJ with v8 as base, and v9 and v17 as versions.  Can you give me an example where the content would vary based on what is stored in v9 and v17? 

It might be that they select v9 when compiled on Java 15? However, since the generated code must be identical to when they compile against v8 it seems unnecessary complex? Theoretically they could take advantage of using some new class format feature but for example Record would be visible in the API and that would make the code fail on a Java 8 compile. I can't think of an example where the v9 or v17 class could influence the bits in the C class file?

I haven't got time to look into the source code but it would be highly illogical imho to compile against the highest matching target version of the compiler. I see no gain and a lot of pitfalls. Then again, it would not be the first time Java surprised me with a willingness to inflict massive pain for for minute gain ... So it would be interesting if you could show this.

> So certainly the visible API (and therefore classes) can diverge between versions

The public API is the visible API for the compiler, it must be **identical** between the versions. There could be (package) private classes and/or methods added/removed in the versions, but the exact public API (and that is what the compiler sees) **must** be 100% identical, no ifs or buts. The implementation of these classes can have wildly different requirements but that is not an issue for the compiler nor the Analyzer. 

It is only important when the rubber meets the road, the runtime. Then the JVM & the framework. The versions can then vary the requirements of the bundle. 

The still open issue is how this is going to be handled by the resource indexer. The only possibility I see is to create a resource for each version but that kind of sucks since the public API is the same and that means you get multiple exporters for the same packages, which sucks. It will slow down the resolver. I am exercising fragment models in my head but have not stumbled on a clean solution yet in that jungle. I am worried we need to handle this at resolver level which would be very disappointing. Anyway, that is for when we got this working.

laeubi:
Not sure if this is a misunderstanding, that a type must be present in **all** versions, but just assume the following jar:

```
a/Test.class
META-INF/versions/11/a/Java11IsCool.class
```

If I use that code with my project and compile it against target >= 11 I can use the `Java11IsCool` type in my code, (technically Test.class can even be omitted its just here for clarification) even though it is not available for any target < 11, so a compiler need to take the versioned items into account.

And even with 

```
a/Test.class
META-INF/versions/11/a/Test.class
META-INF/versions/11/a/Great.class
```

`Test` type might have more methods e.g. here a new method that returns a value of type `Great` ...

pkriens:
From the [Java 17 Jar spec](https://docs.oracle.com/en/java/javase/17/docs/specs/jar/jar.html):

* If the class of the class file is public or protected then that class must preside over a class of the same fully qualified name and access modifier whose class file is present under the top-level directory.  
* **The public API exported by the classes in a multi-release JAR file must be exactly the same across versions**, hence at a minimum why versioned public or protected classes for class files under a versioned directory must preside over classes for class files under the top-level directory.

Therefore, in your example:

* `Java11IsCool` must be (package) private since it is not in the top level. Ergo, it cannot be compiled against when it is on someone's classpath, and
*  `Test` v11 **must** not have different public and protected members than `Test` top level so whatever target you compile it against, you're not allowed to see a different public API.

I'd now bet a large beer that the compiler only reads the top level (the spec is clearly written to allow this) but Java has always been lala land. So I owe you a beer if the compiler could see the versioned directories but from a specification point of view it is _crystal clear_ to me that the Analyzer & the compiler do not need to see the versioned directories. 

For something completely different. I want to merge this change so we can all play with it and get experience. Any objections by anybody?

laeubi:
Well I don't drink beer, but thanks for the offer :-)

Nerveless you might want to look at
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathMultiReleaseJar.java

and further on
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/MultiReleaseJarTests.java

where there are explicit tests that (under different target settings) compilation gives an error or not under exactly the circumstances I have described here. I have no deeper insight about javac ...

So I trust the compiler experts here, that "The public API exported by the classes" means I cannot have a method that return `int` in version 9 but String in version 11, but I can have another public method with a different name that is not present in version 9, because the exported API is still the same (it just has grown).

Nerveless a **compiler** must also be able to compile **private API** so even if our more strict interpretation holds true, a compiler still needs to be MR-Jar aware ... at least from my testing with such jars (with or without OSGi).

> For something completely different. I want to merge this change so we can all play with it and get experience. Any objections by anybody?

My comment regarding duplicate management of target release in EE still holds regardless of the rest of this discussion ...

biz.aQute.bndlib/src/aQute/bnd/build/model/EE.java
pkriens: that does not cover < 9

biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java
pkriens: this is an option for bnd to activate the plugin

biz.aQute.bndlib/src/aQute/bnd/build/model/EE.java
laeubi: Also please look at `test.model.EETest.checkEEHasValidRelease(EE)` that is even covered by a test...

laeubi:
> That is clearly not what the spec says ... 

The spec is actually not clear, but what the spec clearly does not say is that a compiler can ignore any version entries.

> make a test case where we can see a difference.

See testcase here: https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/MultiReleaseJarTests.java

it has a jar and it has expected outcomes that depending on the target level more types are discovered, and obviously they are somewhat "public" because otherwise they can't be referenced in the code sample.

pkriens:
@laeubi it is now in master & snapshot repos. Since I could not easily test this with maven I would appreciate it if you could take it for a spin. As far as I can see this covers Stefan Bischof's primary use case but would be good to know it actually works. So feedback would be welcomed.

laeubi:
I used this repository that has Multi-Release jars + and where I tried to add BND headers:

- https://github.com/laeubi/clickhouse-jdbc/tree/bnd-multirelease
- `export JAVA_HOME=<path to jdk > 11>`
- `mvn clean package -U -DskipTests -Drelease`
- inspecting `clickhouse-jdbc/clickhouse-http-client/target/clickhouse-http-client-0.3.3-SNAPSHOT.jar`

It has a MANIFEST.MF with OSGi headers, and it has `versions/9` / `versions/11` folders but it is **missing the supplemental manifests** in `versions/9/OSGI-INF/MANIFEST.MF` and `versions/11/OSGI-INF/MANIFEST.MF`

PR #5576, opened 2023-03-02
Default properties were not set in the analyzer

pkriens:
The analyzer therefore ignored multi release jars. The analyzer is now fixed up with the default properties __versiondefaults__ is not set, will never overwrite though.

Also fixed a small oversights:

* If the JAR already contains a module-info.class then it will not be overridden.
* The delta calculations will not set the module info flag

---

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/MultiReleaseTest.java
+++ b/biz.aQute.bndlib.tests/test/test/MultiReleaseTest.java
@@ -175,6 +175,35 @@ public class MultiReleaseTest {
 
 	}
 
+	@Test
+	public void testPlainBuilder() throws Exception {
+		try (Builder builder = new Builder()) {
+			builder.setProperty("-includeresource", """
+				sun_1_8/=compilerversions/src/sun_1_8/, \
+				META-INF/versions/9/jdk_9_0/=compilerversions/src/jdk_9_0/, \
+				META-INF/versions/11/jdk_11_0/=compilerversions/src/jdk_11_0/, \
+				META-INF/versions/17/jdk_17/=compilerversions/src/jdk_17/, \
+				META-INF/versions/19/jdk_19/=compilerversions/src/jdk_19/, \
+				""");
+
+			Jar jar = builder.build();
+			assertThat(builder.check()).isTrue();
+
+			assertThat(jar.getResource("META-INF/versions/9/META-INF/MANIFEST.MF")).isNotNull();
+			assertThat(jar.getResource("META-INF/versions/9/module-info.class")).isNotNull();
+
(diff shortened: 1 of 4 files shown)


biz.aQute.bndlib.tests/test/test/MultiReleaseTest.java
```diff
@@ -175,6 +175,35 @@ public void testMultiReleaseAnalysis() throws Exception {
...
+				META-INF/versions/9/jdk_9_0/=compilerversions/src/jdk_9_0/, \
+				META-INF/versions/11/jdk_11_0/=compilerversions/src/jdk_11_0/, \
+				META-INF/versions/17/jdk_17/=compilerversions/src/jdk_17/, \
+				META-INF/versions/19/jdk_19/=compilerversions/src/jdk_19/, \
+				""");
+
+			Jar jar = builder.build();
+			assertThat(builder.check()).isTrue();
+
+			assertThat(jar.getResource("META-INF/versions/9/META-INF/MANIFEST.MF")).isNotNull();
```
laeubi, line 192:
```suggestion
			assertThat(jar.getResource("META-INF/versions/9/OSGI-INF/MANIFEST.MF")).isNotNull();
```

The path for the supplemental manifest is **OSGi-INF** not **META-INF** see:

https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#framework.module-multireleasejar

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -1810,10 +1811,9 @@ protected void begin() {
 		if (inited == false) {
 			inited = true;
 			super.begin();
-
 			updateModified(getBndLastModified(), "bnd last modified");
+			About.fixup(this);
```
laeubi, line 1815: Should it really be part of the Analyzer and not the JPMSPlugin fetching its default value from the About.getDefaults()?

biz.aQute.bndlib.tests/test/test/MultiReleaseTest.java
pkriens: you're right. I'd missed that Will fix it

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
pkriens: No, otherwise this fails for everything

laeubi: Then calling this once in `aQute.bnd.osgi.Processor.getProperties0()` might be more appropriate...

Issue #5570, opened 2023-02-26
NPE in ActivelyClosingClassLoader if the bundle is not started


laeubi:
`ActivelyClosingClassLoader` access the bundle context here

https://github.com/bndtools/bnd/blob/8ab7b919cc0c4eee0a7d6f05b06327b5e81f74d7/biz.aQute.bndlib/src/aQute/bnd/osgi/ActivelyClosingClassLoader.java#L219-L223

As the bundle `biz.aQute.bndlib` is not marked as lazy-start it is possible that the bundle is in `RESOLVED` state and then BundleCOntext is null and a NPE is thrown.

I see the following options:

- Mark the bundle as lazy-start
- check for null and throw a meaningful exception
- just doing nothing and do not try to load from the system bundle

kriegfrj:
Thanks Christoph!

Quick question: Is this a theoretical problem, or has it actually happened to you? If you've seen it manifest, do you have a stack trace you can add to this issue? The reason I ask is because if it's actually happened there may be other contributing factors we might need to look at to see if they need attention too.

> `ActivelyClosingClassLoader` access the bundle context here
> 
> https://github.com/bndtools/bnd/blob/8ab7b919cc0c4eee0a7d6f05b06327b5e81f74d7/biz.aQute.bndlib/src/aQute/bnd/osgi/ActivelyClosingClassLoader.java#L219-L223
> 
> As the bundle `biz.aQute.bndlib` is not marked as lazy-start it is possible that the bundle is in `RESOLVED` state and then BundleCOntext is null and a NPE is thrown.
>

I admit I'm not a bundle lifecycle guru, but my understanding is that if this code is running, and it's not a "lazy start" bundle, then shouldn't the bundle have already been moved into the "ACTIVE" state before the code started executing?
 
> I see the following options:
> 
> * Mark the bundle as lazy-start
> * check for null and throw a meaningful exception
> * just doing nothing and do not try to load from the system bundle

I believe that there is a philosophical objection to lazy-start here. Admittedly my experience with it has been limited, but that limited experience is painful.

The third would change behaviour in a way that is not backward compatible. While I'm not against breaking backward compatibility if there is a good reason, I don't think that this is a sufficiently good reason in this case as there is a simple and viable alternative (option 2).

[...]

PR #5580, opened 2023-03-04
Avoid NullPointerException in case the bundle is not started

laeubi:
Fix https://github.com/bndtools/bnd/issues/5570

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/ActivelyClosingClassLoader.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/ActivelyClosingClassLoader.java
@@ -18,6 +18,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -25,6 +26,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Stream;
 
 import org.osgi.framework.Bundle;
+import org.osgi.framework.BundleContext;
 import org.osgi.framework.FrameworkUtil;
 
 import aQute.bnd.exceptions.Exceptions;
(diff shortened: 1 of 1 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/ActivelyClosingClassLoader.java
```diff
@@ -212,15 +214,18 @@ public Class<?> loadClass(String name) throws ClassNotFoundException {
...
-				}
+				Optional<Class<?>> bundleClass = tryLoadFromBundle(name, bundle, nfe).or(() -> {
+					BundleContext bundleContext = bundle.getBundleContext();
+					if (bundleContext == null) {
+						// not an active bundle, nothing more we can do here...
+						return Optional.empty();
+					}
+					return tryLoadFromBundle(name, bundleContext.getBundle(0), nfe);
+				});
+				return bundleClass.orElseThrow(() -> nfe);
```
pkriens, line 228: I think this is a lot more readable:

```
			try {
				return super.loadClass(name);
			} catch (ClassNotFoundException nfe) {
				Bundle bundle = FrameworkUtil.getBundle(ActivelyClosingClassLoader.class);
				BundleContext c;
				if (bundle == null || (c = bundle.getBundleContext()) == null)
					throw nfe;

				try {
					return bundle.loadClass(name);
				} catch (ClassNotFoundException anotherNfe) {
					Bundle system = c.getBundle(0);
					return system.loadClass(name);
				}
			}
```

In general the rule we have about commenting in bnd is not to do it, the code should speak. So if you want to comment something, try making a temporary variable that elucidates the code. It never ceases to amaze me how many of the comments were not valid anymore after 20 years. Also, private methods should use private comments (/* nor /**) or no comments. Again, the the goal is that the code is clear enough.

Although in general I think it is really good to abstract behavior, in this case, the `tryLoadFromBundle` seems to introduce just more complexity and it does not seem useful outside this class. Notice that if you have a BundleContext, you can always get bundle 0 (unless the framework is starting down but that would be the least of your problems then.)

pkriens (changes requested):
Thanks. I'd like a bit simpler solution since this is such a small problem.

laeubi:
The code presented here is much less error prone than cascading try/catches and clearly outlines the delegation model used here. Beside that is can be easily enhanced e.g. if one likes to extend this to load from other bundles (not just the system bundle) and if there is still an issue it only needs to be fixed in one branch and not in multiple branches.

You can easily see this as in our presented alternative still has an issue that if loading from the system bundle fails, that exception is thrown instead of the original CNF exception like done in the other branch, beside that the current solution hides suppressed exceptions.

So this is what I can offer to _improve BND_ the other one would just be another band-aid (for me) but of course you are free to apply whatever you thing do suffice.

PR #5581, opened 2023-03-05
Add an example to verify maven multi-release jar build with maven

laeubi:
It compiles a class HttpClient in three flavors:
- default version (Java 8) that uses a library to perform some task not part of JDK 8
- version 9 (Java 9) that uses the now build in support added in JDK 9
- version 11 (Java 11) that uses the new Java HttpClient added in JDK 11

This emulates what one often find in external libs where OSGi headers has to be added as part of a maven build.

----

@pkriens (FYI @stbischof) Here we go with an MR-Maven-Build that is a common setup (e.g. on the [clickhouse-jdbc](https://github.com/ClickHouse/clickhouse-java) driver), currently it shows the following issues:

1. BND generates a module-info.class (but only in the versioned folders) even though JPMS is not enabled in the BND instructions
2. Java 9 + Java 11 Manifests import the package `org.apache.commons.io` even though in that version(s) the package is never used.

### DIFF
--- a/maven-plugins/bnd-maven-plugin/src/it/mr-jar/README.txt
+++ b/maven-plugins/bnd-maven-plugin/src/it/mr-jar/README.txt
@@ -1,0 +1,8 @@
+This is an example Multi-Release-Jar bundle.
+
+It compiles a class HttpClient in three flavors:
+- default version (Java 8) that uses a library to perform some task not part of JDK 8
+- version 9 (Java 9) that uses the now build in support added in JDK 9
+- version 11 (Java 11) that uses the new Java HttpClient added in JDK 11
+
+This emulates what one often find in external libs where OSGi headers has to be added as part of a maven build.
--- a/maven-plugins/bnd-maven-plugin/src/it/mr-jar/invoker.properties
+++ b/maven-plugins/bnd-maven-plugin/src/it/mr-jar/invoker.properties
@@ -1,0 +1,7 @@
+invoker.goals=--no-transfer-progress package
+
+# Run mvn with --debug for debug logging
+#invoker.debug=true
+
+# Run mvn in debugging mode and wait for a debugger to attach
+#invoker.environmentVariables.MAVEN_DEBUG_OPTS=-agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=8000
--- a/maven-plugins/bnd-maven-plugin/src/it/mr-jar/pom.xml
+++ b/maven-plugins/bnd-maven-plugin/src/it/mr-jar/pom.xml
@@ -1,0 +1,110 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+	<groupId>bnd.test</groupId>
+	<artifactId>mr</artifactId>
+	<version>0.0.1-SNAPSHOT</version>
+	<name>BND MR-JAR TestCase</name>
+	<description>The simplest form of a MR Jar Project that adds BND headers to an existing library project that produces MR-Content jar.</description>
+
+	<properties>
+		<minJDK>1.8</minJDK>
+		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+		<bnd.version>${bnd.version}</bnd.version>
+	</properties>
+
+	<dependencies>
+		<!-- required for java < 9 -->
+		<dependency>
+			<groupId>commons-io</groupId>
+			<artifactId>commons-io</artifactId>
+			<version>2.11.0</version>
+		</dependency>
+	</dependencies>
+	<build>
+		<plugins>
+			<!-- compile the sources -->
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<version>3.11.0</version>
+				<configuration>
+					<source>${minJDK}</source>
+					<target>${minJDK}</target>
+					<release>8</release>
+				</configuration>
+				<executions>
+					<execution>
+						<id>java9</id>
+						<goals>
+							<goal>compile</goal>
+						</goals>
+						<configuration>
+							<release>9</release>
+							<jdkToolchain>
+								<version>9</version>
+							</jdkToolchain>
+							<compileSourceRoots>
+								<compileSourceRoot>${project.basedir}/src/main/java9</compileSourceRoot>
+							</compileSourceRoots>
+							<multiReleaseOutput>true</multiReleaseOutput>
+						</configuration>
+					</execution>
+					<execution>
+						<id>java11</id>
+						<goals>
+							<goal>compile</goal>
+						</goals>
+						<configuration>
+							<release>11</release>
+							<jdkToolchain>
+								<version>11</version>
+							</jdkToolchain>
+							<compileSourceRoots>
+								<compileSourceRoot>${project.basedir}/src/main/java11</compileSourceRoot>
+							</compileSourceRoots>
+							<multiReleaseOutput>true</multiReleaseOutput>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+			<!-- generate the header -->
+			<plugin>
+				<groupId>biz.aQute.bnd</groupId>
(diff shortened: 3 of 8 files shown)


pkriens:
In my mind the requirements would always be additional and the OSGi spec is moot about this. I also wonder how practical it is in the real world. The analyzer will pick up implementation classes from lower versions that have their own unique names. In practice this will likely leak any dependencies. However, I can see that in this example it makes sense to start the analysis from scratch. That will make it slower since I can't use the delta analysis. Ah well, that is life.

For the module stuff, I assumed it was harmless since an existing module file would not be overridden. However, it is  a bit cleaner to follow the instruction. Although I've not looked if this instruction will require it to be version specific. 

We still have the problem how to handle this on the repository level :-( What a mess.

timothyjward:
>I don't think this is any real issue on the repository level, as MR jars are not allowed to add new capabilities that are visible to the user. It might be an issue for the Resolver Spec though as depending on the EE of the framework one might want to take additional requirements (that can change) into account but the spec currently do not say anything about it and that might then be better brought up for the spec first.

This isn't totally true. The repository cares about requirements as well as capabilities. We will either need to:

a) Create one resource per MR folder, where the osgi.ee requirement has a maximum as well as a minimum if needed
or
b) Update the repository specification XML format to support a MR resource in a better way

Option a) is simpler for tools as it doesn't involve teaching everyone a new format. It does, however, have two issues:

1. Multiple resources will have the same identity and content capabilities - this may make the solution unworkable as those are used to identify the resources...
2. It will explode the resolve complexity as there will suddenly be multiple candidates for every requirement satisfied by a MR bundle. Resolve operations may become unworkably slow.

The best way I can see to help with these two issues is for the runtime part of the repository specification to change as follows:
1. Add a directive (probably better than an attribute?) to the identity capability flagging the MR version. This will help to avoid identity clashes.
2. Add some sort of `atEEVersion` method to the `Repository` which flattens the repository view into the one that would apply for a given ee.

laeubi:
@timothyjward you are right but nothing prevents you from providing two (or more) resources for the same bundle and I must confess i really like the idea because it does not require any change to the spec, it will only has the issue that the osgi.id might not be unique anymore but even that can be handled by appending e.g. the release number to the resource.

On the other hand, a repository is running in a JVM and it would be legal for an implementation to present only that item that matches the running release level.

The xml is another story as it is only *one* possible representation of a serialized repository, lets say a http server is involved one might uses `https://my.server.test/repository.xml?release=14` would work.

Neverless I think this is something best discussed at the osgi-specification and is not directly related to producing a valid MR-OSGi-Bundle.

pkriens:
@timothyjward I think we can solve it as follows:

We encode the primary jar without, the OSGi defined versioned headers: imports & requirements. Then we add a requirement for a new capability, let's call it `mrj`, with a `mrj` and `version` attribute. 

* `mrj` – the bsn of the primary jar
* `version` – the version of the primary jar

Then we create a synthetic resource for the base and each release version. This resource provides the `mrj` capability but that has the requirements ( is also imports) of the release version. Additionally, it requires the EE in such a way that only one of the synthetic resources can be resolved for any given EE range. I.e. the EE requirement must exclude the ranges for which the other synthetic resources match.

image

So how do make a synthetic resource?

* A resource from a repository _must_ have an `osgi.content` capability ... If we would allow resources to not have a `osgi.content` capability by the spec then we can infer that we do not have to download it.
* We could specify a special identity type. For now we could use UNKNOWN. This would be easy to check and then not install the synthetic resource in the framework (or add to -runbundles). 
* We could use the `data:` scheme for the URL and provide a tiny bundle but that could take up a lot of space if MRJs become popular. Also, for some asinine reason the data url is not supported out of the box on Java 17?
* We could give the synthetic resources the same URL + a unique fragment. However, then there is a big chance we download the same URL multiple times.

I guess the type is the best solution and should work out of the box or with little changes?

That said, the hardest part will be to change the cardinality, we now go from 1 file == 1 resource to n resources :-( We use ResourceBuilder all over the place and its `build()` method should now return multiple resources. I guess we need to deprecate this method and add `builds()` to return a list of resources.

I would see adding the concept of MRJ's to the spec as you propose as a failure. We pay a pretty steep price for the complexity of the resources model. Although the MRJ spec is imho asinine (a small benefit, for surprising complexity and wide open for error cases and I expect little benefit in practical use), it was the clear intention of the specification to model that these asinine cases (think fragments!) inside the model and not have to extend it for every new use case. If the model can't handle this, then there are much simpler models that can also not handle this :-)

laeubi:
Can we please first fix the issues with _build_ a jar before solving other use-cases even though they are interesting and useful?

I think I already suggested it somewhere, but maybe one can enable Github discussions what is much more appropriate for such kind of collecting ideas and investigate different proposals.

pkriens:
@laeubi could you retest? I fixed the issues you raised.

laeubi:
The Java 11 versioned manifest looks fine, but the java 9 one looks like this

```
Manifest-Version: 1.0
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=9))"
```

while it must somehow replace the import from the default...

```
Manifest-Version: 1.0
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=9))"
Import-Package:  ??? is empty allowed here ???
```

mabye @tjwatson can answer this, the spec says that the header is only replaced if present, so I don't know how one could clear it out, maybe using something like

`Import-Package: empty;resolution:=optional`

or one would need a special syntx for supplemental manifests like

`Import-Package: -`

pkriens:
I agree that the spec text is a tiny bit ambiguous. As a long term spec writer, I expect (or hope) this was not the intention, for if the 'if present' is interpreted that it is _not_ replaced and the base value must be used, you open a rats nest of special cases, like for example how to represent no imports as we found out. Imho those special cases must be explicitly addressed.

The absolute minute advantage of saving a header value to reuse the base value if they happen to be the same (and the ZIP compression voids any advantage anyway) seems not to weigh up to keeping it simple.

The semantics for the base manifest is that no Import-Package means no import. I do not think we should deviate from this but the spec should clarify this case.

timothyjward:
> If there is a need to handle the special case of removing the Import-Package or Require-Package header entirely then it is best to open a specification issue.

Is it really possible to have a Multi-Release jar with no imports? I thought that the whole point was to have version specific implementation classes on Java 9+, and those are going to (at the very least) import the `java.lang` package. Given that any bundle with a base of Java 9 is supposed to import `java.*` packages now, it seems impossible to get yourself into a situation where this is a problem.

pkriens:
Could you try again? I now generate an empty Import-Package header. The RFC was a tiny minute little bit more clear but I think the spec should call out this case. It just makes it all more complicated for all involved and it has no purpose.

I also updated the Java imports to only depend on the fw version on the classpath.

If you this now works then we can close this issue

laeubi:
I'll check the next days, but actually the test-case should be merged to prevent future regressions...

pkriens:
I've rebased the branch but run in this error, seems to compile on Java 17?

```
assert default_manifest.getValue('Require-Capability').contains('osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"')
       |                |                              |
       |                |                              false
       |                'osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=17))"'
       [Manifest-Version:'1.0', Archiver-Version:'Plexus Archiver', Created-By:'19.0.2 (Eclipse Adoptium)', Built-By:'runner', Build-Jdk:'19.0.2', Bnd-LastModified:'1678892401561', Bundle-Description:'The simplest form of a MR Jar Project that adds BND headers to an existing library project that produces MR-Content jar.', Bundle-ManifestVersion:'2', Bundle-Name:'BND MR-JAR TestCase', Bundle-SymbolicName:'mr', Bundle-Version:'0.0.1.202303151500', Export-Package:'bnd.mr.example;version="0.0.1"', Import-Package:'java.io,java.lang,java.lang.invoke,java.net,org.apache.commons.io;version="[1.4,2)"', Multi-Release:'true', Require-Capability:'osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=17))"', Tool:'Bnd-7.0.0.202303151446-SNAPSHOT']

```

laeubi:
> I've rebased the branch but run in this error, seems to compile on Java 17?

Ill take a look, I''m just not very familiar with the groovy verification stuff, but the compile version should actually not matter here, maybe the target level is not correctly set as the compiler is even Java 19...

laeubi:
I rebased the branch and when I run the test project as a maven build with Java 17 it produces the desired results.

laeubi:
It seems it works now but some workflows have timeouts...

pkriens:
We can declare victory!

PR #5584, opened 2023-03-08
Add section about target-dir

chrisrueger:
* based on forum post https://bnd.discourse.group/t/stop-gradle-from-stepping-on-eclipses-long-toes/295


pkriens:
thanks! this is how open source should work :-)

PR #5612, opened 2023-03-27
Removed unnecessary usages of compiler options

### LINKED ISSUE #5578: build.bnd refers to javac.* as 1.8
the build.bnd file has a reference to 1.8 while we're compiling and building with 17.

@bjhargrave any reasons?

amitjoy:
closes #5578


pkriens:
That was easy!

I guess it gets the VM now from the local jdt files. Not sure I like this :-( However, this is fine. Need to think about it, I think it is nicer if we define this in the build.bnd centrally. But this would require the more bnd centered approach that I want to push more.

PR #5616, opened 2023-03-28
Indexing Multi release jars

pkriens:
This rather big PR updates the code to index the multi release jars.

Some major things to watch out for:

- EE API & handling changed for resolving. We listed the versions as separate capabilities instead of a list and we had redundant methods


pkriens:
# Multi Release Indexing

I've implemented multi-release-jar (MRJ) indexing in the way I outlined. The bundle has now a requirement on a osgi.multirelease capability and this is provided by a supporting resource for each supported VM release version.

This required support to expand a normal Resource to multiple. I introduced a SupportingResource for this reason. It extends Resource but it can provide an auxiliary resources. My experience with the ex-sun goes tells me they will come up with more of these brain haired ideas. Of course separating the concerns and making it more simple & reusable is beneficial as well,

I ran into an issue with the ResourceBuilder/resolver. I found out we were adding the EE as separate capabilities. This made it impossible (I think) to make a resource depend on a range of versions. I fixed that and found out we had 3 methods adding an EE so I consolidated them into one. Currently we add a capability for  every EE name once and the version is then a list. This is a mighty subtle change, and it matches what is in the spec and what Felix is doing, but it might be problematic, so be aware. None of the test cases detected this change though.

I only found 2 places where it was necessary to expand the SupportingResource but I feel a tad uncomfortable with that low number. Those places were the ResourcesRepository and the XMLResourceGenerator. If you come up for more places to look at, let me know.

There was another thing I had to change to make this work. In the resolve context, we were filtering resources that were not either a bundle or a fragment. I now leave the resources in there and do the filtering in the getRunBundles() method. Again, a subtle change. This also did not make any tests fail but if you extensively use the API then you might have more resources than you expected.

This is a pretty big change to some of the most fundamental parts of bnd so if you're a heavy user of the API you might want to check what gets changed.

timothyjward:
Hi Peter - can we hold off merging this until there has been a decision made in https://github.com/osgi/osgi/issues/574 ?

What you have implemented here is one of the possible options being debated, but as you have noted "This is a pretty big change to some of the most fundamental parts of bnd". I'd prefer that this PR didn't have to be unpicked, or that it gets left as a landmine to hurt us in the future, if the consensus is to solve this in a different way.

pkriens:
> Hi Peter - can we hold off merging this until there has been a decision made in https://github.com/osgi/osgi/issues/574 ?

Why would we need a decision? This can be implemented without a spec change? it would be completely against the idea of the cap/req model if we need to modify the spec every time we got a new artifact?

Puzzled ...

timothyjward:
> Why would we need a decision? This can be implemented without a spec change?

Pretty much anything can be implemented without a specification change, but if interoperability matters then we may need one anyway. If other tools are to create resources which bnd can understand, and vice versa, then we need to think about that impact. For example at a glance this change creates:
* [A new namespace](https://github.com/bndtools/bnd/blob/c38857dbcdf4b817bba9d3b99fae141d5f56d415/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/MultiReleaseNamespace.java) that may need to be standardised.
* A new resource type that tools need to be taught to deal with (for example [this commented out code](https://github.com/bndtools/bnd/blob/c38857dbcdf4b817bba9d3b99fae141d5f56d415/biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java#L498-L508). As far as I can tell the resource type is actually missing which [violates the specification of the Identity Namespace](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.namespaces.html#i1776750)

Whatever is done here it needs to work well so that resolve validation will work in things like the Feature Service, as well as in bnd. I fully support the aim to avoid bumping the resolver or repository specifications (and I'm pretty hopeful that we won't have to change either) but that doesn't mean there won't need to be some new things standardised to make sure that MR bundles work properly across all use cases, not just this one. We also want to be sure that whatever we do decide to do is minimally breaking to existing clients, at the very least the solution should comply with the existing specifications.

None of this is intended to say that I think this implementation is wrong in its approach. I just think that it is prejudging the outcome of the bug I linked to, and that it may be problematic if the bug is resolved differently.

pkriens:
If interoperability was in play, clearly we needed a spec. However, if you can achieve that interoperability without a spec than that is imho always superior. There is already more than enough text in the world :-(

- There is absolutely no need to standardize a namespace. A namespace defines semantics and in this case there are no semantics, I even considered a GUID. )Which I still think is kind of attractive since it reduces the complexity by voiding the need for a filter, name, and version, I skipped this idea because it is less readable in the XML.
- The filtering of non-OSGi resources was unnecessarily restricting the resolver scenarios. It restricts all kind of interesting scenarios so I wan that in anyway. It is only when they become bundles that they should be filtered.
- I fixed the missing Identity type. (I am sure it was there some time! :-) I am using 'synthetic', which will be treated as an unknown type. 

I am really surprised that this topic did no have more urgency in the enterprise world. Currently MRJs are causing havoc in the field, it creates deployments that **will** fail and it is really hard to work round. MRJs are becoming more common when this awful idea gains its inevitable popularity. So we need a solution _ASAP_, I am really amazed how long this has been lingering.

So I really appreciate anybody looking at the code and providing feedback but I do need this working in the very near future for my customers and I really need a solution to be part of bnd release 7. I will keep the PR open for feedback until after Friday's bnd meeting.

P.S. The only thing I am a tad uncomfortable about (my weak spot! :-) is the use of the synthetic Identity type. I did try to use actual bundles in two tries. First having the same content capability as the primary and another with a data URL that contained a dummy bundle. I strongly prefer the current solution.

timothyjward:
Hi Peter - here are the results of the discussion on today's call https://github.com/osgi/osgi/issues/574#issuecomment-1488898641

I think that it should not be too difficult to implement this, as it is extremely similar to what you have now. The optimisation in the resolve context to remove resources with an inappropriate ee requirement should also be possible for `BndrunResolveContext`.

pkriens:
This PR does not set a new standard. Yes, it gets written to XML but that XML is 100% valid OSGi. These files will be read and understood by anybody that supports the OSGi resource model, no other knowledge is required. Yes, it contains an unknown type but the whole idea of the resource model was to _not_ have to spec every new resource type!

That said, I am warming up to the idea of using plain bundles restricted by a vm range. We could indeed safely filter out the non-applicable resources without even knowing anything about MRJs at the resolves level, so the resolver only sees one resource. (I added an instruction that can do this now, we could set the default to do this.) This would prevent the overlapping capabilities problem that concerned I think. 

Only wonder if it is allowed to have resources with the same identity bsn & version? I think I saw a problem when the identity was the same.

So let's try this out in real world situations since the performance of the resolver is hard to predict or benchmark. Release 7 is still a few months away so we should commit this PR and then you add the OSGi proposed encoding. The encoding is handled in two relatively small methods in Resource Builder and you can leverage the cardinality change. Lets then measure the performance differences and request feedback. Before the first release candidate we can then decide what to do.

I therefore propose we commit this PR after the Friday meeting (or today if you're OK). We then have time until the release 7 cycle to make a final decision.

timothyjward:
> Only wonder if it is allowed to have resources with the same identity bsn & version? I think I saw a problem when the identity was the same.

I've been fiddling with this for a while, and it turns out that the `Resource#equals()` definition is absolutely critical, both within bnd and the Felix resolver internals. Part of the resolution operation maintains a blacklist based on Resource identity, meaning that if one of the MR resources fails to resolve (which is guaranteed) then Felix will throw them all away. The only way to get past this is to return only the "correct" resource from the `ResolveContext`. Note that this issue is independent of the solution that we choose. If we generate more than one `Resource` that has the same content URL + SHA then those resources are `.equals()` and things start to go badly wrong.

I am fairly close to making the multi-resource scenario work, but the `equals` issue means that it's definitely not as elegant as I would like.

PR #5618, opened 2023-03-29
Do not overwrite source info for packages 

### LINKED ISSUE #5130: [builder] Access to information generated during doExpand()
#### Executive summary

In looking at #3175, I have done a deep-dive on the Bnd builder/analyzer and it seems that the Analyzer generates (nearly) all of the information that I need to make this work well. Unfortunately, it overwrites the necessary information before the caller can get its hands on it.

##### The problem

Suppose I have have two projects:

* `local-proj`, which has the package `my.local.ws.pkg` as a private package;
* `other.proj.one`, which has two source packages `some.priv.pack` and `some.exported.pack` and a `bnd.bnd` file like this:
```
-buildpath: local-proj
-privatepackage: some.priv.pack,my.local.ws.pkg
Export-Package: some.exported.pack;version=1.2.3
```

`Project.build()` ultimately calls `doExpand()`, which builds up the contents of `contained` so that it has the full list of packages that the Jar-under-construction will ultimately contain. After `doExpand()` is called for project `other.proj.one`, the `contained` field looks like this:

`some.priv.pack;-internal-source:="other.proj.one-0.0.0",my.local.ws.pkg;-internal-source:=local-proj-0,some.exported.pack;-internal-source:="other.proj.one-0.0.0"`

Thus the `-internal-source:` directive on each entry correctly identifies the project that each package came from. This is the information I need to correctly build up the correct set of transitive dependencies for #3175.

However, after completing the expand phase, `Project.build()` then calls `calcManifest()`, which in turn calls `analyze()` -> `analyzeBundleClasspath()` -> `learnPackage()`. `dot` is passed as the `Jar` parameter to `learnPackage()`, so when this code runs:

https://github.com/bndtools/bnd/blob/dbf5fb4ddbb5ca0d2cc3292bdff0433e4f781949/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L662-L706

...then the `-internal-source:` directive is overwritten with the result of `getName(jar)` - which evaluates to `dot-0.0.0`. Thus the useful information that I was after is overwritten.

The first lines above seem to attempt to bypass the setting of `-internal-source:` if it is already set "in the manifest". I'm not sure how it could be triggered.

##### The solution

I'm unsure of the best solution, as I'm not sufficiently familiar with the intent of the code in `learnPackage()` and friends. Here are the possible solutions that I can see:

1. Modify `learnPackage()` so that it doesn't overwrite `-internal-source:` if it's already present. I'm just not sure what effect this has on code that comes after.
2. Add an "expand()" event to the `OnWorkspace` event bus that is called after the expand phase but before the analyze phase, which provides a snapshot of the artifact info before it is destroyed by `analyze()`.

Thoughts?

pkriens:
This is from #5130. This fix does not overwrite the internal source info for a package when it has already been set. 

Fixes #5130

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/BuilderTest.java
+++ b/biz.aQute.bndlib.tests/test/test/BuilderTest.java
@@ -54,6 +54,7 @@ import aQute.bnd.header.Parameters;
 import aQute.bnd.osgi.Analyzer;
 import aQute.bnd.osgi.Builder;
 import aQute.bnd.osgi.Constants;
+import aQute.bnd.osgi.Descriptors.PackageRef;
 import aQute.bnd.osgi.Domain;
 import aQute.bnd.osgi.EmbeddedResource;
 import aQute.bnd.osgi.Jar;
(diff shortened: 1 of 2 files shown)


pkriens:
Father Krieg, can you have a check?

pkriens:
@kriegfrj I think the change is innocuous enough to commit and the tests pass. So I am going to commit it to keep a nice small PR list :-) Please reopen when you want to have a discussion around it.

kriegfrj:
Thanks @pkriens , I'll see if I can revive the dependent PRs at some point soon. Blessings!

pkriens:
Yeah, that would be great. You were responsible of the most useful bndtool's features! Every time I can add dependency from the Java editor you pop up in my mind :-) So would be great to see you back ...

kriegfrj:
Well, every time that I use Bndtools I remember you and the rest of the team... the few features that I added wouldn't be much use without the foundation upon which it was built... :smile:

PR #5624, opened 2023-04-01
Skip synthetic resources in repos

pkriens:


pkriens:
@bndtools/maintainers 

Bit of a question. This PR removes the synthetic resources from the repository view. It turns out there are more MRJs than I thought.

image

However, I'm not sure I'd like to see them? I could aggregate this primary resource somehow? Ideas?

PR #5633, opened 2023-04-10
Initial commit for issue-5632.  Introduces service wizard support

scottslewis:
classes in org.bndtools.core.ui.wizards.service package and Java OSGi Service wizard in org.bndtools.core.ui.wizards.service.java package

### DIFF
--- a/bndtools.core/resources/unprocessed/serviceTemplates/java-service-api/bnd.bnd
+++ b/bndtools.core/resources/unprocessed/serviceTemplates/java-service-api/bnd.bnd
@@ -1,9 +1,3 @@
-$if(javaLevel)$
-javac.source=$javaLevel$
-javac.target=$javaLevel$
-javac.compliance=$javaLevel$
-$endif$
-
 -buildpath: \
 	osgi.annotation,\
 	org.osgi.annotation.bundle
--- a/bndtools.core/resources/unprocessed/serviceTemplates/java-service-api/bnd.bnd
+++ b/bndtools.core/resources/unprocessed/serviceTemplates/java-service-api/bnd.bnd
@@ -1,0 +1,11 @@
+$if(javaLevel)$
+javac.source=$javaLevel$
+javac.target=$javaLevel$
+javac.compliance=$javaLevel$
+$endif$
+
+-buildpath: \
+	osgi.annotation,\
+	org.osgi.annotation.bundle
+	
+Bundle-Version: 1.0.0.\${tstamp}
--- a/bndtools.core/resources/unprocessed/serviceTemplates/java-service-api/bnd.bnd
+++ b/bndtools.core/resources/unprocessed/serviceTemplates/java-service-api/bnd.bnd
@@ -8,4 +8,3 @@ $endif$
 	osgi.annotation,\
 	org.osgi.annotation.bundle
 	
-Bundle-Version: 1.0.0.\${tstamp}
--- a/bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/$srcDir$/$basePackageDir$/$serviceName$Consumer.java
+++ b/bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/$srcDir$/$basePackageDir$/$serviceName$Consumer.java
@@ -1,0 +1,40 @@
+package $basePackageName$;
+
+import org.osgi.service.component.annotations.Component;
+import org.osgi.service.component.annotations.Reference;
+
+import $api_package$.$serviceName$Service;
+
+@Component(immediate=true)
+public class $serviceName$Consumer {
+	
+	private $serviceName$Service service;
+	
+	/**
+	 * The bind and unbind methods will be called by the service component runtime
+	 * when a $serviceName$Service reference is detected in the local service registry
+	 * @param s
+	 */
+	@Reference
+	void bind$serviceName$Service($serviceName$Service s) {
+		this.service = s;
+		System.out.println("$serviceName$Service instance bound to consumer");
+	}
+	
+	void unbind$serviceName$Service($serviceName$Service s) {
+		this.service = null;
+		System.out.println("$serviceName$Service instance unbound from consumer");
+	}
+	
+	/**
+	 * The activate method will be called immediately upon the $serviceName$Service 
+	 * reference being bound to this component via the bind method$serviceName$Service
+	 * method.
+	 */
+	void activate() {
+		// test method
+		String myName = "OSGi Service Consumer";
+		System.out.println(myName+" calling $serviceName$Service...");
+		System.out.println("$serviceName$Service responded: "+service.methodName(myName));
+	}
+}
--- a/bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/$srcDir$/$basePackageDir$/$serviceName$Consumer.java
+++ b/bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/$srcDir$/$basePackageDir$/$serviceName$Consumer.java
@@ -2,39 +2,21 @@ package $basePackageName$;
 
 import org.osgi.service.component.annotations.Component;
 import org.osgi.service.component.annotations.Reference;
+import org.osgi.service.component.annotations.Activate;
 
 import $api_package$.$serviceName$Service;
 
 @Component(immediate=true)
 public class $serviceName$Consumer {
 	
-	private $serviceName$Service service;
+	final $serviceName$Service service;
 	
-	/**
-	 * The bind and unbind methods will be called by the service component runtime
-	 * when a $serviceName$Service reference is detected in the local service registry
-	 * @param s
-	 */
-	@Reference
-	void bind$serviceName$Service($serviceName$Service s) {
-		this.service = s;
-		System.out.println("$serviceName$Service instance bound to consumer");
+	@Activate
+	public $serviceName$Consumer(@Reference $serviceName$Service service) {
+		this.service = service;
+		// test service method
+		System.out.println("$serviceName$Consumer calling $serviceName$Service");
+		System.out.println("$serviceName$Service response is "+service.methodName("$serviceName$Consumer"));
 	}
-	
-	void unbind$serviceName$Service($serviceName$Service s) {
-		this.service = null;
-		System.out.println("$serviceName$Service instance unbound from consumer");
-	}
-	
-	/**
-	 * The activate method will be called immediately upon the $serviceName$Service 
-	 * reference being bound to this component via the bind method$serviceName$Service
-	 * method.
-	 */
-	void activate() {
-		// test method
-		String myName = "OSGi Service Consumer";
-		System.out.println(myName+" calling $serviceName$Service...");
-		System.out.println("$serviceName$Service responded: "+service.methodName(myName));
-	}
+
 }
--- a/bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/bnd.bnd
+++ b/bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/bnd.bnd
@@ -3,9 +3,3 @@
 	org.osgi.service.component.annotations,\
 	$api_package$
 
-$if(javaLevel)$
-javac.source=$javaLevel$
-javac.target=$javaLevel$
-javac.compliance=$javaLevel$
-$endif$
-
--- a/bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/bnd.bnd
+++ b/bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/bnd.bnd
@@ -9,4 +9,3 @@ javac.target=$javaLevel$
 javac.compliance=$javaLevel$
 $endif$
 
-Bundle-Version: 1.0.0.\${tstamp}
--- a/bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/bnd.bnd
+++ b/bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/bnd.bnd
@@ -1,0 +1,12 @@
+-buildpath: \
+	osgi.annotation,\
+	org.osgi.service.component.annotations,\
+	$api_package$
+
+$if(javaLevel)$
+javac.source=$javaLevel$
+javac.target=$javaLevel$
+javac.compliance=$javaLevel$
+$endif$
+
+Bundle-Version: 1.0.0.\${tstamp}
--- a/bndtools.core/resources/unprocessed/serviceTemplates/java-service-impl/bnd.bnd
+++ b/bndtools.core/resources/unprocessed/serviceTemplates/java-service-impl/bnd.bnd
@@ -3,9 +3,3 @@
 	org.osgi.service.component.annotations,\
 	$api_package$
 
-$if(javaLevel)$
-javac.source=$javaLevel$
-javac.target=$javaLevel$
-javac.compliance=$javaLevel$
-$endif$
-
--- a/bndtools.core/resources/unprocessed/serviceTemplates/java-service-impl/bnd.bnd
+++ b/bndtools.core/resources/unprocessed/serviceTemplates/java-service-impl/bnd.bnd
@@ -9,4 +9,3 @@ javac.target=$javaLevel$
 javac.compliance=$javaLevel$
 $endif$
 
-Bundle-Version: 1.0.0.\${tstamp}
--- a/bndtools.core/resources/unprocessed/serviceTemplates/java-service-impl/bnd.bnd
+++ b/bndtools.core/resources/unprocessed/serviceTemplates/java-service-impl/bnd.bnd
@@ -1,0 +1,12 @@
+-buildpath: \
+	osgi.annotation,\
+	org.osgi.service.component.annotations,\
+	$api_package$
+
+$if(javaLevel)$
+javac.source=$javaLevel$
+javac.target=$javaLevel$
+javac.compliance=$javaLevel$
+$endif$
+
+Bundle-Version: 1.0.0.\${tstamp}
--- a/bndtools.core/src/org/bndtools/core/ui/wizards/service/NewBndServiceWizard.java
+++ b/bndtools.core/src/org/bndtools/core/ui/wizards/service/NewBndServiceWizard.java
@@ -1,0 +1,195 @@
+package org.bndtools.core.ui.wizards.service;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Map.Entry;
+
+import org.bndtools.core.ui.wizards.shared.ISkippingWizard;
+import org.bndtools.core.ui.wizards.shared.TemplateParamsWizardPage;
+import org.bndtools.templating.Template;
+import org.bndtools.utils.javaproject.JavaProjectUtils;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.SubMonitor;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.ui.wizards.NewJavaProjectWizardPageTwo;
+import org.eclipse.jface.wizard.IWizardPage;
+
+import bndtools.Plugin;
+import bndtools.wizards.project.ProjectTemplateParam;
+
+public class NewBndServiceWizard extends AbstractNewBndServiceMultiProjectWizard implements ISkippingWizard {
+
+	public static final String						DEFAULT_BUNDLE_VERSION	= "0.0.0.${tstamp}";	//$NON-NLS-1$
+
+	protected NewBndServiceWizardPageOne			pageOne;
+	protected TemplateParamsWizardPage				paramsPage;
+
+	protected NewBndServiceWizardNonApiProjectPage	implPage;
+	protected NewBndServiceWizardNonApiProjectPage	consumerPage;
+	protected String								templateName;
+
+	protected TemplateParamsWizardPage createParamsPage() {
+		return new TemplateParamsWizardPage(ProjectTemplateParam.valueStrings());
+	}
+
+	protected NewBndServiceWizardNonApiProjectPage createImplPage() {
+		return new NewBndServiceWizardNonApiProjectPage(
+			new NewBndServiceWizardPageOne(pageOne, ServiceTemplateConstants.DEFAULT_IMPL_SUFFIX, templateName));
+	}
+
+	protected NewBndServiceWizardNonApiProjectPage createConsumerPage() {
+		return new NewBndServiceWizardNonApiProjectPage(
+			new NewBndServiceWizardPageOne(pageOne, ServiceTemplateConstants.DEFAULT_CONSUMER_SUFFIX, templateName));
+	}
+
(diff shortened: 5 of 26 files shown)


bndtools.core/resources/unprocessed/serviceTemplates/java-service-api/bnd.bnd
```diff
@@ -0,0 +1,11 @@
...
+javac.source=$javaLevel$
+javac.target=$javaLevel$
+javac.compliance=$javaLevel$
+$endif$
+
+-buildpath: \
+	osgi.annotation,\
+	org.osgi.annotation.bundle
+	
+Bundle-Version: 1.0.0.\${tstamp}
```
pkriens, line 11: I do not think it is a good idea to set the Bundle-Version in the bnd file. I find that releasing the whole workspace with one version is better.

bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/$srcDir$/$basePackageDir$/$serviceName$Consumer.java
```diff
@@ -0,0 +1,40 @@
...
+		this.service = null;
+		System.out.println("$serviceName$Service instance unbound from consumer");
+	}
+	
+	/**
+	 * The activate method will be called immediately upon the $serviceName$Service 
+	 * reference being bound to this component via the bind method$serviceName$Service
+	 * method.
+	 */
+	void activate() {
```
pkriens, line 34: Hmm, you're using a default in DS which I never really liked. I also think using the constructor activation far superior since you can make the instance fields final. At least add an @Activate ann.

bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/bnd.bnd
```diff
@@ -0,0 +1,12 @@
+-buildpath: \
+	osgi.annotation,\
+	org.osgi.service.component.annotations,\
+	$api_package$
+
+$if(javaLevel)$
```
pkriens, line 6: hmm, I'd prefer not to see these details on project levels but let them be handled on workspace level.

bndtools.core/resources/unprocessed/serviceTemplates/java-service-impl/bnd.bnd
```diff
@@ -0,0 +1,12 @@
+-buildpath: \
+	osgi.annotation,\
+	org.osgi.service.component.annotations,\
+	$api_package$
+
+$if(javaLevel)$
```
pkriens, line 6: same, handle on workspace level

bndtools.core/src/org/bndtools/core/ui/wizards/service/NewBndServiceWizard.java
```diff
@@ -0,0 +1,195 @@
...
+		params.put(ProjectTemplateParam.VERSION.getString(), DEFAULT_BUNDLE_VERSION);
+		// Source Folders
+		IJavaProject javaProject = pageTwo.getJavaProject();
+		Map<String, String> sourceOutputLocations = JavaProjectUtils.getSourceOutputLocations(javaProject);
+		int nr = 1;
+		for (Map.Entry<String, String> entry : sourceOutputLocations.entrySet()) {
+			String src = entry.getKey();
+			String bin = entry.getValue();
+			if (nr == 1) {
+				params.put(ProjectTemplateParam.SRC_DIR.getString(), src);
```
pkriens, line 105: Are you sure to make these come from the workspace?

bndtools.core/src/org/bndtools/core/ui/wizards/service/NewBndServiceWizardPageOne.java
```diff
@@ -0,0 +1,365 @@
...
+	 * service-consumer)
+	 *
+	 * @param templateName the name of the template to look for. May be null if
+	 *            no template to be found/loaded.
+	 */
+	public NewBndServiceWizardPageOne(String templateName) {
+		this(null, null, templateName);
+	}
+
+	public NewBndServiceWizardPageOne(NewBndServiceWizardPageOne parentPageOne, String projectNameSuffix,
```
pkriens, line 70: I didn't look at the details. However, I think it is better to set the bnd model and then from the bndfile generate this information in one place. See WorkspaceSynchronizer.createProject() is where I tried to standardize this for bndtools. My goal has always been to split any functionality in 2 parts. One thing that can be done in bnd generically (which tends to be a lot easier to test and write) and then small mappings to the different drivers like Gradle, bnd command line, Maven, Eclipse, Intellij,, and VSCode whenever I get around. However, this has been an up hill battle :-(

pkriens (commented):
Thanks for doing this. There is a lot of new code in there so I cannot review it in detail but I noticed it is nicely in only few places. Some comments:

It looks fine as far as I can see. Only comments is the general setup stuff. I think we should keep that by default in the workspace build.bnd and the code to set the class path should be centralized, see WorkspaceSynchronizer

But I am ok to merge this if you think it is done. We are in the release 7 snapshot phase so I am fine to experiment, you don't have to make it perfect before we merge. It is a lot easier to get feedback when it is in the snapshot.

Which does mean that at the beginning of the release cycle we should review it of course.

pkriens:
One other issue. For this project you create an API project per service. I kind of really like to have only one API project that contains _all_ APIs. However, in there I create a bnd file for each separate API using the sub bundles feature. I generally also have a JAR with all the APIs. This gives me the advantage that anyone can depend on only one service API but it keeps the number of those tiny pesky API projects small, reducing the 'emotional cost' of creating an API. I use this at all my customers and it works extremely well. It also makes testing with launchpad easy since all the APIs can come from the class path.

Might be too hard but I think when you try it, you'd appreciate it.

bndtools.core/resources/unprocessed/serviceTemplates/java-service-api/bnd.bnd
scottslewis: I don't understand what you mean by '...general setup stuff.  I think we should keep that by default in the workspace build.bnd'.   Doesn't this make everything bound to that one workspace?   In other words...for Bundle-Version and other stuff...where can it go so that a specific workspace isn't required? (i.e. or in all workspaces)?

bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/$srcDir$/$basePackageDir$/$serviceName$Consumer.java
scottslewis: I just forgot the @activate annotation.  I'll add that.   I didn't want to pick the most elegant default...but rather one that wouldn't require much to understand it's behavior (e.g. explicit bind/unbind, possible generalizations like dynamics, etc).  I haven't honestly kept up on all the injection and annotation methods for most recent DS.   Could you point me to an example (from spec or elsewhere) of what you would prefer?

bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/bnd.bnd
scottslewis: Again I don't get what you mean workspace-level...when these service projects could be created in any workspace (?)

bndtools.core/resources/unprocessed/serviceTemplates/java-service-impl/bnd.bnd
scottslewis: Isn't the javaLevel set during project creation?   In that case shouldn't it be project level (as would be expected by wizard user)?

bndtools.core/src/org/bndtools/core/ui/wizards/service/NewBndServiceWizard.java
scottslewis: Same as above...i.e. which workspace?   Also...can't the second java page be used to set these for project creation (admittedlyI I'm not showing that page under typical usage, but it can be made to appear/set for created project(s).

bndtools.core/src/org/bndtools/core/ui/wizards/service/NewBndServiceWizardPageOne.java
scottslewis: I'm still confused.   Ultimately this new service wizard creates three projects.  Won't it be sort of strange for the user if they run this wizard multiple times and it resets workspace-level values each time?  I 'm still confused.

kriegfrj:
I think this is a fantastic initiative and I am really sorry that I do not have the time to have a look at this in detail.

One thing that I would suggest (if you have not done so already) is that you consider making use of the `ExtensionFacade`. This really helps with your development because you can make changes and have them hot-loaded into your Bndtools test Eclipse instance and test them right away, rather than having to restart your test Eclipse instance for every small change. This will help you and those coming after you.

For more details, have a look here: https://github.com/bndtools/bnd/blob/master/DEV_README.md#using-the-extensionfacade

scottslewis:
> I think this is a fantastic initiative and I am really sorry that I do not have the time to have a look at this in detail.
> 
> One thing that I would suggest (if you have not done so already) is that you consider making use of the `ExtensionFacade`. This really helps with your development because you can make changes and have them hot-loaded into your Bndtools test Eclipse instance and test them right away, rather than having to restart your test Eclipse instance for every small change. This will help you and those coming after you.
> 
> For more details, have a look here: https://github.com/bndtools/bnd/blob/master/DEV_README.md#using-the-extensionfacade

Thanks.   The extension facade looks very nice.  I sort of suspected it was in bndtools already, but now I know exactly where it is. 

FWIW, the only reason that I'm using the extension registry for *initial* java service wizard is that there's already quite a large set of ui programmer types who do Eclipse wizard UI creation...and I'm hoping to make it easy for them to reuse the classes in org.bndtools.core.ui.service.wizards (and the bndtools templating) to do additional service wizards themselves...with the extension registry if that's what they are used to...or with the extension facade if more comfortable with that.  I agree that it's very nice to have the dynamics of the extension facade (for testing/debugging especially), but the PDE tooling for creating/completing extensions (i.e. with forms-like ui) is pretty helpful for some devs out there.   And with these classes to use/reuse it's pretty easy to create a new service wizard simply.  So they have a choice.

kriegfrj:
> > I think this is a fantastic initiative and I am really sorry that I do not have the time to have a look at this in detail.
> > One thing that I would suggest (if you have not done so already) is that you consider making use of the `ExtensionFacade`. This really helps with your development because you can make changes and have them hot-loaded into your Bndtools test Eclipse instance and test them right away, rather than having to restart your test Eclipse instance for every small change. This will help you and those coming after you.
> > For more details, have a look here: https://github.com/bndtools/bnd/blob/master/DEV_README.md#using-the-extensionfacade
> 
> Thanks. The extension facade looks very nice. I sort of suspected it was in bndtools already, but now I know exactly where it is.
> 
> FWIW, the only reason that I'm using the extension registry for _initial_ java service wizard is that there's already quite a large set of ui programmer types who do Eclipse wizard UI creation...and I'm hoping to make it easy for them to reuse the classes in org.bndtools.core.ui.service.wizards (and the bndtools templating) to do additional service wizards themselves...with the extension registry if that's what they are used to...or with the extension facade if more comfortable with that. I agree that it's very nice to have the dynamics of the extension facade (for testing/debugging especially), but the PDE tooling for creating/completing extensions (i.e. with forms-like ui) is pretty helpful for some devs out there. And with these classes to use/reuse it's pretty easy to create a new service wizard simply. So they have a choice.

Sorry - yes, for code intended to be used outside of the Bndtools repository itself, that makes sense. I'm sorry I haven't read your code int detail enough to understand that level of finesse. :smile: At least now you are aware of it.

bndtools.core/resources/unprocessed/serviceTemplates/java-service-api/bnd.bnd
pkriens: Yes. The default mode for bndtools is that your workspace is a cohesive set of bundles. This deviates from the Eclipse model where your workspace is an unrelated set of things. That is why a workspace for us is in one directory.

Since the tooling tends to do a release version on the workspace, you're seriously screwed when you keep the versions in the bnd files since you cannot release the same version twice. So all the workspace I'm responsible for are keeping the bundle version in the build.bnd. Saves a lot of pain! :-) The cost that identical bundles appear with multiple versions is something will inevitably cause the world to heat up faster but somehow can't keep me awake at night :-)

A bnd workspace is a bit like the Reactor (?) model of maven. It is a bunch of models that are cohesive.

bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/$srcDir$/$basePackageDir$/$serviceName$Consumer.java
pkriens: `@Activate`

The coolest is constructor injection. It allows you to make your fields final, I've fallen heads over heels in love iwth it. Wonder how I could've lived without! :-)

     
     final $serviceName$Service service;
     @Activate
     public $serviceName$Consumer(@Reference $serviceName$Service service) {
        this.service = service;
     }

I did not remove the `private` by accident, in almost every test situation you want to be able to access it. I am a strong believer in package private. The advantage of private is that you keep your hands free in the future are also valid for package private. And it is less of an eye sore! :-)

bndtools.core/resources/unprocessed/serviceTemplates/java-service-impl/bnd.bnd
pkriens: We tend to set the Java level once in the workspace and not change it in the projects. This makes it a lot easier to manage, and since the projects have a shared release cycle it is not a big deal.

bndtools.core/src/org/bndtools/core/ui/wizards/service/NewBndServiceWizard.java
pkriens: Hmm, there is always a Workspace. You can get it from Central or as a Workspace service. There is only going to be one Workspace object in bndtools. If the workspace changes, we violate the existing singleton to reference the new directories. So only need to get it once. If you're really cool, you can get events when they get updated, see the `on(name)` method.

The `src`, `test`, `target`, etc. macros refer to the relative paths in a project where the working directories reside.

bndtools.core/src/org/bndtools/core/ui/wizards/service/NewBndServiceWizardPageOne.java
pkriens: No, the other way around ... you should _follow_ what is in the workspace, never ever set it. I hope the other explanations made it more clear now.

pkriens:
We are at least 1.5 - 2  months away from a release cycle (I am going hiking for 5 weeks!). As far as I can see this will not affect any core code or replace existing code? So lets get this in so everybody can play with it! I.e. it does not have to be perfect at the start. This stuff is really hard to see without trying it.

@kriegfrj since you're interested, can I ask you to help Scott merging when he is ready? I do not want to have him to wait 5 weeks ...

bndtools.core/resources/unprocessed/serviceTemplates/java-service-api/bnd.bnd
scottslewis: So just to be clear, are you saying that I should just remove the Bundle-Version header from the templates and expect that it will be provided by the cnf/build.bnd for all projects in the workspace?

bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/$srcDir$/$basePackageDir$/$serviceName$Consumer.java
scottslewis: For my information:   How does constructor injection work if the service is dynamic?

bndtools.core/resources/unprocessed/serviceTemplates/java-service-consumer/bnd.bnd
scottslewis: The reason I don't like this...with the current new project wizard and/or the new service projects wizard...is that on the first wizard page the user is asked what/which java level they want for the newly created project/projects.  I would seem to me to be shocking for the user if they were to set to something specific and then not have it set to what they specified for each project...or to have the workspace javalevel be reset by the creation of a new project (and break everything they would have there already).

bndtools.core/resources/unprocessed/serviceTemplates/java-service-impl/bnd.bnd
scottslewis: Yes, but the project creation wizards (even the bnd project creation wizards) explicitly have javalevel setting for that project on the first page (i.e. it's almost the first thing they see when creating a new project).   I think it would be terribly shocking to have that fail to work as expected for someone that's used to having it there for 100 years or so. :).   I suggest that if you don't want to surprise people that have been using the pre-bnd way wrt workspaces for a long time...remove the javaLevel setting from at least the bnd project wizards (you can't of course remove them from the Java project wizards).

bndtools.core/src/org/bndtools/core/ui/wizards/service/NewBndServiceWizard.java
scottslewis: Right...the relative paths *in a project* (not all projects currently in workspace).  I took this code from the Bndtools project wizards FWIW...it's not like I'm inventing a new approach for bndtools by doing things this way.   I'm just copying what's there in bndtools own project creation wizard....mostly so that it's visually and behaviorally familiar to as many users as possible.

As a proposed compromise, I'll remove bundle-version and leave javaLevel and paths...since these are set at the project level in the current project creation UIs.

bndtools.core/src/org/bndtools/core/ui/wizards/service/NewBndServiceWizardPageOne.java
scottslewis: Yeah I didn't think that it would make sense to set it at the workspace level as that work be terrible:  user says use javalevel 17 for this project in creation wizard and it changes things to Java-17 in the workspace...breaking all other projects that are there (and expect to stay at Java 16).

So how about this:   How can I query the bndtools workspace javaLevel?   If I can do that in wizard code, I'll leave out the javaLevel if the project wizard selected level is same as workspace.   If' it's not same, then I'll put in new project bnd.bnd.

Actually, if you will tell me how I can query the src and bin/target paths currently set in workspace I'll do that for these settings as well.   I think that will work well...if users are using your approach (workspace rules all) then I won't pollute the meta-data.  If they are not then then their wishes expressed through wizard UI won't be ignored.

scottslewis:
> 

Ok from the above discussion about workspace vs. project scope for Bundle-Version, javaLevel, and src, bin/target, test, src-gen paths:

1) I'm going to remove Bundle-Version from templates
2)  @pkriens if you will please point me to the programmatic API for the bnd Workspace current javaLevel, along with default, src, bin, test, src-gen/target, etc paths I will leave these blank in the project meta-data if user specifies same as what is currently set in workspace, and only override workspace explicitly if they choose something different from workspace.
3) I need to put copyright headers in all new classes and templates

I agree about getting things out there for usage.   This is all intended to help new bndtools/osgi service users...not any of the existing experts...so it needs usage to point the way forward.

Hiking for 5 weeks?  Are you hiking the [Pacific Crest Trail ](https://www.pcta.org/)...in my part of world?

scottslewis:
I've removed the javaLeval, Bundle-Version and simplified using constructor injection for the consumer.   Turns out I was already defaulting to the workspace-provided values so it was already doing the desired thing.  If the user picks the 'wrong' java level for a project, bndtools puts a little warning on project for them.

scottslewis:
@pkriens  and @kriegfrj:  I've made all the changes suggested and as far as I'm concerned this is ready for merge.  I've removed draft status and all checks are passing.

bndtools.core/resources/unprocessed/serviceTemplates/java-service-api/bnd.bnd
scottslewis: removed

pkriens:
I will give it a run! However, I've very little time left before my vacation. Sure hope @kriegfrj will put his critical eye on it :-)

chrisrueger:
@scottslewis thanks this is great stuff. Could you maybe post some screenshot of the current wizzard?

scottslewis:
> @scottslewis thanks this is great stuff. Could you maybe post some screenshot of the current wizzard?

Sure.  Here you go.  Select Java OSGi Wizard  from New...->Other.,,OSGi Services

wizard1

Then enter api project name (com.mycorp.services.helper) and service name:  Helper

wizard2

Choose Finish and it creates three projects

image

the api project, an impl project, and a consumer project.   API project has HelperService (java interface),  impl project has HelperImpl, and consumer project has HelperConsumer classes.   

For some reason I don't understand I couldn't get the min workspace template to compile these projects because it doesn't seem to have a repo for this annotation:

@org.osgi.annotation.bundle.Export

which is used in the package-info.java of the api project.   

I'm not sure why this would be (that the min workspace template wouldn't have this annotation class...but it doesn't.  Maybe I don't have latest of min workspace.

In any event, to create the above I use the ECF Remote Services workspace template, which is at:

https://github.com/ECF/bndtools.workspace

If any of you can/would like to fix the min workspace template so that it includes repos/bundles that allow the Export annotation to resolve I would appreciate it.   My intention with the service wizard is to use only the most minimal of annotation classes so dependencies of the generated code.  To that end it's only looking for Export and Version.  Finds Version fine from minimal workspace template, but doesn't find org.osgi.annotation.bundle.Export.

pkriens:
@scottslewis Great! But it makes me aware you can select the JRE. This should not be set imho, it should come from the Workspace as I explained. This is information that _must_ be in the build.bnd or bnd.bnd files or the gradle build will fail.

pkriens:
when I get back I'll take a look

scottslewis:
Hi.   I've made some small improvements and simplifications to the original set of classes and created new pull request #5643.  I would appreciate someone merging while Peter is away.

Also:   I have the following question:   I would like have a workspace template be able to add/remove extensions (e.g. new wizards).   I understand how to introduce project (and now service) templates as part of a workspace template, but I don't know of any way to add/remove items from the extension registry.  @kriegfrj is this something that the ExtensionFacade can help with?  Thanks.

kriegfrj:
> Also: I have the following question: I would like have a workspace template be able to add/remove extensions (e.g. new wizards). I understand how to introduce project (and now service) templates as part of a workspace template, but I don't know of any way to add/remove items from the extension registry. @kriegfrj is this something that the ExtensionFacade can help with? Thanks.

Just to clarify: you're talking about adding/removing extensions in the new workspace/project being created by Bnd using these templates?

ExtensionFacade at the moment is designed to be used inside the Bndtools project itself. The principle (and even the code) could be adapted to be used in other projects, but as yet it has not been.

To add/remove items from the extension registry in your new project wizard, you'd probably need to include a plugin.xml as one of the template files containing the necessary entries to register the extension.

scottslewis:
> To add/remove items from the extension registry in your new project wizard, you'd probably need to include a plugin.xml as one of the template files containing the necessary entries to register the extension.

Thanks.   I will give this a try.   I was under the impression that the bundle that has the templates would *not* have it's plugin.xml processed (i.e. any extensions loaded into the extension registry) as part of loading the templates, but perhaps it is.   Is there anything 'special' that has to be done to the templates bundle to make this happen (it's not an AbstractUIPlugin, but rather just a bundle)?

PR #5662, opened 2023-05-11
Updated Java OSGi Service wizard to use service.png for icon

scottslewis:
Simply uses bndtools service.png for the New OSG Service -> Java Service wizard


scottslewis:
@kriegfrj would you please approve this pr?   It just changes the icon that's used for the new OSGi Service Java wizard to the service.png icon that's already in bndtools.  Thanksinadvance.   BTW...the some of the builds are not succeeding...I don't think it has anything to do with the content of this change.

kriegfrj:
Thanks!

Issue #4258, opened 2020-08-10
[p2 repo] P2 Repo difficult to refresh


kriegfrj:
I was trying to refresh a P2Repository (so that the attributes would get indexed and I can take advantage of quick fixes and the like), using the "refresh repositories" button on the repositories view.

Unfortunately, it seems that P2Repository is being quite stubborn about it. I tried a number of approaches, including:

1. Hitting the refresh button
2. Deleting the index.
3. Deleting the index from disk and then hitting refresh
4. Deleting the entire p2 cache directory from disk and hitting refresh
5. Stopping Eclipse, deleting the index, restarting Eclipse.

It took the last one to actually get P2Repository to properly re-index and add the missing attributes.

I suspect that there are two issues here:

1. P2Repo keeps the index in-memory and the file on disk is effectively write-only after startup.
2. P2Repo is looking at the timestamp or checksum of the bundles that it downloads and it is not re-indexing them if they haven't changed. This makes sense as long as the indexing code itself hasn't changed.

On the downside, I found this behaviour quite counter-intuitive and it wasted a fair bit of my time. On the upside, it's not often that we change the indexing code so it's not often that people will run into this behaviour (though a corollary to this is another downside: because it doesn't happen often, people are less likely to become familiar enough with it to remember what to do when it happens).

I think there are a few options available to us to make this smoother for the end-user:
1. P2Repo should keep track of the last modified time of the index the last time that it read it from the disk or wrote it to the disk. During a refresh operation, it could check this timestamp time against the current last modified time to see if the index has been modified by an external source (or deleted). If so, it should discard its in-memory copy and re-read from disk (and re-generate from scratch if the index is missing).
2. Add a "Reload Index & Bundles" menu option like OSGiRepository which does a hard refresh.
3. Ideally, a mechanism for the index to know when it needs to re-index a bundle due to a change in indexing code.

kriegfrj:
Added link to PR that actually fixed this issue (#5683).

PR #5683, opened 2023-06-05
P2 Repositories lost synthetic resources for MRJs

pkriens:
See https://bnd.discourse.group/t/multirelease-jar-and-resolution/308/7 for the sob story

It is complicated.

MRJs are now build as a SupportingResource. This is the primary resource and some other resources you need to properly resolve. For an MRJ this is a synthetic resource that contains a capability unique for the MRJ that requires the EE and the requirements specific for the EE.

I.e. in some places we need to be careful that we expand a Resource. These places are few.

In the P2Indexer, we had an optimization for refresh. The existing Resource objects were reused if the MD5 of a request matched the existing resource. If the existing resource was used, then there were two situations. The existing index had been parsed from the network. In that case things worked fine since the SupportingResource was linked to the synthetic resources.

However, when the index was read from the cache (an XML file) the shit hit the fan. Generally no resources were fetched from the network so only the existing resources were used. However, these resources were read from the XML and had no link to the associated supporting resources. So after constructing a new index from the old resources, it wrote the new index to the XML, losing the supporting resources.

The fix is to remove the optimization scheme. This code was executed in a plugin refresh. Such a large potentially expensive actions should be more specific. So the refresh reinitializes the the plugin but will use the index if present. A menu was added to provide a proper refresh entry that will remove the index file and reread the repo.

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
@@ -2,15 +2,12 @@ package aQute.bnd.repository.p2.provider;
 
 import static aQute.bnd.osgi.repository.ResourcesRepository.toResourcesRepository;
 import static aQute.bnd.osgi.resource.ResourceUtils.toVersion;
-import static java.util.Collections.singleton;
-import static java.util.stream.Collectors.toMap;
 
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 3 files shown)


scottslewis (approved):
Appears reasonable to me.  For my info:   where does this new refresh menu item appear?  Is it on the context menu for the P2 repos only?

kriegfrj:
> The fix is to remove the optimization scheme. This code was executed in a plugin refresh. Such a large potentially expensive actions should be more specific. So the refresh reinitializes the the plugin but will use the index if present. A menu was added to provide a proper refresh entry that will remove the index file and reread the repo.

Does this mean that this PR also fixes #4258?

pkriens:
@scottslewis Can you verify

pkriens:
@kriegfrj the refresh is now manually through a menu entry on the repository entry in the Reposities View.

This will definitely refresh the repo since it throws away the cache. It now also refreshes the workspace and properly locks the workspace. At least, I try to. Please verify

scottslewis:
@pkriens yes I've run the fix in locally built bndtools workspace and it no longer has a problem with resolving with third-party MRJs with my use cases (e.g. jersey-common, jakarta.xml-bind, etc).

> @kriegfrj the refresh is now manually through a menu entry on the repository entry in the Reposities View.

I can also verify that this Refresh menu item is present on P2 Repos and by debug/console output seems to be refreshing the P2 repo to current remote state.

scottslewis:
Please see pr #5685 for suggested UI enhancement for P2Repository Refresh action

kriegfrj:
> @kriegfrj the refresh is now manually through a menu entry on the repository entry in the Reposities View.
> 
> This will definitely refresh the repo since it throws away the cache.

Excellent, I'll mark #4258 as resolved too then. I think that was exactly the same issue that @scottlewis had - Bnd caches the index and doesn't know that the indexing code itself has changed, so it doesn't know that its cache is out of date. #4258 happened to me when we introduced the class index bloom filter - the quick fixes for the P2 resources didn't work until I manually deleted the cached index. With this fix I could have deleted the cache and have it refresh without having to restart Eclipse and do it manually.

> It now also refreshes the workspace and properly locks the workspace. At least, I try to. Please verify

I'll try to verify in coming days.

Issue #5666, opened 2023-05-18
AbstractBndrun.getBundles() should use runtime classpath normalization


marcphilipp:
In order to benefit from [runtime classpath normalization](https://docs.gradle.org/current/userguide/incremental_build.html#sec:task_input_using_classpath_annotations) (e.g. to normalize timestamps in manifest attributes), `AbstractBndrun.getBundles()` should be annotated with `@Classpath` instead of `@PathSensitive(NONE)`.

https://github.com/bndtools/bnd/blob/c6104f28bcbfeaaebf0bd696403bd05e2f037d72/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/AbstractBndrun.java#L112-L116

Example from the JUnit build: https://ge.junit.org/c/akzt6zv7vrnwk/j7wczpxnvx22m/task-inputs?cacheability=cacheable&expanded=WyJzM2RraWlrenNzeXNzLWJ1bmRsZXMiXQ

Would it be possible to change that?

bjhargrave:
But this is not a classpath. It is a collection of bundle which are available to the bndrun task to resolve against or install as bundles. Classpath implies a ordered collection and this collection does have any ordering. So I don't think using `@Classpath` is correct here.

marcphilipp:
Point taken. We could annotate `getBundles()` with `@Internal` instead and define a derived `FileCollection` that applies a consistent ordering. Do you think that would be feasible/acceptable?

marcphilipp:
Related issue: https://github.com/gradle/gradle/issues/23153

bjhargrave:
> We could annotate `getBundles()` with `@Internal` instead and define a derived `FileCollection` that applies a consistent ordering.

In principle, yes. I am just not sure how to handle the sorting of the `getBundles()` file collection into the derived file collection as this would need to be done lazily. 

I assume the derived file collection would then be added to the task inputs via API calls with a ClasspathNormalizer to get the normalization you desire.

pkriens:
@marcphilipp Could you provide a PR now?

jprinet:
I'm on it as discussed with @marcphilipp ☝️

jprinet:
Please find the PR here:
https://github.com/bndtools/bnd/pull/5684

Please note that the ordering is probably not necessary as the collection order will respect the order in which the elements were added into the `bundles`. 
Adding the `sort` operation would only help for cases where adding to the collection would happen in an inconsistent order.
[...]

PR #5684, opened 2023-06-06
Apply Runtime classpath normalization and sort to the bundles input

jprinet:
## Issue
When some files added to the `bundles` input of the `AbstractBndrun` Gradle task are containing some data changing build over build (typical could be a _build time_ in a `Manifest` file for instance), a build cache miss occurs on the Gradle task.

## Reason
the `bundles` input is using [PathSensitivity.NONE](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/PathSensitivity.html#NONE) as file inputs normalization which prevents from using [Runtime normalization](https://docs.gradle.org/current/userguide/build_cache_concepts.html#runtime_classpath).

## Fix
Change the current normalization of the `bundles` property to [@Internal](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Internal.html) and add an extra property with the [@Classpath](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Classpath.html) normalization which allows Runtime normalization.

This extra property is sorted as `@Classpath` normalization takes order into consideration, although the `bundles` items are not necessarily added in a consistent order.

fixes https://github.com/bndtools/bnd/issues/5666

### DIFF
--- a/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/AbstractBndrun.java
+++ b/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/AbstractBndrun.java
@@ -118,11 +118,13 @@ public abstract class AbstractBndrun extends DefaultTask {
 
 	/**
 	 * Wrapper of the {@link #getBundles()} method to allow <code>@Classpath</code> normalization and sorting.
+	 * <p>
+	 * This method is only relevant for Gradle task input fingerprinting and should not be used.
 	 *
 	 * @return The sorted bundles.
 	 */
 	@Classpath
-	public Provider<List<File>> getBundlesSorted() {
+	Provider<List<File>> getBundlesSorted() {
 		return getBundles().getElements().map(
 			c -> c.stream()
 				.map(FileSystemLocation::getAsFile)


gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/AbstractBndrun.java
```diff
@@ -109,12 +112,26 @@ public RegularFileProperty getBndrun() {
...
 		return bundles;
 	}
 
+	/**
+	 * Wrapper of the {@link #getBundles()} method to allow <code>@Classpath</code> normalization and sorting.
+	 *
+	 * @return The sorted bundles.
+	 */
+	@Classpath
+	public Provider<List<File>> getBundlesSorted() {
```
bjhargrave, line 126: Can we reduce the accessibility from `public`? `private` would be preferable if Gradle will use this. This avoids this method becoming API.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/AbstractBndrun.java
```diff
@@ -109,12 +112,26 @@ public RegularFileProperty getBndrun() {
...
+		return getBundles().getElements().map(
+			c -> c.stream()
+				.map(FileSystemLocation::getAsFile)
+				.sorted(Comparator.comparing(File::getName))
```
bjhargrave, line 130: We need to use a better comparator. 

```suggestion
				.sorted(IO.fileComparator(File::getAbsolutePath))
```

This will provide a consistent reproducible ordering over the complete path. See https://github.com/bndtools/bnd/commit/3d5359e677b6317afe1347a8d4b75e8be6f8f13b.

bjhargrave (changes requested):
A couple of comments. Thanks for the PR!

pkriens:
@jprinet Don't forget to sign your commits. See the details of the DCO. Will you also make the method private?

jprinet:
> @jprinet Don't forget to sign your commits. See the details of the DCO. Will you also make the method private?

Just learnt that it was [not possible](https://scans.gradle.com/s/wzr6jrkbnnl7o)

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/AbstractBndrun.java
bjhargrave: If we cannot reduce the accessibility from public, then we should note in the javadoc that this method in not API and is only used by Gradle to access getBundles in a sorted manner for normalization purposes.

bjhargrave (changes requested):

jprinet:
I was able to make it work with a `package` visibility @bjhargrave . Let me know if you are ok with the updated Javadoc.

bjhargrave:
Thanks!

jprinet:
Great, do you have any ETA @bjhargrave for a new plugin release?

pkriens:
The next release is planned next month. I've on my list to start preparing the release candidates.

Issue #5689, opened 2023-06-09
Java 21-ea NPE issue


beikov:
It seems that the byte code parsing is affected by

* https://bugs.openjdk.org/browse/JDK-8271623
* https://bugs.openjdk.org/browse/JDK-8292275

I'm seeing the following error when running the Eclipse Transformer on Java 21:

```
java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:233)
	at aQute.bnd.classfile.ConstantPool.utf8Info(ConstantPool.java:710)
	at aQute.bnd.classfile.MethodParametersAttribute$MethodParameter.write(MethodParametersAttribute.java:78)
	at aQute.bnd.classfile.MethodParametersAttribute.write(MethodParametersAttribute.java:44)
	at aQute.bnd.classfile.Attribute.writeAttributes(Attribute.java:136)
	at aQute.bnd.classfile.MemberInfo.write(MemberInfo.java:55)
	at aQute.bnd.classfile.ClassFile.write(ClassFile.java:134)
	at org.eclipse.transformer.action.impl.ClassActionImpl.apply(ClassActionImpl.java:444)
	at org.eclipse.transformer.action.impl.ActionImpl.basicApply(ActionImpl.java:537)
	at org.eclipse.transformer.action.impl.ActionImpl.apply(ActionImpl.java:505)
	at org.eclipse.transformer.action.impl.ContainerActionImpl.apply(ContainerActionImpl.java:274)
	at org.eclipse.transformer.action.impl.ContainerActionImpl.apply(ContainerActionImpl.java:159)
	at org.eclipse.transformer.action.impl.ActionImpl.apply(ActionImpl.java:622)
	at org.eclipse.transformer.Transformer$TransformOptions.transform(Transformer.java:1781)
	at org.eclipse.transformer.Transformer.run(Transformer.java:1874)
	at org.eclipse.transformer.jakarta.JakartaTransformer.main(JakartaTransformer.java:30)
```

beikov:
Seems 6.4.1 is not affected by this anymore. Sorry for the noise. Should have updated first.

bjhargrave:
Can you please provide a small github repo which demonstrates the problem? Then we can investigate. Thanks.

beikov:
I don't know anything about your test infrastructure, but AFAIU the problem, you only need to compile a class like this:

```java
public class Outer {
  public class Inner { }
}
```

[...]

PR #5690, opened 2023-06-09
classfile: Properly handle unnamed MethodParameter

bjhargrave:
A method parameter may not have a name.

Fixes https://github.com/bndtools/bnd/issues/5689

### DIFF
--- a/biz.aQute.bnd.util/src/aQute/bnd/classfile/MethodParametersAttribute.java
+++ b/biz.aQute.bnd.util/src/aQute/bnd/classfile/MethodParametersAttribute.java
@@ -71,11 +71,12 @@ public class MethodParametersAttribute implements Attribute {
 		static MethodParameter read(DataInput in, ConstantPool constant_pool) throws IOException {
 			int name_index = in.readUnsignedShort();
 			int access_flags = in.readUnsignedShort();
-			return new MethodParameter(constant_pool.utf8(name_index), access_flags);
+			String name = (name_index == 0) ? null : constant_pool.utf8(name_index);
+			return new MethodParameter(name, access_flags);
 		}
 
 		void write(DataOutput out, ConstantPool constant_pool) throws IOException {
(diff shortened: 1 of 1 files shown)


bjhargrave:
@pkriens You may want to cherry-pick this back for a 6.4.2 release.

pkriens:
thanks for doing this man!

jhanders34:
> @pkriens You may want to cherry-pick this back for a 6.4.2 release.

@pkriens any more thoughts about putting this into a 6.4.2 release?

jhanders34:
@pkriens Any more thoughts on getting this put into a 6.4.2 release?

pkriens:
If I get a very good reason why 7.0.0 is not good enough?

6.4.1 is only intended for people that need runtime bundles for Java 1.8. Since this is in a library that is intended to use for development, I am not sure there is a need to have it available on Java 1.8?

jhanders34:
We can make it work with 7.0.0 when it is available.  Will still need to use the 6.x transform bundle when running Eclipse Transformer when testing Java 8 and 11.

pkriens:
There is a 7.1.0 snapshot and a 7.0.0.RC1 right now. All Java 17

Can't you run the 1.8 and 11 code on a Java 17 VM?

jhanders34:
We can run Java 1.8 and 11 code on a Java 17 VM, but our byte code is Java 8, 11, and 17 depending on the level of Java the feature requires.  So we test Java 8 and 11 VMs for our code still because Java 8 and 11 are still supported and aren't end of life yet.

pkriens:
Still not getting it, sorry. A release is a significant amount of work and I was hoping 6.4.1 classic would slowly die. I see this is BJ's project.

@bjhargrave if you say it is very important to have this library in Java 1.8, then I'll make another release of classic. I was thinking calling it like 180.2.0 to move it out of the 6.4 semantics, that was a mistake I realize now.

Issue #5699, opened 2023-06-16
Slow typing in "Add Bundle Requirement" dialog in Eclipse plugins


reckart:
The "Add Bundle Requirement" dialog accessible from the BND Runfile editor from the BND Eclipse plugins feels pretty slow.

In particular, when typing into the "filter bundle name" field, the input blocks until the filter has been applied to the list below. If there are a couple of hundred (I didn't count them, but probably in the upper 3-digit to lower/mid 4-digit count), this feels pretty sluggish.

Maybe the performance of the filtering method can be improved.

It feels like there is no debouncing of typing events at all. Adding a 200ms debounce would likely improve user experience.

A yet better solution though might be to make the filtering asynchronous to the typing into the field such that the user can continue typing while the filtering happens.

kriegfrj:
:+1: I agree about the less-than-stellar user experience. This sounds like a good way of fixing it.

kriegfrj:
Btw, any chance of you having a look at it and submitting a PR yourself, @reckart ?

reckart:
That probably depends on how annoyed I am with it in the future. I hope that I'll be "soon" through with the OSGI-ification of the codebase that I am in the middle of and may not be sufficiently slowed down by the issue as compared to digging into the code for that considering the turnaround times for Eclipse plugin development (at least considering my current level of familiarity with the process) and in particular with bnd (build process based on Gradle, having to install the plugin into Eclipse afterwards for testing, etc.)

I have been looking into other pretty blocking performance issues related to bnd in the past months - but this is probably not one I'll code on...

pkriens:
I am making a quick fix. It annoyed me as well and it turns out there is a FilteredTree that seems to simplify the existing code.

PR #5700, opened 2023-06-16
Slow bundle dialog

pkriens:
Typing in the filter box did a slow update. This PR uses a (copy of) the FilteredTree that is supposed to be much faster. Now also supports wildcards a la globbing

---

### DIFF
--- a/bndtools.core/src/bndtools/editor/project/AvailableBundlesPart.java
+++ b/bndtools.core/src/bndtools/editor/project/AvailableBundlesPart.java
@@ -4,8 +4,6 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.TimeUnit;
 
 import org.eclipse.jface.util.LocalSelectionTransfer;
 import org.eclipse.jface.viewers.IStructuredSelection;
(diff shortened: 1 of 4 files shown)


pkriens:
#5699

bndtools.core/src/bndtools/editor/project/AvailableBundlesPart.java
```diff
@@ -16,89 +14,65 @@
...
+																		public boolean select(Viewer viewer,
+																			Object parentElement, Object element) {
+																			boolean select = false;
+																			if (element instanceof RepositoryBundle) {
+																				RepositoryBundle repoBundle = (RepositoryBundle) element;
+																				RepositoryPlugin repo = repoBundle
+																					.getRepo();
+
+																				if (includedRepos == null) {
 																					select = true;
```
kriegfrj, line 60: Sorry to be painful but it's difficult to review when there's these unrelated whitespace-only changes in the commit. 😦

kriegfrj: (Thanks for taking it on!)

pkriens: I've no idea how it can happen. We automatically format on save and the formatting instructions have not changed. Somehow Eclipse has some randomness in it. I agree it is a pain.

kriegfrj: Well, look what I found: 
Github diff

I'll get to reviewing it then!

kriegfrj:
It looks good to me on the surface of it. I have tried it in my local build and it seems to run noticeably faster even on reasonably small repository bases. 👏 🥳  Ideally it would be good also for this to have some time in pre-release so that the community can put it through its paces before we release, as it touches a fairly core component (though there is always the text file workaround if the GUI fails for some reason).

I have not analyzed the code in detail (well, actually I started and then I read it properly and realised you'd lifted it from Eclipse, so I didn't bother reviewing further. :smile:). A few questions/comments: 
* What was your reasoning behind copying the code from Eclipse rather than referencing it directly? Is this component not available in our current baseline of Eclipse? If so, perhaps it is time to consider lifting the baseline, given that we've already gone from Java 8 to Java 17 in this release?
* If we're going to stick with a copy rather than a reference:
  * Can we document why this decision was made in the commit message and here?
  * Can we add/alter comments to/at the top of the copied files to make it clear at a glance that they are copies (including a reference to the FQN of the original class)?

pkriens:
I copied the code from Eclipse because the filtering cross all levels. It was not possible to only filter on bsn. I spent 2 hours or so because there are options that indicate what gets filtered/is visible etc but in the end I had to make a code change. The standard filtering was akward, it would hide all the versions and no way to open a bsn to see the versions. I tried sub classing but this dragged in 60% of the code.

It is not very big so I do not mind that much.

I can add a reference. I guess it was obvious :-)

kriegfrj:
> I can add a reference. I guess it was obvious :-)

I must be a bit slow or inattentive because I didn't notice at first and wasted a bunch of time reviewing what I thought was your code when it was actually Eclipse's code. :smile: Though it was obvious in hindsight...

pkriens:
My bad ... excuses.

kriegfrj:
> ... excuses.

Ditto for me... sorry, that was meant to be a jab at myself, not at you...

PR #5701, opened 2023-06-16
Use an ArtifactView to gather dependencies for BundleAction

### LINKED ISSUE #5695: Regression with Gradle 8.2-rc-2
After updating the JUnit build from Gradle 8.1 to 8.2-rc-2 (see [draft PR](https://github.com/junit-team/junit5/pull/3351)), the produced manifests no longer include version numbers for imported packages which causes our `verifyOSGi` `Resolve` tasks to fail ([Build Scan](https://ge.junit.org/s/4etbacry2q3fo/failure#1)).

Full diff:
```diff
--- MANIFEST-8.1	2023-06-12 09:41:32
+++ MANIFEST-8.2	2023-06-12 09:41:32
@@ -1,7 +1,7 @@
 Manifest-Version: 1.0
 Build-Date: 2023-06-12
-Build-Revision: e41a174f53d543ecfee683cf854e61067d7bf900 
-Build-Time: 09:39:13.752+0200
+Build-Revision: 34d235eadd8ab0f318541bce13b6b2ab71d40a28 
+Build-Time: 09:33:40.345+0200
 Built-By: JUnit Team
 Bundle-ManifestVersion: 2
 Bundle-Name: JUnit Platform Suite Commons
@@ -15,12 +15,10 @@
 Implementation-Vendor: junit.org
 Implementation-Version: 1.10.0-SNAPSHOT
 Import-Package: org.apiguardian.api;resolution:=optional;version="[1.1
- ,2)",org.junit.platform.commons.util;version="[1.10,2)",org.junit.pla
- tform.engine;version="[1.10,2)",org.junit.platform.engine.discovery;v
- ersion="[1.10,2)",org.junit.platform.launcher;version="[1.10,2)",org.
- junit.platform.launcher.core;version="[1.10,2)",org.junit.platform.su
- ite.api;version="[1.10,2)",org.junit.platform.commons.logging;status=
- INTERNAL;version="[1.10,2)"
+ ,2)",org.junit.platform.commons.util,org.junit.platform.engine,org.ju
+ nit.platform.engine.discovery,org.junit.platform.launcher,org.junit.p
+ latform.launcher.core,org.junit.platform.suite.api,org.junit.platform
+ .commons.logging;status=INTERNAL
 Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
 Specification-Title: junit-platform-suite-commons
 Specification-Vendor: junit.org
```

Do you have any ideas why that would be?

tresat:
This fixes #5695.

Using an `ArtifactView` avoids ordering issues with internal Gradle functionality when the `jar` task is realized prior to the `compileClasspath` configuration's attributes being modified just before resolution.  It makes the attributes used to resolve libraries for the `BundleAction` entirely predictable, and thus prevents a behavior change in Gradle 8.2 from affecting this process.

This also stops BND modifying attributes on Gradle-created configurations, which should be avoided.

If you revert the changes to `BundleTaskExtension` you can see the difference in behavior between Gradle version - this fix makes both behave consistently.  This test reproduces the relevant aspects of an issue identified in the JUnit 5 project here: https://github.com/junit-team/junit5/pull/3351#issuecomment-1591316685.

### DIFF
--- a/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
+++ b/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
@@ -25,6 +25,7 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.Set;
 import java.util.jar.Manifest;
+import java.util.stream.Collectors;
 import java.util.zip.ZipFile;
 
 import aQute.bnd.exceptions.Exceptions;
@@ -42,6 +43,9 @@ import org.gradle.api.Action;
 import org.gradle.api.GradleException;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
+import org.gradle.api.artifacts.ArtifactView;
+import org.gradle.api.artifacts.Configuration;
+import org.gradle.api.attributes.LibraryElements;
 import org.gradle.api.file.ConfigurableFileCollection;
 import org.gradle.api.file.DirectoryProperty;
 import org.gradle.api.file.FileCollection;
@@ -208,9 +212,13 @@ public class BundleTaskExtension {
 		classpath = objects.fileCollection();
 		allSource = objects.fileCollection();
 		outputDirectory = objects.directoryProperty();
+
 		SourceSet mainSourceSet = sourceSets(project).getByName(SourceSet.MAIN_SOURCE_SET_NAME);
 		setSourceSet(mainSourceSet);
-		classpath(mainSourceSet.getCompileClasspath());
+
+		Configuration mainCompileClasspath = task.getProject().getConfigurations().getByName(mainSourceSet.getCompileClasspathConfigurationName());
+		requestJarLibrary(mainCompileClasspath, objects);
+
 		properties = objects.mapProperty(String.class, Object.class)
 			.convention(Maps.of("project", "__convention__"));
 		defaultBundleSymbolicName = task.getArchiveBaseName()
@@ -240,6 +248,25 @@ public class BundleTaskExtension {
 			.property("default Bundle-SymbolicName", getDefaultBundleSymbolicName());
 		task.getInputs()
 			.property("default Bundle-Version", getDefaultBundleVersion());
+	}
+
+	/**
+	 * Adds an {@link ArtifactView} to {@link #classpath(Object...)} that will explicitly request variants of dependencies with the
+	 * {@link LibraryElements#JAR} library element set to {@link LibraryElements#JAR}.
+	 *
+	 * This method ensures that Gradle does not perform an optimization to local project dependencies and request the
+	 * {@link LibraryElements#CLASSES} value instead.  Doing so would add the directory containing the other project's compiled classes
+	 * to the classpath instead of the jar itself.  If the jar is not present, reading information from the {@code MANIFEST.MF}
+	 * file will not be possible, so we want to avoid Gradle applying this optimization.
+	 *
+	 * @param mainCompileClasspath the main source set of the project
+	 * @param objects the object factory of the project
+	 */
+	private void requestJarLibrary(Configuration mainCompileClasspath, ObjectFactory objects) {
+		ArtifactView jarView = mainCompileClasspath.getIncoming().artifactView(viewConfiguration -> {
+			viewConfiguration.attributes(attributeContainer -> attributeContainer.attribute(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE, objects.named(LibraryElements.class, LibraryElements.JAR)));
+		});
+		classpath(jarView.getFiles());
 	}
 
 	/**
@@ -335,8 +362,6 @@ public class BundleTaskExtension {
 			.getTasks()
 			.named(sourceSet.getCompileJavaTaskName(), AbstractCompile.class)
 			.flatMap(AbstractCompile::getDestinationDirectory));
-
-		jarLibraryElements(getTask(), sourceSet.getCompileClasspathConfigurationName());
 	}
 
 	ConfigurableFileCollection getAllSource() {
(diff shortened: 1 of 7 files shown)


bjhargrave:
I will review this week.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
```diff
@@ -242,6 +250,25 @@ public BundleTaskExtension(org.gradle.api.tasks.bundling.Jar task) {
...
+	 * file will not be possible, so we want to avoid Gradle applying this optimization.
+	 *
+	 * @param mainCompileClasspath the main source set of the project
+	 * @param objects the object factory of the project
+	 */
+	private void requestJarLibrary(Configuration mainCompileClasspath, ObjectFactory objects) {
+		ArtifactView jarView = mainCompileClasspath.getIncoming().artifactView(viewConfiguration -> {
+			viewConfiguration.attributes(attributeContainer -> attributeContainer.attribute(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE, objects.named(LibraryElements.class, LibraryElements.JAR)));
+		});
+		classpath(jarView.getFiles());
```
big-guy, line 269: A slightly different variation on this would be to return the FileCollection and put the `classpath(requestJarLibrary(...))` in the caller.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
```diff
@@ -335,8 +362,6 @@ public void setSourceSet(SourceSet sourceSet) {
 			.getTasks()
 			.named(sourceSet.getCompileJavaTaskName(), AbstractCompile.class)
 			.flatMap(AbstractCompile::getDestinationDirectory));
-
-		jarLibraryElements(getTask(), sourceSet.getCompileClasspathConfigurationName());
```
big-guy, line 339: I think we should be trying to get rid of `jarLibraryElements` entirely since it affects project-level resolution and can cause the same problem to happen elsewhere.

This is still being called from here (in two places):
https://github.com/bndtools/bnd/blob/6e8bbb257fe83539f6158628be828fef5c56dce3/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java#L258

I think this means there's a similar problem, but it's much less likely to be a real problem because we always configure the compile tasks.  With the ArtifactView being used for the classpath, I'm not sure if these calls are necessary now.

gradle-plugins/biz.aQute.bnd.gradle/src/test/groovy/aQute/bnd/gradle/TestBundlePlugin.groovy
```diff
@@ -409,4 +409,28 @@ class TestBundlePlugin extends Specification {
...
+		String testProject = "builderplugin5"
+		File testProjectDir = new File(testResources, testProject).canonicalFile
+		assert testProjectDir.isDirectory()
+
+		when:
+		def result = TestHelper.getGradleRunner()
+				.withProjectDir(testProjectDir)
+				.withArguments("--parallel", "--stacktrace", "resolve")
+				.withPluginClasspath()
+				.withDebug(true)
```
big-guy, line 424: This shouldn't be checked in with debug == true

gradle-plugins/biz.aQute.bnd.gradle/src/test/groovy/aQute/bnd/gradle/TestBundlePlugin.groovy
```diff
@@ -409,4 +409,28 @@ class TestBundlePlugin extends Specification {
...
+				.withGradleVersion(gradleVersion)
```
big-guy, line 425: I'm not sure if they'll want to run this with multiple versions in an ad-hoc way.  I think this was useful for diagnosing/fixing the original problem in the RC, but I don't think we should include it here.

gradle-plugins/biz.aQute.bnd.gradle/src/test/groovy/aQute/bnd/gradle/TestBundlePlugin.groovy
```diff
@@ -409,4 +409,28 @@ class TestBundlePlugin extends Specification {
...
+				.forwardOutput()
+				.build()
+
+		then:
+		result.task(":jar") == null // Jar tasks never run
+		result.task(":resolve").outcome == SUCCESS
+
+		where:
+		gradleVersion << ["8.1", "8.2-rc-2"]
```
big-guy, line 434: At the very least, I doubt they'll want to keep testing against a RC

big-guy (changes requested):

gradle-plugins/biz.aQute.bnd.gradle/src/test/groovy/aQute/bnd/gradle/TestBundlePlugin.groovy
tresat: I wanted to provide a test that would demonstrate the change in behavior between Gradle versions if the changes to `BundleTaskExtension` are reverted.  After the fix is accepted and the project updated to build using Gradle 8.2, this could be changed later to just test the latest Gradle version.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
tresat: Yep.  I'm not sure which would read clearer.  I'll leave it to the project maintainers to choose.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
tresat: I tried this and it seems to work.  I made a separate PR here: https://github.com/bndtools/bnd/pull/5702 in case there is another reason it should remain.

bjhargrave:
I commented in #5702 that the solution is incomplete since we need to also address the compile classpath for BndPlugin. I am not sure we should have 2 PRs covering the same topic.

bjhargrave:
Closing as this PR is superseded by #5702.

PR #5702, opened 2023-06-20
gradle: Use an ArtifactView for jar library elements selection

tresat:
Using an ArtifactView avoids ordering issues when the Jar task is
realized prior to the compileClasspath configuration being modified
just before resolution.  It makes the attributes used to resolve
libraries more predictable, and prevents an internal behavior change in
Gradle 8.2 from affecting this process.

This also avoids modifying attributes on Gradle-created configurations,
which should be avoided.

### DIFF
--- a/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java
+++ b/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java
@@ -255,7 +255,8 @@ public class BndPlugin implements Plugin<Project> {
 					AbstractCompile.class, t -> {
 						t.getDestinationDirectory()
 							.fileValue(destinationDir);
-						jarLibraryElements(t, sourceSet.getCompileClasspathConfigurationName());
+						FileCollection jarLibraryElements = jarLibraryElements(t, sourceSet.getCompileClasspathConfigurationName());
+						t.setClasspath(jarLibraryElements.plus(t.getClasspath()));
 					});
 				generateInputAction.ifPresent(compileTask::configure);
 				sourceSet.getOutput()
@@ -280,7 +281,8 @@ public class BndPlugin implements Plugin<Project> {
 					AbstractCompile.class, t -> {
 						t.getDestinationDirectory()
 							.fileValue(destinationDir);
-						jarLibraryElements(t, sourceSet.getCompileClasspathConfigurationName());
+						FileCollection jarLibraryElements = jarLibraryElements(t, sourceSet.getCompileClasspathConfigurationName());
+						t.setClasspath(jarLibraryElements.plus(t.getClasspath()));
 					});
 				sourceSet.getOutput()
 					.dir(Maps.of("builtBy", compileTask.getName()), destinationDir);
@@ -535,8 +537,9 @@ public class BndPlugin implements Plugin<Project> {
 						.withPropertyName("projectFolder");
 					/* bnd can include from -buildpath */
 					t.getInputs()
-						.files(sourceSets.getByName(SourceSet.MAIN_SOURCE_SET_NAME)
-							.getCompileClasspath())
+						.files(tasks.named(sourceSets.getByName(SourceSet.MAIN_SOURCE_SET_NAME)
+								.getCompileJavaTaskName(), AbstractCompile.class)
+							.map(AbstractCompile::getClasspath))
 						.withNormalizer(ClasspathNormalizer.class)
 						.withPropertyName("buildpath");
 					/* bnd can include from -dependson */
--- a/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
+++ b/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
@@ -210,7 +210,7 @@ public class BundleTaskExtension {
 		outputDirectory = objects.directoryProperty();
 		SourceSet mainSourceSet = sourceSets(project).getByName(SourceSet.MAIN_SOURCE_SET_NAME);
 		setSourceSet(mainSourceSet);
-		classpath(mainSourceSet.getCompileClasspath());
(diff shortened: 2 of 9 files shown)


gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java
```diff
@@ -252,11 +251,7 @@ public void execute(Task tt) {
 				sourceSet.getOutput()
 					.setResourcesDir(destinationDir);
 				TaskProvider<AbstractCompile> compileTask = tasks.named(sourceSet.getCompileJavaTaskName(),
-					AbstractCompile.class, t -> {
-						t.getDestinationDirectory()
-							.fileValue(destinationDir);
-						jarLibraryElements(t, sourceSet.getCompileClasspathConfigurationName());
```
bjhargrave, line 258: Removing this is wrong. It causes Bnd workspace builds to fail to see the jars on the classpath when building at is only sees the output folders. Bnd workspace builds (i.e. BndPlugin) need to be able to configure the compile classpath configuration to use jars and not output folders.  This is true for both compilation as well as bundle assembly. In Bnd, a project can build a jar which contains classes which are not built by the project's compilation step (using `-conditionalpackage`, `Export-Package` for example) and so when a bundle depends upon another project, the compile path must utilize the jar built by the other project. Otherwise compilation errors can occur due to types which are not visible in the project's output directory. 

So any solution also needs to address the actual compile classpath used during compilation and not just during jar assembly.

In non-Bnd workspace builds, (using BundleTaskExtension), Bnd is not responsible for managing the compile classpath. BundleTaskExtension only needs a classpath for jar assembly which is defaulted to the main compile classpath.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
```diff
@@ -208,9 +210,13 @@ public BundleTaskExtension(org.gradle.api.tasks.bundling.Jar task) {
 		classpath = objects.fileCollection();
 		allSource = objects.fileCollection();
 		outputDirectory = objects.directoryProperty();
+
 		SourceSet mainSourceSet = sourceSets(project).getByName(SourceSet.MAIN_SOURCE_SET_NAME);
 		setSourceSet(mainSourceSet);
-		classpath(mainSourceSet.getCompileClasspath());
+
+		Configuration mainCompileClasspath = task.getProject().getConfigurations().getByName(mainSourceSet.getCompileClasspathConfigurationName());
+		requestJarLibrary(mainCompileClasspath, objects);
```
bjhargrave, line 218: This can work here because the task extension maintains its own classpath file collection to enable to the user to configure it differently than the main compile classpath.

bjhargrave (changes requested):
The proposed solution does not properly handle the Bnd workspace model builds.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java
tresat: I've updated this PR with a possible fix for this which doesn't modify the existing Gradle configurations, but ensures the jar are on the compilation classpaths used by the tasks.

Do you have a test case where requesting only classes would break and jars are required to pass that I could test this against?  This passes `:gradle-plugins:build`.

Also, the instructions in the README suggest running `:build` and `:publish`, but these tasks don't exist in the root project.  This looks like an oversight.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java
```diff
@@ -902,6 +905,32 @@ public void execute(Task tt) {
...
+			try {
+				attributes.attribute(LibraryElements.LIBRARY_ELEMENTS_ATTRIBUTE, project.getObjects().named(LibraryElements.class, LibraryElements.JAR));
+				task.getLogger()
+						.info("Added libraries jars to to {}:{}", project.getPath(), compileClasspath.getName());
+			} catch (IllegalArgumentException e) {
+				task.getLogger()
+						.info("Unable to add libraries jars to to {}:{}", project.getPath(), compileClasspath.getName(), e);
+			}
+		});
+		task.setClasspath(task.getClasspath().plus(librariesView.getFiles()));
```
big-guy, line 931: We shouldn't this.  IIUC, the jar produced by projects using BndPlugin is different from just the class files, so the whole point is to only have the jar.

If that's the case, then we can just leave the method

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java
big-guy: IIUC, just the class files are no longer appropriate.  In that case, we should just move `jarLibraryElements` outside the block that configures the task so that it's always done regardless if the compile task is configured or not.  That would mean you can't use the task in `jarLibraryElements`. It looks like it's just for logging, so I think that's OK.

big-guy (changes requested):

bjhargrave:
I am working on improvements to this PR and will soon push them.

bjhargrave:
> I am working on improvements to this PR and will soon push them.

Slightly delayed as there were a number of issue related to updating the build to Gradle 8.1.1 which I first have to resolve.

bjhargrave:
I have updated this PR with the additional changes. I tested locally on 8.2-rc-2 successfully but the test case should not hard code the gradle version number.

bjhargrave (approved):

pkriens:
@bjhargrave reason you did not merge it?

bjhargrave:
> @bjhargrave reason you did not merge it?

Just allowing some time for feedback from @tresat and/or @big-guy. 

I will merge it later.

tresat:
Thanks for addressing this!

PR #5722, opened 2023-07-18
Allow to configure base via coordinates (and CLI argument)

### LINKED ISSUE #5721: bnd-baseline-maven-plugin: Allow to override "base" via command line
Currently [base](https://github.com/bndtools/bnd/tree/master/maven-plugins/bnd-baseline-maven-plugin#configuration-properties) only accepts a complex object which can only be configured in pom.xml and not overridden per command line parameter.
For certain edge cases (e.g. where there is an inofficial release in some corporate repository added as mirror) there should be the possibility to easily override the base version to compare to without touching the pom.xml. That parameter should accept Maven coordinates in the common format `<groupId>:<artifactId>:<version>`.

kwin:
This closes #5721


kwin:
Yes, it is ready

pkriens:
thanks

Issue #5726, opened 2023-07-19
how to change ~/.bnd to another folder?


AlBundy33:
I'm using the current bndtools-plugin from eclipse marketplace and want to use another folder than ~/.bnd
I want to use another drive because I'm on windows.

I already started eclipse with `-Duser.home=D:/test` but the directory is still created in `C:/Users/xyz/.bnd`

pkriens:
The calculation is taking place in the aQute.lib.io.IO class:
```
		File getHome() {
			File home = testFile(getSystemEnv("HOME"));
			if ((home == null) || !home.isDirectory()) {
				home = testFile(System.getProperty("user.home"));
			}
			assert home != null;
			return home;
		}
```
Now looking at it, I am not sure the home directory is consistently found, there seems to be other home calculators (bnd is >23 years old) but as far as I can see if they are not specifically activated, the path should be `~/.bnd` and that should result in using the above code.

Let me know which files placed in there that give you problems.

AlBundy33:
currently I have this files in ~/.bnd
![image](https://github.com/bndtools/bnd/assets/6294586/11a1d5a4-831c-437a-bdc3-14e1eb342be4)
but in windows `%HOME%` points of course to my actual home directory.

calling `SET HOME=` before starting eclipse (with `-Duser.home=D:/test`) seems to work becuase `D:/test/.bnd` was created with `urlcache`.
Not sure where default-ws comes from - this is currently missing in D:/test - but ~/.bnd doesnot exists as well.
So maybe I pressed to much buttons in my first try. 😄 

Maybe a new variable BND_HOME would be nice to change the path permanently.

pkriens:
You can set `bnd.home.dir` in the commandline:

```
public class Home {
	public static final String	USER_HOME_BND_DEFAULT		= "~/.bnd";
	public static final String	USER_HOME_BND_SYSTEM_PROP	= "bnd.home.dir";

	public static final String getUserHomeBnd() {
		return System.getProperty(USER_HOME_BND_SYSTEM_PROP, USER_HOME_BND_DEFAULT);
	}

}

**Comment by AlBundy33** [comment]
Thanks - I'll try that.
Advantage of a new environment variable like BND_HOME would be that this can be configured on system-level (~/.profile on linux or system/user settings on windows).

**Comment by AlBundy33** [comment]
Starting eclipse with something like
```
-Duser.home=C:/temp/home
-Dbnd.home.dir=C:/temp/home/.bnd_temp
```
[...]

PR #5727, opened 2023-07-21
Cleaned up the home directory and added a BND_HOME env var

pkriens:
---


pkriens:
Fixes https://github.com/bndtools/bnd/issues/5726

Issue #5568, opened 2023-02-25
bnd-testing-maven-plugin:7.0.0-SNAPSHOT:testing goal fails for certain tests


wborn:
I gave the 7.0.0-SNAPSHOT builds a test on openhab-core but it seems some tests now fail which work fine with 6.4.0.

It looks like there are some classloader issues:

```
[INFO] --- bnd-testing-maven-plugin:7.0.0-SNAPSHOT:testing (default) @ org.openhab.core.ephemeris.tests ---
[INFO] Matching glob *
[INFO] Testing /home/wouter/git/openhab/openhab-core/itests/org.openhab.core.ephemeris.tests/itest.bndrun:
org.openhab.core.ephemeris [org.openhab.core.ephemeris.internal.EphemerisManagerImpl] WARN : Erroneous day set definition dayset-weekend: No enum constant java.time.DayOfWeek.MONDAX
jollyday [de.jollyday.util.ClassLoadingUtil] WARN : Could not load class with current threads context classloader. Using default. Reason: ClassNotFoundException: de.jollyday.impl.DefaultHolidayManager cannot be found by biz.aQute.tester.junit-platform_7.0.0.202302132255-SNAPSHOT
jollyday [de.jollyday.util.ClassLoadingUtil] WARN : Could not load class with current threads context classloader. Using default. Reason: ClassNotFoundException: de.jollyday.datasource.impl.XmlFileDataSource cannot be found by biz.aQute.tester.junit-platform_7.0.0.202302132255-SNAPSHOT
jollyday [de.jollyday.util.XMLUtil] WARN : Could not create JAXB context using the current threads context classloader. Falling back to ObjectFactory class classloader.
jollyday [de.jollyday.util.ClassLoadingUtil] WARN : Could not load class with current threads context classloader. Using default. Reason: ClassNotFoundException: de.jollyday.parser.impl.ChristianHolidayParser cannot be found by biz.aQute.tester.junit-platform_7.0.0.202302132255-SNAPSHOT
jollyday [de.jollyday.util.ClassLoadingUtil] WARN : Could not load class with current threads context classloader. Using default. Reason: ClassNotFoundException: de.jollyday.parser.impl.FixedParser cannot be found by biz.aQute.tester.junit-platform_7.0.0.202302132255-SNAPSHOT
[...]

PR #5743, opened 2023-08-07
tester: Use CompositeClassLoader for TCCL when loading tests

bjhargrave:
We use CompositeClassLoader to composite the tester bundle's class
loader with the original TCCL when loading and executing tests. This
will provide visibility classes and resource from the original TCCL.

Fixes https://github.com/bndtools/bnd/issues/5568


kriegfrj:
Before we go down this path I want to revisit https://github.com/bndtools/bnd/commit/859636bd1d12f925174ab5c299c1633320d3d801 and see if it (or this) was really necessary, because I don't think that they should be. I've backed out that change in my local commit and I can see that it causes the tests to fail, but I think that these failures might be due to the isolation between the Eclipse launcher and the Launchpad system-under-test.

kriegfrj:
Referr #5744 for the follow-up, alternate solution that superseded this one.

PR #5765, opened 2023-09-05
Adds support for extra staging repository

### LINKED ISSUE #5761: Add attribute on Maven Release Repository to check release on another repository
When preparing a release for Maven Central, you usually have to go through a staging repo. This is empty by definition. A release to maven central for e.g. sonartype will not do the release. We already check if an artifact exits in the target repo, but for this case we would need to check a different repo (maven central). Thus an attribute is needed to set a different the repo to check against.

juergen-albert:
fixes #5761


pkriens:
Looks good to me. Seems pretty straightforward and I do not think there is any backward compatibility issue?

pkriens:
Can you merge today? I want to release RC1 tomorrow.

juergen-albert:
If you don't see any issues, I'm happy to merge this.

PR #5771, opened 2023-09-08
Removes Require-Capability for osgi.service AnyService

### LINKED ISSUE #5750: Use of AnyService seems causes osgi.service capability
With bnd 6.4.0 I always get osgi.service;filter:="(objectClass=org.osgi.service.component.AnyService)";effective:=active, which does not really made sense. I think there should be no service requirement for such a case.

I could not test it in a different WS and a newer bnd version yet.

juergen-albert:
fixes #5750

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/component/DSAnnotationTest.java
+++ b/biz.aQute.bndlib.tests/test/test/component/DSAnnotationTest.java
@@ -4395,7 +4395,6 @@ public class DSAnnotationTest {
 			assertOk(b);
 			Attributes a = getAttr(jar);
 			checkProvides(a);
-			checkRequires(a, ComponentConstants.COMPONENT_SPECIFICATION_VERSION, AnyService.class.getName());
 
 			//
 			// Test all the defaults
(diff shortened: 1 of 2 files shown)


biz.aQute.bndlib.tests/test/test/component/DSAnnotationTest.java
```diff
@@ -4444,6 +4443,25 @@ public void anyservice() throws Exception {
...
+
+			Jar jar = b.build();
+			assertOk(b);
+			Attributes a = getAttr(jar);
+			checkProvides(a);
+
+			jar.getManifest()
+				.write(System.out);
+
+			assertThat(a.getValue(Constants.REQUIRE_CAPABILITY)).doesNotContain("osgi.service");
```
bjhargrave, line 4461: This check it too broad. It expects no osgi.service requirement. Changes to the test component in the future could break this. Perhaps just ensuring the require capability header does not contain 'AnyService.class.getName()' is better.

```suggestion
			assertThat(a.getValue(Constants.REQUIRE_CAPABILITY)).doesNotContain(AnyService.class.getName());
```

bjhargrave (changes requested):

juergen-albert:
@bjhargrave It would be nice if you could have a look, as there was one test that explicitly checked that the AnyService Requirement was set.

biz.aQute.bndlib.tests/test/test/component/DSAnnotationTest.java
juergen-albert: Ahh, you are already on it. Thanks heaps.

PR #5774, opened 2023-09-13
Fix a few broken links

wborn:
This makes it easier to find the linked content.


pkriens:
wow, those are old files.

thanks

Issue #5783, opened 2023-09-26
BND tools will be broken on 2023-12


tjwatson:
If PR https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/810 remains then BND tools will be broken on 2023-12 as several bundles will not resolve:

```
g! ss -s INSTALLED
"Framework is launched."


id	State       Bundle
14	INSTALLED   bndtools.builder_6.4.0.202211291949
15	INSTALLED   bndtools.core_6.4.0.202211291949
16	INSTALLED   bndtools.core.services_6.4.0.202211291949
18	INSTALLED   bndtools.m2e_6.4.0.202211291949
19	INSTALLED   bndtools.release_6.4.0.202211291949
88	INSTALLED   org.bndtools.launch_6.4.0.202211291949
89	INSTALLED   org.bndtools.remoteinstall_6.4.0.202211291949
91	INSTALLED   org.bndtools.templating.gitrepo_6.4.0.202211291949
g! diag 15
bndtools.core [15]
  Unresolved requirement: Import-Package: org.eclipse.jdt.internal.corext.refactoring.rename; bundle-symbolic-name="org.eclipse.jdt.ui"; bundle-version="[3.21.0,4.0.0)"
```

The `org.eclipse.jdt.internal.corext.refactoring.rename` package is no longer exported by `org.eclipse.jdt.ui`

bjhargrave:
This should be fixed before 7.0.0 is final.

bjhargrave:
So I think the issue is that the `org.eclipse.jdt.internal.corext.refactoring.rename` package had moved from bundle `org.eclipse.jdt.ui` to `org.eclipse.jdt.core.manipulation`. Perhaps bundle `org.eclipse.jdt.ui` should be configures to reexport the `org.eclipse.jdt.internal.corext.refactoring.rename` package to avoid breaking other users?

tjwatson:
> So I think the issue is that the `org.eclipse.jdt.internal.corext.refactoring.rename` package had moved from bundle `org.eclipse.jdt.ui` to `org.eclipse.jdt.core.manipulation`. Perhaps bundle `org.eclipse.jdt.ui` should be configured to reexport the `org.eclipse.jdt.internal.corext.refactoring.rename` package to avoid breaking other users?

I would agree if this wasn't considered an internal package by JDT

bjhargrave:
[...]

PR #5784, opened 2023-09-26
Handle the internal refactoring packages moving bundles

tjwatson:
Fixes issue #5783

Since all `org.eclipse.jdt.internal.corext.refactoring` packages moved I updated the `eclipse.importpackage` rule to not add `bundle-symbolic-name` or `bundle-version` attributes for all of them.

This does result in many warnings about unused imports for `org.eclipse.jdt.internal.corext.refactoring.*` during the build.  Not sure if there is a better way to avoid that.  The other option is to place the rule in the `bnd.bnd` for only `bndtools.core` since that is the only bundle I see that actually uses anything from `org.eclipse.jdt.internal.corext.refactoring` but maybe that will change in the future?

### DIFF
--- a/cnf/includes/bndtools.bnd
+++ b/cnf/includes/bndtools.bnd
@@ -3,10 +3,22 @@ Bundle-Copyright: Copyright (c) Neil Bartlett (2009, ${tstamp;yyyy}) and others.
 Bundle-DocURL: https://bndtools.org/
 
 # Import-Package clauses for Eclipse packages
+#
 # We use bundle-symbolic-name and bundle-version because Eclipse is
 # terrible at managing packages and Require-Bundle is too promiscuous.
+#
+# For org.eclipse.jdt.internal.corext.refactoring we do not use
+# bundle-symbolic-name or bundle-version.  This package is not
+# considered API by the Eclipse project and therefore is free
+# to move to another bundle.
 eclipse.importpackage: \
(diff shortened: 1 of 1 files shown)


pkriens:
Thanks for doing this!

Can I merge this? @bjhargrave want to approve?

pkriens:
If this is merged, I will add it to rc3

bjhargrave:
> This does result in many warnings about unused imports for `org.eclipse.jdt.internal.corext.refactoring.*` during the build

You should add a `-fixupmessages.eclipserefactor` instruction to the bnd file to elide the warning. (See `biz.aQute.repository/bnd.bnd` for some examples. You cannot use a plain `-fixupmessages` since this file is included in many projects and could conflict with those project's bnd files.) 

Also, it would probably be good to add a comment in the bnd file to explain to the future why this package needs special treatment.

bjhargrave:
Oops, you missed your Signed-off-by on the second commit. Also, please squash into a single commit as that will make it easier for Peter to cherry-pick onto the release branch for 7.0.0.

tjwatson:
> Oops, you missed your Signed-off-by on the second commit. Also, please squash into a single commit as that will make it easier for Peter to cherry-pick onto the release branch for 7.0.0.

Done

bjhargrave (approved):

bjhargrave:
Peter, you should make sure this fix still works on the current Eclipse release before cherry-picking to the 7.0.0 release branch :-)

pkriens:
Ok, we can merge it here, I try it out, and will create an RC3 with this cherry picked?

Issue #5793, opened 2023-10-03
CapabilityComparator.compare violates contract


mnlipp:
When running the resolver (bnd 6.4.0), I sometimes get 
```
Exception during resolution.
java.lang.IllegalArgumentException: Comparison method violates its general contract!
	at java.base/java.util.TimSort.mergeHi(TimSort.java:903)
	at java.base/java.util.TimSort.mergeAt(TimSort.java:520)
	at java.base/java.util.TimSort.mergeForceCollapse(TimSort.java:461)
	at java.base/java.util.TimSort.sort(TimSort.java:254)
	at java.base/java.util.Arrays.sort(Arrays.java:1308)
	at java.base/java.util.stream.SortedOps$SizedRefSortingSink.end(SortedOps.java:353)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:510)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:682)
	at biz.aQute.resolve.AbstractResolveContext.findProvidersFromRepositories(AbstractResolveContext.java:286)
	at biz.aQute.resolve.AbstractResolveContext.lambda$0(AbstractResolveContext.java:244)
	at java.base/java.util.HashMap.computeIfAbsent(HashMap.java:1228)
	at biz.aQute.resolve.AbstractResolveContext.findProviders0(AbstractResolveContext.java:213)
	at biz.aQute.resolve.AbstractResolveContext.findProviders(AbstractResolveContext.java:153)
	at biz.aQute.resolve.ResolveProcess$1.findProviders(ResolveProcess.java:151)
	at org.apache.felix.resolver.Candidates.populate(Candidates.java:208)
	at org.apache.felix.resolver.ResolverImpl.getInitialCandidates(ResolverImpl.java:543)
	at org.apache.felix.resolver.ResolverImpl.doResolve(ResolverImpl.java:432)
	at org.apache.felix.resolver.ResolverImpl.resolve(ResolverImpl.java:421)
	at org.apache.felix.resolver.ResolverImpl.resolve(ResolverImpl.java:375)
[...]

Issue #4172, opened 2020-06-18
Resolver runs in circles till it eats all the memory


juergen-albert:
Taking the discussion from #4154 to a new Bug, so we can look at this independently. This may end up becoming a bug for the resolver, but we can do the analysis first. 

The current assumption is, that the resolver tears itself apart, when 
a. it has two many candidates for certain requirements
b. the Repositories have Artifacts with "unclean" Metadata (like Require Bundle)

An example of a failing resolve is: https://gitlab.com/gecko.io/geckoEMF-Tooling/-/tree/failing_resolve

Here the resolve dies for https://gitlab.com/gecko.io/geckoEMF-Tooling/-/blob/failing_resolve/org.gecko.emf.osgi.codegen/codegen.bndrun if we remove the runblacklist the resolve works just fine.

juergen-albert:
After comparing a few resolver logs I may have found some clues.

The resolver stumbles seems to run circles with requirements for the package org.osgi.util.function and org.osgi.util.promise because they are included and exported in a lot of bundles.

I've seen thousends of entries with: 

```
DEBUG: Candidate permutation failed due to a conflict between imports; will try another if possible. (Uses constraint violation. Unable to resolve resource org.apache.felix.scr [org.apache.felix.scr version=2.1.16.v20200110-1820] because it is exposed to package 'org.osgi.util.promise' from resources org.eclipse.osgi.util [org.eclipse.osgi.util version=3.5.300.v20190708-1141] and org.eclipse.osgi.util [org.eclipse.osgi.util version=3.5.300.v20190708-1141] via two dependency chains.

Chain 1:
  org.apache.felix.scr [org.apache.felix.scr version=2.1.16.v20200110-1820]
    import: (&(osgi.wiring.package=org.osgi.util.promise)(version>=1.0.0)(!(version>=2.0.0)))
     |
    export: osgi.wiring.package: org.osgi.util.promise
  org.eclipse.osgi.util [org.eclipse.osgi.util version=3.5.300.v20190708-1141]

Chain 2:
  org.apache.felix.scr [org.apache.felix.scr version=2.1.16.v20200110-1820]
    import: (&(osgi.wiring.package=org.osgi.service.component.runtime)(version>=1.4.0)(!(version>=1.5.0)))
     |
    export: osgi.wiring.package=org.osgi.service.component.runtime; uses:=org.osgi.util.promise
  org.apache.felix.scr [org.apache.felix.scr version=2.1.14]
[...]

PR #5807, opened 2023-10-12
Changed the comparators for Resource, Capability, and Requirement

pkriens:
---

### DIFF
--- a/aQute.libg/src/aQute/lib/comparators/Comparators.java
+++ b/aQute.libg/src/aQute/lib/comparators/Comparators.java
@@ -1,0 +1,187 @@
+package aQute.lib.comparators;
+
+import java.lang.reflect.Array;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Map;
+
+/**
+ * Some utilities to make comparisons easier to do, especially with null
+ * handling
+ */
+public interface Comparators {
+
+	/**
+	 * Compare two objects for null.
+	 * <ul>
+	 * <li>if they are both null, 0
+	 * <li>if a and !b -> 1
+	 * <li>if !a and b -> -1
+	 * <li>Otherwise Integer.MAXVALUE
+	 * </ul>
+	 * <p>
+	 * To use this, compare the result with <=1 and return if so.
+	 *
+	 * @param <T> the shared type
+	 * @param a
+	 * @param b
+	 * @return the result
+	 */
+
+	static <T> int compareNull(T a, T b) {
+
+		if (a != null) {
+			if (b != null)
+				return Integer.MAX_VALUE;
+			else
+				return 1;
+		} else if (b != null)
+			return -1;
+		else
+			return 0;
+	}
+
+	/**
+	 * Compare two objects for null and equality.
+	 * <ul>
+	 * <li>if they are both null, 0
+	 * <li>if a and !b -> 1
+	 * <li>if !a and b -> -1
+	 * <li>if a.equals(b) -> 0
+	 * <li>Otherwise Integer.MAXVALUE
+	 * </ul>
+	 * <p>
+	 * To use this, compare the result with <=1 and return if so.
+	 *
+	 * @param <T> the shared type
+	 * @param a
+	 * @param b
+	 * @return the result
+	 */
+
+	static <T> int compareNullEquals(T a, T b) {
+		int n = compareNull(a, b);
+		if (isDecided(n))
+			return n;
+		else
+			return a.equals(b) ? 0 : Integer.MAX_VALUE;
+	}
+
+	/**
+	 * Check if the result of compareNull is decided, i.e. -1 >= compare <= 1
+	 */
+
+	static boolean isDecided(int compare) {
+		return compare != Integer.MAX_VALUE;
+	}
+
+	/**
+	 * Compare two objects.
+	 * <ul>
+	 * <li>Test for null as in {@link #compareNull(Object, Object)}
+	 * <li>Test for same class, if not class names are compared
+	 * <li>Test if they are comparable, use the object comparison
+	 * <li>If array, compare the fields of the array with this method in order,
+	 * if all match, the longer is higher
+	 * <li>If Iterable, compare the members of the Iterable with this method in
+	 * order, longer is higher
+	 * <li>returns 0 in a desperate attempt. This can make objects return 0 that
+	 * are not equals.
+	 *
+	 * @param <T> the shared type of the objects
+	 * @param a
+	 * @param b
+	 * @return 0, 1 a>b, -1 a<b
+	 */
+	@SuppressWarnings({
+		"rawtypes", "unchecked"
+	})
+	static <T> int compare(T a, T b) {
+		int compare = compareNull(a, b);
+		if (isDecided(compare))
+			return compare;
+
+		assert a != null && b != null;
+
+		Class<? extends Object> ca = a.getClass();
+		Class<? extends Object> cb = b.getClass();
+		if (ca != cb) {
+			assert ca != null && cb != null;
+			return ca.getName()
+				.compareTo(cb.getName());
+		}
+
+		assert ca == cb;
+
+		if (Comparable.class.isAssignableFrom(ca)) {
+			Comparable<T> cmpa = (Comparable<T>) a;
+			return cmpa.compareTo(b);
+		}
+
+		if (ca.isArray()) {
+			int na = Array.getLength(a);
+			int nb = Array.getLength(b);
+			int l = Math.min(na, nb);
+			for (int n = 0; n < l; n++) {
+				Object va = Array.get(a, n);
+				Object vb = Array.get(b, n);
+				compare = compare(va, vb);
+				if (compare != 0)
+					return compare;
+			}
+
+			return Integer.compare(na, nb);
+		}
+
+		if (a instanceof Collection) {
+			Collection la = (Collection) a;
+			Collection lb = (Collection) b;
+
+			int na = la.size();
+			int nb = lb.size();
+			int l = Math.min(na, nb);
+
+			Iterator ia = la.iterator();
+			Iterator ib = lb.iterator();
+
+			for (int i = 0; i < l; i++) {
+				assert ia.hasNext() && ib.hasNext();
+				Object oa = ia.next();
+				Object ob = ib.next();
+				compare = compare(oa, ob);
+				if (compare != 0)
+					return compare;
+			}
+
+			return Integer.compare(na, nb);
+		}
+
+		return 0;
+	}
+
+	/**
+	 * Compare sequential fields in a Map with {@link #compare(Object, Object)}.
+	 * The given fields are retrieved from the map and then compared. The first
+	 * non-0 value defines the result.
+	 *
+	 * @param a the a value or null
+	 * @param b the b value or null
+	 * @param keys the array of keys to use
+	 * @return 0, 1 a>b, -1 a<b
+	 */
+	static int compare(Map<String, Object> a, Map<String, Object> b, String... keys) {
+		int compare = compareNull(a, b);
+		if (isDecided(compare))
+			return compare;
+
+		for (String key : keys) {
+			Object va = a.get(key);
+			Object vb = b.get(key);
+			compare = compare(va, vb);
+			if (compare != 0)
+				return compare;
+		}
+		return 0;
+	}
+
+}
--- a/aQute.libg/src/aQute/lib/comparators/Comparators.java
+++ b/aQute.libg/src/aQute/lib/comparators/Comparators.java
@@ -2,14 +2,76 @@ package aQute.lib.comparators;
 
 import java.lang.reflect.Array;
 import java.util.Collection;
+import java.util.Comparator;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.function.IntSupplier;
 
 /**
  * Some utilities to make comparisons easier to do, especially with null
  * handling
  */
 public interface Comparators {
+
+	/**
+	 * When you compare, the nulls can be quite annoying to handle. For this
+	 * interface, we use a magic value to mark outside the normal 1,0,-1 values
+	 * for compare to indicate that both values are not null and should be
+	 * compared with traditional means.
+	 */
+	int COMPARISON_REQUIRED = Integer.MAX_VALUE;
+
+	/**
+	 * Protect a comparator from being called with null
+	 * <ul>
+	 * <li>if they are both null, 0
+	 * <li>if a and !b -> 1
+	 * <li>if !a and b -> -1
+	 * <li>Otherwise result of the nonNulComparator.compare(a,b)
+	 * </ul>
+	 *
+	 * @param <T> the shared type
+	 * @param a
+	 * @param b
+	 * @param nonNullComparator
+	 * @return the result
+	 */
+
+	static <T> int compare(T a, T b, Comparator<T> nonNullComparator) {
+
+		if (a != null) {
+			if (b != null)
+				return nonNullComparator.compare(a, b);
+			else
+				return 1;
+		} else if (b != null)
+			return -1;
+		else
+			return 0;
+	}
+
+	/**
+	 * Shield the nonNullComparator from null objects. This is a convenience
+	 * method so you do not have to invent new parameter names but can use the
+	 * direct parameters from the invoked method.
+	 * <ul>
+	 * <li>if they are both null, 0
+	 * <li>if a and !b -> 1
+	 * <li>if !a and b -> -1
+	 * <li>Otherwise nonNullComparator.getAsInt()
+	 * </ul>
+	 *
+	 * @param <T> the shared type
+	 * @param a
+	 * @param b
+	 * @return the result
+	 */
+
+	static <T> int compare(T a, T b, IntSupplier nonNullComparator) {
+		return compare(a, b, (aa, bb) -> nonNullComparator.getAsInt());
+	}
 
 	/**
 	 * Compare two objects for null.
@@ -28,17 +90,8 @@ public interface Comparators {
 	 * @return the result
 	 */
 
-	static <T> int compareNull(T a, T b) {
-
-		if (a != null) {
-			if (b != null)
-				return Integer.MAX_VALUE;
-			else
-				return 1;
-		} else if (b != null)
-			return -1;
-		else
-			return 0;
+	static <T> int comparePresent(T a, T b) {
+		return compare(a, b, () -> COMPARISON_REQUIRED);
 	}
 
 	/**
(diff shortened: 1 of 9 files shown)


pkriens:
Fixes  #5807

pkriens:
@mnlipp could you take a look at these changes?

mnlipp:
> @mnlipp could you take a look at these changes?

I will. Just have to solve some other problem first in order to test it in my environment.

kriegfrj:
> Fixes #5807

Haven't had a detailed look at the fix, but the "Fixes #5807" text should go in one or more of the commits too, no?

pkriens:
@kriegfrj I guess :-) When I get this one merged, I'll close #5807 

Thanks for engaging.

mnlipp:
In order to test your version, I've upgraded everything in my environment to 7.0.0. Nice effect: resolution works again without problems (for whatever reason; I assume that there is no difference between the 6.4.x comparator and the 7.0.0 comparator).

Then I've build the p2 repository from your PR and installed it into eclipse. Now the resolution fails (doesn't terminate). I've appended the last entries from the log. I won't claim that I understand this. Failure is somehow always related to the felix webconsole, no matter how long I wait. I've tried putting some things on the blacklist that are not part of my previous (7.0.0) successful solution (e.g. `org.apache.tomcat-servlet-api` -- I don't need that, so I assumed that it "prunes" some possibilities) but this didn't help. (Interestingly, `org.apache.tomcat-servlet-api [org.apache.tomcat-servlet-api` still appears in the log even when it is on the blacklist, no idea if this is okay or another problem.)

```
...
DEBUG: Candidate permutation failed due to a conflict between imports; will try another if possible. (Uses constraint violation. Unable to resolve resource org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6] because it is exposed to package 'javax.servlet.http' from resources org.apache.tomcat-servlet-api [org.apache.tomcat-servlet-api version=9.0.80] and org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0] via two dependency chains.

Chain 1:
  org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6]
    import: (&(osgi.wiring.package=javax.servlet.http)(version>=3.0.0)(!(version>=5.0.0)))
     |
    export: osgi.wiring.package: javax.servlet.http
  org.apache.tomcat-servlet-api [org.apache.tomcat-servlet-api version=9.0.80]

Chain 2:
  org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6]
    require: (&(osgi.implementation=osgi.http)(version>=1.1.0)(!(version>=2.0.0)))
     |
    provide: osgi.implementation;uses:='javax.servlet,javax.servlet.http,org.osgi.service.http.context,org.osgi.service.http.whiteboard';osgi.implementation='osgi.http';version:Version='1.1.0'
  org.apache.felix.http.bridge [org.apache.felix.http.bridge version=4.1.2]
    import: (&(osgi.wiring.package=javax.servlet.http)(version>=3.1.0)(!(version>=4.0.0)))
     |
    export: osgi.wiring.package: javax.servlet.http
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0])
DEBUG: Candidate permutation failed due to a conflict between imports; will try another if possible. (Uses constraint violation. Unable to resolve resource org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6] because it is exposed to package 'javax.servlet.http' from resources org.apache.tomcat-servlet-api [org.apache.tomcat-servlet-api version=9.0.80] and org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0] via two dependency chains.

Chain 1:
  org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6]
    import: (&(osgi.wiring.package=javax.servlet.http)(version>=3.0.0)(!(version>=5.0.0)))
     |
    export: osgi.wiring.package: javax.servlet.http
  org.apache.tomcat-servlet-api [org.apache.tomcat-servlet-api version=9.0.80]

Chain 2:
  org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6]
    require: (&(osgi.implementation=osgi.http)(version>=1.1.0)(!(version>=2.0.0)))
     |
    provide: osgi.implementation;uses:='javax.servlet,javax.servlet.http,org.osgi.service.http.context,org.osgi.service.http.whiteboard';osgi.implementation='osgi.http';version:Version='1.1.0'
  org.apache.felix.http.bridge [org.apache.felix.http.bridge version=4.1.2]
    import: (&(osgi.wiring.package=javax.servlet.http)(version>=3.1.0)(!(version>=4.0.0)))
     |
    export: osgi.wiring.package: javax.servlet.http
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0])
DEBUG: Candidate permutation failed due to a conflict between imports; will try another if possible. (Uses constraint violation. Unable to resolve resource org.apache.felix.http.jetty [org.apache.felix.http.jetty version=4.2.18] because it is exposed to package 'javax.servlet' from resources org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0] and org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=1.2.0] via two dependency chains.

Chain 1:
  org.apache.felix.http.jetty [org.apache.felix.http.jetty version=4.2.18]
    import: (&(osgi.wiring.package=javax.servlet)(version>=3.1.0)(!(version>=4.0.0)))
     |
    export: osgi.wiring.package: javax.servlet
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0]

Chain 2:
  org.apache.felix.http.jetty [org.apache.felix.http.jetty version=4.2.18]
    require: (&(osgi.contract=JavaServlet)(version=3.1))
     |
    provide: osgi.contract: JavaServlet
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=1.2.0])
DEBUG: Candidate permutation failed due to a conflict between imports; will try another if possible. (Uses constraint violation. Unable to resolve resource org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6] because it is exposed to package 'javax.servlet' from resources org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0] and org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=2.1.0] via two dependency chains.

Chain 1:
  org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6]
    import: (&(osgi.wiring.package=javax.servlet)(version>=3.0.0)(!(version>=5.0.0)))
     |
    export: osgi.wiring.package: javax.servlet
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0]

Chain 2:
  org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6]
    require: (&(osgi.implementation=osgi.http)(version>=1.1.0)(!(version>=2.0.0)))
     |
    provide: osgi.implementation;uses:='javax.servlet,javax.servlet.http,org.osgi.service.http.context,org.osgi.service.http.whiteboard';osgi.implementation='osgi.http';version:Version='1.1.0'
  org.apache.felix.http.bridge [org.apache.felix.http.bridge version=4.1.2]
    import: (&(osgi.wiring.package=javax.servlet.http)(version>=3.1.0)(!(version>=4.0.0)))
     |
    export: osgi.wiring.package: javax.servlet.http; uses:=javax.servlet
    export: osgi.wiring.package=javax.servlet
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=2.1.0])
DEBUG: Candidate permutation failed due to a conflict between imports; will try another if possible. (Uses constraint violation. Unable to resolve resource org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6] because it is exposed to package 'javax.servlet' from resources org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0] and org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=2.1.0] via two dependency chains.

Chain 1:
  org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6]
    import: (&(osgi.wiring.package=javax.servlet)(version>=3.0.0)(!(version>=5.0.0)))
     |
    export: osgi.wiring.package: javax.servlet
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0]

Chain 2:
  org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6]
    require: (&(osgi.implementation=osgi.http)(version>=1.1.0)(!(version>=2.0.0)))
     |
    provide: osgi.implementation;uses:='javax.servlet,javax.servlet.http,org.osgi.service.http.context,org.osgi.service.http.whiteboard';osgi.implementation='osgi.http';version:Version='1.1.0'
  org.apache.felix.http.bridge [org.apache.felix.http.bridge version=4.1.2]
    import: (&(osgi.wiring.package=javax.servlet.http)(version>=3.1.0)(!(version>=4.0.0)))
     |
    export: osgi.wiring.package: javax.servlet.http; uses:=javax.servlet
    export: osgi.wiring.package=javax.servlet
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=2.1.0])
DEBUG: Candidate permutation failed due to a conflict between imports; will try another if possible. (Uses constraint violation. Unable to resolve resource org.apache.felix.http.jetty [org.apache.felix.http.jetty version=4.2.18] because it is exposed to package 'javax.servlet' from resources org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0] and org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=1.2.0] via two dependency chains.

Chain 1:
  org.apache.felix.http.jetty [org.apache.felix.http.jetty version=4.2.18]
    import: (&(osgi.wiring.package=javax.servlet)(version>=3.1.0)(!(version>=4.0.0)))
     |
    export: osgi.wiring.package: javax.servlet
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0]

Chain 2:
  org.apache.felix.http.jetty [org.apache.felix.http.jetty version=4.2.18]
    require: (&(osgi.contract=JavaServlet)(version=3.1))
     |
    provide: osgi.contract: JavaServlet
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=1.2.0])
DEBUG: Candidate permutation failed due to a conflict between imports; will try another if possible. (Uses constraint violation. Unable to resolve resource org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6] because it is exposed to package 'javax.servlet' from resources org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0] and org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=2.0.0] via two dependency chains.

Chain 1:
  org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6]
    import: (&(osgi.wiring.package=javax.servlet)(version>=3.0.0)(!(version>=5.0.0)))
     |
    export: osgi.wiring.package: javax.servlet
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0]

Chain 2:
  org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6]
    require: (&(osgi.implementation=osgi.http)(version>=1.1.0)(!(version>=2.0.0)))
     |
    provide: osgi.implementation;uses:='javax.servlet,javax.servlet.http,org.osgi.service.http.context,org.osgi.service.http.whiteboard';osgi.implementation='osgi.http';version:Version='1.1.0'
  org.apache.felix.http.bridge [org.apache.felix.http.bridge version=4.1.2]
    import: (&(osgi.wiring.package=javax.servlet.http)(version>=3.1.0)(!(version>=4.0.0)))
     |
    export: osgi.wiring.package: javax.servlet.http; uses:=javax.servlet
    export: osgi.wiring.package=javax.servlet
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=2.0.0])
DEBUG: Candidate permutation failed due to a conflict between imports; will try another if possible. (Uses constraint violation. Unable to resolve resource org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6] because it is exposed to package 'javax.servlet' from resources org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0] and org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=2.0.0] via two dependency chains.

Chain 1:
  org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6]
    import: (&(osgi.wiring.package=javax.servlet)(version>=3.0.0)(!(version>=5.0.0)))
     |
    export: osgi.wiring.package: javax.servlet
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=3.0.0]

Chain 2:
  org.apache.felix.webconsole [org.apache.felix.webconsole version=4.9.6]
    require: (&(osgi.implementation=osgi.http)(version>=1.1.0)(!(version>=2.0.0)))
     |
    provide: osgi.implementation;uses:='javax.servlet,javax.servlet.http,org.osgi.service.http.context,org.osgi.service.http.whiteboard';osgi.implementation='osgi.http';version:Version='1.1.0'
  org.apache.felix.http.bridge [org.apache.felix.http.bridge version=4.1.2]
    import: (&(osgi.wiring.package=javax.servlet.http)(version>=3.1.0)(!(version>=4.0.0)))
     |
    export: osgi.wiring.package: javax.servlet.http; uses:=javax.servlet
    export: osgi.wiring.package=javax.servlet
  org.apache.felix.http.servlet-api [org.apache.felix.http.servlet-api version=2.0.0])

```

chrisrueger:
> (doesn't terminate)

This sounds a bit like 
- https://github.com/bndtools/bnd/issues/4172#issuecomment-646128590 
- https://github.com/bndtools/bnd/pull/4409
- https://issues.apache.org/jira/browse/FELIX-6358

Sorry, I cannot be more helpful here.

mnlipp:
> Sorry, I cannot be more helpful here.

Thanks. Eventually, I'll have to blacklist until things work, done that before. In this special context, however, the question is whether this PR "breaks" something. I would have expected that modifying the comparator has effects on the time required to find the solution. I wouldn't have been surprised if the solution found was slightly different But I find it surprising that the "known solution" isn't found any more.

pkriens:
@tjwatson could you take a quick look at the diagnosis and see if you recognize the problem?

@mnlipp can you share the the whole debug log?

mnlipp:
Log file attached. (Project is https://github.com/mnlipp/jgrapes-osgi/commits/master (968cf2d6640e7b923fc139258c91a61582cdda1f), resolving `org.jgrapes.osgi.demo.console/democonsole.bndrun`.)

[tmp13821316911941492298.log.gz](https://github.com/bndtools/bnd/files/12913922/tmp13821316911941492298.log.gz)

pkriens:
Hmm. These are you providers for the javax.servlet.http package:

image

Http servlet might not be the absolute worst API in Java, from an OSGi point of view developers have tried since the heydays of EE server to give OSGi a really hard time making every bad semantic versioning and API decision that mankind could possibly come up with. 

You're torturing the resolver beyond my imagination :-) 

* Two versions of Tomcat Servlet API,
* Ten versions of the Felix http API? 
* Three times httplite? 
* And last and least, 3x javax servlet API. And I am ignoring the Jakarta jars.

Eighteen choices! And then I am not counting that all those JARs lie through their teeth that they support multiple versions for these packages since semantic versioning was not on the radar in the EE world. So the resolver probably sees 40 or 50 candidates. That is really not what it was designed for.

I am really happy that you made me cleanup the comparators, they were really awfully bad. However, one of the wisest lessons of Elon Musk is not to optimize a solution, but try to get rid of the problem altogether. It has been my motto all my life, a better design is when it has fewer potential error cases.

I think we need to figure out why the solution does not resolve on Apache/Eclipse and I've asked Eclipse's @tjwatson to look over our shoulders. 

From a bndtools point of view I think the solution for you is trivial, you have to clean up your repositories. The problem you run into is closely associated with [#4172](https://github.com/bndtools/bnd/issues/4172#issuecomment-646128590), you're having WAY too many options for the resolver. Also in bndtools, the resolution took > 20 seconds on an M2 chip. I haven't got a single bndrun file that takes more than 3 secs to resolve. 

By having this many options for the servlet (and associated) packages you are begging on your knees  for resolver issues.

I really appreciate helping to make the resolver better and the work you did. Very welcome. But I suggest we move the log and a commit record of your project and attach it to [#4172](https://github.com/bndtools/bnd/issues/4172#issuecomment-646128590) because I am pretty sure you've run into this bug.

On your site, get rid of all but one servlet API JAR and your resolve will be sub second and more reliable ... Imho, keeping the repository's small saves a tremendous amount of issues, a lot of complicated things tend too get trivial.

What do you think?

tjwatson:
Can you describe what you are fixing here.  It says fixing #5807, but that is this PR.

There were previous issues in BND where the comparator for BND resources was clearly wrong and that caused issues with BND and the resolver.  But I find it hard to tell (only looking for a little bit) what this PR is fixing or the behavior it changed.

pkriens:
@tjwatson sorry. the bug is #5793 but I don't think you're that concerned about it. I've changed the comparators in bnd since they were terrible and did not give consistent ordering. However, @mnlipp, who reported a sorting error due to the instability of the ordering, now has a resolution that succeeds on bndtools but fails on Eclipse.

I was just asking if you could take a look at the log output in this bug and see if you maybe recognize the issue, the resolver loops forever. In the mean time I found out the project had 18 (!) jars with the servlet packages in myriad versions. It looks to me like the Apache Felix https://github.com/bndtools/bnd/issues/4172#issuecomment-646128590 problem?

mnlipp:
> What do you think?

Two things:

(1) As I don't have a [release manager](https://bnd.discourse.group/t/memory-usage-explosion-with-bnd-resolver-plugin/307/2?u=mnlipp) for my OSS projects, I rely on the conventional strategy of trying to use the latest version of each bundle. (These _may_ introduce new bugs but they _should_ fix known security issues.) In order to automate this, my repository plugin scans the maven repositories starting with some explicitly given groups and lower bounds for artifacts and generates an index of these and all their dependencies (yes, the resulting index/repository is too big). All providers for the javax.servlet.http package that are in the "IndexedMaven" repository are actually dependencies of some other bundles, none was included explicitly.

From time to time, I raise the lower bounds for the artifacts to reduce the number of bundles in the (generated) repository, but specific selections for my application (when required) are mostly managed by blacklisting bundles. This might push the resolver to its limits, but I've always thought that this is what it was made for. Yes, it takes some time, but I know that the underlying problem is difficult and, hey, how often do you have the opportunity to really see your hardware investment paying off.

Unless this PR introduces some "fatal" issue, I will eventually get the resolution working again by adjusting my repository. And I understand that I should try to do this (and will within the next days) in order to provide additional input for evaluating this PR.

(2) I still consider it strange that this PR *breaks* the resolution. I only had a superficial look at the code, and from this it looks like a rewrite. So, complicated as my resolution scenario may be, I wonder if it happens to be the missing (unit)test that reveals a regression bug introduced by this PR. And if this is the case, the PR might eventually not only break my resolving but also other's. If I was responsible for QA, I'd at least insist on some more non-trivial scenarios to be tested with this PR in place and check if they still resolve.

tjwatson:
> I think we need to figure out why the solution does not resolve on Apache/Eclipse and I've asked Eclipse's @tjwatson to look over our shoulders.

I think I misunderstood the ask of me.  So this is an issue trying to install the latest local build of the development branch+PR changes into an Eclipse installation and the resolution error is coming from that?  Or is this error happening from some resolution driving from BND which is not using this PR?

pkriens:
@tjwatson sorry for not being more clear, my brain turning mush :-(

* A bndrun file is resolved
* Turned into p2 repo
* Installed in Eclipse
* Resolve never ends ...

@mnlipp This is the story, isn't it?

The thing is, I've changed the comparators of Resource, Capability, and Requirement so it is kind of a conincedence that it now fails. However, there are 18 Jars with the servlet packages which each have numerous versions so it looks like a prime candidate for the resolver bug. I just wanted you to take a quick look if the log looked familiar with the resolver bug where a resolution never ends.

pkriens:
@mnlipp I just believe in a very clean work floor to keep productivity up. That said, in maven they at least pick the _first_ encountered version of an artifact. You seem to include multiple versions of the same artifact. This would already solve this problem I think.

>  I still consider it strange that this PR breaks the resolution. 

I do not think this PR breaks the resolution in any way. Let's assume the bundles we give Eclipse are not a valid solution. The resolver in Eclipse should then tell is why it is bad. There is just no input we could possible give to Eclipse that would warrant not resolving. I.e this by definition is a bug in Eclipse/Apache resolver setup. No input should be able to create this behavior.

I always hate these kind of 'coincidences' since they are extremely rare in our world. When two things happen at the same time they are almost always related. However, in this case Eclipse/Apache _is_  by definition wrong by not resolving or claiming there is no resolution. Those are the only two valid outcomes of the API. Without this resolve bug out of the way, there is nothing we can do. I am also pretty confident that if bndtools found a solution, there is _is_ a solution. I've not printed out the wiring yet to inspect it by hand. Kept back by my PTSD of servlets, it feels like having to dive in a cave filled with snakes :-(

pkriens:
So @mnlipp, what do we do? I am inclined to submit this PR and see what happens? 

@bndtools/maintainers any feedback to not merge this?

mnlipp:
I am, just now, trying to get the resolution working again (sorry, didn't have time to do it earlier). I've reduced the number of servlet apis. Now (2 minutes ago) I get
```
Resolution failed. Capabilities satisfying the following requirements could not be found:
    [<<INITIAL>>]
      ⇒ osgi.identity: (osgi.identity=de.mnl.osgi.log4j2osgi)
          ⇒ [de.mnl.osgi.log4j2osgi version=1.2.0.ga]
              ⇒ osgi.wiring.package: (&(osgi.wiring.package=org.apache.logging.log4j)(version>=2.16.0)(!(version>=3.0.0)))
                  ⇒ [org.apache.logging.log4j.api version=2.21.0]
                      ⇒ bnd.multirelease: (&(bnd.multirelease=org.apache.logging.log4j.api)(&(version>=2.21.0)(version<=2.21.0)))
```

I've never seen this before and I'm currently trying to understand it (log4j-api with version 2.21.0 *is* in my repository). I don't know yet whether this is related to this PR and wasn't "visible" before because of the never ending resolution or whether this is a completely independent issue. It's up to you to wait some more hours and see if something else shows up in my "special environment" or submit the PR now...

mnlipp:
> so to make you co-responsible when it fails :-), shall we merge? 

Well, go ahead.

tjwatson:
> The thing is, I've changed the comparators of Resource, Capability, and Requirement so it is kind of a conincedence that it now fails. However, there are 18 Jars with the servlet packages which each have numerous versions so it looks like a prime candidate for the resolver bug. I just wanted you to take a quick look if the log looked familiar with the resolver bug where a resolution never ends.

I haven't had time to dig into this in details.  But it does sound like the resolver is simply overwhelmed trying to solve the NP-hard problem with the number of possible choices.  There are two things that usually help when installing such atrocities.

1. Start eclipse with `-clean` so that it re-installs all the bundles and resolves from scratch.  This makes it possible to make more optimized decisions early such that it can find a solution faster.
2. Add the property `equinox.resolver.revision.batch.size=1` to the `config.ini` file of your eclipse installation.

pkriens:
This is mostly in the resolve context where we use it to order. It would not surprise me if we sort more often than necessary. Looking at what I know about the resolve process, would surprise me if the comparator would make a visible difference.

mnlipp:
Please note that I didn't systematically measure the time! My remark was based on an observation made when switching from one version to the other, which implied a restart of Eclipse. Although I am quite sure that I also had (re-)started Eclipse before the switch to verify the "old" resolution, all kinds of causes may have had an effect on that single observation. 

Especially that fact that the result of the old (re-tested) resolve process was what was already given while this PR version found a slightly different result may have had a significant effect. Better measurements would be required before pondering too much about this.

tjwatson:
> I agree... I don't think it's worth the effort to benchmark formally/accurately. But I do think that it is worth us as alpha testers keeping an eye on it in case any significant performance changes are noted by any of us - that might warrant closer investigation.

Perhaps, but this is largely a concern of the Felix resolver implementation and its use in Equinox vs. anything BND tools can do.

pkriens:
wow, this is shaping up to become a nice tool!

I think is ok. There are a lot of rewrapped jars out there. The resolver will pick them more or less randomly. As long as the contracts are fulfilled this is ok.

Issue #5816, opened 2023-10-16
bndrun editor: Dragging bundle from Browse Repos to Run Bundles / Run Requirements should use stricter version-range instead of open range version='1.2.3'


chrisrueger:
At EclipseCon2023 @juergen-albert pointed out in his presentation that in OSGI 

version='1.2.3' basically means "this version and everything above". Doing this could be dangerous,as it might pull in higher versions you do not expect.

See https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#i2655136
<img width="1301" alt="image" src="https://github.com/bndtools/bnd/assets/188422/e4ce5180-3ff7-442a-a68c-14efd72d838a">

## Problem

When I drag a bundle in the bndrun-editor to e.g. the Run Bundles then it uses the '1.2.3' version instead of a range.

<img width="927" alt="image" src="https://github.com/bndtools/bnd/assets/188422/f99bd45d-1ae7-4157-8c9a-ef3e8888f448">

**Actual result:**
`org.apache.commons.collections4;version='4.1.0'`

**Expected / suggested result:**
`org.apache.commons.collections4;version='[4.1.0,4.1.1)'` or `org.apache.commons.collections4;version='[4.1.0,4.1.0]'`

Having a stricter and smaller range could prevent problems and unexpected events.

Edit: Just want to add, that I think this is not a bug. But since bndtools should "do good" things, I think the current behavir might lead to unexpected events. So consider this a "Usability improvement".

pkriens:
a) the idea of the resolver is to only supply the functional bundles, the bundles you care about to be in your application. It is hard to imagine your customer cares about apache collections? The idea of the resolver is that you specify the bundles the customer is interested, they tend to be leaf bundles that nobody depends on. The resolver than drags in the rest.

b) I am not a fan of these small ranges until you get into problems. The more freedom the resolver has, the faster and better solutions it finds. This is not maven where you want a hard reference to a a specific JAR.

c) I could live with restricting it to the next major range but definitely not by default to the next micro, that is way too restrictive and makes life downstream too complicated.

chrisrueger:
> c) I could live with restricting it to the next major range but definitely not by default to the next micro, that is way too restrictive and makes life downstream too complicated.

👍

I think this was my intention... I just didn't write it😃 my bad. 
My point was, that i was surprised that version=1.2.3 is not an exact match. 
Usually we use a range until the next major when editing by hand, but i just noticed when dragging in the ui.

[...]

PR #5823, opened 2023-10-18
use VersionRange when adding bundle to Run Bundles

chrisrueger:
Related to #5816 

This PR changes the behaviour of the .bndrun Editor when adding a bundle from a Repo to the Run Bundles panel,
a) via drag&drop
b) via the +-Button in the Run Bundles panel

The goal is that a version-range is used from the bundleVersion (inclusive) to the next major version (exclusive)

Before this change:
- bundle was added like `org.apache.felix.scr;version='2.2.2'` (which means "[2.2.2 and everything above](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#d0e2221)", which could be surprising, as it looks like an exact version match if you don't know the spec)

After this change:
- bundle gets added like org.apache.felix.scr;version='[2.2.2,3.0.0)'

image

image

Gets added as this.
image

Now it looks similar to what happens in the Run Requirements:
`bnd.identity;id='org.apache.felix.gogo.runtime';version='[1.1.0,2.0.0)'`

image

### DIFF
--- a/bndtools.core/src/bndtools/model/repo/RepositoryBundleUtils.java
+++ b/bndtools.core/src/bndtools/model/repo/RepositoryBundleUtils.java
@@ -9,19 +9,32 @@ import aQute.bnd.version.Version;
 import aQute.bnd.version.VersionRange;
 
 public class RepositoryBundleUtils {
-	private static final String VERSION_LATEST = "latest";
 
+	/**
+	 * Converts a RepositoryBundle into a versionioned clause. There are 2
+	 * cases: a) If the bundle comes from a workspace repository, then
+	 * bsn;version=snapshot is returned. b) Otherwise only the bsn without a
+	 * version is returned.
+	 *
+	 * @param bundle
+	 * @return a version as described above
+	 */
 	public static VersionedClause convertRepoBundle(RepositoryBundle bundle) {
 		Attrs attribs = new Attrs();
 		if (RepoUtils.isWorkspaceRepo(bundle.getRepo())) {
-			attribs.put(Constants.VERSION_ATTRIBUTE, VERSION_LATEST);
+			attribs.put(Constants.VERSION_ATTRIBUTE, Constants.VERSION_ATTR_SNAPSHOT);
 		}
 		return new VersionedClause(bundle.getBsn(), attribs);
 	}
 
 	/**
 	 * Converts a RepositoryBundleVersion into a version or version range. e.g.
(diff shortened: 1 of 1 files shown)


bndtools.core/src/bndtools/model/repo/RepositoryBundleUtils.java
```diff
@@ -17,26 +19,55 @@ public static VersionedClause convertRepoBundle(RepositoryBundle bundle) {
...
 			StringBuilder builder = new StringBuilder();
 
 			builder.append(bundleVersion.getVersion()
 				.getMajor());
 			builder.append('.')
 				.append(bundleVersion.getVersion()
 					.getMinor());
+
+			// TODO why is this check?
 			if (phase != DependencyPhase.Build)
```
chrisrueger, line 47: I'm glad to get some hints, since this is my first code PR in bndtools and I am new to the Eclipse GUI development.

Maybe you could help me with what this check means. I left it in place, and the version check also works with it. 
But not sure why it is there.

chrisrueger: @pkriens 
I found a commit by @njbartlett  

Commit 814c0946d33e170be8e17bf5a8458a07cb507573

> fix #681: Dragging version of dependency to -buildpath => use Major.minor segments only

image

I tried dragging an version-entry from the Repository browser directly in the `-buildpath` and can confirm that this only uses the Major.minor segment.

image

What is the reason for using only the **Major.minor segment** here?

In case it is not needed anymore, the check `if (phase != DependencyPhase.Build)` could be removed, which would make the code simpler (avoiding the StringBuilder and use the `bundleVersion.getVersion()` directly.

Thoughts?

kriegfrj: Perhaps the reasoning is that two bundles with the same major.minor should be identical, from a build perspective. Micro version number pertains to implementation details, not to the API - so the API should be identical between two bundles that have the same major.minor version. Just a thought.

chrisrueger: Thanks @kriegfrj , I adjusted the code to handle both cases more explicit as it was before and added some comments. So the -buildpath logic in bnd.bnd stays as is, based on your statement. The  other case (.bndrun Editor) now adds a range.

chrisrueger:
Thanks a lot for merging @pkriens

Issue #4031, opened 2020-05-07
[bnd-run-maven-plugin] Run in test phase with target/classes


kenwenzel:
This issue is based on the discussion at [https://groups.google.com/d/msg/bndtools-users/MMBdqN-wU78/x1GKtbH2AwAJ](https://groups.google.com/d/msg/bndtools-users/MMBdqN-wU78/x1GKtbH2AwAJ).

For fast turnarounds it may sometimes be desirable to run the OSGi framework with exploded bundles. For example, web applications with HTML templates and JS files don't require a restart of the bundle if just a single line of code changes.
This way of running an application is the default behaviour of Eclipse PDE that simply handles all workspace projects as exploded bundles when starting the OSGi framework.

I tried to do the same with bnd-run-maven-plugin but it fails due to a check in FileSetRepository:
[https://github.com/bndtools/bnd/blob/3a4ea3ea7cba5359f0212487691693d0f488b1b5/biz.aQute.repository/src/aQute/bnd/repository/fileset/FileSetRepository.java#L98](https://github.com/bndtools/bnd/blob/3a4ea3ea7cba5359f0212487691693d0f488b1b5/biz.aQute.repository/src/aQute/bnd/repository/fileset/FileSetRepository.java#L98)

Is it possible to change the respective code for allowing bnd-run-maven-plugin to execute within `mvn test`?

bjhargrave:
> Is it possible to change the respective code for allowing bnd-run-maven-plugin to execute within `mvn test`?

Did you try this change to see if it even works? I suspect there are many other parts of launching that expect bundles to be files. This is probably just the first place.

kenwenzel:
Not yet. What is the best way to build and test the Maven plugins? How can I use Gradle to install them within my local Maven repository?

bjhargrave:
When you build the project with gradle, it will install the jars into the local .m2/repo. 

See https://github.com/bndtools/bnd/blob/master/.github/scripts/build.sh for the targets to build.

PR #5829, opened 2023-10-20
[bndtools] Launch to bndrun

kriegfrj:
This PR is a proof-of-concept. It will take a PDE `.launch` file and convert it into a `.bndrun`. The final `.bndrun` might still need some manual tweaking. It also assumes that you've done the work to convert the PDE target platform definition into Bnd respositories in your workspace.

There are two parts:

1. `biz.aQute.bnd.pde`, which contains the actual utility that does the conversion;
2. A context menu addition so that when you right-click on an `xxx.launch` file, you will have a menu item "Generate bndrun...". If you select it, it will create a corresponding `xxx.bndrun` file using the `biz.aQute.bnd.pde` library.

I am interested in feedback of all types. Is this a useful addition? Is the name appropriate? Is it in the right package? Should it be in its own bundle or rolled into another?


chrisrueger:
Thanks a lot @kriegfrj . Could you maybe describe the usecase / purpose a bit more? Who would be the user of this feature and when would it be used? That context might help with feedback.

kriegfrj:
Sure, 

Use case is for someone who wants to convert a project from PDE to Bndtòols. Part of that process includes having to convert PDE launch files into bndrun equivalents, for launching the frameworks. This tool automates a large part of that step.

peterkir:
Are the PRs can be checked as builded binaries (which can be installed and used in Eclipse) or do I have to test it from the source?

kriegfrj:
> Are the PRs can be checked as builded binaries (which can be installed and used in Eclipse) or do I have to test it from the source?

The binaries are not publically available. There's two ways you could test it:

1. From the Bndtools Eclipse test instance:
   1. Check out this branch of Bnd in your Bnd workspace in Eclipse.
   2. Browse to the `bndtools.core` project and run the `bndtools.xxx.bndrun` launch file (where xxx corresponds to the OS that you're running on). This will launch another Eclipse instance that has the current bundles installed.
2. In your own Eclipse installation:
   1. Check out this branch of Bnd.
   2. Build the p2 repo locally (`./gradlew p2` from the project root directory at the command line).
   3. In your Eclipse instance, do "Install new software" and add your local P2 repo.
   4. Install Bndtools from the local P2 repo and restart.

The first is probably easier.

kriegfrj:
Btw, thanks for being brave enough to consider giving it a try, @peterkir !

pkriens:
@kriegfrj you're still planning to move this forward?

kriegfrj:
Has anyone tried this out?

chrisrueger:
> Has anyone tried this out?

No, sorry, since I don't have any PDE knowledge. But we might include @laeubi in the discussion. He was involved some [bnd - PDE collaboration work](https://bnd.discourse.group/t/announcement-bnd-and-pde-cooperation/372) and might be a good PDE person.

kenwenzel:
Is this also related to run the OSGi container with extracted bundles in `target/classes`?

https://github.com/bndtools/bnd/issues/4031

pkriens:
@kriegfrj I want to close this to inactivity. Feel free to reopen it when you want to work on it again. Hope one day you'll be back ...

pkriens:
Missed the question of @kenwenzel ...

Anybody can answer this?

laeubi:
The referenced issue
- https://github.com/bndtools/bnd/issues/4031

talks about `bnd-run-maven-plugin` this is about eclipse `launch` config to `bndrun` conversation so I highly doubt it solves anything or is related to that topic.

Beside that, Equinox (not sure about Felix) already supports using "exploded bundles", so it is more how bnd handles it and how maven works, running tests directly from `target/classes` is more a hack than something maven-devs would recommend (even though it works in most cases).

kenwenzel:
@laeubi If you execute `mvn test` then all unit tests are run based on the contents of `target/classes`.

I'm just looking for a bnd-based alternative for the `maven-osgi-exec-plugin`
https://github.com/BestSolution-at/maven-osgi-plugin/tree/master/maven-osgi-exec-plugin
that uses Equinox SimpleConfigurator to launch the OSGi framework based on a `bundles.info` file.
This mimics the standard PDE launch behavior.

The big advantage is, for example, that a web application could be started where HTML or JS files are modified and copied into `target/classes` (e.g. via Maven) and these changes are directly visible in the application without creating jar files for each small modification.
I'm aware that this is not completely aligned with OSGi-based module deployment but it allows for fast development cycles.

laeubi:
> @laeubi If you execute mvn test then all unit tests are run based on the contents of target/classes.

As said, this is a workaround maven uses because you don't have the package phase run (what likely will change with maven 4)

> I'm aware that this is not completely aligned with OSGi-based module deployment but it allows for fast development cycles.

As said, this is completely possible (and PDE/Tycho are working exactly that way usually), but in fact from a maven perspective the packaged jar is "the artifact" to use after the package phase (so after that you won't see` target/classes` on the classpath and also bnd maven plugin does not.

I don't say its useful /doable (the easiest way for your usecase would be to put the maven output folder on the classpath of the bundles using a equinox specific dev-entry) but not related to the work of this PR...

The main question for me would be: Why do you start a maven run at all and not launch it directly in the IDE (what uses references to jars and works already to hot-replace them AFAIK).

laeubi:
>  Hot code replacement 

Hot code replace != hot jar replace :-)

Anyways, due how maven work this won't work in a traditional maven build as there is no one who rebuilds the jar (in contrast to an IDE that can rebuild it on code change)

If you are using Equniox framework you can emulate "the PDE way" by specify a so called "dev classpath" with:

https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/misc/runtime-options.html#osgidev

> osgi.dev {-dev}
>     if set to the empty string, dev mode is simply turned on. This property may also be set to a comma-separated class path entries which are added to the class path of each plug-in or a URL to a Java properties file containing custom classpath additions for a set of plug-ins. For each plug-in requiring a customized dev time classpath the file will contain an entry of the form
> 
>         <plug-in id>=<comma separated list of classpath entries to add>
> 
>     where plug-in id "*" matches any plug-in not otherwise mentioned.

so in your case `osgi.dev=<id of your bundle>=${project.basedir}/target/classes` (or maybe even using `src/main/resources` would be better) would put your folder in question to the classpath of the specified bundle.

kenwenzel:
@laeubi Thanks for the hint on `osgi.dev`. Based on this discussion I assume that bnd will not support any launching with exploded bundles in the future?!

@pkriens I think you can close this PR now.

laeubi:
I defiantly can't speak for bnd or the future, I just can give some insight into the technical difficulties (from maven an/or OSGi), e.g. Tycho does add `osgi.dev` entries by default because that's how it works in PDE, but thats diverges from how maven works by default (both bnd and tycho execute at the integration-test phase where a packaged jar is actually the default).

What might also work is using bnd-run and use jar references to start the application from the local maven repo, and then run a `mvn install` in a second console to install changed jars on the fly...

kriegfrj:
@kenwenzel :
> Is this also related to running the OSGi container with extracted bundles in target/classes?

The short answer to your question is "no". :smile:

@laeubi:
> Hot code replace != hot jar replace :-)
> 

This is true. In fact, I've found in some cases where the two seem to step on each other's toes...

kriegfrj:
@pkriens 
> @kriegfrj I want to close this to inactivity. Feel free to reopen it when you want to work on it again. Hope one day you'll be back ...

It's been a while, but from memory this works. I just wanted some other people to try it out before merging. However, after this sitting here for so long and no one really trying it out, I can see that this was a rather forlorn hope. :smile:

So I'm going to propose another approach:

* It seems that no one is going to use it unless it actually gets merged.
* There seems little harm in merging it:
   * Its effects are confined to its own unique workflow which is quite parallel and distinct from an existing Bnd workflow.
   * The user always the opportunity to review any generated bndrun file before using it.

Hopefully, this way we might get some actual users and some real-world feedback, ideas and (hopefully) contributions.

Thoughts?

pkriens:
Thanks for engaging!

chrisrueger:
@pkriens @kriegfrj I just tried to update my Eclipse instance to latest Snapshot (as I have done a lot in the recent weeks, because I am always on the latest snapshot). After this PR got merged I get this error:

image

```
Cannot complete the install because one or more required items could not be found.
Software being installed: Bndtools 7.1.0.DEV-202411050747-g7d340b3 (bndtools.main.feature.feature.group 7.1.0.DEV-202411050747-g7d340b3)
Missing requirement: bndtools.core.services 7.1.0.202411050747-SNAPSHOT (bndtools.core.services 7.1.0.202411050747-SNAPSHOT) requires 'java.package; biz.aQute.bnd.pde.launch2bndrun [1.0.0,2.0.0)' but it could not be found
Cannot satisfy dependency:
From: Bndtools 7.1.0.DEV-202411050747-g7d340b3 (bndtools.main.feature.feature.group 7.1.0.DEV-202411050747-g7d340b3)
To: org.eclipse.equinox.p2.iu; bndtools.core.services [7.1.0.202411050747-SNAPSHOT,7.1.0.202411050747-SNAPSHOT]
```


What I did:

- Eclipse / About Eclipse 

then

image

and then

image


Then it takes longer than usual

image


and then the error appears

kriegfrj:
> Are the PRs can be checked as builded binaries (which can be installed and used in Eclipse) or do I have to test it from the source?

@peterkir , as you seemed to express an interest in trying this feature but didn't want to have to test it from the source...

This feature is now available in the Snapshot build of Bndtools. So if you're still keen to try it out, you can install the snapshot build of Bndtools and give it a try now.

Issue #5843, opened 2023-10-26
PropertiesLineReader has now way to read the comment


laeubi:
When `PropertiesLineReader` encounters a comment it currently just returns (and throws away everything), so after that the caller needs to recover the data by some other means.

Beside that `key()` method returns the key of the last seen item but should be `null` at this point

pkriens:
Are you going to work on this class? Could you then just open an (empty if you want) PR? I want to keep the issues razor sharp focused on problems that need urgent fixing.

laeubi:
See
- https://github.com/bndtools/bnd/pull/5850

Issue #5839, opened 2023-10-25
PropertiesLineReader never updates lastValue


laeubi:
[PropertiesLineReader](https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/properties/PropertiesLineReader.java) has a method value() but the field **is always null** as it is never assigned any value.

So any call to `PropertiesLineReader#value()` always throws exception.

laeubi:
If one is working on it, it would be great to have not only `IRegion` for whole but also for key / value separately.

pkriens:
Will you be submitting a PR?

laeubi:
See
- https://github.com/bndtools/bnd/pull/5850

PR #5850, opened 2023-10-28
Improve PropertiesLineReader and add testcase

laeubi:
Fixes https://github.com/bndtools/bnd/issues/5839
Fixes https://github.com/bndtools/bnd/issues/5843


pkriens:
If you fix the DCO, I will merge it asap

Issue #5844, opened 2023-10-26
[pde+bndtools] extract the bnd instructions (BndSourceEditorPage) into a bndtools.editor bundle


laeubi:
bndtools and pde both now include an editor that shows a source tab.
Unless 
- https://github.com/bndtools/bnd/issues/5833

is done and adopted, it seems a good idea to share these codes (currently there is an attempt to copy the basic parts [here](https://github.com/eclipse-pde/eclipse.pde/pull/823)).

For that purpose i'd like to suggest a new  `bndtools.editor` bundle that contains these basic parts in a way that it only depends on

- bndlib
- eclipse
- bndtools.api

that way PDE can include it with a consumer version range making it possible to nicely coexits with bndtools install.

Beside the editor, such bundle should at best also contain a [content-type definition](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fextension-points%2Forg_eclipse_core_contenttype_contentTypes.html) so PDE+BND can share the same content type so extension of one plugin can be applied to the other easily (this probably would also fix current [PDE vs BNDTools Editor confusion](https://github.com/eclipse-pde/eclipse.pde/issues/672) as one can define proper priority rules then).

pkriens:
Do you intend to make a PR or should we move this to abeyance  until someone wants to work on this?

PR #5852, opened 2023-10-29
Add help buttons for different views

chrisrueger:
See #5844 

The goal of this PR is to make it easier to get context sensitive help inside Bndtools using "help" buttons to each View linking to the relevant sections of the manual. 

Each button should also have a helpful tooltip summarizing the basic concept of the view ideally being helpful already. A click should open the webbrowser with the relevant page.

This would  bridge the gap between Eclipse and the manual and allow us to improve especially those entry-pages in the manual, since we now know they are most likely clicked from bndtools in Eclipse. 

## Current status of this PR

I provided the following Help buttons:

### Bndtools Explorer
image

Link: https://bndtools.org/manual/packageexplorer.html

### Repositories View
image

Link: https://bndtools.org/manual/repositories-view.html

### Resolution View
image

Link: https://bndtools.org/manual/resolution-view.html

### Bnd Editor for `bnd.bnd` files
image

Link: https://bndtools.org/manual/bndeditor.html

### Bnd Editor for `.bndrun` files
image

Link: https://bndtools.org/manual/bndeditor.html#run

### Workspace Bnd Editor e.g. cnf/build.bnd

image

Link: https://bndtools.org/manual/bndeditor.html

### Welcome Screen

Now the following sections of the Eclipse Welcome Screen contain a Bndtools relevant entry linking to a page on the bndtools.org website

image 

**Icons**
The help icon is currently from Eclipse internally and might be a bit big. But maybe this is a good thing, as it catches the eye more.
If you can tell me where to get a better icon which is allowed to be used, please let me know. @pkriens


chrisrueger:
Setting to Draft PR again because I will add the `-runbundles` required for Eclipse Welcome / Intro screen and add some entries for Bndtools there. see #5853

pkriens:
> If you can tell me where to get a better icon which is allowed to be used, please let me know. @pkriens

If you can find a better icon with open source use, just add it to the bndtools.core properties, see org.bndtools.core.ui.icons.Icons class. You could also scale them.

I thought we had a list of open source projects we used in bndtools but can't find it anymore. Maybe time to add it.

chrisrueger:
> I thought we had a list of open source projects we used in bndtools but can't find it anymore. Maybe time to add it.

Maybe you mean this in the main README?
image

chrisrueger:
I think I am done so far. Would be glad to get feedback. 
I updated the PR description with screenshots of the touched editors / views.

bndtools.core/src/bndtools/editor/common/Buttons.java
```diff
@@ -0,0 +1,81 @@
...
+import org.eclipse.jface.action.Action;
+import org.eclipse.jface.action.ActionContributionItem;
+import org.eclipse.jface.action.IAction;
+import org.eclipse.swt.program.Program;
+
+/**
+ * Helper containing help buttons for different parts / views which link to the
+ * user manual.
+ */
+public final class Buttons {
```
pkriens, line 13: Isn't this HelpButtons? You could call it Buttons but then imho it should contain all buttons and the functions should not be called *Help*

BTW, I like how you separate the data from the the code. However, if you make a function like createHelpButton, next time consider moving it in a utility in bndtools.utils package. Note that it is not related to Help, it is just a useful function to make an action. This makes this class smaller and more readable and might be used in other places.

* You could make this an interface so you do not have to clutter the class with 'public static final', that is implied in an interface.
* We're using Java 17 so you can use the """ comments to make the text easier to read

I am ok to keep it here, these changes are not so important but for future PRs.

pkriens (approved):
I am ok to merge

bndtools.core/src/bndtools/editor/common/Buttons.java
chrisrueger: Thanks @pkriens for the suggestions. During dev I always wonder how to call what and where to put what :)
I did the renaming to HelpButtons and moved the methods to EditorUtils (feel free to put it somewhere else if you like).

For the """ Textblocks: I tried but decided against it because the code formatter always changed the indentation which aftected display of the tooltips negatively. Maybe this is an item for another PR.

pkriens: About where to put what ... I generally have the rule of cohesion. Each aggregate (method,class,package,bundle) should have a clear focus. A class with a util (like the action methods) and the data is not cohesive. And whenever you can make a function reusable without effort, do it. It might never be reused but it will at least be easier to understand without the context of parts that are not relevant for that function.

Anyway, thanks for all the work!

PR #5859, opened 2023-10-31
Repositories View: MavenRepo: Add more copy to clipboard action menu entries on right click

chrisrueger:
This PR adds more Copy to Clipboard actions to a revision of a leaf entry of a Repository item:

image

For example for an entry of `com.fasterxml.jackson.core.jackson-databind` : 

## Add GAV to Clipboard

`com.fasterxml.jackson.core:jackson-databind:2.13.4` is added to clipboard

## Add tooltip to Clipboard

The content of this tooltip when you hover over a leaf-entry is added:

image

```
com.fasterxml.jackson.core.jackson-databind
General data-binding functionality for Jackson: works on core streaming API
com.fasterxml.jackson.core.jackson-databind
SHA-256: C9FAFF420D9E2C7E1E4711DBEEBEC2506A32C9942027211C5C293D8D87807EB6
Coordinates: com.fasterxml.jackson.core:jackson-databind:2.13.4
```

is added to the clipboard.

I found this helpful because the tooltip currently cannot be copy pasted and the tooltip contains useful information.

## Add Compile Dependencies to Clipboard

```
com.fasterxml.jackson.core:jackson-annotations:2.13.4
com.fasterxml.jackson.core:jackson-core:2.13.4
```

is added to the clipboard.
It is basically similar to **Add Compile Dependencies**

I wanted to add "Runtime Dependencies" too, but always got an empty result. Didn't have time to investigate. 
If you find it useful too, it's easy to add too.

## Feedback

Consider this PR just a suggestion. My main goal was to have **Add GAV to Clipboard** because this is handy when researching a dependency on google. The idea for the other two entries just came by playing around with it. 
We can remove them if they are too much. Ideally I would like to have a sub-menu "Copy to Clipboard" with those entries,  but didn't know how to do that.

I can also add testcases, when we know what to keep and what not.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/service/Actionable.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/service/Actionable.java
@@ -14,7 +14,22 @@ import java.util.Map;
 public interface Actionable {
 	/**
 	 * Return a map with command names (potentially localized) and a Runnable.
-	 * The caller can execute the caller at will.
+	 * The caller can execute the caller at will. Note: The command names can be
+	 * hierarchical using
+	 *
+	 * <pre>
+	 *  ::
+	 * </pre>
+	 *
+	 * (space, double-colon, space) as the delimiter. This can be used by
+	 * callers to build hierarchical menues. The last part of such a
+	 * hierarchical label can be used as the actual label e.g. of a button.
+	 * <h2>Examples</h2>
+	 * <ul>
+	 * <li>Label 1</li>
+	 * <li>Sub Menu :: Label 1 in Submenu</li>
+	 * <li>Sub Menu :: Label 2 in Submenu</li>
+	 * </ul>
 	 *
 	 * @param target the target object, null if commands for the encompassing
 	 *            entity is sought (e.g. the repo itself).
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
@@ -832,9 +832,7 @@ public class MavenBndRepository extends BaseRepository implements RepositoryPlug
 				return actions.getProgramActions((String) target[0]);
 			case 2 :
 				Archive archive = getArchive(target);
-				String bsn = (String) target[0];
-				Version version = (Version) target[1];
-				return actions.getRevisionActions(archive, bsn, version, registry.getPlugin(Clipboard.class));
+				return actions.getRevisionActions(archive, registry.getPlugin(Clipboard.class));
 			default :
 		}
 		return null;
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
@@ -68,6 +68,7 @@ import aQute.bnd.service.Refreshable;
 import aQute.bnd.service.Registry;
 import aQute.bnd.service.RegistryPlugin;
 import aQute.bnd.service.RepositoryPlugin;
+import aQute.bnd.service.clipboard.Clipboard;
 import aQute.bnd.service.maven.PomOptions;
 import aQute.bnd.service.maven.ToDependencyPom;
 import aQute.bnd.service.release.ReleaseBracketingPlugin;
@@ -831,7 +832,9 @@ public class MavenBndRepository extends BaseRepository implements RepositoryPlug
 				return actions.getProgramActions((String) target[0]);
 			case 2 :
 				Archive archive = getArchive(target);
-				return actions.getRevisionActions(archive);
+				String bsn = (String) target[0];
+				Version version = (Version) target[1];
+				return actions.getRevisionActions(archive, bsn, version, registry.getPlugin(Clipboard.class));
 			default :
 		}
 		return null;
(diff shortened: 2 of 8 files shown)


pkriens:
Great! love this work.

I agree it would be nice to have submenus but this would require a bigger change in the support. This code is not allowed to touch Eclipse code in any way because it is part of bndlib. We could make the menu names hierarchical but then we probably want to move the menu builder out of the RepositoriesView and put in some util. I'd say nice to have but there are more interesting things to do right now.

* Add copy: bsn;version=version ?
* The copy tooltip & bsn/version could go to the repositories view actions? That would benefit everybody

kriegfrj:
> Great! love this work.
> 
> I agree it would be nice to have submenus but this would require a bigger change in the support. This code is not allowed to touch Eclipse code in any way because it is part of bndlib. We could make the menu names hierarchical but then we probably want to move the menu builder out of the RepositoriesView and put in some util. I'd say nice to have but there are more interesting things to do right now.
> 
> * Add copy: bsn;version=version ?
> * The copy tooltip & bsn/version could go to the repositories view actions? That would benefit everybody

A possible alternative to submenus: a single menu item for `Copy...` which pops up a simple dialogue box to prompt for which details you want to copy?

chrisrueger:
> could go to the repositories view actions? That would benefit everybody

Do you mean, move the code for adding the menu entries e.g. here in `bndtools.views.repository.RepositoriesView.createContextMenu()`? @pkriens 

image

And remove it from `aQute.bnd.repository.maven.provider.RepoActions` again? 
Just want to clarify if I understand you correctly.

That way we would also have the ability to use Eclipse Code and do more fancy stuff, right?

chrisrueger:
Here is an updated version in https://github.com/bndtools/bnd/pull/5859/commits/1999d8fc4a411cd05c762c7ee5a5d3b03f6986bc where I managed to get a "Copy to clipboard" submenu working. 

The submenu has a **Generic part** and a **Repo-specific part**:

- **Generic part:** the  submenu no works on all _leaf_ nodes of the Respository view if the entry is _instanceof RepositoryBundleVersion_ which should also work for  repositories other than Maven
- **Repo-specific part:** Each repository can still contribute own entries to that Submenu if the label `startsWith("Copy")` (not sure if this hack is reasonable...) see `aQute.bnd.repository.maven.provider.RepoActions`

image


**1. Copy bsn+version**

`com.fasterxml.jackson.core.jackson-databind;version=2.13.4`

**2. Copy info ** 

`RepositoryBundleVersion [version=2.13.4, bundle=RepositoryBundle [repo=MavenBndRepository [localRepo=/Users/christoph/.m2/repository, storage=Maven Central, inited=true, redeploy=false], bsn=com.fasterxml.jackson.core.jackson-databind]]
`
**3. Copy tooltip**
```
com.fasterxml.jackson.core.jackson-databind
General data-binding functionality for Jackson: works on core streaming API
com.fasterxml.jackson.core.jackson-databind
SHA-256: C9FAFF420D9E2C7E1E4711DBEEBEC2506A32C9942027211C5C293D8D87807EB6
Coordinates: com.fasterxml.jackson.core:jackson-databind:2.13.4
```

**4. Copy GAV (only when it is a RepositoryBundleVersion from a MavenRepository )**

`com.fasterxml.jackson.core:jackson-databind:2.13.4`

**5. Copy  compile Dependencies**
```
com.fasterxml.jackson.core:jackson-annotations:2.13.4
com.fasterxml.jackson.core:jackson-core:2.13.4

```

pkriens:
Yes, you can change the Repositories View to your delight. The only thing that is paramount is that the bndlib code never touches Eclipse code ... Adjacent, the Repositories view must never know anything about the instance classes of the RepositoryPlugin interfaces. I.e. the Repository View is **not** allowed to cast the RepositoryPlugin to a BndRepositoryPlugin. The interface must fully go through the `actions()` method. Once the Repositories View starts to know about implementation details of the plugins it will quickly become a mess. i really like the Map<String,Runnable> interface we have now.

I'd be ok to do string matching on the label. We already to checkmarks etc. Making the labels hierarchical would be anice extension. However, if we go that route it would be nice to extract this actions -> menu handling in a separate class so we could potentially use it in other places. The `Actionable` interface is quite useful in other views.

chrisrueger:
Ok 

> Making the labels hierarchical would be anice extension. However, if we go that route it would be nice to extract this actions -> menu handling in a separate class so we could potentially use it in other places.

Ok I will give this some thought. 
A slash (`/`)  could be used for hierarchy. Do we want to support unlimited hierarchy layers? or just one additional (as in 1 submenu)?

**Label Example 1:** 
- Copy to Clipboard / Copy info
- Copy to Clipboard / Copy GAV
- First layer / Second Layer / Third layer

or 

**Label Example 2: (double colon as delimiter)** 
- Copy to Clipboard :: Copy info
- Copy to Clipboard :: Copy GAV
- First layer :: Second Layer :: Third layer

Maybe double colon (::) is less likely  to be used in the label text than slash? Thoughts?

chrisrueger:
@pkriens I think this is ready for review and feedback.
I added support for hierarchical menues using ` :: ` as the delimiter. 

This 

```
Copy to Clipboard :: Copy bsn;version
Copy to Clipboard :: Copy info
Copy to Clipboard :: Copy GAV
Clear from Cache
Delete from Index
```

would roughly contribute to this menu and submenu:

image

I also added the `Copy to Clipboard :: Copy Info` on all levels of RepositoryView now:
- Repo
- Bundle
- BundleRevision

Note:  I added a note about the hierarchy delimiter to the Javadoc of `Actionable.actions()`. If you don't like that , just let me know.

pkriens:
Are we ready to go?

chrisrueger:
> Are we ready to go?

If you have no objections, yes.

chrisrueger:
Ah wait... found something. Would fix that first. I give update @pkriens

pkriens:
@chrisrueger sorry, saw your second message too late.

pkriens:
I guess you have to make a new PR :-(

chrisrueger:
> I guess you have to make a new PR :-(

Yeah sorry, was a bit of last minute. 
Done https://github.com/bndtools/bnd/pull/5862

Issue #5833, opened 2023-10-21
Provide a Language Server for bnd instructions


laeubi:
The [Language Server protocol](https://langserver.org/) is used between a tool (_the client_) and a language smartness provider (_the server_) to integrate features like auto complete, go to definition, find all references and alike into the tool.

Like discussed on [EclipseCon2023](https://www.eclipsecon.org/2023) BND instructions seem very well suited to be provided as a LSP so it can be used in an IDE agnostic way.

For Eclipse there is https://github.com/eclipse/lsp4e to easily integrate such servers.

FYI @maho7791 @juergen-albert @stbischof  ...

@mickaelistria can you give some advice here how to best start using [lsp4j](https://github.com/eclipse-lsp4j/lsp4j)?

mickaelistria:
I think the current documentation in LSP4J ( https://github.com/eclipse-lsp4j/lsp4j/blob/main/documentation/README.md ) already gives good advice. I suggest you get started from there, and if you have further questions, you may try opening issues against LSP4J to ask for details.

pkriens:
I did some work in this a couple of years ago. I maintain a LSP in the Alloys projects and it was easier than I thought. Although I was highly disappointed how many functions still required client side code.

There is a Red Hat project that uses LSP for Gradle + Maven. Seeing it was OSGi based I thought it would be easy to integrate bnd support there. I think this would still be a very valid approach but the code base, based on Maven, did not build out of the box and I could not get it to work quickly. They do not have some nice extension mechanism like services that I could see. However, combining effort with this would imho be the best approach.

If bnd(tools) has any future, then I think it would be in this direction. I am interested in pursuing this further if some other people are interested.

laeubi:
@pkriens I think a standalone LanguageServer would be sufficient here, I would expect it to have similar features like current BNDEditor but still things can be made on top of this on clientside so to summarize:

- Autocompletion works for all known bnd instructions
- Syntax highlight
- Some error checking e.g. invalid values used for an instruction

maho7791:
Hi Peter, thanks for this hint. We currently trying different approaches with the LSP and wanna start something within next year. Guido is already trying to get firm with all that stuff. We are really interested to get something work in that direction

laeubi:
> I think the current documentation in LSP4J (https://github.com/eclipse-lsp4j/lsp4j/blob/main/documentation/README.md ) already gives good advice.

[...]

PR #5861, opened 2023-11-01
[bnd] Language Server for bnd

pkriens:
This is a branch for the language server


laeubi:
See
- https://github.com/bndtools/bnd/issues/5833

as a first step one would need the following compile dependencies

```
<dependency>
    <groupId>org.eclipse.lsp4j</groupId>
    <artifactId>org.eclipse.lsp4j</artifactId>
    <version>0.21.1</version>
</dependency>
<dependency>
    <groupId>org.eclipse.lsp4j</groupId>
    <artifactId>org.eclipse.lsp4j.jsonrpc</artifactId>
    <version>0.21.1</version>
</dependency>
```

@pkriens as I'm not familiar with bnd project structure, can you add these so they are available in that project you created? I think there is no IP issues to consider as these are EPL2.0 already?
I then can migrate my current sever side attempts here:

- https://github.com/eclipse-pde/eclipse.pde/pull/821

Integration with bndtools can then be part of

- https://github.com/bndtools/bnd/issues/5844

pkriens:
done.

pkriens:
I get the vibe that this is not achieving a lot of love ... closing it until we get activitiy. 

I'll keep the branch alive.

PR #5862, opened 2023-11-03
Fix empty bundle info

chrisrueger:
Fixes a small bug just found after https://github.com/bndtools/bnd/pull/5859 was merged.
It fixes the problem, that copy to clipboard on a RepoBundle did not copy anything to the clipboard.

image

@pkriens could you merged that too please?

### DIFF
--- a/bndtools.core/src/bndtools/views/repository/RepositoriesView.java
+++ b/bndtools.core/src/bndtools/views/repository/RepositoriesView.java
@@ -1161,6 +1161,13 @@ public class RepositoriesView extends ViewPart implements RepositoriesViewRefres
 						info.append(tooltipContent);
 						clipboard.copy(info.toString());
 					}
+			else {
+						// bundle does not seem to have a tooltip
+						// let's just add general bundle info
+						info.append(rb.toString());
+						clipboard.copy(info.toString());
+					}
+
 				} catch (Exception e) {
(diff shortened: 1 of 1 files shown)


pkriens:
I'll merge it once it finishes ok

chrisrueger:
Thanks for merging. So just let me know if you notice something which needs improvement / fixing.

pkriens:
You're doing some really nice work! Looking for something a tiny bit bigger? :-) 

One UI thing that I think would be really useful is to make the Resolve Window detached. It is currently modal and that is a real pita. First easy step could be to disconnect the view with the final resolution so you can compare different resolutions.

No pressure though, love what you're doing here.

chrisrueger:
> a tiny bit bigger? :-)
> 
> One UI thing that I think would be really useful is to make the Resolve Window detached. It is currently modal and that is a real pita. First easy step could be to disconnect the view with the final resolution so you can compare different resolutions.

Haha :) those were my first steps in Eclipse GUI development ever... (comming from a web app background) So I was glad the topics were small ;) But once I found the right place for my breakpoint it got better. 
Regarding your proposal: I try to get an understanding. My problem is that I am not yet familiar with all the GUI terms, patterns, classes and do's and don't. But let's see. I'll open a PR once I have something.

pkriens:
The menu entries work like a charm!

PR #5863, opened 2023-11-03
make Resolve Window non-modal

chrisrueger:
now you can click on Resolve multiple times which opens a new window everytime, so you can e.g. compare different results

Based on comment in https://github.com/bndtools/bnd/pull/5862#issuecomment-1792268413

### DIFF
--- a/bndtools.core/src/bndtools/editor/BndEditor.java
+++ b/bndtools.core/src/bndtools/editor/BndEditor.java
@@ -386,6 +386,11 @@ public class BndEditor extends ExtendedFormEditor implements IResourceChangeList
 									return button;
 								}
 							};
+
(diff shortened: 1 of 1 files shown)


chrisrueger:
@pkriens Not sure if this is what you had in mind in https://github.com/bndtools/bnd/pull/5862#issuecomment-1792268413

image

pkriens:
Just gonna try it, looks wonderful. that seems easier than I thought :-)

pkriens:
I think you need to probably look what happens if you include the optional requirements. This will I think edit the BndEditModel but this can of course be edited from the main editor window as well now.

You need to make sure that if it can only be updated from one window. There is properties support in this class but optimistic locking is fine with me as well. If it has changed, you refuse to update.

PR #5865, opened 2023-11-05
Conflict handling when having multiple Resolve-results open and changes happen in resolve window(s) vs main window

chrisrueger:
Based on https://github.com/bndtools/bnd/pull/5863#issuecomment-1792931273

> I think you need to probably look what happens if you include the optional requirements. This will I think edit the BndEditModel but this can of course be edited from the main editor window as well now.
> 
> You need to make sure that if it can only be updated from one window. There is properties support in this class but optimistic locking is fine with me as well. If it has changed, you refuse to update.

## Preview

When any other UI part / view changes something to the underlying BndEditModel (e.g. in the main window I edit and save the `.bndrun` file or I update optional requirements in one of the resolve windows then this change is detected and further updates fail.

image

The same behavior applies when clicking the "Update" Button when the model has changed.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
@@ -14,6 +14,7 @@ import java.io.StringReader;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


pkriens:
Can I merge?

chrisrueger:
> Can I merge?

Yes.

PR #5875, opened 2023-11-08
[tester ws] Java 17 tidy-up

kriegfrj:
A long-overdue tidy up of the workspace used for testing the tester functionality, including making it work now with Java 17 as the default EE.


pkriens:
Thanks!

PR #5882, opened 2023-11-12
Add indirection to fetch the Workspace instead of using static singleton

laeubi:
Currently there is a lot of code that is bound to the "Central" class what has several problems when it comes to reuse this outside of bndtools.

This adds two ways of getting out of this reference chain for components that should be reusable:

1) From a project context use the Eclipse Adapter Pattern to adapt an IProject > Workspace
2) For cases where a global workspace model is desired and no project context exits, the "central" Workspace is provided as a service in the runntime.

### DIFF
--- a/bndtools.core/src/bndtools/Activator.java
+++ b/bndtools.core/src/bndtools/Activator.java
@@ -12,14 +12,18 @@ import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.resource.ImageDescriptor;
 import org.eclipse.swt.widgets.Display;
 import org.eclipse.ui.plugin.AbstractUIPlugin;
+import org.osgi.framework.Bundle;
 import org.osgi.framework.BundleContext;
+import org.osgi.framework.ServiceFactory;
 import org.osgi.framework.ServiceRegistration;
 import org.osgi.service.url.URLConstants;
 import org.osgi.service.url.URLStreamHandlerService;
 
 import aQute.bnd.build.ReflectAction;
+import aQute.bnd.build.Workspace;
 import aQute.bnd.osgi.Processor;
 import aQute.bnd.service.action.Action;
+import bndtools.central.Central;
 
 /**
  * The activator class controls the plug-in life cycle
@@ -33,6 +37,8 @@ public class Activator extends AbstractUIPlugin {
 	public static volatile Activator						instance;
 	BundleContext											context;
 	private ServiceRegistration<URLStreamHandlerService>	dataUrlHandlerReg;
+
+	private ServiceRegistration<Workspace>					workspaceRegistration;
 
 	/*
 	 * (non-Javadoc)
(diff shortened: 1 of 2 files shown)


laeubi:
It seems bndtools already will (maybe) register a service .. so I think that's good enough as an contract (for me) to build on...

kriegfrj:
@laeubi, you might be interested in #5036. I'm hoping to get it to a point where I can merge it this week.

laeubi:
@kriegfrj this looks very interesting/helpful! But you should really consider integrating your ResourceChangeWhiteboard into regular eclipse platform so more people can use it, we recently made some efforts in that area already.

kriegfrj:
> @kriegfrj this looks very interesting/helpful! But you should really consider integrating your ResourceChangeWhiteboard into regular eclipse platform so more people can use it, we recently made some efforts in that area already.

I agree that core (somewhere) would be a better place for the `ResourceChangeWhiteboard`. I simply don't have the spare development capacity to familiarise myself with Eclipse project/building to do this. I would be grateful if you (or someone else in the core team) could do it - it would be a simple job for someone already familiar with building/deploying/testing the Resources plugin.

laeubi:
@kriegfrj while it could potentially work it has some drawbacks:

1. due to legal reasons pushing code not written by me requires a quite lengthy process to get approval
2. If comments or enhancements are arise one needs to address them indirectly with back and forth and even mix foreign and own contributions (see 1)
3. Next time you might again struggle to find someone to get the code to the "right" place

So I really like to encourage you to just give it a try, things have greatly been simplified over the past moths with the move to github and alike so here is a quick introduction:

- Download the Eclipse Installer and setup an development IDE, it takes some time but is all automatic: https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#creating-an-eclipse-development-environment
- Setup and configure your Eclipse account: https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#setting-up-your-eclipse-and-github-account
- Now fork the repository https://github.com/eclipse-platform/eclipse.platform and add it as a new remote to your above checkout
- Create a new branch, add some code (at best show the usage on one example in the platform code as well) and push it to your fork, then open a PR and we will see how we can get this in.

If you have any problems me and also @HannesWell and other committers are there to help you on your way for your fist Eclipse contribution and you'll see how things getting easier after the first few steps.

kriegfrj:
> @kriegfrj while it could potentially work it has some drawbacks:
> 
> 1. due to legal reasons pushing code not written by me requires a quite lengthy process to get approval

Bndtools is already under Eclipse 2.0 license, so this shouldn't be too hard?

> 2. If comments or enhancements are arise one needs to address them indirectly with back and forth and even mix foreign and own contributions (see 1)
> 3. Next time you might again struggle to find someone to get the code to the "right" place
> 

These points I sympathise with.

> If you have any problems me and also @HannesWell and other committers are there to help you on your way for your fist Eclipse contribution and you'll see how things getting easier after the first few steps.

Thank you so much for the offer. I'll see what I can do.

> Alternatively we can put this in a separate bundle so it is easy to share between the environments. I think bndtools.api could be used for this purpose since it is not allowed to touch the rest of the bndtools code, it just provides base Eclipse - bnd functions. If we need another bundle for pure non-bnd Eclipse functions, then lets make it. I think we should at least finish this work here.

I like this idea. The other problem is if I go the Eclipse route, then my other changes that depend on it won't be able to be released until that version of Eclipse gets released *and* we up the baseline for Bndtools. This approach gives us a way to do a proof-of-concept and then migrate to core Eclipse once we're happy.

PR #5883, opened 2023-11-12
Fix typo error message, add help to Resolution Wizard, Method renaming

chrisrueger:
A few tweaks. See commit messages for details.
- added the word "not" in error message
- the ResolutionWizzard had a Help-Button (lower left) which was without function. I think it was there by default ,but nobody added functionality. I now linked to the manual (Resolving section)
- give two button methods from earlier  more appropriate names

Can be merged if no objections.


pkriens:
Why did you rename the createButton to createHelpButton? there is nothing 'helpy' about these buttons?

chrisrueger:
> Why did you rename the createButton to createHelpButton? there is nothing 'helpy' about these buttons?

It is. The buttons have a hard coded help icon.

PR #5887, opened 2023-11-13
Remove migrated pdetools.pde project

laeubi:
This removes the code migrated to PDE here:
- https://github.com/eclipse-pde/eclipse.pde/pull/919

once the PDE part is merged this can be merged as features from this project will then be provided by PDE directly


laeubi:
I could need some help how to make the build happy ...

pkriens:
It craters at building the p2 repository. My assumption is that you did not remove the bndtools.pde feature that lists this as bundle/plugin? You can run gradle in the bndtools.p2 project with the p2 target to see if this is the case. The output is probably somewhere uploaded but this might be faster.

Which raises a question. I'd prefer if you deprecate these files/project/feature. I'd like to wait one or two releases before removing anything permanently. I haven't looked into details but do we handle the case some people have an older Eclipse/PDE installed when installing bndtools 7.1.0? I do not want to force them to upgrade.

PR #5889, opened 2023-11-14
Add first experiments showing basic autocompletion and token support

laeubi:
These are my first experiments/attempts of a Bnd Language server implementation migrated to bnd see

- https://github.com/eclipse-pde/eclipse.pde/pull/821

for client side, this would require some adjustment to bndtools source editor but might better be  a separate PR...


pkriens:
We never use the copyright headers since they are kind of annoying at the top and we seem to have been able to convince even IBM they were needed. They also need to be policed because they could say something different than our license we maintain at the top.

We generally also do not put our names in source files. Almost all code becomes touched by others or is copied from others. We have the log in git, with the signed commits, so I do not see the use of it and prefer not to start this imho awkward convention.

laeubi:
Headers removed...

pkriens:
You want me to automatically merge these?

laeubi:
Would be good so others can build on top of it.

pkriens:
Yes, but it would become part of https://github.com/bndtools/bnd/pull/5861

We now have two drafts for the same purpose.

laeubi:
I don't understand this targets the branch you mentioned so it will be merged there, and then this one will vanish so whats wrong with that?

laeubi:
> We now have two PRs for the same work. This is confusing.

No, my PR is extending your PR ...

> So if we can merge yours, we are back to 1 PR where we can have discussions.

That's how feature branches work, and you wanted a feature branch for that ... I can't merge it as I'm not a repro owner and is is usually good to review changes against the feature branch if working with different people on the same feature so I'm not understand why/what should be made different here?

pkriens:
No, my PR compares the branch against master. So I want it to continuously follow. I want your changes in that PR and then your PR closed.

First time I tried this but I want to keep the list of PRs as small as possible. And this is a working branch so there is no use to keep a draft. That is why I want to merge it to the branch.

laeubi:
I set it to "ready" all checks are green, but its up to you to press the merge button:

> Only those with [write access](https://docs.github.com/articles/what-are-the-different-access-permissions) to this repository can merge pull requests.

PR #5894, opened 2023-11-15
open jar on dblclick on AdvancedSearch entry

chrisrueger:
* before this change: double clicking on a result entry  of an Advanced search (e.g. by package name) nothing happened
* after this change: the jar of the bundle opens. this is now same behavior as double clicking on a bundleVersion in the repo browser without advanced search filtering

## Testing

* filter the RepoBrowser e.g. by a specific package name
* then double click on one of the search results
* Expected Result: The Jar Viewer should open this entry. 

image

image

### DIFF
--- a/bndtools.core/src/bndtools/model/repo/RepositoryResourceElement.java
+++ b/bndtools.core/src/bndtools/model/repo/RepositoryResourceElement.java
@@ -3,12 +3,13 @@ package bndtools.model.repo;
 import java.util.Objects;
 
 import org.bndtools.utils.resources.ResourceUtils;
+import org.eclipse.core.runtime.IAdaptable;
 import org.osgi.framework.Version;
 import org.osgi.resource.Resource;
 
 import aQute.bnd.service.RepositoryPlugin;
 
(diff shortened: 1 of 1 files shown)


bndtools.core/src/bndtools/model/repo/RepositoryResourceElement.java
```diff
@@ -41,6 +42,20 @@ public Resource getResource() {
 		return resource;
 	}
 
+	@SuppressWarnings("unchecked")
+	@Override
+	public <T> T getAdapter(Class<T> adapter) {
+
+		// this is triggered e.g. when you doubleclick on an
```
pkriens, line 49: In general we do not provide this comment. This is about a completely different part of the code and things might change in the future that invalidates this comment. The IAdaptable is the contract and that is documented in the code. It is fine to add this extra information in the commit.

I admit of not commenting public APIs in our code but not commenting non-public API is an explicit strategy in bnd.

pkriens: no hurry

PR #5904, opened 2023-11-27
Prefer the minus instruction over the header like variant

laeubi:
Currently if one creates a model from scratch the BndEditModel prefer the header-like one over the minus instruction.

To not confuse users with the old variant, this one give precedence over the header-like and also deprecate the old header like variant.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
@@ -1214,24 +1214,24 @@ public class BndEditModel {
 			setEntries(resourceEntries2, Constants.INCLUDE_RESOURCE);
 		}
 
-		if (hasIncludeResourceInstruction()) {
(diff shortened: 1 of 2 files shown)


laeubi:
@pkriens sorry for the inconvenience and thanks for the fixup!

I must confess I have a bit of a hard time with Bndtools what sometimes makes it hard to to be finally sure everything is fine:

1. I get a lot of dirty changes in Git staging view sometime (sometimes not), it also show some new file
2. BndExplorer shows "green" even if I see build error marker on a project, sometimes it is read for now obvious reason
3. Sometimes it takes very (very very) long after changing something quite minimal and it seems there is some action in progress but I can't see it (or its progress)

So I somewhat hoped that even if I do not catch all issues right now in the IDE the build will complain (what seems not to be the case).

Is there a recommended workflow to "clean" the BndWorkspace (e.g. should one hit "reaload workspace" once in a while for that purpose?).

PR #5907, opened 2023-11-28
Improve code templates 

chrisrueger:
Related to https://github.com/bndtools/bndtools.workspace.min/pull/11#issuecomment-1830576497 

It tries to address (some) of the suggestions by @pkriens from [here](https://github.com/bndtools/bndtools.workspace.min/pull/11#issuecomment-1830270542):

> If we update the templates (and we should) some thinks to consider:
> 
> - Specify the framework without version, let bnd pick one

addressed in this PR

> - You might want to make the bndrun template use cache. This works great in general and saves lots of problems with git merge.

@pkriens do you have some pointer what that means? Do you mean https://bnd.discourse.group/t/the-resolve-cache-mode/302 ?

> - No versions, use default version and let the repositories control the versions avaialble

addressed in this PR. Let me know if I forgot something

> - Remove Bundle-Version everywhere, should be defined in build.bnd

addressed in this PR

> - No specification of java anywhere

addressed in this PR

> - No Export-Package or Private-Package headers, defaults should be used.

TODO

> - Need to update our Junit storyhere. Probably need to work with @kriegfr

TODO 

> - remove the enroute stuff

TODO 

> - Might throw out some

TODO

PR #5915, opened 2023-12-01
Eclipse hack boot delegation

### LINKED ISSUE #5911: File / New / Bnd Osgi Workspace: NoClassDefFoundException javax/net/ssl/SSLContext in GitCloneTemplate.generateOutputs() in Git git = cloneCmd.call();
Eclipse SDK
Version: 2022-09 (4.25)
Build id: I20220831-1800
Java 17.0.9

while working on the templates today I discovered an error when I wanted to create a Bnd Workspace via a workspace template on github.

1. File / New / Bnd Osgi Workspace

image

2. Choose one of the existing ones (also my own didn't work)

image

3. Error on next page

image

The error **only** happened in the Eclipse instance I started from the Eclipse instance containing the bndtools workspace.

It works **fine** when I do it with my normal Eclipse where I am developing for bndtools (Eclipse 2023-09)

## Debugging

- I happens in `org.bndtools.templating.jgit.GitCloneTemplate.generateOutputs(Map<String, List<Object>>, IProgressMonitor)`

image

image

## What did I want to do

I actually wanted to test https://github.com/bndtools/bnd/pull/5907 with the update template in
https://github.com/bndtools/bndtools.workspace.min/pull/11 and https://github.com/bndtools/workspace/pull/24

pkriens:
Eclipse seems to use boot delegation for javax.*. Since we're not doing this in our debug setup, we get errors. This patch just does the same brain damaged thing
Eclipse is doing to mitigate even brain deader
bundles. Yes, I am frustrated, bnd fixed this problem 20 years ago!
Fixes #5911

---

### DIFF
--- a/bndtools.core/bndtools.shared.bndrun
+++ b/bndtools.core/bndtools.shared.bndrun
@@ -22,7 +22,7 @@
 	eclipse.product=org.eclipse.sdk.ide,\
 	osgi.console=,\
 	osgi.instance.area.default=../bndtools.test/workspace,\
-	org.osgi.framework.bootdelegation='javax.net,javax.management,org.xml.sax,org.xml.sax.helpers,javax.xml,javax.xml.parsers',\
+	org.osgi.framework.bootdelegation='javax.*,org.xml.*',\
 	logback.configurationFile=${fileuri;${.}/logback.xml}
 
 # Keep sorted so that we can diff


chrisrueger:
Great, thanks. I can confirm https://github.com/bndtools/bnd/issues/5911 is fixed.
Also I notice that now the internal code templates of https://github.com/bndtools/bnd/pull/5907 (org.bndtools.templates.osgi) show up in the debug instace. I think this was not the case before.

PR #5918, opened 2023-12-02
Add a MakeJar to create a plain-old-jar file as a resource

laeubi:
Currently one can already create a new bundle using additional bnd instructions file but this has some caveats when one only wants to package some stuff into a jar and also some processing overhead.

This adds a new MakeJar that do what the name suggest and create a dumb jar from an input directory.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/make/MakeJar.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/make/MakeJar.java
@@ -1,0 +1,61 @@
+package aQute.bnd.make;
+
+import java.io.ByteArrayOutputStream;
+import java.io.File;
+import java.util.Map;
+import java.util.jar.Attributes;
+import java.util.jar.Manifest;
+
+import aQute.bnd.osgi.About;
+import aQute.bnd.osgi.AbstractResource;
+import aQute.bnd.osgi.Builder;
+import aQute.bnd.osgi.Constants;
+import aQute.bnd.osgi.Jar;
+import aQute.bnd.osgi.Resource;
+import aQute.bnd.service.MakePlugin;
+
+public class MakeJar implements MakePlugin {
+
+	@Override
+	public Resource make(Builder builder, String destination, Map<String, String> argumentsOnMake) throws Exception {
+		String type = argumentsOnMake.get("type"); //$NON-NLS-1$
+		if (!"jar".equals(type)) { //$NON-NLS-1$
+			return null;
+		}
+		String input = argumentsOnMake.get("input"); //$NON-NLS-1$
+		if (input == null) {
+			builder.error("No input specified on a make instruction for %s, args=%s", destination, argumentsOnMake); //$NON-NLS-1$
+			return null;
+		}
+		File folder = builder.getFile(input);
+		if (!folder.isDirectory()) {
+			return null;
+		}
+		boolean noExtra = Boolean.parseBoolean(argumentsOnMake.get("noExtra")); //$NON-NLS-1$
+		boolean reproducible = Boolean.parseBoolean(argumentsOnMake.get("reproducible")); //$NON-NLS-1$
+		return new AbstractResource(System.currentTimeMillis()) {
+
+			@Override
+			protected byte[] getBytes() throws Exception {
+				try (Jar jar = new Jar(folder)) {
+					Manifest manifest = new Manifest();
+					Attributes mainAttributes = manifest.getMainAttributes();
+					mainAttributes.put(Attributes.Name.MANIFEST_VERSION, "1.0"); //$NON-NLS-1$
+					if (!noExtra) {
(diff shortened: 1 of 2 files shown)


laeubi:
> A Jar is created ahead of time. We generally try to delay the work until we actually write the output.

Good point, even though the `JarResource` seems to be made for this, what do you suggesting instead? As Bnd at this points really want the resource because it is missing I assume it can't delay much here anyways.

> It is also never closed. This is not a big deal in this case because it is a directory but a Jar is an AutoCloseable. It is inside knowledge that I know that a directory is ok, on a Windows a Jar on a file locks it up.

What files should it look up? It does only operate on memory resources here as far as I can see.

> modify MakeCopy to accept directories

I'm not sure I understand the `MakeCopy` to propose something like that, also not I don't want to copy that folder into the (bundle) jar, I want it to pack the contents into a new jar and then put it into the bundle-jar, not sure if MakeCopy really do that?

pkriens:
> What files should it look up? It does only operate on memory resources here as far as I can see.

As I said, a folder is ok. However, a zip file is opened as a stream and the entries are in the resource. However, these are internal details you should never rely on. Jar declares AutoCloseable so it *must* be life cycle managed.

MakeCopy does not handle folders because it is logically impossible to expand the structure in the jar, it can only provide a resource == single entry in the jar.

Currently a folder is invalid and will likely give weird errors since it tries to turn it into a URL. So we can safely turn a folder into a jar/zip entry/Resource. In bnd, that does not sound illogical. Then we could:

* Use JarResource (and close the Jar)
* Use FileResource (needs a small change but if done well, I think it makes a lot of sense). 
* Make a special FolderResource

The simplest solution is probably a new FolderResource. I am not fond of the JarResource since Jars are big objects and prefer to have it created when needed. It is generally used when you already have a Jar at hand. We could use FileResource but it is one of the workhorses and BJ added quite a bit of complexity do handle caching and buffers. So a new FolderResource might be the simplest solution.

laeubi:
> As I said, a folder is ok. However, a zip file is opened as a stream and the entries are in the resource. 

But that the point the "Jar" is a class it is not a file so it can't have anything opened as it is only a structure in memory ... and if you look at close it closes the zipfile (what is null as the jar is not created from a zip) and it closes all resources , what is a noop as well as we only have fileresources that have a pointer to the file but not the file content and then throws away all added resources so after that it is completely empty so can't be used as a jar anymore... In contrast the JarResource can close the jar.

Nerveless I can change that without a problem, but if you are so concerned you might be interested to fix `aQute.bnd.make.MakeBnd` what EXACTLY do use the same pattern with a Jar+JarResource, guess how I found about it ;-)

> Currently a folder is invalid and will likely give weird errors since it tries to turn it into a URL. So we can safely turn a folder into a jar/zip entry/Resource. In bnd, that does not sound illogical

What I don't understand is what a "FolderResource" should be transformed to a byte stream (what a resource actually is) without specify if it should be jar, gzip, tar, ...

So I still think we probably talk about different things here but not sure... a folder can already be added to a jar as individual Resources.

pkriens:
> But that the point the "Jar" is a class it is not a file so it can't have anything opened as it is only a structure in memory ... 
See ZipResource and Jar.buildFromZip.

we spent months trying to fix issues with this because it blocked builds on Windows. As I said, a folder would be ok *but you're not allowed to assume that ever*. that is internal knowledge. Follow the API! It says it is an AutoCloseable so you **must** close it.

> Nerveless I can change that without a problem, but if you are so concerned you might be interested to fix aQute.bnd.make.MakeBnd what EXACTLY do use the same pattern with a Jar+JarResource, guess how I found about it ;-)

Not really. See line 36 of MakeBnd:

https://github.com/bndtools/bnd/blob/58b881801bcf9e3c0d9d53f4c90e46cb416ed6e6/biz.aQute.bndlib/src/aQute/bnd/make/MakeBnd.java#L36

> What I don't understand is what a "FolderResource" should be transformed to a byte stream (what a resource actually is) without specify if it should be jar, gzip, tar, ...

As I said, I think it is logical in bnd to make it a zip. But be my guest to support more formats ... The make copy gets the information to look at the extension of the file name ...

You're costing me way more time then anybody else. The common feeling I get is that you keep second guessing me. It might help is both if you assume I am right in OSGi and bnd issues.

laeubi:
> It might help is both if you assume I am right in OSGi and bnd issues.

I try to not assume anything beforehand, nevertheless if your the only one that is always right its a bit annoying to ask people to open PRs instead of issues just to tell them you don't like the effort put into PRs.... maybe it could also help to not assume everyone else is a complete idiot regarding OSGi + bnd.

Anyways I'm now made the resource completely lazy and `Autoclose` the jar as soon as possible... and I don't do this to annoy anyone but because I think it might be useful for others.

pkriens:
I've just committed a change, you can now use 
```
-includeresource foo.jar=@some/directory/
```

PR #5928, opened 2023-12-07
Resolution view: allow reqs/caps filtering on the tooltip text

chrisrueger:
this allows more fine grained Glob filtering e.g. on the resource name e.g. using "FROM:*mypackage" which allows to filter reqs and caps based on their underlying resources

## Example

In the example below I have selected
- a bundle
- and the complete Maven Central repo in the repo browser
- the requirements I have filtered so that I only see requirements coming from my bundle which contains the letters "obs" in the resource name

image

This is very handy in the use case where I am particularly interested in e.g. requirements of a particular resource. 
E.g. in my case I wanted to find out all reqs of my bundle which are not satisfied by Maven Central so that I can drag them into .bnd Editor "Customize Imports" and mark them as _optional_

### DIFF
--- a/bndtools.core/src/bndtools/views/resolution/ResolutionView.java
+++ b/bndtools.core/src/bndtools/views/resolution/ResolutionView.java
@@ -36,6 +36,7 @@ import org.eclipse.jface.action.Action;
 import org.eclipse.jface.action.IAction;
 import org.eclipse.jface.action.IToolBarManager;
 import org.eclipse.jface.action.MenuManager;
+import org.eclipse.jface.action.Separator;
 import org.eclipse.jface.dialogs.ErrorDialog;
 import org.eclipse.jface.resource.ImageDescriptor;
 import org.eclipse.jface.util.LocalSelectionTransfer;
@@ -128,7 +129,11 @@ public class ResolutionView extends ViewPart implements ISelectionListener, IRes
 	private Label				reqsLabel;
 	private Label				capsLabel;
 
+	// hideSelfImportsFilter => show only resolved requirements (strange name)
 	private ViewerFilter		hideSelfImportsFilter;
+
+	// the oposite of the hideSelfImportsFilter filter
+	private ViewerFilter				hideOptionalRequirements;
 	private ViewerFilter		filterShowCapsProblems;
 
 	private boolean				inputLocked	= false;
@@ -147,6 +152,7 @@ public class ResolutionView extends ViewPart implements ISelectionListener, IRes
 	private final FilterPanelPart		capsFilterPart								= new FilterPanelPart(
 		Plugin.getDefault()
 			.getScheduler());
+	private static final String			SEARCHSTRING_HINT							= "Enter search string (Space to separate terms; '*' for partial matches)";
 
 	private CapReqMapContentProvider	reqsContentProvider;
 	private CapReqMapContentProvider	capsContentProvider;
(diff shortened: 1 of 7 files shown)


pkriens:
In BndtoolsExplorer we first added the concept of filter modifiers, like your `FROM:` here. I think we need to be consistent here. I am ok to move them to constants somewhere in a public API package. However, foremost we need to be consistent, which would make it `:from`.
https://github.com/bndtools/bnd/blob/4d1c6b08957f853691d906113414d657db91e6ea/bndtools.core/src/bndtools/explorer/BndtoolsExplorer.java#L510-L529

pkriens:
And notice that they can be abbreviated. I.e. in BndtoolsExplorer, `:e` works as well

pkriens:
And last but not least, you need to sign off the commits. (You can set that to be done automatically)

chrisrueger:
**Latest status:**

We now have:
- a new filter button to hide optional requirements
- improved placeholder text in search inputs indicating wildcards and spaces


image

With this I set it ready for review and if no objects, ready to merge.

bndtools.core/src/bndtools/views/resolution/ResolutionView.java
```diff
@@ -128,7 +129,11 @@ public class ResolutionView extends ViewPart implements ISelectionListener, IRes
 	private Label				reqsLabel;
 	private Label				capsLabel;
 
+	// hideSelfImportsFilter => show only resolved requirements (strange name)
```
pkriens, line 132: Please no such comments. Feel free to change it for a better name but these comments tend to stay for millennia ...

Base idea is to not comment code, only API. If you want to make things more clear, use better names and/or create intermediate variables that show what you want to convey.

This is an old code base, started far before I fully developed all those ideas, and also not with the intention to be a public API, so you will see many violations.

bndtools.core/src/bndtools/views/resolution/ResolutionView.java
```diff
@@ -128,7 +129,11 @@ public class ResolutionView extends ViewPart implements ISelectionListener, IRes
...
 	private ViewerFilter		hideSelfImportsFilter;
+
+	// the oposite of the hideSelfImportsFilter filter
```
pkriens, line 135: Same

pkriens (commented):
I am going to merge to speed this up, I am curious :-)

You can remove those comments in another PR you;re working on. No hurry.

bndtools.core/src/bndtools/views/resolution/ResolutionView.java
chrisrueger: > Please no such comments. Feel free to change it for a better name but these comments tend to stay for millennia

Ok sorry, I will create another PR before I forget it. 
While we are on this: Could you explain me the reasoning behind the current name "hideSelfImportsFilter"?
Because I was a bit confused by this, as the tooltip of the button said "Show resolved". 
So I would rename it from  `hideSelfImportsFilter` to `hideUnresolved` (in case we keep the negated naming).
Or would that be wrong?

bndtools.core/src/bndtools/views/resolution/ResolutionView.java
chrisrueger: ok

chrisrueger:
Ok thanks. No problem. I left the variable name.

Issue #5632, opened 2023-04-10
Enhance wizard-based creation of OSGi services


scottslewis:
As per discussion here

https://bnd.discourse.group/t/osgi-services-creation-wizards/218/12

It would be useful to for bndtools to support wizard-based creation of OSGi services.   

As an initial contribution, I've created a new package/API:   org.bndtools.core.ui.wizards.service  for building wizards for creating OSGi services with (by default) three projects:  a service-api project, a service-impl project, and service-consumer project.    These wizard classes allow the use of named bndtools templates  for creating the service-api, service-impl, and service-consumer projects via a single wizard.

Further, I've created a Java OSGi Service wizard in new package:  org.bndtools.core.ui.wizards.service.java and referenced in _plugin.xml.   This uses the new package/API above and 3 built-in templates (in new bundtools.core/resources/unprocessed/serviceTemplates) in bndtools.core.

scottslewis:
Please see draft pull request #5633

scottslewis:
Please see new pull request #5643  for simplification of new service wizard support

scottslewis:
@kriegfrj  I would appreciate your merging of the pr in bndtools.workspace repo here

https://github.com/bndtools/workspace/pull/18

All it does is add the bundle annotation bundle to the workspace template...so that the java wizard contribution (associated with this issue) will work as expected by users when using the bndtools.workspace workspace template.

scottslewis:
@kriegfrj @pkriens 

I've completed the technical work to close this issue except for two things:

1) The modification of the bndtools workspace to include the org.osgi.annotation.bundle (so that if the new java osgi service wizard is used, the wizard-create classes will all compile...since the @Version annotation is used by the wizard-created code).   The pull request for this is change is here https://github.com/bndtools/workspace/pull/18   (in the bndtools/workspace repo).

2) I think it would be useful to add this same org.osgi.annotation.bundle (from maven central) to the workspace.minimal workspace template.    However, currently this minimal template has absolutely nothing in it...meaning that I would have to add maven central repo and the inclusion of the org.osgi.annotation.bundle (as with the above workspace).   This seems still pretty minimal (I don't see how anyone could actually use the current minimal workspace template for any real project)....but I leave this to the committers to decide.   If you will approve a pull request for the minimal workspace template to add the necessary annotation classes then please let me know and I'll submit one.

[...]

PR #5932, opened 2023-12-11
Fix for workspace error/refresh issue in Java OSGi Service wizard

scottslewis:
Previous to this change, the use of the Java OSGi Service Wizard would result in a *red* workspace error marker.  This was spurious and it's necessary to force refresh the bndtools Workspace so that the dependencies across the 3 create projects are refreshed.


scottslewis:
@pkriens I removed the use of the UI thread for the workspace refresh.   Seems the unreliability/timing issue was being created  by using the canBeForked == true in project creation (done by superclass)

pkriens:
That is great. However, looking a bit further I noticed that you're not locking the workspace? 

You're calling a method that absolutely needs to be `writeLocked(...)`.

We've been awful in this but a workspace can definitely cause weird errors. Sorry for not noticing this in the previous comment.

scottslewis:
> That is great. However, looking a bit further I noticed that you're not locking the workspace?
> 
> You're calling a method that absolutely needs to be `writeLocked(...)`.
> 
> We've been awful in this but a workspace can definitely cause weird errors. Sorry for not noticing this in the previous comment.

Hi @pkriens .  Would you or someone mind just putting the appropriate code (e.g. writeLocked) in the appropriate location and merging this?   I don't have time right now to fumble around with the bndtools workspace synchronization needs.  Thanksinadvance.

scottslewis:
@pkriens ping.   See previous message about bndtools synchronization.  I would like to get this and issue #5632 completely into main branch so that I can test snapshots in subsequent days.  When is the next release target date?

pkriens:
I will fix it. It is quite simple, there is a lambda taking readLock and writeLock method on Workspace.

Issue #5890, opened 2023-11-14
Different build behavior if release.dir contains symbolic link or not 


gg-dim:
We have a bnd workspace project that is build via 
`./gradlew  release -Drelease.dir=/var/jenkins/repo/release/org.gecko.emf --info --stacktrace`
from within the /var/jenkins/.../workspace.

If the /var/jenkins/repo is a symbolic link (`ln -s`)  to another directory, the generated index.xml contains relative urls like:
`<attribute name="url" value="../../../../../jenkins/repo.gecko/release/org.gecko.emf/org.gecko.emf.osgi.example.model.basic/org.gecko.emf.osgi.example.model.basic-5.0.0.jar"/>
`
Without the symbolic link the same entry looks like:
`<attribute name="url" value="org.gecko.emf.osgi.example.model.basic/org.gecko.emf.osgi.example.model.basic-5.0.0.jar"/>`

We would expect, that the url is in both cases the same (the one without ../).

pkriens:
@juergen-albert could you take a look?

pkriens:
Can you make a minimal github repo? Your boss said you should do it! :-)

maho7791:
I provided a PR, that fixed the problem under linux.:

I sym-linked a folder as myrelease folder into the cnf via 

`ln -s /home/test/tmp/bndtools /devel/git/bndtools-7.0.0/bndtools.test/workspace/cnf/myrelease`

The Repo setup is:
```
-plugin.1: aQute.bnd.deployer.repository.LocalIndexedRepo;\
		readonly            =false;\
		pretty              =false;\
		name                =MyRelease;\
		local               =${build}/myrelease

-plugin.2: aQute.bnd.deployer.repository.LocalIndexedRepo;\
		readonly            =false;\
		pretty              =false;\
		name                =MyRelease2;\
		local               =/home/test/tmp/bndtools
```

XML before:

 ```
<capability namespace="osgi.content">
      <attribute name="size" value="491" type="Long"/>
      <attribute name="mime" value="application/vnd.osgi.bundle"/>
      <attribute name="url" value="../../../../git/bndtools-7.0.0/bndtools.test/workspace/cnf/myrelease/test/test-2.0.0.jar"/>
      <attribute name="osgi.content" value="8C168E7B386C11534EB43209C18D6A1D40CE38772E27AE7D11149BC86472DCD1"/>
    </capability>
```

**MyRelease works, MyRelease2 doesn't work**

After Patch:

  ```
[...]

PR #5944, opened 2023-12-23
#5890 distinguish base path for repo path for sym-links

maho7791:
refs #5890

- added check for symbolic links for the base path in LocalIndexRepositories


pkriens:
Thanks. I assumed you tested it?

PR #5958, opened 2024-01-06
Fix Gradle plugin version refs in README

Goooler:
The version of 7.1.0 hasn't been released, we shouldn't ref to it in docs.

### DIFF
--- a/gradle-plugins/README.md
+++ b/gradle-plugins/README.md
@@ -1,5 +1,8 @@
 # Gradle Plugins
 
+[![Maven Central](https://img.shields.io/maven-central/v/biz.aQute.bnd.builder/biz.aQute.bnd.builder.gradle.plugin)](https://central.sonatype.com/artifact/biz.aQute.bnd.builder/biz.aQute.bnd.builder.gradle.plugin)
+[![Plugin Portal](https://img.shields.io/gradle-plugin-portal/v/biz.aQute.bnd.builder)](https://plugins.gradle.org/plugin/biz.aQute.bnd.builder)
+
 Bnd includes Gradle plugins for [Gradle][1] users to build Bnd projects in [Bnd Workspace builds][20] as well as in [non-Bnd Workspace builds][21].
 A typical Gradle build is a non-Bnd workspace build.
 A Bnd Workspace build uses the information specified in the Bnd Workspace's `cnf/build.bnd` file and each project's `bnd.bnd` file to configure the Gradle projects and tasks.
--- a/gradle-plugins/README.md
+++ b/gradle-plugins/README.md
@@ -35,7 +35,7 @@ To make the Bnd Builder Gradle Plugin available to your build, use the following
 ```groovy
 pluginManagement {
   plugins {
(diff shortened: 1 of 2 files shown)


gradle-plugins/README.md
```diff
@@ -33,10 +33,8 @@ You can apply the `biz.aQute.bnd.builder` plugin to your project and/or you can
 To make the Bnd Builder Gradle Plugin available to your build, use the following in `settings.gradle`:
 
 ```groovy
-pluginManagement {
```
bjhargrave, line 36: You miss the point of this. It is in settings and thus configures the version of the plugin so that later use in build.gradle files do not need to state the version. This change is wrong.

bjhargrave (commented):
This is wrong. The README reflects the current branch which is building 7.1. When this branch is tagged as `7.1.0`, then the readme will be correct in the branch. With this change, then the readme in each tag will be for the prior version and not the version built by the tag.

So I don't approve of this PR.

Goooler:
> This is wrong. The README reflects the current branch which is building 7.1. When this branch is tagged as 7.1.0, then the readme will be correct in the branch. With this change, then the readme in each tag will be for the prior version and not the version built by the tag.

I was setting up bnd Gradle plugin for a new project, just copy-paste the example from README to my project, and found the current version is missing. The versions reflected in README should be available on Maven repos instead of being about to be released.

bjhargrave:
> The versions reflected in README should be available on Maven repos instead of being about to be released.

See 3rd paragraph of readme:

> This README represents the capabilities and features of the Bnd Gradle Plugins in the branch containing this README. So for the master branch, this will be the [latest development SNAPSHOT build](https://github.com/bndtools/bnd/tree/master/gradle-plugins#using-the-latest-development-snapshot-build-of-the-bnd-gradle-plugins). See the appropriate Git tag for the README for the Bnd Gradle Plugin version you are using.

bjhargrave:
When a new feature is added, the readme will be updated as part of the PR adding the new feature. If the readme must reflect the prior release, then we cannot update the readme until after the release branch is cut for the next release which means the readme updates will be forgotten.

Goooler:
Yeah, maybe we can add badges to reflect the latest version, even if developers forget to updating that.

gradle-plugins/README.md
Goooler: It could be placed either inside `pluginManagement` or not, I'm reverting this change.

Goooler: Got it.

bjhargrave:
> Yeah, maybe we can add badges to reflect the latest version

This is nice, but it only references one of the 3 plugins. There are also biz.aQute.bnd and biz.aQute.bnd.workspace plugins.

(And all the other changes need to be removed from this PR.)

bjhargrave:
> This is nice, but it only references one of the 3 plugins. There are also biz.aQute.bnd and biz.aQute.bnd.workspace plugins.

Actually, this does not help for the readme for a git tag since it always refers to the latest version rather than the version documented by the readme in the tag.

PR #5964, opened 2024-01-11
BndPomRepository should support both bsn and gav like MavenBndRepository

### LINKED ISSUE #5961: BndPomRepository cannot resolve OSGi dependencies by GAV like MavenBndRepository
I was using `MavenBndRepository` so far. 
Yesterday I experimented with BndPomRepostory ([converted central.mvn to pom.xml](https://bnd.discourse.group/t/pro-con-of-using-pom-xml-vs-central-mvn-files-for-maintaining-dependencies-in-bndtools-mavenbndrepository-vs-bndpomrepository/407)) and I was getting errors.

**Eclipse Problems View: Bndtools Path Problem Marker**

`com.couchbase.client:core-io;version=1.7.23 Not found in [bnd-cache [/myproject/plugins/cnf/cache/7.1.0/bnd-cache r/w=false], BndPomRepository [name=mavenpom, localRepo=/Users/christoph/.m2/repository, location=cnf/cache/pom-mavenpom.xml]]
`
**Eclipse Problems View: Bndtools Problem Marker**

`${repo} macro refers to an artifact com.google.http-client:google-http-client-latest (HIGHEST) that has an error: com.google.http-client:google-http-client;version=latest Not found in [bnd-cache [/myproject/plugins/cnf/cache/7.1.0/bnd-cache r/w=false], BndPomRepository [name=mavenpom, localRepo=/Users/christoph/.m2/repository, location=cnf/cache/pom-mavenpom.xml]]
`

The thing is that my `/.m2/repository` definitely contains both.

Here is a `bnd.bnd` where those errors appear when I use `BndPomRepo`.

```
-buildpath: \
	com.couchbase.client:core-io;version=1.7.23,\

-includeresource: \
	@${repo;com.google.api-client:google-api-client;latest},\

```

Both things work **fine** when using MavenBndRepository.

I noticed the BndPomRepository builds an .xml index file and for the first entry above I see this:

image

## Now the strange thing how I can get it to work:

When I use only the `bsn` without the `groupId` then the error goes away.

```
-buildpath: \
	core-io;version=1.7.23,\

-includeresource: \
	@${repo;com.google.api.client.googleapis;latest},\

```

Notice the difference... there is no colon separating `groupId` and `artifactId`.

## Question

Is this expected behavior, that BndPomRepository behaves different?

Any idea where to start debugging this? 
Maybe differences in MavenBndRepository.get() vs BndPomRepository.get() ?

Here is my repository config ( only have one of each active)

```
-plugin.6b.Central: \
	aQute.bnd.repository.maven.pom.provider.BndPomRepository; \
		snapshotUrl=http://repository.apache.org/snapshots/; \
		releaseUrl=https://repo.maven.apache.org/maven2/; \
		pom=${.}/pom.xml; \
		transitive=false; \
		name="mavenpom"
		
-plugin.6.Central: \
		snapshotUrl=http://repository.apache.org/snapshots/; \
		releaseUrl=https://repo.maven.apache.org/maven2/; \
		index=${.}/central.mvn; \
		readOnly=true; \
		name="Maven Central"
```

Content of `central.mvn`:

`com.couchbase.client:core-io:1.7.23`

Content of corresponding `pom.xml` (which was created from `central.mvn`)

```
<dependency>
            <groupId>com.couchbase.client</groupId>
            <artifactId>core-io</artifactId>
            <version>1.7.23</version>
        </dependency>
```

chrisrueger:
Closes #5961 

This PR should fix the problem that for a valid OSGi bundle MavenBndRepository.get(bsnOrGav) can be used with bsn or gav and returns something. 
BndPomRepostory.get() instead only works with the bsn for a valid OSGi bundle. (Non OSGi bundles which do not have a bsn meta data can only be looked up by gav wich works as expected)

This PR's goal is to make both repos behave the same and support bsn or gav for OSGi bundles.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BridgeRepository.java
@@ -51,7 +51,8 @@ public class BridgeRepository {
 	}
 
 	private final Repository								repository;
-	private final Map<String, Map<Version, ResourceInfo>>	index	= new HashMap<>();
+	private final Map<String, Map<Version, ResourceInfo>>	bsnIndex	= new HashMap<>();
+	private final Map<String, Map<Version, ResourceInfo>>	gavIndex	= new HashMap<>();
 
 	@ProviderType
 	public interface InfoCapability extends Capability {
@@ -225,37 +226,50 @@ public class BridgeRepository {
 	}
 
 	private void index(Resource r) {
-		String bsn;
-		Version version;
+
 
 		if (ResourceUtils.hasType(r, ResourceBuilder.SYNTHETIC))
 			return;
 
+		// index both separately: bsn + gav
+
 		IdentityCapability bc = ResourceUtils.getIdentityCapability(r);
 		if (bc != null) {
+
+			String bsn;
+			Version version;
+
 			bsn = bc.osgi_identity();
 			version = bc.version();
+
+			bsnIndex.computeIfAbsent(bsn, k -> new HashMap<>())
+				.put(version, new ResourceInfo(r));
+
 		} else {
 			logger.debug("No identity for {}, trying info", r);
-			InfoCapability info = getInfo(r);
-			if (info == null) {
-				// No way to index this
-				logger.debug("Also no info, giving up indexing");
-				return;
-			}
+		}
+
+		InfoCapability info = getInfo(r);
+		if (info != null) {
+
+			String bsn;
+			Version version;
 
 			bsn = info.name();
 			version = info.version();
 			if (version == null)
 				version = Version.LOWEST;
+
+			gavIndex.computeIfAbsent(bsn, k -> new HashMap<>())
+				.put(version, new ResourceInfo(r));
+		}
+		else {
+			// No way to index this
+			logger.debug("No info, giving up indexing");
+			return;
 		}
 
-		Map<Version, ResourceInfo> map = index.get(bsn);
-		if (map == null) {
-			map = new HashMap<>();
-			index.put(bsn, map);
-		}
-		map.put(version, new ResourceInfo(r));
+
 	}
 
 	public Resource get(String bsn, Version version) throws Exception {
@@ -267,7 +281,15 @@ public class BridgeRepository {
 	}
 
 	public ResourceInfo getInfo(String bsn, Version version) throws Exception {
-		Map<Version, ResourceInfo> map = index.get(bsn);
+		Map<Version, ResourceInfo> map = bsnIndex.get(bsn);
+		if (map == null)
+			return null;
+
+		return map.get(version);
+	}
+
+	public ResourceInfo getInfoByGAV(String gav, Version version) throws Exception {
+		Map<Version, ResourceInfo> map = gavIndex.get(gav);
 		if (map == null)
 			return null;
 
@@ -277,7 +299,7 @@ public class BridgeRepository {
 	public List<String> list(String pattern) throws Exception {
 		List<String> bsns = new ArrayList<>();
 		if (pattern == null || pattern.equals("*") || pattern.equals("")) {
-			bsns.addAll(index.keySet());
+			bsns.addAll(bsnIndex.keySet());
 		} else {
 			String[] split = pattern.split("\\s+");
 			Glob globs[] = new Glob[split.length];
@@ -285,7 +307,7 @@ public class BridgeRepository {
 				globs[i] = new Glob(split[i]);
 			}
 
-			outer: for (String bsn : index.keySet()) {
+			outer: for (String bsn : bsnIndex.keySet()) {
 				for (Glob g : globs) {
 					if (g.matcher(bsn)
 						.find()) {
@@ -299,7 +321,7 @@ public class BridgeRepository {
 	}
 
 	public SortedSet<Version> versions(String bsn) throws Exception {
-		Map<Version, ResourceInfo> map = index.get(bsn);
+		Map<Version, ResourceInfo> map = bsnIndex.get(bsn);
 		if (map == null || map.isEmpty()) {
 			return new TreeSet<>();
 		}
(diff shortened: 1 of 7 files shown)


chrisrueger:
## 1st draft, base for discussion.

- I ported `MavenBndRepositoryTest.testGetViaBSNAndGAV()` to `PomRepositoryTest.testGetViaBSNAndGAV()` to reproduce the problem
- next commits are a try to fix it. Not sure if it is a good impl, but at least based for discussion

I hooked into the indexing of `BridgeRepository` because this seems to handle the  indexing of 3-digit versions which allows you to lookup for "1.2.0" as well as "1.2"  (e.g. for commons-cli:commons-cli:1.2 (commons-cli-1.2.jar) which physically just defines a 2-digit version.

pkriens:
I think you're making it too hard but I might be wrong.

A non-OSGi artifact has _nothing_ to do with the BridgeRepository. They lack any OSGi metadata. Trying to support them as a first class citizen there feels uncomfortable because they wreak havoc on many different places, the bsn is a pretty big invariant of OSGi. This is a hack only the Maven repositories need, it shouldn't touch code unrelated.

That said, to be able to wrap them, and sometimes use in unit tests, we need a way to address them. However, we should keep in mind that they are not the common way to address bundles. 

This only needs to be supported in the RepositoryPlugin::get method. It is rather easy to check the bsn for a ':' and if it has, you can directly construct the GA, get the versions, and compare the versions to find the best matching.  The BndPomRepository already has a list of archives from the POM.

Something like this:
```java
	@Override
	public File get(String bsn, Version version, Map<String, String> properties, DownloadListener... listeners)
		throws Exception {
		if (!init()) {
			return null;
		}

		Archive archive;
		if (isMavenGAV(bsn)) {
			Archive spec = Archive.valueOf(bsn + ":" + version);
			archive = archives.stream()
				.filter(a -> matches(a, spec))
				.findAny()
				.orElse(null);
			if (archive == null)
				return null;
		} else {
			ResourceInfo resource = bridge.getInfo(bsn, version);
			if (resource == null) {
				archive = trySources(bsn, version);
				if (archive == null)
					return null;
			} else {

				String from = resource.getInfo()
					.from();
				archive = Archive.valueOf(from);
			}
		}
		Promise<File> p = repoImpl.getMavenRepository()
			.get(archive);

		if (listeners.length == 0)
			return p.getValue();

		Map<String, String> attrs = archive.attributes();
		new DownloadListenerPromise(reporter, name + ": get " + bsn + ";" + version, p, attrs, listeners);
		return repoImpl.getMavenRepository()
			.toLocalFile(archive);
	}
```

Notice that the IndexFile in the MavenBndRepository is basically collapsed into the BndPomRepository. The `archives` is the index.

chrisrueger:
Thanks @pkriens for the review. 
I will rollback the BridgeRepo code. 

> The `archives` is the index.

I think I had something similar in between. I saw the `archives` were null or so.. Not sure anymore, I will double check.

Thanks a lot. I post an update.

chrisrueger:
@pkriens I have changed it to something similar to your suggestion. 
In order to access the `archives` found no other way than populating them first in `PomRepository` 

https://github.com/bndtools/bnd/pull/5964/files#diff-3a6aadeb70271d493c56e3f477c4b1ddf7edd867b4b0cfefef9b6bbb20f355fdR94-R95

so I can do `pomRepo.archives.stream()...`. ([see](https://github.com/bndtools/bnd/pull/5964/files#diff-63147ca4b34e8e6e058a4a1b685827b57b58e3346f5a488c9d372998bc89e78fR317-R320))

The `archives` in `BndPomRepository` itself were always null when using a `pom.xml`. They were only populated when setting revisions (`configuration.revision()`).

chrisrueger:
Furthermore I added the same GAV handling to `pomRepo.versions(bsn)` because I noticed that this was also related to my initial reported problem and `pomRepo.get()` was not enough. 

After fixing .versions(bsn) too all the errors below in my project went away.

image

pkriens:
Yeah, much better. It is now restricted to the BndPomRepository and does not pollute the generic code. Thanks!

pkriens:
pressed the wrong button ... sorry

chrisrueger:
@pkriens 

While playing with my fix in this PR I noticed the following behavior? 

When the `pom.xml` contains two versions of the same dep like this:

```
<dependency>
    <groupId>org.apache.felix</groupId>
    <artifactId>org.apache.felix.gogo.shell</artifactId>
    <version>1.1.4</version>
</dependency>
<dependency>
    <groupId>org.apache.felix</groupId>
    <artifactId>org.apache.felix.gogo.shell</artifactId>
    <version>1.1.0</version>
</dependency>

```

I get this when I want to start my `.bndrun` (`-runbundles org.apache.felix.gogo.shell;version='[1.1.4,1.1.5)',\`)

image


It works fine (no error) if I reverse the order and put 1.1.4 **last**:

```
<dependency>
    <groupId>org.apache.felix</groupId>
    <artifactId>org.apache.felix.gogo.shell</artifactId>
    <version>1.1.0</version>
</dependency>

<dependency>
    <groupId>org.apache.felix</groupId>
    <artifactId>org.apache.felix.gogo.shell</artifactId>
    <version>1.1.4</version>
</dependency>

```

or remove 1.1.0

So it seems: last version wins.

In MavenBndRepository (`.mvn`) I do not have this problem. Since I generated the `pom.xml` from my `.mvn` file, it coindicently ended up in that order in the `pom.xml` but I just noticed it now with my first steps with `BndPomRepo`

Question:
Is this expected behavior of BndPomRespository? 
Or should I file a bug and dig deeper? Any idea where I could start search putting my breakpoint to search the needle in the haystack?

pkriens:
Well, officially you cannot have 2 versions in maven so from a pom pov it is problematic. From a BndPomRepository pov it is obviously wrong ... We should be able to support multiple versions. I think this is the culprit:

https://github.com/bndtools/bnd/blob/bc9ecf9f2ba2193959121273a675d6c749364c60/biz.aQute.repository/src/aQute/maven/provider/POM.java#L512

It explicitly removes multiple versions. I'd expect it would only show the first so it could be something else

chrisrueger:
Thanks @pkriens I will create an issue out of it and try digging in during the next days.

PR #5969, opened 2024-01-15
Issue/5828 quick fix version

### LINKED ISSUE #5828: [bndtools] package-info.java quickfix uses deprecated Version annotation
When you export a package in Bndtools and there is no package-info, it will warn you and give you the opportunity to create one for your automatically. However, the automatically created one uses the Bnd annotation, which is now (since Bnd 7.0.0) deprecated in favour of the OSGi standard Version annotation. Should generate it using the new annotation.

pkriens:
This massive amount of work was originally intended to fix #5828. However, on inspection, this issue was based on the more or less deprecated Export-Package header semantics and the associated editor GUI. Clearly this editor needs an overhaul but we've more or less standardized on doing exports via the @Export and @Version annotations in `package-info.java`. For this, I added a new wizard NewPackageWizard and added this to the list of 'new' menu entries in the Bndtools perspective. (File/New should now show it.). It might be necessary to reset the Bndtools perspective to see them. After this I got busy with some quick fixes for package-info.java. 
This piqued my interest in the quick fixes and before long I'd developed a nice library that makes quick fixes significantly easier to do. It is type safe and uses modern java with streams, monads, and optionals. So I added the following refactorings:

* Numbers
  * Convert decimal number to hex
  * Convert decimal number to binary
  * Convert hex numbers to decimal
  * Convert binary numbers to decimal
  * Add/Update _ (spaced 3 for decimal, 4 for hex & binary)
  * Remove underscores
* Strings
  * To upper case
  * To lower case
  * TextBox (""") to string literal (")
  * String literal to Text Box
  * Wrapping of text box content
* Versions
  * Bump major
  * Bump minor
  * Remove qualifier
* Diverse
  * Assign constructor parameter to final field
* Gogo Command
  * Add Gogo command annotation
  * Add a method as a function (updates GogoCommand.function value)
  * Remove a method from the functions
  * Remove GogoCommand annotation and all other related annotations
* Component 
  * Add Component
  * Remove Component (and all related annotations)
  * Add Reference to field, method, or parameter in Activate method/constructor
  * Add/remove Activate/Deactivate annotations (with cleanup)

One of the dangers is that I created too many functions. I therefore need some feedback how they are working for you.


pkriens:
I will commit this version before a review. Reason is that this is a lot of quite complex code and experience shows that it will not happen or take a long time. I think we need to try these changes out and I am ready to fix them.

pkriens:
I am also very open to add more quick fixes. It might even be an idea to allow quick fixes to come from the repositories so people can make custom versions. For an organization it might be very interesting to have custom quick fixes.

Its now really easy! :-)

chrisrueger:
> * Assign constructor parameter to final field

That would be a good addition to Eclipse JDT in general. @laeubi @HannesWell

chrisrueger:
Maybe a bit early: 
Where do you would like to collect Bug-Reports / suggestions? A collector issue?
 
e.g. I installed snapshot today and e.g. the "Add @Component" Quickfix replaces all current import statements with `import org.osgi.service.component.annotations.Component;`. I would have expected an addition.

Furthermore:
What do you think about a list of all quickfixes at bndtools.org ? I could help with this.

pkriens:
a) ok, open an issue and we use it as a collection issue
b) The PR contains a list. Let's get some experience before we document it. If you're too early with documentation you write it 3x

laeubi:
> > * Assign constructor parameter to final field
> 
> That would be a good addition to Eclipse JDT in general. @laeubi @HannesWell

Everyone can contribute/propose enhancements here:

- https://github.com/eclipse-jdt/eclipse.jdt.core
- https://github.com/eclipse-jdt/eclipse.jdt.ui

beside that JDT offers assign a constructor parameter to a field already, if you enable the save action to make fields final if possible you archive a similar goal.

pkriens:
There is a difference between automatic final _because the code does not modify it_ and _it must be final because I say so_. The automatic cleanup is imho not so useful because it is a consequence of the code, not a check on the code. And I don't think private variables are useful either; I've not been able to figure out how to change those to package private.

Issue #5967, opened 2024-01-15
Occasional ConcurrentModificationException


juergen-albert:
One of our builds fails on github from time to time with an ConcurrentModificationException. One Example beeing: https://github.com/geckoprojects-org/org.gecko.emf.utils/actions/runs/7528737524/job/20492000254

I can't reproduce it locally and it only seems to happen during the github actions.

The Stacktrace:

```
Caused by: java.util.ConcurrentModificationException
	at aQute.bnd.osgi.Processor.addAll(Processor.java:229)
	at aQute.bnd.osgi.Processor.getInfo(Processor.java:214)
	at aQute.bnd.gradle.BndPlugin.checkProjectErrors(BndPlugin.java:986)
	at aQute.bnd.gradle.BndPlugin.checkErrors(BndPlugin.java:978)
	at aQute.bnd.gradle.BndPlugin$2.execute(BndPlugin.java:470)
	at aQute.bnd.gradle.BndPlugin$2.execute(BndPlugin.java:466)
	at org.gradle.api.internal.AbstractTask$TaskActionWrapper.execute(AbstractTask.java:802)
	at org.gradle.api.internal.AbstractTask$TaskActionWrapper.execute(AbstractTask.java:775)
	at org.gradle.api.internal.tasks.execution.TaskExecution$3.run(TaskExecution.java:242)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:29)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$1.execute(DefaultBuildOperationRunner.java:26)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:66)
	at org.gradle.internal.operations.DefaultBuildOperationRunner$2.execute(DefaultBuildOperationRunner.java:59)
```

pkriens:
I took a quick look. It seems that we check errors in each task while we also still execute that might generate errors/warnings.

I think the only reasonable solution is to actually start using the Workspace lock in the Gradle bnd plugin. We can execute each Task in a read lock and then collect the errors in a write lock. This seems doable, I only count about 8 tasks. Since we could still have concurrent errors (although I expect this is extremely unlikely because we create dedicated processors like Builder for each task) the read lock would officially not suffice. This is a common problem so maybe we should at least make the list objects concurrent.

@bjhargrave, got a minute to shine your light on this?

[...]

PR #5986, opened 2024-01-23
Issue/5967 concurrent error handling

pkriens:
#5967 indicated that in Gradle and maybe also in Maven the error/warning message handling was not properly handling concurrent access in getInfo. 

This PR has put the core message handling of Processor in a new class MessageReporter. This class should be concurrent safe for normal usage. It should at least not throw ConcurrentModificationExceptions ...

The behavior is slightly changed. The previous code handled out modifiable versions of its own error/warnings. This was used in several places to change these lists from the outside. One place was `Processor.check(patterns)`. A lot of tests depended on this behavior so that was explicitly coupled.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/MessageReporter.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/MessageReporter.java
@@ -149,7 +149,6 @@ public class MessageReporter {
 			}
 		}
 
-
 	}
 
 	class Cache {
@@ -200,17 +199,25 @@ public class MessageReporter {
 			}
 			String s = Processor.formatArrays(format, args);
 			Message m = new Message(counter.getAndIncrement(), s, true);
-			messages.putIfAbsent(s, m);
+			putMessage(s, m);
 			return m;
 		} finally {
 			processor().signal();
 		}
 	}
 
+	void putMessage(String s, Message m) {
+		ConcurrentHashMap<String, Message> current;
+		do {
+			current = messages;
+			current.putIfAbsent(s, m);
+		} while (current != messages);
+	}
+
 	SetLocation warning(String format, Object... args) {
 		String s = Processor.formatArrays(format, args);
 		Message m = new Message(counter.getAndIncrement(), s, false);
-		messages.putIfAbsent(s, m);
+		putMessage(s, m);
 		return m;
 	}
 
@@ -239,7 +246,7 @@ public class MessageReporter {
 			.sorted()
 			.map(m -> new Message(counter.getAndIncrement(), m, actualPrefix))
 			.forEach(m -> {
-				messages.putIfAbsent(m.message, m);
+				putMessage(m.message, m);
 			});
 	}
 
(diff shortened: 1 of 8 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/MessageReporter.java
```diff
@@ -0,0 +1,279 @@
...
+
+	SetLocation warning(String format, Object... args) {
+		String s = Processor.formatArrays(format, args);
+		Message m = new Message(counter.getAndIncrement(), s, false);
+		putMessage(s, m);
+		return m;
+	}
+
+	void getInfo(Reporter from, String prefix) {
+		String actualPrefix = prefix == null ? prefix = processor().getBase() + " :" : prefix;
```
bjhargrave, line 225: Don't use `prefix =` in the true side of the ternary operator. This is a non-obvious side effect.

biz.aQute.bndlib/src/aQute/bnd/osgi/MessageReporter.java
```diff
@@ -0,0 +1,279 @@
...
+
+		MessageReporter other;
+		if (from instanceof Processor processor) {
+			other = processor.reporter;
+		} else if (from instanceof MessageReporter mr) {
+			other = mr;
+		} else {
+			// backward compatible reporters
+			addAll(true, prefix, from);
+			addAll(false, prefix, from);
```
bjhargrave, line 235:
```suggestion
			addAll(true, actualPrefix, from);
			addAll(false, actualPrefix, from);
```

biz.aQute.bndlib/src/aQute/bnd/osgi/MessageReporter.java
```diff
@@ -0,0 +1,279 @@
...
+			Message m = new Message(counter.getAndIncrement(), newMessage, error);
+			Location location = reporter.getLocation(message);
+			if (location != null)
+				location.to(m);
+		}
+	}
+
+	ConcurrentHashMap<String, Message> clear() {
+		ConcurrentHashMap<String, Message> previous = messages;
+		messages = new ConcurrentHashMap<>();
```
bjhargrave, line 268: If you care, there is still a race condition here where errors can get lost.

pkriens: no, see the `putMessage()`. It is possible (but very unlikely) that the same message will be posted twice, which I think is acceptable.

To prevent this, it would not blocking between the `putMesssage()` and the `clear()` which I do not think is worth the overhead? But I am easy to convince here if you think that would be worth it?

biz.aQute.bndlib/src/aQute/bnd/osgi/MessageReporter.java
pkriens: Ouch. Thanks.

pkriens:
@bjhargrave I guess the BndPlugin does not need an update now? Now the getInfo() and the error/warning methods are thread safe, it should just work?

Shall I merge this?

bjhargrave:
> I guess the BndPlugin does not need an update now? Now the getInfo() and the error/warning methods are thread safe, it should just work?
> 
> Shall I merge this?

Agree.

pkriens:
Thanks for still participating in your free time, highly appreciated.

PR #5990, opened 2024-01-24
Add a verify goal to the bnd-resolver-maven-plugin

### LINKED ISSUE #5984: Add a "validate" target for the bnd-resolver-maven-plugin
The current bnd-resolver-maven-plugin does validation using `failOnChanges` but this is a brittle solution. If different JVMs or different architectures are used then resolve output can be subtly different, e.g. by including an API jar such as `org.osgi.service.cm`.

We should add a `validate` target which creates a repository and run requirements from the `-runbundles`, checking that the -run bundles are locatable and will resolve.

timothyjward:
The  goal provides a way to validate that a set of run bundles is a valid resolution. It does this by creating a set of run requirements requiring each run bundle by identity and performing a resolve operation. This resolve operation forbids any capabilities that come from bundles not in the list of run requirements. If the resolution succeeds then it is further checked against the orginal run requirements.

If these checks pass then we know that:

 * The resolution is valid
 * The resolution satisfies the original run requirements

This is useful for CI when we want to test that a checked-in bndrun file has been correctly resolved.

Fixes #5984

### DIFF
--- a/maven-plugins/bnd-resolver-maven-plugin/src/it/parent/pom.xml
+++ b/maven-plugins/bnd-resolver-maven-plugin/src/it/parent/pom.xml
@@ -7,6 +7,10 @@
 	<version>0.0.1</version>
 	<packaging>pom</packaging>
 
+	<properties>
+		<it.resolve.skip>false</it.resolve.skip>
+	</properties>
+
 	<build>
 		<pluginManagement>
 			<plugins>
@@ -55,6 +59,9 @@
 							<goals>
 								<goal>resolve</goal>
 							</goals>
+							<configuration>
+								<skip>${it.resolve.skip}</skip>
+							</configuration>
 						</execution>
 					</executions>
 				</plugin>
--- a/maven-plugins/bnd-resolver-maven-plugin/src/it/verify/pom.xml
+++ b/maven-plugins/bnd-resolver-maven-plugin/src/it/verify/pom.xml
@@ -1,0 +1,37 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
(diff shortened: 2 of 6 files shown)


maven-plugins/bnd-resolver-maven-plugin/src/it/parent/pom.xml
```diff
@@ -55,6 +59,9 @@
 							<goals>
 								<goal>resolve</goal>
 							</goals>
+							<configuration>
+								<skip>${it.resolve.skip}}</skip>
```
bjhargrave, line 63:
```suggestion
								<skip>${it.resolve.skip}</skip>
```

maven-plugins/bnd-resolver-maven-plugin/src/it/verify/pom.xml
```diff
@@ -0,0 +1,37 @@
...
+	<modelVersion>4.0.0</modelVersion>
+
+	<parent>
+		<groupId>biz.aQute.bnd-test</groupId>
+		<artifactId>resolver-test</artifactId>
+		<version>0.0.1</version>
+		<relativePath>../parent</relativePath>
+	</parent>
+
+	<artifactId>resolve</artifactId>
```
bjhargrave, line 13:
```suggestion
	<artifactId>verify</artifactId>
```

bjhargrave (commented):
Don't you need a `postbuild.groovy` file to confirm the goal did something?

Also this PR should update `maven-plugins/bnd-resolver-maven-plugin/README.md` to specify the new goal and its configuration parameters.

pkriens:
Can this be merged? @bjhargrave  you're satisfied? I have no special knowledge here so I am trusting Tim here.

bjhargrave:
I am ok as long as the readme (documentation) update is also included. Without any doc, no one will know that the goal exists or how to use it. Bnd has done a decent job of documenting the maven and gradle plugins. We need to continue.

timothyjward:
> Bnd has done a decent job of documenting the maven and gradle plugins. We need to continue.

Agreed. Given that we seem to have consensus I'll write up what we have.

timothyjward:
Docs done - it also helpfully identified a property I could remove from the `VerifierMojo`, along with some newly dead code.

PR #5996, opened 2024-01-25
Avoid creating an error when a Collection field has a field-option of REPLACE

### LINKED ISSUE #5994: Invalid error on DS field replacement
My DS component includes the following:

    @Reference(cardinality = ReferenceCardinality.MULTIPLE, policy = ReferencePolicy.DYNAMIC)
    volatile Set<MyService> services = new HashSet<>();

As far as I am aware this is perfectly legal [according to the DS spec](https://docs.osgi.org/specification/osgi.cmpn/7.0.0/service.component.html#service.component-reference.field.option)

Despite this the following code in bnd triggers an error. I think this check needs to be removed.
https://github.com/bndtools/bnd/blob/58a13dcff87929498cc5f1eb6c786f94c0f35197/biz.aQute.bndlib/src/aQute/bnd/component/DSAnnotationReader.java#L1036-L1040

timothyjward:
The existing code enforced a restriction which is not present in the specification. Specifically it prevented Collection type fields from using the  field-option, which is a valid use of the annotations.

Fixes #5994


bjhargrave (commented):
See https://github.com/bndtools/bnd/issues/5994#issuecomment-1910764215

No change should be made here.

PR #5999, opened 2024-01-29
Improve error reporting for DS annotations

timothyjward:
The error messages when handling field/constructor injection with Collection types are somewhat confusing, in that they don't explain what the correct usage would be. This is particularly problematic with collection subclasses. This commit attempts to improve the error messages, and to improve the naming of an internal field which is used to track the fact that a DS reference is a collection type which is not `Collection` or `List`

See also #5994


pkriens:
LGTM, do I need to merge?

PR #6000, opened 2024-01-29
Fixes for the bnd-resolver-maven-plugin verification

timothyjward:
The test case used to test the verifier was too simple, and failed to identify that the wrong requirement was being tested in the ResolutionCallback. This commit makes the test more extensive, and fixes the bug.


rotty3000 (approved):
Nice catch! LGTM

PR #6002, opened 2024-01-30
Ensure bnd-resolver-maven-plugin verify handles multi-release jars

timothyjward:
Multi-Release Jars are modelled in bnd using SupportingResource children. When verifying we only care about the *real* resources which end up in the run bundles so we must

1. Permit SupportingResource children to pass through the ResolveCallback
2. Remove SupportingResource children from the resolve result before comparing it

### DIFF
--- a/maven-plugins/bnd-resolver-maven-plugin/src/it/verify-multirelease/invoker.properties
+++ b/maven-plugins/bnd-resolver-maven-plugin/src/it/verify-multirelease/invoker.properties
@@ -1,0 +1,7 @@
+invoker.goals=--no-transfer-progress package
+
+# Run mvn with --debug for debug logging
+#invoker.debug=true
+
+# Run mvn in debugging mode and wait for a debugger to attach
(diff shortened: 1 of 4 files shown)


timothyjward:
I've been attempting to make use of the verify goal in Eclipse sensiNact, which has helpfully been identifying bugs and allowing me to provide more extensive testing.

I think this one definitely needs @pkriens to take a quick look as he's the expert on `SupportingResource`

pkriens:
Why do you care about the SupportingResource? What is happening when you just ignore it and treat it as a normal resource?
I do not understand the problem so that might be the issue but this code should definitely NOT become common. A Supporting Resource is a resource. The idea is that only the Repository implementations should be aware of the fact that it needs support.
So I am missing something but we should try to find a solution that does not require instanceof for that will explode.

timothyjward:
>I believe that there should be a way to do step 2 using only the rules from the RunResolution, but it will require a new method in that class.

My proposal for this is in e85642c338556629301844e5100ba33c828107a9

timothyjward:
All cleaned up after discussion with Peter. I will squash and merge to avoid `SupportingResource` being in the history.

PR #6004, opened 2024-01-31
Update dependabot.yml to ignore test dependency

timothyjward:
I believe this will stop dependabot complaining about the version of Jackson used in the resolver verification tests.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,6 +14,10 @@ updates:
     directory: "/maven-plugins"
     schedule:
       interval: "daily"
+    ignore:
+      - dependency-name: "com.fasterxml.jackson.core:jackson-databind"
+        # This version is used in the resolver tests and should not be updated
+        versions: ["2.11.1"]
 
   # Maintain dependencies for gradle
   - package-ecosystem: "gradle"


pkriens:
I find it hard to believe you think this is a good solution? :-) it means every time dependabot has a new dependency that occurs in your test, we can add it here until over time all test dependencies have found their way here. This does not sound like the Tim I worked with!

Could we use another pom name? In maven you can do `mvn clean test -Dmaven.source=alternate.xml`. I see that some people want dependabot to scan these alternates, implying they are not scanned today. Exactly what we need? Alternatively, you could also just resolve the bndrun file, save the results, and then verify the just resolved file?

There are complaints of the lack of an exclusion in the dependabot configuration but Github product management seems not very keen following this up.

timothyjward:
> I find it hard to believe you think this is a good solution? :-) it means every time dependabot has a new dependency that occurs in your test, we can add it here until over time all test dependencies have found their way here.

We need a stable test in order to provide coverage for the originally identified problem. In this case the problem was that we used the wrong requirement in a lambda, but that didn't matter unless the resolution contained a chain with import package dependencies. There are two ways to supply this test case: 

1. Create some test bundles which we make as part of the build
1. Use some existing bundles which have the right properties

Jackson databind is a very good fit for option 2, and saves us work in our build, it also guarantees that we can find external dependencies from the pom file when verifying.

Using Jackson In this case means that the test case has a single dependency in its pom. We don't want the version to change over time as this would change the test, which is not related to the version of Jackson, but to the shape of its dependencies. There is (and will be) no further filtering needed in the dependabot configuration because the change here includes every dependency from the pom.

> Could we use another pom name? In maven you can do mvn clean test -Dmaven.source=alternate.xml. I see that some people want dependabot to scan these alternates, implying they are not scanned today. Exactly what we need? 

This adds complexity to the test case, or requires that we change every single test case for the `bnd-resolver-maven-plugin`. I'm not even sure how we would get the invoker plugin to do this.

> Alternatively, you could also just resolve the bndrun file, save the results, and then verify the just resolved file?

This doesn't satisfy the test case, which needs to use the pom dependencies to find the resources to verify. We need a pom and it needs to have the dependency in it.

> This does not sound like the Tim I worked with!

I'd like to think that I was always pragmatic. In this case I don't believe that we should be changing a useful test case to satisfy a warning from a tool which doesn't understand the context of this dependency. There is no leakage of integration test dependencies at any scope from the released maven plugins. If there were then we would long ago have needed to update from the ancient versions of `org.apache.felix.framework` and `org.apache.felix.eventadmin` used elsewhere in the same test suite.

I will reiterate that following dependabot's recommendations will change what this test is testing. If we were to move to `2.16.1` then the test would cover multi-release bundles, which are covered in a different test case. I do not think that we want dependabot to start writing our tests for us.

rotty3000:
Adding support for @timothyjward 's solution as there is overwhelming precedent for this in bnd such as [this example](https://github.com/bndtools/bnd/tree/master/biz.aQute.bndlib.tests/testresources) (check all the libraries in there...). They are no different; merely stable examples of specific problems bnd is attempting to solve.

The only difference _here_ is that dependabot got tripped up by the fact it couldn't distinguish between a "real" project dependency and a "test" dependency.

pkriens:
In bnd we have lots of file based repositories for exactly this kind of problem. BJ went out of its way even to change many tests to use file based repositories instead of relying on the current build. Why would maven be different? 

As far as I can see there are many solutions:

* make your own file based repository
* make an intermediate parent pom with a name `x.pom` and refer to it with `relativePath`
* copy a pom from a directory that is excluded from pom scanning. 
* use scope=import
* make a special test module that generates a fixed dependency that is used as dependency in the resolve test

I tried to show this but for some reason the maven Indexer test fails on my system and I do not have time to figure out the mysteries of maven testing.

The problem is that it looks innocent now but it will be confusing if someone actually wants to use jackson, and once you open this gate, it will be opened further over time because it seems common practice.

pkriens:
Can we fix this: https://github.com/bndtools/bnd/security/dependabot

PR #6023, opened 2024-02-07
Slow IO

### LINKED ISSUE #6014: [win] slow IO.getBasedFile
since [JDK ](https://bugs.openjdk.org/browse/JDK-8207005) disable the file canonicalization cache File.getCanonicalPath is very slow on windows.
example sampling during building eclipse platform workspace.
image

the zipslip check in https://github.com/bndtools/bnd/blame/f6371af9346e75f0170987ceb52ce40cf5a8fa9b/aQute.libg/src/aQute/lib/io/IO.java#L850
could be much faster to just return child in the normal case that the filename does not contain an directory escape:
`if (!file.contains("..")) return getFile(base, file);`
which avoid costly file name capitalization at all.

pkriens:
- fixed the getBasedFile to not use canonical files since they were not needed
- Made classes for Windows and Other oses Fixes #6014

---

### DIFF
--- a/aQute.libg/src/aQute/lib/io/Other.java
+++ b/aQute.libg/src/aQute/lib/io/Other.java
@@ -1,0 +1,49 @@
+package aQute.lib.io;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Path;
+
+import aQute.lib.io.IO.OS;
+
+class Other implements OS {
+
+	@Override
+	public File getBasedFile(File base, String subPath) throws IOException {
+		String use;
+		if (subPath.contains("..")) {
+			Path normalizedPath = Path.of(subPath)
+				.normalize();
+			if (normalizedPath.getNameCount() > 0 && normalizedPath.getName(0)
+				.equals(IO.DOTDOT)) {
+				throw new IOException("io.sub.up invalid path, will escape the designated directory. path='" + subPath
+					+ "', base='" + base + "', normalized='" + normalizedPath + "'");
+			}
+			use = normalizedPath.toString();
+		} else
+			use = subPath;
+		return new File(base, use);
+	}
+
+	@Override
+	public String getenv(String string) {
+		return System.getenv(string);
+	}
+
+	@Override
+	public String toSafeFileName(String string) {
+		StringBuilder sb = new StringBuilder();
+
+		for (int i = 0; i < string.length(); i++) {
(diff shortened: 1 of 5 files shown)


aQute.libg/src/aQute/lib/io/Other.java
```diff
@@ -0,0 +1,49 @@
...
+
+class Other implements OS {
+
+	@Override
+	public File getBasedFile(File base, String subPath) throws IOException {
+		String use;
+		if (subPath.contains("..")) {
+			Path normalizedPath = Path.of(subPath)
+				.normalize();
+			if (normalizedPath.getNameCount() > 0 && normalizedPath.getName(0)
```
jukzi, line 17: index 0 only will not check against  something/../../escaped

aQute.libg/src/aQute/lib/io/Windows.java
```diff
@@ -0,0 +1,75 @@
...
+	public File getBasedFile(File base, String subPath) throws IOException {
+		String use;
+		Matcher matcher = WINDOWS_BAD_FILE_NAME_P.matcher(subPath);
+		if (matcher.find()) {
+
+			try {
+				Path normalizedPath = Path.of(subPath)
+					.normalize();
+
+				if (normalizedPath.startsWith(IO.DOTDOT)) {
```
jukzi, line 26: that "startsWith" will not catch `something/../../escaped`

jukzi (changes requested):
you are better off resolving relative against base an then check if its still legal

pkriens:
> you are better off resolving relative against base an then check if its still legal

I disagree. It is the responsibility of the caller to make sure the base is ok. Eclipse file systems and Java file systems and the OS file systems interact in horrible ways. See how hard the testing of this is, especially on windows.

The thing I learned here is that the `new File(base,path)` will ensure that any absolute file will **always** be a child of base. From the doc:

> If the {@code child} pathname string is absolute then it is converted into a **relative** pathname in a system-dependent way.

I.e. unless base is `null`, `new File(base,child)` will therefore always return a child of base.

jukzi (approved):
> I.e. unless base is `null`, `new File(base,child)` will therefore always return a child of base.

not really intuitive.
Thanks for caring about the issues. looks like the tests are sufficient

pkriens:
thanks for the collaboration

PR #6027, opened 2024-02-08
print diff if bndrun verification fails

### LINKED ISSUE #5985: Better print out if bndrun verification fails
If the resolver plugin is set to verify and fails it prints out the list of actual `-runbundles`  and the list of calculated bundles. They are two separate lines and finding the different is quite hard. 

We should find a better printout two compare the two if possible.

chrisrueger:
Closes #5985 

print a readable diff sentence if  existing runbundles and calculated runbundles are different.

e.g. 

```
Diff [osgi.enroute.junit.wrapper;version='[4.12.0,4.12.1)', test.simple;version=snapshot] exist in calculated runbundles but missing in existing runbundles
```


chrisrueger:
@juergen-albert please review, if that solves your problem or if this is totally wrong

@pkriens does that additional output do any harm? e.g. Are there tools expecting exactly the existing output or something like that?

pkriens:
> Are there tools expecting exactly the existing output or something like that?

I hope not. And if there are, they should be put out of their misery :-)

pkriens:
I think you're going to have the same problem as Tim had: dependabot. This will update your version numbers and then your tests will fail, looking at the version numbers it looks like our own build dependencies. In this case, using one of the XML files as repository might work better since those versions will not change.

chrisrueger:
> I think you're going to have the same problem as Tim had: dependabot. This will update your version numbers and then your tests will fail, looking at the version numbers it looks like our own build dependencies. In this case, using one of the XML files as repository might work better since those versions will not change.

Thanks for pointing that out. I simplified the test and removed bundle and version stuff.  Since it was an existing testcase which also only checks only parts of the message I think this is sufficient.

pkriens:
@juergen-albert if you're ok with this then you can merge it.

pkriens:
Lets merge this, if @juergen-albert doesn't like it, we can change it. ok?

juergen-albert:
Sorry for my silence on this. I'm still fighting with the github notification settings, to not get too much noise. Long story short, I missed this,

+1 from me. Thanks @chrisrueger

PR #6029, opened 2024-02-12
refreshFromModel() before resolving

### LINKED ISSUE #6028: RunRequires list can be emptied if you click Resolve button too fast after opening a .bndrun file
I noticed a race-condition problem in this scenario:

1. Open a .bndrun
2. Click the resolve button while the Run Requirements list is not finished loading

image

Actual result:
- resolving takes place (this is OK), BUT afterwards the runrequirements are still empty and if you save then you end up with a .bndrun file which has the runrequirements deleted

Expected result:
- resolving should wait until the runrequirements are loaded and not delete them. It should basically work as I would have waited myself 1-2 seconds more.

This happened to me quite often yet. It happens since our .bndrun is quite large and takes some time to load and I click very fast sometimes. 

I have a PR ready.

chrisrueger:
This fixes the problem that the runrequirements get deleted if you click the resolve button too fast when the runrequirements part has not been loaded yet.

Closes #6028


pkriens:
Thanks! I love having you around!

Issue #6033, opened 2024-02-15
[Maven] Export in IDE doesnt work


maho7791:
With Maven workspaces the export doesnt work in den bndrun editor

maho7791:
Export runs successfully but doesn't export something

pkriens:
what is the state of this?

maho7791:
Should be fixed with PR: https://github.com/bndtools/bnd/pull/6041

PR #6041, opened 2024-02-23
refs #6033 - Add EXPORT to MavenRunProvider

maho7791:
possible fix for #6033

- added EXPORT to MavenRunPOrovider
- without the UI export exports an empty jar with just the launcher
- with EXPORT switch, the bundles from the bndrun will be exported as well


pkriens:
@maho7791 can we merge these PRs?

maho7791:
I just kept this in case @timothyjward wanted to look over it. I already use a fixed version.

timothyjward:
Looks fine - it's not a problem that I encounter really

Issue #6048, opened 2024-03-13
Ability to configure threshold for semantic versioning failures


reckart:
## Proposed change

Add the ability to configure a threshold for the semantic versioning check in the `bnd-baseline-maven-plugin` similar to  [semanticVersionLevel](https://siom79.github.io/japicmp/MavenPlugin.html) in japicmp.

## Context

I am trying to use the `bnd-baseline-maven-plugin` in my project to monitor my package-level `@Version` annotations.

However, I am running into a problem now. As part of my build, I have another tool that injects the current project version as a class-level annotation (`MyVersionAnnotation `) into my class files.

This seems to clash with the `bnd-baseline-maven-plugin` which then fails suggesting me that I should bump the micro level.

```
[ERROR] ===============================================================
[ERROR]   Name                 Type       Delta      New        Old        Suggest    
[ERROR]   mymodule             BUNDLE     MAJOR      7.0.0.202403111508 6.5.0      -          
[ERROR] ===============================================================
[ERROR]   Name                 Type       Delta      New        Old        Suggest    If Prov.
[ERROR] * mymodule             PACKAGE    MICRO      6.5.0      6.5.0      6.5.1      -
[ERROR]   MICRO                PACKAGE    mymodule
[ERROR]     CHANGED            CLASS      mymodule.MyClass
[ERROR]       CHANGED          ANNOTATED  MyVersionAnnotation
[ERROR]         REMOVED        PROPERTY   version='6.5.0'
[ERROR]         ADDED          PROPERTY   version='7.0.0-SNAPSHOT'
[ERROR]     ADDED              ANNOTATED  org.osgi.annotation.bundle.Export
[ERROR]     ADDED              ANNOTATED  org.osgi.annotation.versioning.Version
[ERROR]       ADDED            PROPERTY   value='6.5.0'
```

I don't want to bump the micro level just because this annotation changes. 

## Alternatives considered

- Add a parameter to exclude annotations in general or certain annotations from the comparison.
- Add `@BaselineIgnore` to the `@MyVersionAnnotation` - which I tried but didn't work

PR #6049, opened 2024-03-13
Issue #6048: Ability to configure threshold for semantic versioning failures

reckart:
- Added new parameter to baseline mojo
- Added additional test to BaselineTest
- Added additional baseline plugin integration test

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java
@@ -1,20 +1,25 @@
 package aQute.bnd.differ;
 
+import static aQute.bnd.service.diff.Delta.ADDED;
+import static aQute.bnd.service.diff.Delta.MAJOR;
+import static aQute.bnd.service.diff.Delta.MICRO;
+import static aQute.bnd.service.diff.Delta.MINOR;
+import static aQute.bnd.service.diff.Delta.REMOVED;
+import static aQute.bnd.service.diff.Delta.UNCHANGED;
+
 import java.io.IOException;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Formatter;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 import java.util.jar.Manifest;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import aQute.bnd.header.Attrs;
 import aQute.bnd.header.OSGiHeader;
 import aQute.bnd.header.Parameters;
 import aQute.bnd.osgi.Constants;
 import aQute.bnd.osgi.Descriptors;
+import aQute.bnd.osgi.Instruction;
 import aQute.bnd.osgi.Instructions;
 import aQute.bnd.osgi.Jar;
 import aQute.bnd.osgi.Processor;
@@ -26,9 +31,7 @@ import aQute.bnd.service.diff.Type;
 import aQute.bnd.unmodifiable.Sets;
 import aQute.bnd.version.Version;
 import aQute.libg.generics.Create;
-import aQute.service.reporter.Reporter;
-
-import static aQute.bnd.service.diff.Delta.*;
+import aQute.service.reporter.Reporter;;
 
 /**
  * This class maintains
@@ -89,21 +92,6 @@ public class Baseline {
 	 * @throws Exception
 	 */
 	public Set<Info> baseline(Jar newer, Jar older, Instructions packageFilters) throws Exception {
-		return baseline(newer, older, packageFilters, IGNORED);
-	}
-
-	/**
-	 * This method compares a jar to a baseline jar and returns version
-	 * suggestions if the baseline does not agree with the newer jar. The
-	 * returned set contains all the exported packages.
-	 *
-	 * @param newer
-	 * @param older
-	 * @return null if ok, otherwise a set of suggested versions for all
-	 *         packages (also the ones that were ok).
-	 * @throws Exception
-	 */
-	public Set<Info> baseline(Jar newer, Jar older, Instructions packageFilters, Delta threshold) throws Exception {
 		Tree n = differ.tree(newer);
 		Parameters nExports = getExports(newer);
 		Tree o = differ.tree(older);
@@ -111,15 +99,10 @@ public class Baseline {
 		if (packageFilters == null)
 			packageFilters = new Instructions();
 
-		return baseline(n, nExports, o, oExports, packageFilters, threshold);
+		return baseline(n, nExports, o, oExports, packageFilters);
 	}
 
 	public Set<Info> baseline(Tree n, Parameters nExports, Tree o, Parameters oExports, Instructions packageFilters)
-		throws Exception {
-		return baseline(n, nExports, o, oExports, packageFilters, IGNORED);
-	}
-
-	public Set<Info> baseline(Tree n, Parameters nExports, Tree o, Parameters oExports, Instructions packageFilters, Delta threshold)
 		throws Exception {
 		diff = n.diff(o);
 		Diff apiDiff = diff.get("<api>");
@@ -150,6 +133,8 @@ public class Baseline {
 
 			if (!packageFilters.matches(pdiff.getName()))
 				continue;
+
+			var threshold = getThreshold(packageFilters, pdiff);
 
 			final Info info = new Info();
 			infos.add(info);
@@ -186,7 +171,8 @@ public class Baseline {
 							break;
 					}
 
-					if (diff.getDelta().compareTo(threshold) < 0) {
+					if (threshold.map(d -> diff.getDelta()
+						.compareTo(d) < 0).orElse(false)) {
 						return true;
 					}
 
@@ -265,7 +251,8 @@ public class Baseline {
 				default -> MAJOR;
 			};
 
-			if (content.compareTo(threshold) < 0) {
+
+			if (threshold.isPresent() && content.compareTo(threshold.get()) < 0) {
 				content = UNCHANGED;
 			}
 
@@ -308,6 +295,25 @@ public class Baseline {
 		}
 
 		return infos;
+	}
+
+	private Optional<Delta> getThreshold(Instructions packageFilters, Diff pdiff) {
+
+		var matcher = packageFilters.matcher(pdiff.getName());
+		if (matcher == null || matcher.isNegated())
+			return Optional.empty();
+
+		var attrs = packageFilters.get(matcher);
+		assert attrs != null : "guaranteed by the matcher != null";
+		var threshold = attrs.getOrDefault(Constants.DIFFPACKAGES_THRESHOLD, "MICRO")
+			.toUpperCase().trim();
+		try {
+			return Optional.of(Delta.valueOf(threshold));
+		}
+		catch (IllegalArgumentException e) {
+			bnd.error("baseline.threshold baseline threshold specified as [%s] but does not correspond to a Delta enum - ignoring", threshold);
+			return Optional.empty();
+		}
 	}
 
 	/**
--- a/biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java
@@ -131,10 +131,11 @@ public class Baseline {
 				.startsWith("java."))
 				continue;
 
-			if (!packageFilters.matches(pdiff.getName()))
+			var matcher = packageFilters.matcher(pdiff.getName());
+			if (!packageFilters.isEmpty() && (matcher == null || matcher.isNegated()))
 				continue;
 
-			var threshold = getThreshold(packageFilters, pdiff);
+			var threshold = getThreshold(packageFilters, matcher);
 
 			final Info info = new Info();
 			infos.add(info);
@@ -171,8 +172,7 @@ public class Baseline {
 							break;
 					}
 
-					if (threshold.map(d -> diff.getDelta()
-						.compareTo(d) < 0).orElse(false)) {
+					if (threshold != null && diff.getDelta().compareTo(threshold) < 0) {
 						return true;
 					}
 
@@ -252,7 +252,7 @@ public class Baseline {
 			};
 
 
-			if (threshold.isPresent() && content.compareTo(threshold.get()) < 0) {
+			if (threshold != null && content.compareTo(threshold) < 0) {
 				content = UNCHANGED;
 			}
 
@@ -297,23 +297,20 @@ public class Baseline {
 		return infos;
 	}
 
-	private Optional<Delta> getThreshold(Instructions packageFilters, Diff pdiff) {
-
-		var matcher = packageFilters.matcher(pdiff.getName());
-		if (matcher == null || matcher.isNegated())
-			return Optional.empty();
-
+	private Delta getThreshold(Instructions packageFilters, Instruction matcher) {
+		if (matcher == null)
+			return null;
 		var attrs = packageFilters.get(matcher);
 		assert attrs != null : "guaranteed by the matcher != null";
 		var threshold = attrs.getOrDefault(Constants.DIFFPACKAGES_THRESHOLD, "MICRO")
-			.toUpperCase().trim();
+			.toUpperCase();
 		try {
-			return Optional.of(Delta.valueOf(threshold));
+			return Delta.valueOf(threshold);
 		}
 		catch (IllegalArgumentException e) {
 			bnd.error("baseline.threshold baseline threshold specified as [%s] but does not correspond to a Delta enum - ignoring", threshold);
-			return Optional.empty();
 		}
+		return null;
 	}
 
 	/**
--- a/biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java
@@ -1,12 +1,5 @@
 package aQute.bnd.differ;
 
-import static aQute.bnd.service.diff.Delta.ADDED;
-import static aQute.bnd.service.diff.Delta.MAJOR;
-import static aQute.bnd.service.diff.Delta.MICRO;
-import static aQute.bnd.service.diff.Delta.MINOR;
-import static aQute.bnd.service.diff.Delta.REMOVED;
-import static aQute.bnd.service.diff.Delta.UNCHANGED;
-
 import java.io.IOException;
 import java.util.Collection;
 import java.util.Collections;
@@ -34,6 +27,8 @@ import aQute.bnd.unmodifiable.Sets;
 import aQute.bnd.version.Version;
 import aQute.libg.generics.Create;
 import aQute.service.reporter.Reporter;
+
+import static aQute.bnd.service.diff.Delta.*;
 
 /**
  * This class maintains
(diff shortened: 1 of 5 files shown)


pkriens:
Please sign the commits: https://github.com/bndtools/bnd/pull/6049/checks?check_run_id=22621136417

pkriens:
First glance, it looks very good.

pkriens:
I did a closer look over the code and my problem is that it will have to be done very differently in the Gradle plugin & the bnd/Eclipse codebase. I think there is an easier way to achieve this which is also a bit more flexible and in line with other bnd code.

The baseline code takes an Instructions that can filter on packages, in bnd this is the [-diffpackages][1] instruction. This is the common [SELECTOR][2] Currently, it is only used to filter on the element name, e.g. package name. However, an Instructions is a `Map<Instruction,Attrs>`. If we make the `threshold` an attribute on the associated Attrs, we 

* do not have to change the API, and
* it works the same across the Gradle & bnd & maven drivers, and 
* you can selectively apply it on package name level.

If you want to keep the maven code simple, you could just make an instructions that matches all packages:
```
Instructions packageFilter = new Instructions("*;threshold=MICRO");
```
and pass this `packageFilter` instead of the null to the `baseline()` method. You could also allow the maven user to set the `packageFilter` so that they can leverage the selective application if they want to. This would be my preference since it would provide the same behavior across drivers.

The baseline method can then be changed to:
```
		for (Diff pdiff : apiDiff.getChildren()) {
			if (pdiff.getType() != Type.PACKAGE) // Just packages
				continue;

			if (pdiff.getName()
				.startsWith("java."))
				continue;

			Delta threshold = getThreshold(packageFilters, pdiff);
			if (threshold == null)
				continue;
```
With the additional methods like
```
	private Delta getThreshold(Instructions packageFilters, Diff pdiff) {

		Instruction matcher = packageFilters.matcher(pdiff.getName());
		if ( matcher == null || matcher.isNegated())
			return null;

		Attrs attrs = packageFilters.get(matcher);
		assert attrs != null : "guaranteed by the matcher != null";
		String delta = attrs.getOrDefault(Constants.DIFFPACKAGES_THRESHOLD, "MICRO")
			.toUpperCase();
		return getDelta(delta, Delta.MICRO);
	}

	private Delta getDelta(String delta, Delta deflt) {
		for (Delta d : Delta.values()) {
			if (d.name()
				.equals(delta))
				return d;
		}
                if ( bnd != null)
		  bnd.error(
			"baseline.threshold baseline threshold specified but does not compare to a Delta enum, using %s: %s", deflt,
			delta);
		return deflt;
	}
```
The rest of the code in Baseline is then exactly as you wrote it.

I know it sucks to take other drivers into account but I spent many a day struggling with maven for this reason even though I never use it ...

[1]: https://bnd.bndtools.org/instructions/diffpackages.html
[2]: https://bnd.bndtools.org/chapters/820-instructions.html#selector

reckart:
I don't really care if the threshold is a separate parameter or can be embedded into the selector. In fact, your selector approach would allow for a more fine-grained control which seems to be a good thing.

The PR should be updatable by you. You already seem to have the code ready. Would you mind pushing it into the PR?

reckart:
It's fine if you push only your changes to `Baseline` and the build fails. I can do the update to the Maven module (including tests) then.

pkriens:
Nope, that is not how it is going to work :-)

This is the merged Baseline class but it needs to be tested and it involves maven and there is a lot on my plate right now.
```
package aQute.bnd.differ;

import static aQute.bnd.service.diff.Delta.ADDED;
import static aQute.bnd.service.diff.Delta.MAJOR;
import static aQute.bnd.service.diff.Delta.MICRO;
import static aQute.bnd.service.diff.Delta.MINOR;
import static aQute.bnd.service.diff.Delta.REMOVED;
import static aQute.bnd.service.diff.Delta.UNCHANGED;

import java.io.IOException;
import java.util.Collection;
import java.util.Collections;
import java.util.Formatter;
import java.util.Map;
import java.util.Set;
import java.util.jar.Manifest;

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

import aQute.bnd.header.Attrs;
import aQute.bnd.header.OSGiHeader;
import aQute.bnd.header.Parameters;
import aQute.bnd.osgi.Constants;
import aQute.bnd.osgi.Descriptors;
import aQute.bnd.osgi.Instruction;
import aQute.bnd.osgi.Instructions;
import aQute.bnd.osgi.Jar;
import aQute.bnd.service.diff.Delta;
import aQute.bnd.service.diff.Diff;
import aQute.bnd.service.diff.Differ;
import aQute.bnd.service.diff.Tree;
import aQute.bnd.service.diff.Type;
import aQute.bnd.unmodifiable.Sets;
import aQute.bnd.version.Version;
import aQute.libg.generics.Create;
import aQute.service.reporter.Reporter;;

/**
 * This class maintains
 */
public class Baseline {
	private final static Logger			logger			= LoggerFactory.getLogger(Baseline.class);
	private final static Set<String>	BASELINEIGNORE	= Sets.of("aQute.bnd.annotation.baseline.BaselineIgnore");

	public static class Info {
		public String				packageName;
		public Diff					packageDiff;
		public Collection<String>	providers;
		public Map<String, String>	attributes;
		public Version				newerVersion;
		public Version				olderVersion;
		public Version				suggestedVersion;
		public Version				suggestedIfProviders;
		public boolean				mismatch;
		public String				warning	= "";
		public String				reason;
	}

	public static class BundleInfo {
		public String	bsn;
		public Version	olderVersion;
		public Version	newerVersion;
		public Version	suggestedVersion;
		public boolean	mismatch;
		public String	reason;
	}

	final Differ		differ;
	final Reporter		bnd;
	final BundleInfo	binfo	= new BundleInfo();

	Diff				diff;
	Set<Info>			infos;
	String				bsn;
	Version				newerVersion;
	Version				olderVersion;
	Version				suggestedVersion;
	String				releaseRepository;

	public Baseline(Reporter bnd, Differ differ) throws IOException {
		this.differ = differ;
		this.bnd = bnd;
	}

	/**
	 * This method compares a jar to a baseline jar and returns version
	 * suggestions if the baseline does not agree with the newer jar. The
	 * returned set contains all the exported packages.
	 *
	 * @param newer
	 * @param older
	 * @return null if ok, otherwise a set of suggested versions for all
	 *         packages (also the ones that were ok).
	 * @throws Exception
	 */
	public Set<Info> baseline(Jar newer, Jar older, Instructions packageFilters) throws Exception {
		Tree n = differ.tree(newer);
		Parameters nExports = getExports(newer);
		Tree o = differ.tree(older);
		Parameters oExports = getExports(older);
		if (packageFilters == null)
			packageFilters = new Instructions();

		return baseline(n, nExports, o, oExports, packageFilters);
	}

	public Set<Info> baseline(Tree n, Parameters nExports, Tree o, Parameters oExports, Instructions packageFilters)
		throws Exception {
		diff = n.diff(o);
		Diff apiDiff = diff.get("<api>");
		infos = Create.set();

		bsn = getBsn(n);

		newerVersion = getVersion(n);
		olderVersion = getVersion(o);

		final boolean binfoMismatch = mismatch(olderVersion, newerVersion);

		boolean firstRelease = false;
		if (o.get("<manifest>") == null) {
			firstRelease = true;
			if (newerVersion.equals(Version.emptyVersion)) {
				newerVersion = Version.ONE;
			}
		}
		Delta highestDelta = UNCHANGED;
		for (Diff pdiff : apiDiff.getChildren()) {
			if (pdiff.getType() != Type.PACKAGE) // Just packages
				continue;

			if (pdiff.getName()
				.startsWith("java."))
				continue;

			Delta threshold = getThreshold(packageFilters, pdiff);
			if (threshold == null)
				continue;

			final Info info = new Info();
			infos.add(info);
			info.reason = getRootCauses(pdiff);
			info.packageDiff = pdiff;
			info.packageName = pdiff.getName();
			info.attributes = nExports.get(info.packageName);
			logger.debug("attrs for {} {}", info.packageName, info.attributes);

			info.newerVersion = getVersion(info.attributes);
			info.olderVersion = getVersion(oExports.get(info.packageName));
			Delta delta = pdiff.getDelta();
			if (delta == UNCHANGED) {
				info.suggestedVersion = info.olderVersion;
				// Fix previously released package containing version qualifier
				if (info.olderVersion.getQualifier() != null) {
					info.suggestedVersion = bump(MICRO, info.olderVersion, 1, 0);
					info.warning += "Found package version with qualifier. Bumping micro version";
				} else if (!info.newerVersion.equals(info.olderVersion)) {
					info.warning += "No difference but versions are not equal";
				}
			} else if (delta == REMOVED) {
				info.suggestedVersion = null;
			} else if (delta == ADDED) {
				info.suggestedVersion = info.newerVersion;
			} else { // We have an API change
				// Process @BaselineIgnore annotations
				delta = pdiff.getDelta(diff -> {
					switch (diff.getDelta()) {
						case UNCHANGED :
						case IGNORED :
							return false;
						default :
							break;
					}

					if (diff.getDelta()
						.compareTo(threshold) < 0) {
						return true;
					}

					switch (diff.getType()) {
						case PACKAGE :
						case INTERFACE :
						case ANNOTATION :
						case CLASS :
						case ENUM :
						case FIELD :
						case METHOD :
							boolean ignore = diff.getChildren()
								.stream()
								.filter(child -> (child.getType() == Type.ANNOTATED)
									&& BASELINEIGNORE.contains(child.getName()))
								.flatMap(child -> child.getChildren()
									.stream())
								.filter(child -> child.getType() == Type.PROPERTY)
								.map(Diff::getName)
								.filter(property -> property.startsWith("value='"))
								.map(property -> property.substring(7, property.length() - 1))
								.anyMatch(version -> {
									try {
										return Version.valueOf(version)
											.compareTo(info.olderVersion) > 0;
									} catch (Exception e) {
										bnd.exception(e,
											"BaselineIgnore unable to compare specified version %s to baseline package version %s",
											version, info.olderVersion);
										return false;
									}
								});
							return ignore;
						default :
							return false;
					}
				});

				info.suggestedVersion = bump(delta, info.olderVersion, 1, 0);

				if (info.newerVersion.compareTo(info.suggestedVersion) < 0) {
					// our suggested version is greater than the new version!
					info.mismatch = mismatch(info.olderVersion, info.newerVersion);

					// We can fix some major problems by assuming
					// that an interface is a provider interface
					if (delta == MAJOR) {

						info.providers = Create.set();
						if (info.attributes != null)
							info.providers
								.addAll(Processor.split(info.attributes.get(Constants.PROVIDER_TYPE_DIRECTIVE)));

						// Calculate the new delta assuming we fix all the major
						// interfaces by making them providers
						Delta otherDelta = pdiff.getDelta(diff -> {
							if (diff.getType() == Type.INTERFACE && diff.getDelta() == MAJOR) {
								info.providers.add(Descriptors.getShortName(diff.getName()));
								return true;
							}
							return false;
						});

						if (otherDelta != MAJOR) {
							info.suggestedIfProviders = bump(MINOR, info.olderVersion, 1, 0);
						}
					}
				}
			}
			Delta content = switch (delta) {
				case IGNORED, UNCHANGED -> UNCHANGED;
				case ADDED -> MINOR;
				case CHANGED -> MICRO; // cannot happen
				case MICRO, MINOR, MAJOR -> delta;
				case REMOVED -> MAJOR;
				default -> MAJOR;
			};

			if (content.compareTo(threshold) < 0) {
				content = UNCHANGED;
			}

			if (content.compareTo(highestDelta) > 0) {
				highestDelta = content;
			}
		}
		// If this is a first release, or the base has a different symbolic
		// name, then the newer version must be ok. Otherwise the version bump
		// for bundles with the same symbolic name should be at least as big as
		// the biggest semantic change
		if (firstRelease || !bsn.equals(getBsn(o))) {
			suggestedVersion = newerVersion;
		} else {
			suggestedVersion = bumpBundle(highestDelta, olderVersion, 1, 0);
			if (suggestedVersion.compareTo(newerVersion) < 0)
				suggestedVersion = newerVersion;
		}

		binfo.bsn = bsn;
		binfo.suggestedVersion = suggestedVersion;
		binfo.olderVersion = olderVersion;
		binfo.newerVersion = newerVersion;

		if (newerVersion.getWithoutQualifier()
			.equals(olderVersion.getWithoutQualifier())) {
			// We have a special case, the current and repository revisions
			// have the same version, this happens after a release, only want
			// to generate an error when they really differ.

			if (getDiff().getDelta() == UNCHANGED)
				return infos;
		}

		// Ok, now our bundle version must be >= the suggestedVersion
		if (newerVersion.getWithoutQualifier()
			.compareTo(getSuggestedVersion()) < 0) {
			binfo.mismatch = binfoMismatch;
			binfo.reason = getRootCauses(apiDiff);
		}

		return infos;
	}

	private Delta getThreshold(Instructions packageFilters, Diff pdiff) {

		Instruction matcher = packageFilters.matcher(pdiff.getName());
		if (matcher == null || matcher.isNegated())
			return null;

		Attrs attrs = packageFilters.get(matcher);
		assert attrs != null : "guaranteed by the matcher != null";
		String delta = attrs.getOrDefault(Constants.DIFFPACKAGES_THRESHOLD, "MICRO")
			.toUpperCase();
		return getDelta(delta, Delta.MICRO);
	}

	private Delta getDelta(String delta, Delta deflt) {
		for (Delta d : Delta.values()) {
			if (d.name()
				.equals(delta))
				return d;
		}
		bnd.error("baseline.threshold baseline threshold specified but does not compare to a Delta enum, using %s: %s",
			deflt, delta);
		return deflt;
	}


	/**
	 * "Major version zero (0.y.z) is for initial development. Anything may
	 * change at any time. The public API should not be considered stable."
	 *
	 * @see <a href="https://semver.org/#spec-item-4">SemVer</a>
	 */
	private boolean mismatch(Version older, Version newer) {
		return older.getMajor() > 0 && newer.getMajor() > 0;
	}

	private String getRootCauses(Diff apiDiff) {
		try (Formatter f = new Formatter()) {
			getRootCauses(f, apiDiff, "");
			return f.toString();
		}
	}

	/*
	 * Calculate the root causes for this diff. This means that we descent
	 * through diffs with the same level (MAJOR/MINOR/MICRO) until we find an
	 * add/remove. The first add/remove is seen as the root cause.
	 */
	private void getRootCauses(Formatter f, Diff diff, String path) {
		for (Diff child : diff.getChildren()) {
			String cpath = path + "/" + child.getName();
			if (child.getDelta() == diff.getDelta()) {
				getRootCauses(f, child, cpath);
			} else {
				if (child.getDelta() == ADDED) {
					f.format("+ %s\n", cpath);
				} else if (child.getDelta() == REMOVED) {
					f.format("- %s\n", cpath);
				}
			}
		}
	}

	/**
	 * Gets the generated diff
	 *
	 * @return the diff
	 */
	public Diff getDiff() {
		return diff;
	}

	public Set<Info> getPackageInfos() {
		if (infos == null)
			return Collections.emptySet();
		return infos;
	}

	public String getBsn() {
		return bsn;
	}

	public Version getSuggestedVersion() {
		return suggestedVersion;
	}

	public void setSuggestedVersion(Version suggestedVersion) {
		this.suggestedVersion = suggestedVersion;
	}

	public Version getNewerVersion() {
		return newerVersion;
	}

	public Version getOlderVersion() {
		return olderVersion;
	}

	public String getReleaseRepository() {
		return releaseRepository;
	}

	public void setReleaseRepository(String releaseRepository) {
		this.releaseRepository = releaseRepository;
	}

	private Version bump(Delta delta, Version last, int offset, int base) {
		return switch (delta) {
			case UNCHANGED, IGNORED -> last;
			case MINOR -> new Version(last.getMajor(), last.getMinor() + offset, base);
			case MAJOR -> new Version(last.getMajor() + 1, base, base);
			case ADDED -> last;
			default -> new Version(last.getMajor(), last.getMinor(), last.getMicro() + offset);
		};
	}

	private Version getVersion(Map<String, String> map) {
		if (map == null)
			return Version.LOWEST;

		return Version.parseVersion(map.get(Constants.VERSION_ATTRIBUTE));
	}

	private Parameters getExports(Jar jar) throws Exception {
		Manifest m = jar.getManifest();
		if (m == null)
			return new Parameters();

		return OSGiHeader.parseHeader(m.getMainAttributes()
			.getValue(Constants.EXPORT_PACKAGE));
	}

	private Version getVersion(Tree top) {
		Tree manifest = top.get("<manifest>");
		if (manifest == null) {
			return Version.emptyVersion;
		}
		for (Tree tree : manifest.getChildren()) {
			if (tree.getName()
				.startsWith(Constants.BUNDLE_VERSION)) {
				return Version.parseVersion(tree.getName()
					.substring(15));
			}
		}
		return Version.emptyVersion;
	}

	private String getBsn(Tree top) {
		Tree manifest = top.get("<manifest>");
		if (manifest == null) {
			return "";
		}
		for (Tree tree : manifest.getChildren()) {
			if (tree.getName()
				.startsWith(Constants.BUNDLE_SYMBOLICNAME) && tree.getChildren().length > 0) {
				return tree.getChildren()[0].getName();
			}
		}
		return "";
	}

	private Version bumpBundle(Delta delta, Version last, int offset, int base) {
		return switch (delta) {
			case MINOR -> new Version(last.getMajor(), last.getMinor() + offset, base);
			case MAJOR -> new Version(last.getMajor() + offset, base, base);
			case ADDED -> new Version(last.getMajor(), last.getMinor() + offset, base);
			default -> new Version(last.getMajor(), last.getMinor(), last.getMicro() + offset);
		};
	}

	public BundleInfo getBundleInfo() {
		return binfo;
	}
}

**Comment by pkriens** [comment]
I'll update the docs if this is in.

**Comment by reckart** [comment]
It's exactly how it works. I'll take your code now, copy/paste it into my workspace and continue from there. The difference is only that you don't get a commit on the PR. The branch is nothing more than a collaborative scratchpad.

**Comment by reckart** [comment]
@pkriens ok, here we go with the updated code and tests

File: biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java [hunk 1]
```diff
@@ -136,6 +134,8 @@ public Set<Info> baseline(Tree n, Parameters nExports, Tree o, Parameters oExpor
 			if (!packageFilters.matches(pdiff.getName()))
```
pkriens, line 134: Not really happy how you change the code I gave you. This is less efficient (you search the Instructions twice now) and though Optional can be nice in APIs, this is not a proper use case imho. You get multiple boxing/unboxing and a simple comparison becomes numerous calls with intermediate objects.
It is not a major deal here but in a tool like bnd you'd like to keep operations that are called often as efficient as possible. And this does not even increase readability by adding more code.

biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java
```diff
@@ -285,6 +297,25 @@ public Set<Info> baseline(Tree n, Parameters nExports, Tree o, Parameters oExpor
...
+	private Optional<Delta> getThreshold(Instructions packageFilters, Diff pdiff) {
+
+		var matcher = packageFilters.matcher(pdiff.getName());
+		if (matcher == null || matcher.isNegated())
+			return Optional.empty();
+
+		var attrs = packageFilters.get(matcher);
+		assert attrs != null : "guaranteed by the matcher != null";
+		var threshold = attrs.getOrDefault(Constants.DIFFPACKAGES_THRESHOLD, "MICRO")
+			.toUpperCase().trim();
```
pkriens, line 309: trim() is unnecessary. This is always properly handled by the Parameters parser.

pkriens (approved):
I am good to go.

biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java
reckart: I find your approach less intelligible because `getThreshold()` suddenly has an implicit side-effect, namely that if there is no threshold, there is no rule. And for some reason I find difficult to trace down, that logic is broken as evidenced by the fact that when I use your code as-is, it breaks a bunch of test cases in `BaselineTest`.

I'll remove the optional and arrange the code slightly differently.

reckart:
I have the impression the build failure of this PR atm has nothing to do with the changes here. The master seems to have failed in the same way.

pkriens:
there were communication failures. Its ok now.

I am good to go, shall I merge?

reckart:
Let me just push the changes to remove the optional etc. I just need to wait for the local tests to complete.

reckart:
@pkriens changes pushed - I hope that addresses your comments sufficiently.

pkriens:
thanks, I did not require it but appreciate it. I will merge if the action is finished.

reckart:
Thank you for the merge 👍

pkriens:
thank you, nice feature to add to bnd :-)

PR #6069, opened 2024-04-03
BndEditor build.bnd: Plugin list should detect plugins by prefix

chrisrueger:
Currently the Plugin list is empty when plugins are defined like:

```
-plugin.0.Main....
-plugin.1.Eclipse....
-plugin.9.Baseline....
```
image

This PR is supposed to change it, so that it becomes:

image

Additionally the MavenBndRepository should be a proper editor like this:

image


biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
```diff
@@ -967,14 +969,67 @@ public void setTestSuites(List<String> suites) {
 	}
 
 	public List<HeaderClause> getPlugins() {
-		return doGetObject(Constants.PLUGIN, headerClauseListConverter);
+		// return all plugins
```
pkriens, line 972: Not sure if this works with inheritance of properties & bnd file properties? It is a powerful but nasty problem
   
     file 1->* 'plugin.*' property 0->* clause 

To make this work I think you need to model this as an object. When you edit a file, you can edit the local ones in the BndEditoModel but I think the inherited ones should be visible greyed out?

bndtools.core/_plugin.xml
```diff
@@ -871,6 +871,23 @@
 				description="Disable verification of server's X509 certificate (insecure, for testing only!)" />
 		</plugin>
 
+		<plugin class="aQute.bnd.repository.maven.provider.MavenBndRepository"
```
pkriens, line 874: I had no idea this existed!

pkriens (commented):
Quite amazing you got this to work in so little code.

However, I think we need a generic solution in the BndEditModel for merged & inherited properties. Inherited properties should be greyed out but when you add properties it is important to realize they exist and what their value is.  (Double clicking could open an editor on their file.)

It seems to screem for a MergedProperty class that handles this generically and can parameterize the type specific interactions? This is a common problem and I think it is a good idea to see how for example `-buildpath` would look like in this way?

Fantastic work! This has been a sore point forever.

bndtools.core/_plugin.xml
chrisrueger: You mean those sections in the `_plugin.xml` ?

Yes, that is how I came to this PR? I was trying the +-Button of the Plugins list, because I wanted to see what this is... because I have never seen anything in the Plugins list  of my build.bnd. And tadaaa I saw that I could add repositories. Unfortunately I did not see the MavenRepo we are using. 
 
We can later add also support for other repositories if needed. The xml I created from `aQute.bnd.repository.maven.provider.Configuration`. I guess  we can do the same for `aQute.bnd.repository.maven.pom.provider.PomConfiguration`.

chrisrueger:
@pkriens I have pushed another commit

> Inherited properties should be greyed out

image

image

This is a "hacky" attempt to implement 
- showing inherited plugins, 
- but grey out the ones which are inherited from another `.bnd` file
- editing and removing is `disabled` for such greyed out entries
- a tooltip is shown for the greyed out entries (see screenshot)

Although I created a class `MergedHeaderClause` which "knows" if the headerClause is from the `local` file or `inherited` I have a strong feeling it is not what you had in mind with

> It seems to screem for a MergedProperty class that handles this generically and can parameterize the type specific interactions? This is a common problem and I think it is a good idea to see how for example `-buildpath` would look like in this way?


But let's discuss based on this and make it better from there.

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
```diff
@@ -967,14 +969,117 @@ public void setTestSuites(List<String> suites) {
...
+	 * property key of the bnd file e.g.
+	 * <code>-plugin.1.Test, -plugin.2.Maven </code> The value is a List of
+	 * plugins, although usually it is just a 1-element list. But it is also
+	 * possible to specify multiple plugins under a single key, thus it is a
+	 * list.
+	 *
+	 * @return a map with a property keys and their plugins.
+	 */
+	public Map<String, List<MergedHeaderClause>> getPluginsProperties() {
+		// return all plugins
```
pkriens, line 1010: I think this logic should be moved to Processor, next to `mergeProperties` so it is easier to track when we make changes in one or the others. Since this is processor, I think we need a call like:

    Processor.getMergedParts(String keyStem) : MergedPart(Processor p, String keyStem, String actualKey, String value)

We could even rewrite ( one day) the mergedProperties with this method. In the BndEditModel, you then turn each one into a number of constituent clauses that can be edited if the processor == the BndEditModel's processor. This would be similar to your MergedHeaderClause, but I would make it something like:

     MergedHeaderClause( MergedPart part, HeaderClause), `isLocal()` can check the part if it is the same Processor.

chrisrueger:
@pkriens I experimented a bit with the `Processor.getPropertyKeys(Predicate)` and `Processor.getMergePropertyKeys(stem)`

Here is what I did in BndEditModel:

```
Processor owner = getOwner();
List<PropertyKey> propertyKeysO = owner.getPropertyKeys(key -> key.startsWith(Constants.PLUGIN));
```

in getPropertyKeys() I added some debug output inside the `while` loop to see whats going on:

```
System.out.println(
				"level: " + localLevel + " propFile: " + localRover.getPropertiesFile() + " localRover: " + "id=("
					+ System.identityHashCode(localRover) + " " + localRover + " keys:" + list);
```

image


```
Processor owner = getMergePropertyKeys id:1334679764
level: 0 propFile: /cnf/build.bnd localRover: id=(1334679764 build.bnd keys:[-plugin.4.Test, -plugin.6.Central, -plugin.foo, -plugin.7.Local]
level: 1 propFile: null localRover: id=(546604699 Workspace [bndtools.core] keys:[-plugin.6.Central, -plugin.foo, -plugin.4.Test, -plugin.7.Local]
level: 2 propFile: null localRover: id=(1039960084  keys:[]
level: 3 propFile: null localRover: id=(1192082665  keys:[]
```

- You see it traverses a chain of `4` levels
- level 0 and level 1 contain the same keys (but different order) - odd?
- level 1 Processor (the parent) is "Workspace" ... the other processes don't print a name in `toString` and it has all the properties again. (is this expected?)

Important: 
- `key=`-plugin.foo`` is coming from an `-include:foo/bar.bnd`. So it is **not** defined in `build.bnd` 

It seems that `rover.stream(false) // local only` is not only pulling local stuff, but also included stuff. 
ok I can understand that an `-include` might be `local` for a `Processor` but in terms of "what is inside the actual `build.bnd` file, it is not local. 


## Why is this a problem? 

`PropertyKey.isLocalTo(processor)` seems to have a different meaning of `local` than my previous approach:

`boolean isLocal = doGetObject(pk.key(), headerClauseListConverter) != null;`

The latter gives me "true" for everything physically in the `build.bnd` file. 
`key=`-plugin.foo`` is not `physically` in the build.bnd file (it is in `-include:foo/bar.bnd`). So it is not local for the purpose of the editor. 

That means, the new stuff does not give me an advantage over my previous approach. I was hoping the the new `PropertyKey.isLocalTo()` is the same as `boolean isLocal = doGetObject(pk.key(), headerClauseListConverter) != null;`

## What's next

Just wanted to write this down.  It helps me to think this more through. 
Let's discuss my findings during the next days/

- Maybe I am just using it wrong.
- or we are using the wrong words (locally in the physical file build.bnd vs locally for the current processor, which also reads `-include`s)
- or my expectations are wrong

chrisrueger:
Thinking out loud.... Adding to the above:
It seems the `BndEditModel.properties` is special. Because it is the only properties object I found in all the above which is purely loaded `un-process(or)ed` from the `build.bnd` 

`properties.load(inputStream);`

image


... so it contains the pure content of that file...without any inheritance or include loading.

Maybe the `Processor` or `PropertyKey` also needs something like those unprocessed properties.

pkriens:
I had the nagging feeling this would happen. The same is also true for the `ext` directory, they are 'included' in the `build.bnd`.

The only way to detect, as we had discussed, is to look in the edit model. It actually reflects the contents in the editor. So you cannot use isLocal in your case, the original method you had was better.

I'll fix the NPE and see why it is null.

pkriens:
I prefer not to add the information if a property is from the file or not. This would involve a lot of work and somehow it makes intuitive sense to use the BndEditModel for this since it is really on the file. Trying to track the pedigree of properties feels like a rats nest.

chrisrueger:
> I prefer not to add the information if a property is from the file or not. This would involve a lot of work and somehow it makes intuitive sense to use the BndEditModel for this since it is really on the file. Trying to track the pedigree of properties feels like a rats nest.

Understandable and totally fine for me.

But, the only thing I am wondering now:

- for my **specific problem** I do not get much value from the new `PropertyKey`. In fact also the duplicates in the list cause more problems. the "old" `Set<String> processor.getPropertyKeys(true)` removed the duplicates. At least in my case this is what I needed. I could then filter out the local ones.

Do you see use-cases where the new List<PropertyKeys> (which contains duplicates... from different processors) has a value? Maybe you already have something in mind where you need exactly this behavior. 

> BndEditModel for this since it is really on the file

Given this statement, I wonder if we should reevaluate the requirement to show _greyed out_ (non-in-the-file) plugins in the plugin list. 

Although I can understand the wish to see them, I don't know if it somehow blurs the line between being 

- a) "local on the file" 
vs. 
- b) showing an `effective` combined list of plugins where only a subset is `local` to the file. 

b) adds a certain complexity and derives from the current "what you see is what you get" behavior of the BndEditor. 
Maybe this is a bit philosophical, but maybe it is ok for an "editor" to only show the `local` list for the file you have opened.  Maybe the `effective` list is something which is interesting to see "somewhere", but maybe not in an "editor"?

I don't want to discuss away what we have. I already have the 'greyed out' stuff implemented. I just see that it is more complex than the initial approach (only local stuff).

chrisrueger:
@pkriens see cd6147f with my additions of stuff we talked about earlier. 

image

chrisrueger:
I did the following additions:

- added more repository plugins to the list: POMRepository and P2Repository and moved MavenBndRepo to the top (since this is mentioned in most tutorials, it may be the first / default? )

image

- make the Help Button work and open the Manual page in the browser. For that I introduced a `helpUrl` parameter in _plugin.xml where the repositories are defined

image

I thinks this is ready for review / merge now. @pkriens

chrisrueger:
Hi @pkriens , how do we want to proceed here?

pkriens:
Hi, I haven't forgotten you. However, I spent a lot of time making the BndEditModel clean. I have a test case in the resolver where I can't figure out why I get 4 bundles in the old case and only 2 in the new case. The Processor is identical and the debug trace is identical :-( Drives me crazy.

Then there was an issue with project creation which I urgently needed to fix.

I am ok to merge this. I will then take this and integrate the new BndEditModel and then we can see where we are. Ok?

chrisrueger:
@pkriens no problem. Merging is totally fine if you don't see any obvious things. Thanks.

PR #6073, opened 2024-04-04
Add support for PropertyKeys, where these records pair the Processor/key

pkriens:
@chrisrueger is working on editing the plugins. Since merged properties, the edit panes in the build.bnd 
editor and the bnd.bnd editor are a bit of a mess. 

This change in processor adds the concept of a PropertyKey, where the Property Key links the Processor that defines the property and the actual key. So a merged property is translated to a properly sorted set of ProperyKeys and can then be edited with the BndEditModel.

Unfortunately, the BndEditModel had no proper knowledge of the Processor. For this reason, a findProcessor method was added to Workspace and Project to find the a Processor that is properly linked with inheritance.

---


pkriens:
@chrisrueger I am now off, but if this has an ok build, feel free to  merge tonight if you're moonlighting ....

PR #6082, opened 2024-04-10
Update of BndEditModel

pkriens:
Now properly reflects the inheritance structure of Processor chains

@chrisrueger I made a few changes to your code, please verify
---

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
@@ -29,7 +29,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
@@ -175,7 +174,7 @@ public class Processor extends Domain implements Reporter, Registry, Constants, 
 	}
 
 	public Processor(Processor parent) {
-		this(parent, parent.getProperties0(), true);
+		this(parent, parent.getRawProperties(), true);
 	}
 
 	public Processor(Properties props, boolean wrap) {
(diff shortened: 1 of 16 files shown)


biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
```diff
@@ -1061,12 +1066,19 @@ public int compareTo(PropertyKey o) {
 		 * @param keys sorted keys defined by {@link #compareTo(PropertyKey)}
```
chrisrueger, line 1066: @pkriens I see you are sorting in the method now.
Maybe then the "sorted" for `keys` should be removed from the javadoc

chrisrueger (approved):
LGTM. The changes to my code seem minor. `test.ProcessorTest.testPropertyKeysFindVisibles()` seems still working.
Besides my comment about the javadoc of `findVisibles()` I have nothing else.

I checked out the code locally and did a few manuial tests. still working.

chrisrueger:
I noticed one new warning recently, but could not figure out when it came in:

image

```
Description	Resource	Path	Location	Type
No translation found for macro: .	bnd.bnd	/bndtools.core	line 1	Bndtools Problem Marker

```

pkriens:
I saw the warning. I could also not find it. Will investigate

pkriens:
Found the warning. You added two lines to the `_plugin.xml` with ${.}:
```
		    <property name="pom" type="string" description="Points to a pom.xml file. This is exclusive with revision. Can be a comma separated list of files." default="${workspace}/cnf/pom.xml" />
```

The ${.} is a very special macro. It is handled with included files during inclusion because we know the file location then. In the pre processing of a file with the macro processor we have no idea where we are.

We might want to think about supporting this in the `-includeresource` process ...

pkriens:
Misunderstood the use of the ${.}. It had to be escaped and i think you do not want to have cnf in the path. ${.} refers to the directory of the preprocessed file.

chrisrueger:
> Misunderstood the use of the ${.}. It had to be escaped and i think you do not want to have cnf in the path. ${.} refers to the directory of the preprocessed file.

Thanks for finding this.

PR #6112, opened 2024-05-05
Add search bar to Unresolved requirements in ResolutionFailedPanel

chrisrueger:
This PR adds a search bar to this view:

image

I found that helpful, because before I could never make sense of that list, because I couldn't "see" / or "find" anything (can't see the forest for the trees). 

Now the search bar at least allows to search for anything from the error message above, and maybe find what is of interest for you. 

Note on search bar positioning: The search bar currently is at the bottom, because I could't find a way to add it below the "Unresolved requirements" header. I managed to add it right _above_ the "Unresolved requirements" header, but there I did not like it because it does not look like it belongs to the "Unresolved requirements".

Maybe because it is a collapsible tree or some view panel which does not allow this. I don't know.
I personally don't find that bad, but it is a bit different than other views where the search bar is at the top.


pkriens:
LGTM, can we merge?

chrisrueger:
Yes

PR #6116, opened 2024-05-08
Add support for repository tags

### LINKED ISSUE #5966: Add runrepos blacklist
One case I often run into is, that the resolver also pulls artifacts from the baseline repo, which is usually only for baselining. I need a mechanism to generically exclude this repo. Right now, we only have a whitelist, but a Blacklist would be much better, as repos usually change over time.

chrisrueger:
Closes #5966

Based on brainstorming here https://bnd.discourse.group/t/baseline-repo-exclude-from-the-build-path/391/6

## In this PR 

- allow to add a property `tags="baseline,release, foo"` to a Repository in `build.bnd`
- default tag in all repositories without a tag is an empty set of tags for backwards compatibility
- in a `.bndrun` if you remove the `-runrepos` you get all repositories which have either no tags set or the 'resolve' tag
- To exclude a repo from resolution in the example, you would not add the `resolve` (but instead any other tag e.g. `<<EMPTY>>` and remove the `-runrepos` from the `.bndrun`. 
- If you have specified `-runrepos` then today's behavior is still in place that the given repo names are used

e.g. a `.bndrun` file without  `-runrepos`  would include the following repos in the resolution:

```
aQute.bnd.repository.maven.provider.MavenBndRepository;\
        tags = "resolve"; \
        name="Maven Central A";\
        releaseUrl="${mavencentral}";\
        snapshotUrl="${ossrh}";\
        index="${.}/central.mvn";\
        readOnly=true,\
```

because it has the `resolve` tag

and 

```
aQute.bnd.repository.maven.provider.MavenBndRepository;\
        name="Maven Central B";\
        releaseUrl="${mavencentral}";\
        snapshotUrl="${ossrh}";\
        index="${.}/central.mvn";\
        readOnly=true,\
```

because there is no tags

but **not**

```
aQute.bnd.repository.maven.provider.MavenBndRepository;\
        tags = "someothertag"; \
        name="Maven Central";\
        releaseUrl="${mavencentral}";\
        snapshotUrl="${ossrh}";\
        index="${.}/central.mvn";\
        readOnly=true,\
```

because it has a **no** `resolve` tag.

To exclude the baseline repo in the example, you would not add the `resolve` tag to the baseline-repo, and then it won't be considered for Resolving.  I have not used baselining yet, so I am not familiar with the topic and what areas are not covered by this.

The current state of the PR just lay the foundation for tagging. More features could be added like allowing for NOT or more sophisticated filter expressions.


chrisrueger:
## Summary from the [bnd zoom call](https://bnd.discourse.group/t/you-wanna-be-a-bnd-developer/443) today:

- revert the Respositories Panel support for tags (e.g. `@resolve`) in `-runrepos`... just keep it a list of workspace names and a way to be specific
- start a step back with an assumed empty `-runrepos` list or no `-runrepos` at
  - in this case the repostories should be fetched via something like `repos = getPlugins(..., "resolve")` - the second param should be the tag `resolve` which should be one a handful of fixed core-tags
  - every Repository should automatically have this tag `resolve` tag

That's it so far. I will post an update once I have this implemented and from there we can discuss the next step. 

Thanks @pkriens for the input.

chrisrueger:
> * revert the Respositories Panel support for tags (e.g. `@resolve`) in `-runrepos`... just keep it a list of workspace names and a way to be specific
> * start a step back with an assumed empty `-runrepos` list or no `-runrepos` at
>   
>   * in this case the repostories should be fetched via something like `repos = getPlugins(..., "resolve")` - the second param should be the tag `resolve` which should be one a handful of fixed core-tags
>   * every Repository should automatically have this tag `resolve` tag

@pkriens I have implemented the above. 

- if `-runrepos` is empty (or not in the `.bndrun` file) the UI shows all repos having the 'resolve' tag
  - if you don't touch any checkbox, then it stays that way
  - if you touch a checkbox then the repo-names are written to `.bndrun` and basically override the 'resolve' tag behavior. 
  - I guess to make use of the new tagging the recommendation would be to remove `-runrepos` from `.bndrun` to make use of the tags

Example: this is my Repositories list with no `-runrepos` specified

image

It is basically all repos with the resolve tag (repos have not tags specified, so the default 'resolve' is used)

Resolution:
- basically uses a similar behavior
- if `-runrepos` is empty (or not in the `.bndrun` file) resolution uses all repos having the 'resolve' tag. Otherwise the `-runrepos` are used

Quoting commit b9a574f

- repos now get the 'resolve' tag by default if not specified
- .bndrun: empty -runrepos will be populated with all repos having the 'resolve' tag
- .bndrun: resolution consider all repos having the 'resolve' tag. that means you can exclude a repo from resolution by manually assigning it a different tag (e.g. to exclude the baseline repo from resolution, you should give the baseline-repo e.g. the tag 'baseline' and make sure it does NOT have the 'resolve' tag

 

Let's discuss

biz.aQute.bndlib/src/aQute/bnd/service/Tagged.java
```diff
@@ -0,0 +1,60 @@
...
+import java.util.Set;
+
+/**
+ * Allows to add tags to implementing classes. Originally intended for tagging
+ * repositories.
+ */
+public interface Tagged {
+
+
+	enum RepoTags {
```
pkriens, line 17: Why not move this to the RepositoryPlugin? The repositories can know about tags but I am a bit uncomfortable with tags knowing about repositories. Seems to break through the layering. And I always like to consider what happens when we get 50 subsystems like repository, do we have to list them all here?

Advantage is of course that it is easier to handle name clashes, but it always amazes me how easy it is to make names pretty unique without having to resort to UUIDs. France has 60,000 cities and villages, and although there are 25 Beaulieu's (beautiful place), clashes are actually quite rare.

biz.aQute.bndlib/src/aQute/bnd/service/Tagged.java
```diff
@@ -0,0 +1,60 @@
...
+		 */
+		resolve
+		// add more if neded e.g. relase, baseline
+	}
+
+	/**
+	 * Each repo has by default the tag {@link RepoTags#resolve} if not tags are
+	 * set at the repo definition in build.bnd That means it is consider
+	 */
+	Set<String> DEFAULT_REPO_TAGS = Set.of(resolve.name());
```
pkriens, line 31: this should also got toe the RepositoryPlugin interface.

biz.aQute.bndlib/src/aQute/lib/deployer/FileRepo.java
```diff
@@ -252,6 +259,7 @@ interface Config {
 	String									name;
 	boolean									inited;
 	boolean									trace;
+	String									tags;
```
pkriens, line 262: I'd make this Set<String>.

I am leaning to make this a domain object, like `Tags` but there are two forces. Tags membership testing could be highly optimized and maybe one day we need logic operators like union and subset. It is also more efficient since you now parse the string every time someone calls getTags. 

Then again, Set<String> is very close to the concept of this object and is standard. And then we could make Tags implement Set<String> ... Tough :-)

Anyway, I do not think we should keep the raw string that would require parsing all the time.

biz.aQute.repository/test/aQute/bnd/repository/maven/provider/WorkspaceTest.java
```diff
@@ -57,6 +64,37 @@ public void testEnv() throws Exception {
 		assertNotNull(workspace);
 		assertNotNull(repo);
 		System.out.println(workspace.getBase());
+
+		// check repo tags
+		// repos should have the 'resolve' tag by default if no tag is specified
+		List<Repository> repos = workspace.getPlugins(Repository.class);
+		List<Repository> resolveRepos = workspace.getPlugins(Repository.class, RepoTags.resolve.name());
```
pkriens, line 71: The enum usage instead of a String constant does not really help, don't you agree?

biz.aQute.bndlib/src/aQute/bnd/osgi/repository/BaseRepository.java
```diff
@@ -25,16 +25,18 @@
 
 import aQute.bnd.exceptions.Exceptions;
 import aQute.bnd.osgi.resource.ResourceUtils;
+import aQute.bnd.service.Tagged;
 
 /**
  * WARNING ! Not tested
  */
-public abstract class BaseRepository implements Repository {
+public abstract class BaseRepository implements Repository, Tagged {
```
pkriens, line 33: Maybe RepositoryPlugin should implement Tagged, with a default implementation of no tags?

chrisrueger: Thanks Peter. Good Points. I'll give it another try when I'm back.

biz.aQute.bndlib/src/aQute/bnd/service/Tagged.java
chrisrueger: Removed the RepoTags enum and added `Constants.REPOTAGS_RESOLVE` and `aQute.bnd.service.RepositoryPlugin.DEFAULT_REPO_TAGS`

Thus, 
- Tags don't know about Repos
- but Repos know about tags

biz.aQute.bndlib/src/aQute/bnd/service/Tagged.java
chrisrueger: Done

biz.aQute.bndlib/src/aQute/lib/deployer/FileRepo.java
chrisrueger: I made it the new Tags class which is a Set.

biz.aQute.bndlib/src/aQute/bnd/service/Tags.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.Collection;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+/**
+ * A set of tags. A tag is a string-token which can be attached to an entity for
+ * categorization and filtering. Typically these entities then implement the
+ * {@link Tagged} interface.
+ */
+public final class Tags extends LinkedHashSet<String> {
```
pkriens, line 15: Not a good idea to extend this class because it is read-write. I'd use an internal set and implement Set<String>. Also make sure the strings are sorted to keep the build repeatable.

biz.aQute.bndlib/src/aQute/bnd/service/RepositoryPlugin.java
```diff
@@ -18,7 +19,15 @@
...
  */
-public interface RepositoryPlugin {
+public interface RepositoryPlugin extends Tagged {
+
+	/**
+	 * Each repo has by default the tag {@link Constants#REPOTAGS_RESOLVE} if no
+	 * tags are set at the repo definition in build.bnd That means it is
+	 * consider
+	 */
+	Tags DEFAULT_REPO_TAGS = Tags.of(Constants.REPOTAGS_RESOLVE);
```
pkriens, line 29: Hmm, this requires collaboration of the plugins. If a user implemented RepositoryPlugin he will not automatically get this default. 
The only way I know how to handle these backward compatibility cases is to treat the empty set as explicitly different from the side of the caller has a behavior for the empty set, not the callee because it has no clue how it is going to be used.

Once there is a single, even a dummy, tag, the caller can assume the user is aware of tags and is in charge. The empty set indicates the user is unaware or the plugin does not yet participate. In both case the default behavior of the _caller_ should apply.

To have a default content, you remove the possibility on the caller side to see the distinction between repos that have not (yet) been tagged and tagged. I think this is a crucial distinction?

biz.aQute.bndlib/src/aQute/bnd/service/Tags.java
chrisrueger: Ok I will change it and implement set using a SortedSet internally.

biz.aQute.bndlib/src/aQute/bnd/service/RepositoryPlugin.java
chrisrueger: > If a user implemented RepositoryPlugin he will not automatically get this default.

I added a default method in `RepositoryPlugin.java` for this purpose:

https://github.com/chrisrueger/bnd/blob/add-repository-tags/biz.aQute.bndlib/src/aQute/bnd/service/RepositoryPlugin.java#L292

Maybe we should have a call about this, to get me on the same page:) My current assumption about "default-behavior" was, that all repositories by default should behave as if they have the "resolve" tag - because they should take part in the resolution. I think I interpreted the sentence somewhere... don't know where anymore. 
But anyway I am absolutely open and agree that we should first define what "default" means and then implement the right thing. 

> I think this is a crucial distinction?

Yes.  Let's stare at code together in call quick call. Today does not work for me within the next days it is possible.

biz.aQute.bndlib/src/aQute/bnd/service/Tags.java
chrisrueger: Added https://github.com/bndtools/bnd/pull/6116/commits/7857d51a22937b0a338938715dc826721b437154

chrisrueger:
@pkriens I have pushed the changes from our call today. 
In the UI I have implemented a simple Combobox mechanism where you can provide the options using a comma-separated string. 

see 7f04909

image

Issue #6131, opened 2024-05-24
Compile error on JDK 21: aQute/lib/collections/SortedList doesn't fully implement SortedSet/List interfaces


neeshy:
Due to conflicting default method implementations:

```
aQute.libg/src/aQute/lib/collections/SortedList.java:31: error: types List<T> and SortedSet<T> are incompatible;
public class SortedList<T> implements SortedSet<T>, List<T> {
       ^
  both define reversed(), but with unrelated return types
  where T is a type-variable:
    T extends Object declared in class SortedList
```

Should be as simple as implementing these methods:

```
T getFirst()
T getLast()
void addFirst(T e)
void addLast(T e)
T removeFirst()
T removeLast()
SortedList<T> reversed()
```

pkriens:
Can you provide a PR?

neeshy:
Sure: #6134

PR #6134, opened 2024-05-28
Implement new JDK21 SortedSet and List methods for SortedList

neeshy:
As per #6131

### DIFF
--- a/aQute.libg/src/aQute/lib/collections/SortedList.java
+++ b/aQute.libg/src/aQute/lib/collections/SortedList.java
@@ -358,11 +358,19 @@ public class SortedList<T> implements SortedSet<T>, List<T> {
 		return get(0);
 	}
 
+	public T getFirst() {
+		return first();
+	}
+
 	@Override
 	public T last() {
 		if (isEmpty())
 			throw new NoSuchElementException("last");
 		return get(size() - 1);
+	}
+
+	public T getLast() {
+		return last();
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


kriegfrj (changes requested):
Thanks for the update, @neeshy ! Looks like the compile fails because on Java 17 these methods do not exist. I think that we still need to preserve Java 17 compile compatibility. So perhaps a solution here will be to comment out the @Override line.

I suspect that this change will also require you to do a minor version update in `package-info.java` from 4.2.0 to 4.3.0:

https://github.com/bndtools/bnd/blob/master/aQute.libg/src/aQute/lib/collections/package-info.java

The build never got that far because of the compile error though.

neeshy:
I've removed the `@Override` lines and it seem to build fine without the version update in `package-info.java`.

chrisrueger:
> it seem to build fine without the version update in `package-info.java`.

Unfortunately not.
See failing build https://github.com/bndtools/bnd/actions/runs/9265278739/job/25492215091?pr=6134

kriegfrj:
Also, when you re-submit this fix, can you please massage your commit message to include "Fixes #6131" on the bottom line, so that the full message looks something like this:

```
[libg] Implement new JDK21 SortedSet and List methods for SortedList

Fixes #6131
```

That way we can use the commit history to refer back to all of this discussion for historical context if necessary, and as an added bonus when we merge it it will automatically close the associated issue. Adding the `[libg]` at the front, while not as important or enforced, is something that I have personally done in the past as it also allows us to see at a glance which part of the workspace the commit is for.

neeshy:
Ah, I was building it through my package manager which I can only assume suppresses that check. Pushed with the suggested changes.

kriegfrj:
Thank you so much, @neeshy ! Congratulations on earning your honorary "first contribution" badge!

PR #6135, opened 2024-05-28
Manual: Add section about Resolving Multi-Release JAR Files

chrisrueger:
Based on and almost quote of answer by @juergen-albert  in https://bnd.discourse.group/t/questions-about-resolving-errors/450/3

I found your answer very good and thought this should be in the manual. 
Open for feedback or additions.


juergen-albert:
LGTM. I'd suggest @pkriens takes a final look, as he wrote most of the code around it.

pkriens:
This is not completely correct. There is only one _bundle_ resource. This was a discussion when we developed the multi release jar support. @timothyjward wanted to model it as different bundles. 


A Multi-Release JAR (MRJ) has directories in `META-INF/versions/<release>`. When the JAR is deployed on a VM with a given release _R_, that VM will preferentially load file resource from its own release _R_ and then down to the main area. This means that a JAR can have different content depending on the VM it is deployed. For OSGi, this means that a bundle resource can have different requirements based on the runtime VM. The capabilities are the same since the public API must not change.

The problem is that this makes some requirements dependent on the VM. To model this, we introduced synthetic resource for each supported VM release that can only resolve on that release and when the release specific requirements of the bundle can be resolved.

This is only about modeling a JAR as a resource in an OSGi repository. Nothing is changed in the original JAR.

So we first treat the _multi release_ bundle as any other bundle when we turn it into a resource, the release directories are ignored. If there are none, we're actually done.

Otherwise, we add a single requirement to a `bnd.multirelease` capability unique for this bundle. This namespace only has the properties `bnd.multirelease=<bsn>` and `version=<version>`. By requiring another resource, we can generate a synthetic resource for each VM release that the bundle supports. This is depicted here, where mrj is bnd.multirelease.

image

For the `osgi.identity` capability, we have to create a resource name that is unique. To keep it readable, the name is the original bsn + "__" + release. The version is the original bundle's version. We introduced a new `type` for these synthetic resources: `bnd.synthetic`.

We also add the requirements of the bundle if it would run on that release.

The synthetic resources are added to the same repository of the original bundle.

rotty3000:
I might be misunderstanding the image @pkriens, but are the versions in the green boxes flipped; where:
- `v17` -> `J2SE [17, ~)`
- `v8` -> `J2SE [1.8, 17)`
?

pkriens:
No, the versions are not flipped :-) A J8 bundle runs on J8 and any later VM. However, when your J8 bundle runs on a J17, it might get different requirements due the J17 VM loading different file resources than an earlier VM. 

So the synthetic resource represents _the bundle's requirements on a restricted range of releases_. If your bundle has `META-INF/versions/{8, 9, 12, 16, 21}` there will be a resource for `[8,9)` that is used to resolve on a Java 8 VM, a resource for `[9,12)` that is resolved for J9, J10, J11, a resource for J12,J13,J14, J15, and a resource for running on J16, J17, J18, J19, J20, and finally a resource for J21 and later. 

I think the code that does this is [highly readable](https://github.com/bndtools/bnd/blob/152980f983172976e8d8efac7dbef0081a9001e8/biz.aQute.bndlib/src/aQute/bnd/osgi/resource/ResourceBuilder.java#L1060-L1140)

pkriens:
You're right. Will fix it.

chrisrueger:
thanks @pkriens I have taken your answer and updated the text. see 4589c54 and added my example and the image.
Feel free to adjust this as needed. I feel that it could need an additional summary like sentence which nails it for a 5-year old but I am not capable of producing this sentence :) 

image

I asked ChatGPT to summarize what you have written, but it basically repeated the same what you said, just structured a little bit different.

chrisrueger:
@pkriens any idea why the DCO failed? I have signed off as usual.

pkriens:
> Commit sha: [4589c54](https://github.com/bndtools/bnd/pull/6135/commits/4589c549a5c7ec8d594d51ad884d34545809494c), Author: Christoph Rueger, Committer: Christoph Rueger; Expected "Christoph Rueger [chrisrueger@gmail.com](mailto:chrisrueger@gmail.com)", but got "Christoph [chrisrueger@gmail.com](mailto:chrisrueger@gmail.com)".

Not sure what it means but I guess you can fix it? Otherwise just merge it .... if it happens again we can fix it.

Issue #6065, opened 2024-03-27
bndtools should assist with importing projects from known bnd workspaces


joe-chacko:
### Problem description
In the bnd workspace model, when bnd is "in charge" of eclipse project dependencies, there are various usability issues:

1. Importing the projects can take a very long time.
2. If the projects are imported out of dependency order, errors can show up due to missing dependencies. Resolving these errors can take a lot of trial and error with cleaning, rebuilding, and even closing and re-opening projects.
3. Even once the projects are imported, the workspace can be slow to load, slow to rebuild, and very susceptible to incoming changes from version control that alter the dependencies.

### Proposed solution

The eclipse bndtools plugin should offer a project import wizard that looks at a specific bnd workspace on the filesystem and allows the user to select which projects to work on. It would then import all the dependent projects too (as known to bnd), and optionally the using projects too (the ones that depend on the selection).

### Proof of Concept

While working on OpenLiberty, we developed some tooling that used bnd to discover and navigate the dependency tree and allow the importing of related projects too. This brought down the time to start working on a new workspace with *some* projects loaded from several hours to 1 minute.

The eclipse plugin we created did the following as a project import wizard:
* asked the user to identify the bnd workspace
* discovered the projects in it
* presented these to the user to select from
* automatically imported depended-upon projects
* optionally imported projects that depended on the selected projects (including their dependencies)

Although our implementation is specific to OpenLiberty, the approach is easily generalised to the bnd workspace model.

Here is the (liberty-specific) workspace location dialog:
![Screenshot 2024-03-27 at 12 22 11](https://github.com/bndtools/bnd/assets/143064/3de72560-053f-4490-8c8f-afcdd7847609)

Here is the project import page. I have selected three projects to import, and requested the users of those projects be imported too:
![Screenshot 2024-03-27 at 12 22 19](https://github.com/bndtools/bnd/assets/143064/4084b05d-703d-4939-b15c-8a34c25c937a)

One minute later, we had a working workspace, compiling without errors, with 144 projects imported:
![Screenshot 2024-03-27 at 12 33 12](https://github.com/bndtools/bnd/assets/143064/e2165d60-beb8-428e-9c3c-489f62fd6c74)

joe-chacko:
Cool — it will take me some time to port it, but good to know the idea has legs.

The screenshots shown are for a new Eclipse workspace. It is in a separate location from the bnd workspace.
[...]

PR #6143, opened 2024-06-05
Include ELPH within the bnd repo

habiblawal1:
Following from issue #6065  this is the initial PR to include the elph tool into bnd. Elph is used currently to automatically import depenent projects for a bnd workspace into an eclipse workspace.


pkriens:
Ehhhhh .... this not a bndtools project?

pkriens:
Can't use this until it is a bndtools project. I'd really like to have this but I do not have time to port it myself.

PR #6161, opened 2024-06-20
Fixes #6159

### LINKED ISSUE #6159: ANSI Console doesn't work with bndtools
There is a `bndtools.core.console.AnsiHandler` in bndtools.core, that causes the Eclipse Console to clear randomly when using ANSI coding.

Is this handler still in use?  

Disabling this handler in the plugin.xml  makes the console output look ok.

https://github.com/bndtools/bnd/blob/35d63c0ac6982114f1627a641355a596826eead6/bndtools.core/_plugin.xml#L4

and

https://github.com/bndtools/bnd/blob/master/bndtools.core/src/bndtools/core/console/AnsiHandler.java

The Eclipse console supports ANSI natively.

maho7791:
- deactivate extension point
- mark Handler as deprecated


pkriens:
We have git as history. You can remove the plugin element. If there is no baselining errors involved, also remove the class. otherwise, I am ok to deprecate it.

thanks.

maho7791:
I'll do so

pkriens:
LGTM

PR #6163, opened 2024-06-24
build.bnd: avoid resolving default containing ${.} to absolute path

chrisrueger:
This PR basically changes the default value for a repo index file when adding / editing repos via plugin editor. The previous default was not good, because it contained the `${.}` which causes some surprising behavior.

Not sure if this is related to https://github.com/bndtools/bnd/commit/f37f8574b7e887644c036cea073d19d853072159

the ${.} in the default-value caused to expand the value always to an absolute path on the harddrive of the developer editing and saving the file. if you accidentally commit this, then it wil fail on other machines. Let's just define the name, relative to the build.bnd . it can still be changed manually

**Steps to reproduce:**

1. Add a  MavenBndRepository to the `build.bnd` a config like below

Notice an `index=${.}/central.maven; \` is added (because this was the default value of the index property). 

```
-plugin.6.Centralstaging: \
		releaseUrl          ='https://repository.apache.org/content/repositories/staging/';\
		readOnly=true; \
		index=${.}/central.maven; \
		name="Maven Central Staging"
```

2. Then save, close build.bnd and reopen again. (in the source view there is still `index=${.}/central-staging.maven;`)
3. Click on "Edit" (or double click) on the created  MavenBndRepo

**Actual result:** 
`index=${.}/central-staging.maven;` is now expanded to `/some/absolute/local/path/central-staging.maven`

image

(Sorry the screenshot contains a `central-staging.maven`, but it is actually a `central.maven`... I just don't have another screenshot of it).

**Expected result**
I did not want to see the expanded path, but just the value from the source editor.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
@@ -1003,17 +1003,26 @@ public class BndEditModel {
 	 * possible to specify multiple plugins under a single key, thus it is a
 	 * list.
 	 *
-	 * @return a map with a property keys and their plugins.
+	 * @return a map with a property keys and their plugins ( macros like (${.})
+	 *         are returned not expanded)
 	 */
 	public Map<String, List<BndEditModelHeaderClause>> getPluginsProperties() {
-		return getProperties(Constants.PLUGIN);
+		return getPropertiesMapInternal(Constants.PLUGIN, false);
 	}
 
-	private Map<String, List<BndEditModelHeaderClause>> getProperties(String stem) {
+
+	/**
+	 * @param stem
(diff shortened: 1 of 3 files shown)


chrisrueger:
@pkriens 
What about an approach like this:

- adding a method `BndEditModel.getPropertiesRaw()` which does not call `replaceHere(...)`
- this method is called from the PluginEditor UI dialog only
- all other existing callers call `BndEditModel.getProperties()` which behaves as it is

image

chrisrueger:
@pkriens I added a change similar as suggested in my Screenshot. The UI now basically says _"give me the properties with macros not expanded"_. 

See commit description of f20abf718c7a9574dbafea7e42a7f5e4fa477997 for details.

pkriens:
men, you're on fire ...

Looks good to me, lets merge and see?

PR #6165, opened 2024-06-25
fix source lookup for bundle classpath lib jars during debugging

chrisrueger:
## What problem does it solve?

This PR solves the problem that the source lookup during a debug session did not work when _stepping into a method_ (F5) of a class which was coming from a jar inside the bundle classpath which was included via 

```
-buildpath: \
	foo.bar;version=latest,\
-includeresource: \
	${repo;foo.bar;latest}; lib:=true
```

Basically my bundle looks like this:

image

the `+library.2.11.0.jar` is a non-OSGI dep and because classloading issues we included it directly to have it on the BCP.
But for classes from `+library.2.11.0.jar` the source code is not found during debugging when you `step into a method`.

then you get this when you "Step into a method" (F5)

image

I found out that the reason was, that those deps / jars in the Bundle Classpath are not considered for source lookup during debugging.

Note: F3 (Open Implementation) works fine and it goes to the correct source file - but that does not help while debugging, because you want to step through the code with F5.

## What does the Solution do?

Basically when we have a project at hand (in the loop of the code I touched) we not only add the project itself to the array of `ISourceContainer`s but now also add all bundles of this project's `-buildpath` to this array.

After that I could successfully F5 into the method of dependencies which are on the BCP via `-includeresource: ${repo;bsn;latest}; lib:=true,\` and a corresponding `-buildpath` entry.
Before this was not working.

I did not see a noticable performance hit, although now more bundles are added to that array which are searched. I suspect (but don't know) that on a higher level there is some caching and filtering of duplicates already. e.g. because I noticed that this specific method is only run once.


chrisrueger:
@pkriens Maybe you have a hint:

- I have an `aQute.bnd.build.Project` at hand.
- in this project there is a bnd.bnd file like

```
-buildpath: \
	foo.bar;version=latest,\
-includeresource: \
	${repo;foo.bar;latest}; lib:=true
```

Is there a way I can get get these jar files from the buildpath referenced bz `-includeresource` ?

Right now I am hacked it by just using all the buildpath entries. But that is too much. 

```
Collection<Container> buildpath = project
	.getBuildpath();
for (Container bp : buildpath) {
	File file = bp.getFile();
	if (file.getName()
		.endsWith(".jar")) {
		additionalSourceContainers
			.add(new ExternalArchiveSourceContainer(file.toString(), false));
	}

}
```

I would like to filter only those entries which are defined in `-includeresource:`

pkriens:
finding the referenced repo entries in other headers is tricky. The macro processing completely bypasses dependency logic. After the macro has done its work, it is a normal file. All repo info is gone.

A way to make this work is to override the `${repo}` macro in a temporary Processor subclass.
```
class RepoCollector extends Processor {
    final Set<BundleId> ids = new LinkedHashSet<>();

    RepoCollector(Processor parent) { super(parent); }

    public String _repo(String[] args) {
               BundleId id = toId(args);
               ids.add(id);
               return // orginal version of repo in Workspace (which should be in the parent chain)
   }
}
```

If you do this, you might want to make a method in Project, looks useful for other things.

chrisrueger:
Thanks @pkriens for the hint. I try to wrap my head around it.

chrisrueger:
Thanks @pkriens . Your hint with `_repo()` gave the initial spark. 
But unfortunatelly I did not find out where to hook in your RepoCollector Processor suggestion. 

So I took a different approach, since I found out `_repo()` triggeres the `${repo;}` macro processing and `Project.add()` gets called with those results. 

This is how the new method `Project.getRepoRefs()` came to this world.

https://github.com/bndtools/bnd/pull/6165/files#diff-a01878046dba7ceb8f43d338872d8b16e659f050c099b70671419d4b778a9a93R950-R964

This populates `Project.repoRefs` and is called by the source lookup stuff in `bndtools.launch.sourcelookup.containers.BndrunDirectiveSourceContainer.addRepoRefs(Set<ISourceContainer>, Project)` which basically gets called when you _step into a method_ with `F5` during debugging.

synesty-cr:
@pkriens I rework some things as discussed in yesterday's call. 
I refactored Project a bit to allow more resuse and get a `Collection<Container>`. 
I decided to move some method to the new RepoCollector and let Project use RepoCollector too. Makes `Project._repo()` simpler too. 
That feels "cleaner" imho since "repo collecting" logic is now in a class whose name reflects this intention.  See a64093f

Also added a testcase.

chrisrueger:
@pkriens I rework some things as discussed in yesterday's call. 
I refactored Project a bit to allow more resuse and get a `Collection<Container>`. 
I decided to move some method to the new RepoCollector and let Project use RepoCollector too. Makes `Project._repo()` simpler too. 
That feels "cleaner" imho since "repo collecting" logic is now in a class whose name reflects this intention.  See a64093f

Also added a testcase. 

Let's discuss 😄

chrisrueger:
@pkriens could we try merging and see how it goes?

pkriens:
you got my permission to try these things out. Beauty of Github is that we can always revert.

PR #6170, opened 2024-06-30
Document and show repo tags in the UI

chrisrueger:
## In this PR

- add documentation to the manual about our repository tags and how they affect resolution introduced in https://github.com/bndtools/bnd/pull/6116  (`-runrepos`)
- show tags in repo browser as comma separated list 
image


chrisrueger:
@juergen-albert since you brought up the usecase of excluding the baseline repo, maybe you could do a quick review of the docs
@pkriens thoughts?

juergen-albert:
The documentation looks solid. Digging into this again, I have one remark and one possible change request.

Remark: We should not use `<<EMTPY>>` as this would indicate to me, that the Repository is empty.  I'm not sure what we should write. Ideally something that indicates that indicates that it is used for resolve by default. Not sure how this could look. An alternative could also be `<<NO TAGS>>`.

Possible change request or something for an additional issue:  It would be nice to have something like `-runrepos`, where we can define the tags I want to use. For a test e.g. I can then use `-runrepostags: resolve, test` (even though the instruction looks rather awkward.

chrisrueger:
> Remark: We should not use `<<EMTPY>>` as this would indicate to me, that the Repository is empty. I'm not sure what we should write. Ideally something that indicates that indicates that it is used for resolve by default. Not sure how this could look. An alternative could also be `<<NO TAGS>>`.

Thanks @juergen-albert . The empty was a suggestion by @pkriens when we had a call about it. <<EMPTY>> is just a placeholder which could have been "foobar" as well. It just means "something else than 'resolve'". 

My suggestion: for the sake of this current PR we keep `<<EMPTY>>` and I will create a new Issue / PR for the change and also update the docs afterwards. Until 7.1. is still in the making I think we are still ok to do such change. I think your suggestion `<<NO TAGS>>` works too, but since it is highly subjective and I suggest a separate issue. I have a no real preference. 



> Possible change request or something for an additional issue: It would be nice to have something like `-runrepos`, where we can define the tags I want to use. For a test e.g. I can then use `-runrepostags: resolve, test` (even though the instruction looks rather awkward.

Same here. I would suggest creating a new issue. I had something like this in the beginning but removed it during the discussion about the first version of this feature.  Basically what I added was that you could specify a tag in -runrepos like this: `-runrepos: @resolve, @test, @myothertag`
I think in a new issue we can still discuss the details.

juergen-albert:
Okay. Then +1 from me!

juergen-albert:
And thanks heaps again. This feature is highly appreaciated.

chrisrueger:
Thanks for merging @juergen-albert  👍

chrisrueger:
added the link to the docu to the [7.1.0 release notes](https://github.com/bndtools/bnd/wiki/Changes-in-7.1.0#tagging-of-repositories)

PR #6186, opened 2024-07-17
upgrade SWT to 3.126 to fix MacOS / JDK bug

### LINKED ISSUE #6185: Issues with bndtools Eclipse dev instance on MacOS Sonoma 14.3+ with temurin JDK > 17.0.09+9
Based on https://github.com/eclipse-platform/eclipse.platform.swt/issues/1012#issuecomment-2232936903 and comments below.

The problem: new contributors will have a hard time figuring out what's wrong because the dev child eclipse instance shows a broken Ui without buttons etc.

This happens under the following setup:
- MacOS Sonoma 14.5 (23F79)

and any Eclipse Temurin JDK > 17.0.09+9 such as:

- OpenJDK Runtime Environment Temurin-17.0.11+9 
- as well as OpenJDK Runtime Environment Temurin-21.0.3+9 

Then, when launching bndtools Eclipse dev instance via `bndtools.cocoa.macosx.aarch64.bndrun` it looks like this:

image

image

## Workarounds 

- it works e.g. with JDK Temurin 17.0.9+9 or other JDKs as pointed out [here](https://github.com/eclipse-platform/eclipse.platform.swt/issues/1012#issuecomment-2233058695).

Relates to https://github.com/bndtools/bnd/issues/5815

chrisrueger:
Closes #6185 

It seems we can stay on older Eclipse 4.25 but use a higher SWT version which has fixed the SWT bug (see https://github.com/eclipse-platform/eclipse.platform.swt/issues/1012).

With this fix I was able to use Eclipse with latest Temurin JDKs like

- OpenJDK Runtime Environment Temurin-17.0.11+9
- OpenJDK Runtime Environment Temurin-21.0.3+9

and it looks as usual again: 

Bildschirmfoto 2024-07-17 um 16 36 23

Bildschirmfoto 2024-07-17 um 16 36 56

### DIFF
--- a/bndtools.core/bndtools.cocoa.macosx.aarch64.bndrun
+++ b/bndtools.core/bndtools.cocoa.macosx.aarch64.bndrun
@@ -259,8 +259,8 @@
 	org.eclipse.platform;version='[4.25.0,4.25.1)',\
 	org.eclipse.sdk;version='[4.25.0,4.25.1)',\
 	org.eclipse.search;version='[3.14.200,3.14.201)',\
-	org.eclipse.swt;version='[3.121.0,3.121.1)',\
-	org.eclipse.swt.cocoa.macosx.aarch64;version='[3.121.0,3.121.1)',\
+	org.eclipse.swt;version='[3.126.0,3.126.1)',\
+	org.eclipse.swt.cocoa.macosx.aarch64;version='[3.126.0,3.126.1)',\
 	org.eclipse.team.core;version='[3.9.500,3.9.501)',\
 	org.eclipse.team.ui;version='[3.9.400,3.9.401)',\
 	org.eclipse.text;version='[3.12.200,3.12.201)',\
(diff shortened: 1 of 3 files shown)


chrisrueger:
I did some local testing with 

- Temurin-17.0.11+9
- Temurin-21.0.3+9

and I could not see any issues. Tested the usual screens like e.g. 

- bnd editor
- resolution view
- Repo Brower + Advanced Search
- .bndrun editor
- build.bnd editor
- resolve dialog in bndrun editor
- + Button to add bundles to runbundles / runrequirements

chrisrueger:
Did another test with an Intel Mac (but MacOs 13.x Ventura)
Ventura did not have the initial problem reported here.
But I can confirm that this PR did not break this.

chrisrueger:
@pkriens I would like to merge this. The problem should happen for 7.0.0 bndtools users using a Eclipse Temurin JDK in combination with latest MacOS Sonoma. In this case at least we could tell them to use the latest 7.1 Snapshot as another workaround (instead of using a special JDK which does not have this problem).

pkriens:
I am ok to merge. Thanks for the work!

PR #6190, opened 2024-07-21
Adapted the sameExactVersion function to also check if the extension of two Maven artifacts are equal.

### LINKED ISSUE #6189: [Maven] When a dependency has multiple artifacts, the non-jar can hide the jar dependency, resulting in resolve issues
The bnd Maven plugins, in their processing of Maven project dependencies, when a dependency has multiple artifacts (like a zip in addition to the jar), the presence of the zip as dependency may hide the jar dependency when the Maven plugins are constructing a repository from the Maven project. As a consequence, the jar that is present as a dependency, may be absent from the repository, and as such resolve operations fail, that would otherwise work if the zip artifact was not referenced, or referenced after the jar.

glimmerveen:
This should fix #6189


glimmerveen:
Thank you for merging this @pkriens !

PR #6191, opened 2024-07-22
fix new project templates not showing up

chrisrueger:
This fixes the problem, that newly created "project template"-projects required a Eclipse restart to show up in the **New Bnd Project** wizard.

See https://bnd.discourse.group/t/enabling-template-repositories-by-default/460/5

This PR makes sure that new Project templates always / immediately show up in the "New Bnd OSGi Project" wizard, after creation.

image

### DIFF
--- a/bndtools.core/src/bndtools/central/EclipseWorkspaceRepository.java
+++ b/bndtools.core/src/bndtools/central/EclipseWorkspaceRepository.java
@@ -19,10 +19,11 @@ import aQute.bnd.osgi.Constants;
 import aQute.bnd.osgi.repository.AbstractIndexingRepository;
 import aQute.bnd.osgi.repository.WorkspaceRepositoryMarker;
 import aQute.bnd.osgi.resource.ResourceBuilder;
+import aQute.bnd.service.Refreshable;
 import aQute.lib.io.IO;
 
 public class EclipseWorkspaceRepository extends AbstractIndexingRepository<IProject, File>
-	implements WorkspaceRepositoryMarker {
+	implements WorkspaceRepositoryMarker, Refreshable {
 	EclipseWorkspaceRepository() {
 		super();
 		Central.onCnfWorkspace(this::initialize);
@@ -78,5 +79,16 @@ public class EclipseWorkspaceRepository extends AbstractIndexingRepository<IProj
 	@Override
 	public String toString() {
 		return NAME;
+	}
+
+	@Override
+	public boolean refresh() throws Exception {
+		initialize(null);
+		return true;
+	}
+
+	@Override
+	public File getRoot() throws Exception {
+		return null;
 	}
 }
--- a/bndtools.core/src/bndtools/wizards/project/NewBndProjectWizard.java
+++ b/bndtools.core/src/bndtools/wizards/project/NewBndProjectWizard.java
@@ -34,7 +34,9 @@ import org.eclipse.jface.wizard.IWizardPage;
 import org.eclipse.ui.IWorkbench;
 
 import aQute.bnd.build.Project;
+import aQute.bnd.exceptions.Exceptions;
 import bndtools.Plugin;
+import bndtools.central.Central;
 
 class NewBndProjectWizard extends AbstractNewBndProjectWizard implements ISkippingWizard {
 
@@ -52,6 +54,13 @@ class NewBndProjectWizard extends AbstractNewBndProjectWizard implements ISkippi
 	@Override
 	public void init(IWorkbench workbench, IStructuredSelection currentSelection) {
 		super.init(workbench, currentSelection);
+
+		try {
+			// refresh, to not miss recently created project templates.
+			Central.refreshPlugins();
+		} catch (Exception e) {
+			throw Exceptions.duck(e);
+		}
 
 		BuiltInTemplate baseTemplate = new BuiltInTemplate(EMPTY_TEMPLATE_NAME, DEFAULT_TEMPLATE_ENGINE);
 		baseTemplate.addInputResource(Project.BNDFILE, new StringResource("")); //$NON-NLS-1$
--- a/bndtools.core/src/bndtools/wizards/project/NewBndProjectWizard.java
+++ b/bndtools.core/src/bndtools/wizards/project/NewBndProjectWizard.java
@@ -34,7 +34,6 @@ import org.eclipse.jface.wizard.IWizardPage;
 import org.eclipse.ui.IWorkbench;
 
 import aQute.bnd.build.Project;
-import aQute.bnd.exceptions.Exceptions;
 import bndtools.Plugin;
 import bndtools.central.Central;
 
@@ -59,7 +58,10 @@ class NewBndProjectWizard extends AbstractNewBndProjectWizard implements ISkippi
 			// refresh, to not miss recently created project templates.
 			Central.refreshPlugins();
 		} catch (Exception e) {
-			throw Exceptions.duck(e);
+			Plugin.getDefault()
+				.getLog()
+				.log(new Status(IStatus.ERROR, Plugin.PLUGIN_ID, 0,
+					String.format("Unable to refresh Plugins in NewBndProjectWizard"), e));
 		}
 
 		BuiltInTemplate baseTemplate = new BuiltInTemplate(EMPTY_TEMPLATE_NAME, DEFAULT_TEMPLATE_ENGINE);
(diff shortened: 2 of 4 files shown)


bndtools.core/src/bndtools/wizards/project/NewBndProjectWizard.java
```diff
@@ -53,6 +55,13 @@ class NewBndProjectWizard extends AbstractNewBndProjectWizard implements ISkippi
 	public void init(IWorkbench workbench, IStructuredSelection currentSelection) {
 		super.init(workbench, currentSelection);
 
+		try {
+			// refresh, to not miss recently created project templates.
+			Central.refreshPlugins();
+		} catch (Exception e) {
+			throw Exceptions.duck(e);
+		}
```
chrisrueger, line 63: @pkriens just a note about why I have put it here: I'd prefer to put it at the end of `bndtools.wizards.project.AbstractNewBndProjectWizard.performFinish()` but I had (racy) issues that the new added Project needs to finished building before the repo-refresh worked for the `EclipseWorkspaceRepository`. 
This sometimes worked, sometimes not. E.g. when I had a debugger enabled then it mostly worked, because it had time to build. 

Question:
Do we have any mechanism to executed `Central.refreshPlugins();` after the new created project has been built?

bndtools.core/src/bndtools/central/EclipseWorkspaceRepository.java
```diff
@@ -79,4 +80,15 @@ protected BiFunction<ResourceBuilder, File, ResourceBuilder> indexer(IProject pr
...
+
+	@Override
+	public boolean refresh() throws Exception {
+		initialize(null);
+		return true;
+	}
+
+	@Override
+	public File getRoot() throws Exception {
+		return null;
```
chrisrueger, line 92: @pkriens also here I did not know what else to return other than null. Is there a "root" **File** for EclipseWorkspaceRepository ?

pkriens: I recall this was ok. Remains of the old days.

bndtools.core/src/bndtools/wizards/project/NewBndProjectWizard.java
```diff
@@ -53,6 +54,16 @@ class NewBndProjectWizard extends AbstractNewBndProjectWizard implements ISkippi
 	public void init(IWorkbench workbench, IStructuredSelection currentSelection) {
 		super.init(workbench, currentSelection);
 
+		try {
+			// refresh, to not miss recently created project templates.
+			Central.refreshPlugins();
```
pkriens, line 59: hmm. question is why this did not happen. Not keen on having this called in lots of places. 

Then again if this fixes the problem now, I am ok. However, it needs to be called in a background job & you need to get the workspace lock.

pkriens (changes requested):

bndtools.core/src/bndtools/wizards/project/NewBndProjectWizard.java
chrisrueger: Thanks Peter for your reply. 

> hmm. question is why this did not happen. Not keen on having this called in lots of places.

I can understand that. But I already cut my teeth on it. See my comment above. 
I tried debugging it, but without success. 
I tracked it down to the following problem:

When you create a new "Template Project" (which you want to see later in the list), then the following needs to happen:
- the new "Template projects" needs to be built (.jar generated)
- then (and only then) `Central.refreshPlugins();` needs to be called. 

If Central.refreshPlugins(); is called too early (when the .jar of the new "Template project" does not exist yet), then the new Template is not visible the "File / New / New Bnd OSGi Project" Wizard. 

**Workaround**: Manually clicking the "Refresh Repos" button in the UI before doing "File / New / New Bnd OSGi Project" Wizard.
image

In this case it works 100% of the time.  But this is manual, and you need to remember it.
If you could help me find the place in the code where we could do it programmatically after building, I am happy to remove the call to `Central.refreshPlugins();` where I have added it now. 

**Question:**
Do we have any mechanism to executed `Central.refreshPlugins();` after the new created project has been built?


> Then again if this fixes the problem now, I am ok. However, it needs to be called in a background job & you need to get the workspace lock.

Could you elaborate on why? Just want to understand it. 

- background job: I think I would like this to happen synchronously in the current thread, to make sure that it has finished indexing the new project exactly where I added it - at least this is the reason why I have put it there. Will a background job do this too? Or is it async and finishes "sometime" in the future? 

- workspace lock: The "File / New / New Bnd OSGi Project" Wizard is a modal overlay in the foreground. When this is opened I guess nobody could do anything in the workspace at the same time. But maybe I don't understand the purpose of this workspace lock. 


But of  course I can look into all these things. Just wanted to clarify.

chrisrueger: Just an additional thought:
If `Central.refreshPlugins();` is too expensive. 
For the purpose of just getting a "fresh view" of the available "Project templates" in the current workspace, it would be maybe enough to just call `EclipseWorkspaceRepository.refresh()`. 

Maybe this is cheaper, and fixes the problem too. `Central.refreshPlugins();` would be more generic affecting all repos, but maybe too much for the purpose.

chrisrueger: Thinking a bit more about it - Here is a suggestion:

- I remove the problematic code line in `NewBndProjectWizard.java`
```
// refresh, to not miss recently created project templates.
Central.refreshPlugins();
```

- I keep the other code in `EclipseWorkspaceRepository`, because then the Repo-Refresh button works, and you could at least make the "Project template" visible without an Eclipse restart. 

image

We could then document this behavior somewhere. Maybe I add it to the help-text which is displayed in the "Project template project" wizard (`"Please make sure to refresh the Repository Browser..."`). 

image


I just commited this proposal.

chrisrueger:
@pkriens I took the liberty to merge this, since I removed the expensive call to `Central.refreshPlugins();` So this is more or less a bugfix + some a Wording solution

bndtools.core/src/bndtools/wizards/project/NewBndProjectWizard.java
scottslewis: FWIW, my opinion is that any additional required actions by the user is going to make this bndworkspace workspace creation/template-based project/service/etc creation dance even more complicated and error prone...especially for new bndtools users.   

I'm not exactly sure whether Christoph's proposal requires user action (if not please ignore), but if it does I would advise against it.   Seems to me that another way to deal with this is to either enhance the bnd project/service, etc wizard superclasses to make the necessary calls, or add new bndtools refresh methods to capture the required semantics.

FWIW I believe I struggled with this for some time when introducing the New OSGI Service Wizards (based upon templates).

chrisrueger: Hi @scottslewis ,

> FWIW, my opinion is that any additional required actions by the user is going to make this bndworkspace workspace creation/template-based project/service/etc creation dance even more complicated and error prone...especially for new bndtools users.
> I'm not exactly sure whether Christoph's proposal requires user action (if not please ignore), but if it does I would advise against it.

Unfortunatelly it does require user action 😞  . The state I merged this afternoon _requires user action_ in the way that I now show a help text that a click on the "Refresh Button of the Repositories" view is required to make the new template show up. 
But this is a slight improvement over the situation before which required a Eclipse restart.
My [change here](https://github.com/bndtools/bnd/pull/6191/files#diff-9d66588331a8171dacc618eeb5f4ee17ed6b33d858adc35eb51d5f471504910c) fixes the problem, that at least now "Refresh Button of the Repositories" works instead of Eclipse restart. 

That being said:
I know this is not ideal and I am glad for any for suggestions and adress and improve it in another issue. 
If you have something in mind let's try it in a issue / PR. 

During my work in this PR here my goal was to find a way that the new Template shows up automatically. 
But i failed to get it working reliably. It was kinda random sometimes. The closes I came was [this](https://github.com/bndtools/bnd/pull/6191/commits/6bc014081cd00ef1f4c235bedb5dc1a684449346#diff-75991c25d92615f5a66ab554df5a7ab4362d0d9e81b3e82eec600abc2a293328R60), but it might be a bit too expensive doing it there. 

As stated above somewhere, I was searching for a mechanism to executed Central.refreshPlugins(); **after** the new created project **has been built**? (_after has been built_ is important, because the _.jar_ artifact of the new project needs to be generated by bnd before it can be picked up by the refresh of [EclipseWorkspaceRepository](https://github.com/chrisrueger/bnd/blob/7c0b73abe0363a7d18a79eb135adda301519611c/bndtools.core/src/bndtools/central/EclipseWorkspaceRepository.java#L85)). If the refresh happens earlier before the _.jar_ is built, then the refresh has no effect. 

Long story short: I would appreciate any pointers of how this could be improved. You are totally right that is is not ideal and I would like to improve it too.

PR #6201, opened 2024-08-05
checks Project validity and handles projects without bnd.bnd file

juergen-albert:
Fixes a Bug, where the created project for the code generator in maven is invalid, if no bnd file exists.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
@@ -1,6 +1,7 @@
 package aQute.bnd.build;
 
 import static aQute.bnd.build.Container.toPaths;
+
 import java.io.BufferedInputStream;
 import java.io.BufferedReader;
 import java.io.File;
@@ -234,7 +235,7 @@ public class Project extends Processor {
 		if (getBase() == null || !getBase().isDirectory())
 			return false;
 
-		return getPropertiesFile() == null || getPropertiesFile().isFile();
+		return getPropertiesFile() == null || (!mustFileExist() || getPropertiesFile().isFile());
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


chrisrueger:
@juergen-albert  Looks like tests fail and there is a file in the stacktrace which was touched by this PR:

```
Caused by: java.lang.NullPointerException: Cannot invoke "java.io.File.getAbsoluteFile()" because "propertiesFile" is null
    at aQute.bnd.osgi.Processor.setProperties (Processor.java:919)
    at aQute.bnd.osgi.Processor.forceRefresh (Processor.java:892)
    at aQute.bnd.maven.generate.plugin.BndContainer.init (BndContainer.java:245)
    at aQute.bnd.maven.generate.plugin.BndContainer.generate (BndContainer.java:138)
```

Maybe `aQute.bnd.osgi.Processor.forceRefresh()` needs a null check of `propertiesFile` before calling `setProperties(propertiesFile, base);`?

or earlier before calling `aQute.bnd.osgi.Processor.forceRefresh()` in your `aQute.bnd.maven.generate.plugin.BndContainer.init (BndContainer.java:245)`


just guessing

juergen-albert:
@chrisrueger I was a bit too quick with opening the PR, but it is fixed now.

@pkriens To make it work, I had to make https://github.com/bndtools/bnd/blob/035bab9ea70ab28a7d16cf08b89bc0507b5852b5/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java#L122 accessible by the Project. I think it makes sense, to look at this marker when checking for the validity of a Project, but I'm hesitant to just merge this, without some review from you. My previous attempt with initializing the Project with null as bnd file (which would be seen as valid) caused NPEs on other operations during refresh.

PR #6214, opened 2024-08-13
Refresh the EclipseWorkspaceRepository in RepositoriesViewRefresher

### LINKED ISSUE #6213: New "Project templates" projects should always / immediately show up in the "New Bnd OSGi Project" wizard, after creation.
This issue originated from PR  https://github.com/bndtools/bnd/pull/6191#issue-2422816266 but this PR did only solve the problem partly, far from ideal which @scottslewis rightly pointed out. 

## The problem

Newly created "project template"-projects still require a Refresh of Repositories (before #6191 even an Eclipse restart was required) to show up in the New Bnd Project wizard. 

PR #6191 allows at least the the Refresh button in the Repository browser works and we show a help text, to at least mention that this needs to be done if a new "project template"-projects is not showing up.

image

## Goal

New Project templates should always / immediately show up in the "New Bnd OSGi Project" wizard, after creation. 
No manual user action needed.

Quoting from [here](https://github.com/bndtools/bnd/pull/6191#discussion_r1714359075):

> I was searching for a mechanism to executed `Central.refreshPlugins();` after the new created project has been built? (after has been built is important, because the `.jar` artifact of the new project needs to be generated by bnd before it can be picked up by the refresh of [EclipseWorkspaceRepository](https://github.com/chrisrueger/bnd/blob/7c0b73abe0363a7d18a79eb135adda301519611c/bndtools.core/src/bndtools/central/EclipseWorkspaceRepository.java#L85)). If the refresh happens earlier before the `.jar` is built, then the refresh has no effect.

So 
- something to hook into after the new added project has been built successfully. 
- which calls `Central.refreshPlugins();` (which triggers a call to [EclipseWorkspaceRepository#refresh()](https://github.com/bndtools/bnd/blob/master/bndtools.core/src/bndtools/central/EclipseWorkspaceRepository.java#L85)
- no user action / button clicking needed

So if there is anybody knowing pointers for how to accomplish the quoted requirement above it would be great.

chrisrueger:
Closes #6213 

This is an improvement of commit 0e1fac345 in PR #6191

We now hook into the RepositoriesViewRefresher which already gets triggered after projects have been built to update the Repositories list. The only thing which is missing here is a Refresh of the EclipseWorkspaceRepository, which is needed so that new "Project template projects" show up immediately in the "New Project" wizard.

The advantage over PR is than no manual user action is needed (e.g. no click on "Refresh button in Repositories browser")

## Testing

1. Open **File / New / Bnd OSGi Project** and create a Template project e.g. mustache

image

image

Give it a name and click finish.

2. quickly Open **File / New / Bnd OSGi Project** again and check if the new template appears. 

image

### DIFF
--- a/bndtools.core/src/bndtools/central/RepositoriesViewRefresher.java
+++ b/bndtools.core/src/bndtools/central/RepositoriesViewRefresher.java
@@ -29,6 +29,7 @@ import org.osgi.framework.FrameworkUtil;
 import org.osgi.framework.ServiceRegistration;
 
 import aQute.bnd.build.Workspace;
+import aQute.bnd.build.WorkspaceRepository;
 import aQute.bnd.osgi.Jar;
 import aQute.bnd.service.RepositoryListenerPlugin;
 import aQute.bnd.service.RepositoryPlugin;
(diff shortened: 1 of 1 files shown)


chrisrueger:
@scottslewis would be great to get your opinion / review

@pkriens thoughts? I picked the `RepositoriesViewRefresher` because I just found that it is called already during build process and as far as I see it it does already some updating of repositories similar what I need to do.

I decided against `Central.refreshPlugins();` since it might be too much. 
Thus I call `Central.getEclipseWorkspaceRepository().refresh();` only, since this is the only thing which needs to be called for the purpose of this PR / issue.

chrisrueger:
I merged that and try it out in the next days. Let me know if you experience any issues.

scottslewis:
Sorry @chrisrueger, I've been occupied with other things.  This refresh approach works for me.

PR #6222, opened 2024-08-19
bnd-maven plugin: delete existing MANIFEST.MF from previous build

### LINKED ISSUE #6221: [bnd-maven-plugin] Running `bnd-process` without `mvn clean` causes different behavior
### Version
- 6.4.0
- 7.1.0-SNAPSHOT

### Description
It looks like the behavior of bnd-maven-plugin's bnd-process differs when the Maven build is executed a second time, without `mvn clean`, respectively when `target/classes/META-INF/MANIFEST.MF` exists from a previous run.

### Reproduction steps
1. Download the attached [bnd-test.zip](https://github.com/user-attachments/files/16646669/bnd-test.zip) and extract it
2. Run `mvn process-classes`
3. Create a copy of `target/classes/META-INF/MANIFEST.MF` named `clean-MANIFEST.MF`
(but don't delete `MANIFEST.MF`)
4. Run `mvn process-classes` again

Now compare `clean-MANIFEST.MF` with `MANIFEST.MF`, you should see the following:
(`clean-MANIFEST.MF` being _old_, `MANIFEST.MF` being _new_)
```diff
  Manifest-Version: 1.0
  Bundle-ManifestVersion: 2
  Bundle-Name: my-app
  Bundle-SymbolicName: my-app
  Bundle-Version: 1.0.0.SNAPSHOT
- Export-Package: example.a;uses:="example.b";version="1.0.0",example.b;
+ Export-Package: example.a;version="1.0.0";uses:="example.b",example.b;
   version="1.0.0",example.c;version="1.0.0"
- Import-Package: example.b,example.c;version="${range;[==,=+)}"
+ Import-Package: example.b;version="[1.0,2)",example.c;version="[1.0,1.
+  1)"
  Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
```

| | `clean-MANIFEST.MF` | `MANIFEST.MF` |
| - | - | - |
| `Export-Package` | first lists `uses` then `version` | first lists `version` then `uses` |
| `Import-Package` | `range` macro was not expanded (separate bug?) | `range` macro was expanded, as desired |
| `Import-Package` | `example.b` has no version range (see also #6220) | `example.b` has a version range |

chrisrueger:
Based on comment https://github.com/bndtools/bnd/issues/6221#issuecomment-2296567709

This fixes a problem where a 'mvn process-classes' picked up `target/classes/META-INF/MANIFEST.MF` from a previous build and produced a new MANIFEST.MF based on the results of the previous build... which should not happen.

We now delete the existing `target/classes/META-INF/MANIFEST.MF` if it exists.

Now this gives consistent results for the following case:

```
mvn clean process-classes
mvn clean process-classes
```
or 

`mvn clean process-classes`
followed by a
`mvn process-classes`

@pkriens I wonder how this PR affects the statement in 

https://github.com/bndtools/bnd/blob/c1bb1658ee3cdd4d7052a3722af3afb107f96828/maven-plugins/bnd-maven-plugin/README.md?plain=1#L111

### DIFF
--- a/maven-plugins/bnd-maven-plugin/README.md
+++ b/maven-plugins/bnd-maven-plugin/README.md
@@ -92,6 +92,28 @@ The `bnd-process` is not executed by default, therefore at least one explicit ex
 </plugin>
 ```
 
+**Deleting existing MANIFEST.MF of previous builds:** 
+Note that `process-classes` uses content in the `classesDir` as input which could also contain a `META-INF/MANIFEST.MF` file (e.g. from a previous build). In some cases this could lead to surprising results when `process-classes` is executed multiple times. You can set the `deleteExistingManifest` configuration to `true` to delete an existing MANIFEST.MF at `manifestPath` prior to execution to avoid that it is used as input to the bnd process.
+
+```xml
+<plugin>
+    <groupId>biz.aQute.bnd</groupId>
+    <artifactId>bnd-maven-plugin</artifactId>
+    <executions>
+        <execution>
+            <id>bnd-process</id>
+            <goals>
+                <goal>bnd-process</goal>
+            </goals>
+            <configuration>
+            	<deleteExistingManifest>true</deleteExistingManifest>
+            </configuration>
+        </execution>
+    </executions>
+</plugin>
+```
+
+
 ### Configuration Parameters
 
 | Configuration Parameter | Description                                                                                                                                                                                                                                                                       |
@@ -99,6 +121,8 @@ The `bnd-process` is not executed by default, therefore at least one explicit ex
 | `bndfile`               | File path to a bnd file containing bnd instructions for this project. The file path can be either absolute or relative to the project directory. _Defaults to `bnd.bnd`_.                                                                                                         |
 | `bnd`                   | Bnd instructions for this project specified directly in the pom file. This is generally be done using a `<![CDATA[  ]]>` section. If the projects has a `bndfile` configuration property or a file in the default location `bnd.bnd`, then this configuration element is ignored. |
 | `manifestPath`          | Specify the path to store the generated manifest file. _Defaults to `${project.build.outputDirectory}/META-INF/MANIFEST.MF`._                                                                                                                                                     |
+| `deleteExistingManifest`          | Whether to delete an existing file at `manifestPath` before execution. _Defaults to `false`._
+|
 | `classesDir`            | The directory where the `maven-compiler-plugin` places its output. _Defaults to `${project.build.outputDirectory}`._                                                                                                                                                              |
 | `includeClassesDir`     | Include the entire contents of `classesDir` in the bundle. *Defaults to `true`*.                                                                                                                                                                                                  |
 | `outputDir`             | The directory where this goal will store its output. _Defaults to `${project.build.outputDirectory}`._                                                                                                                                                                            |
@@ -429,6 +453,8 @@ The `bnd-process-tests` is not executed by default, therefore at least one expli
 | `skip`                  | Skip the goal. _Defaults to `false`._ Override with property `bnd-tests.skip` or `maven.test.skip`.                                                                                                                                                                               |
 | `skipIfEmpty`           | Skip processing if `includeClassesDir` is `true` and the `classesDir` is empty. _Defaults to `false`._ Override with property `bnd.skipIfEmpty`.                                                                                                                                  |
 | `manifestPath`          | Specify the path to store the generated manifest file. _Defaults to `${project.build.testOutputDirectory}/META-INF/MANIFEST.MF`._                                                                                                                                                 |
+| `deleteExistingManifest`          | Whether to delete an existing file at `manifestPath` before execution. _Defaults to `false`._
+|
 | `outputDir`             | The directory where this goal will store its output. _Defaults to `${project.build.testOutputDirectory}`._                                                                                                                                                                        |
 | `packagingTypes`        | The list of maven packaging types for which the plugin will execute. *Defaults to `jar,war`*. Override with property `bnd.packagingTypes`.                                                                                                                                        |
 | `outputTimestamp`       | Timestamp for [reproducible][1] output archive entries, either formatted as ISO 8601 `yyyy-MM-dd'T'HH:mm:ssXXX` or as an int representing seconds since the epoch. _Defaults to `${project.build.outputTimestamp}`_.                                                              |
--- a/maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
+++ b/maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
@@ -96,6 +96,12 @@ public abstract class AbstractBndMavenPlugin extends AbstractMojo {
 	boolean					includeClassesDir;
 
 	/**
+	 * Whether to delete an existing {@link #getManifestPath()} before execution.
+	 */
+	@Parameter(defaultValue = "false")
+	boolean					deleteExistingManifest;
+
+	/**
 	 * The directory where the webapp is built when packaging is {@code war}.
 	 */
 	@Parameter(alias = "warOutputDir", defaultValue = "${project.build.directory}/${project.build.finalName}")
@@ -252,6 +258,16 @@ public abstract class AbstractBndMavenPlugin extends AbstractMojo {
 			List<Builder> subs = builder.getSubBuilders();
 			if ((subs.size() != 1) || !builder.equals(subs.get(0))) {
 				throw new MojoExecutionException("Sub-bundles not permitted in a maven build");
+			}
+
+			if (deleteExistingManifest) {
+				// delete existing MANIFEST.MF
+				// e.g. from previous build to avoid that this MANIFEST.MF
+				// becomes input to the bnd process (which could lead to
+				// suprising results in some cases)
+				File existingMetaInf = getManifestPath();
+				logger.info("Delete existing manifest: {}", existingMetaInf);
+				IO.delete(existingMetaInf);
 			}
 
 			// always add the outputDirectory to the classpath, but
--- a/maven-plugins/README.md
+++ b/maven-plugins/README.md
@@ -65,8 +65,10 @@ A plugin to generate sources and resources.
 
 # Building the Maven Plugins
 
-You must first run `./gradlew :build` to build the Bnd artifacts and install them in your local maven repo.
-You can then run `./mvnw install` to build the Bnd Maven plugins.
+You must first run `./gradlew :build` (or `./gradlew build -x test` skip tests for faster local builds) to build the Bnd artifacts and install them in your local maven repo.
+You can then run `./mvnw install` to build the Bnd Maven plugins (or `mvn install -Dinvoker.skip=true` to skip tests for faster local builds).
+
+
 
 ---
 


Marcono1234:
For the Maven plugin, does this effectively turn `Analyzer#getManifestInfoFromClasspath` into dead code?

Is it actually guaranteed that `target/classes/META-INF/MANIFEST.MF` comes from a previous build by bnd-maven-plugin? I have not tested it yet, but what if a user has a `src/main/resources/META-INF/MANIFEST.MF` and relies on bnd-maven-plugin enhancing / extending it?

chrisrueger:
> but what if a user has a `src/main/resources/META-INF/MANIFEST.MF` and relies on bnd-maven-plugin enhancing / extending it? (Not sure though if that is an intended / supported use case.)

This is a very good question I was concerned about. 
If that is the case, then the question is how to distinguish between those cases.

maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
```diff
@@ -258,6 +258,15 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 			// handle projects with no output directory, like
 			// 'test-wrapper-bundle'
 			if (classesDir.isDirectory()) {
+
+				// delete existing MANIFEST.MF from previous build
+				// target/classes/META-INF is output and should not be picked up
+				File existingMetaInf = new File(classesDir, "META-INF/MANIFEST.MF");
+				if (existingMetaInf.exists()) {
+					logger.info("Delete existing manifest from previous build: " + existingMetaInf);
+					IO.delete(existingMetaInf);
```
bjhargrave, line 267: This does break a potential use case where the user caused a manifest file to be written to target/classes which they want to be input to the bnd process. For example, they could unzip an existing jar into target/classes and run bnd to make it a bundle. This change would prevent that use case. I think you probably need to make some configuration option which can be set to always delete any extant manifest in target/classes.

bjhargrave (changes requested):

maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
chrisrueger: Thanks. I will add a config option.

maven-plugins/bnd-maven-plugin/README.md
```diff
@@ -92,13 +92,37 @@ The `bnd-process` is not executed by default, therefore at least one explicit ex
...
 | `bndfile`               | File path to a bnd file containing bnd instructions for this project. The file path can be either absolute or relative to the project directory. _Defaults to `bnd.bnd`_.                                                                                                         |
 | `bnd`                   | Bnd instructions for this project specified directly in the pom file. This is generally be done using a `<![CDATA[  ]]>` section. If the projects has a `bndfile` configuration property or a file in the default location `bnd.bnd`, then this configuration element is ignored. |
 | `manifestPath`          | Specify the path to store the generated manifest file. _Defaults to `${project.build.outputDirectory}/META-INF/MANIFEST.MF`._                                                                                                                                                     |
+| `deleteExistingManifest`          | Whether to delete an existing `META-INF/MANIFEST.MF` inside the `classesDir` directory. Defaults to `false`
```
bjhargrave, line 124:
```suggestion
| `deleteExistingManifest`          | Whether to delete an existing file at `manifestPath`. Defaults to `false`
```

maven-plugins/bnd-maven-plugin/README.md
```diff
@@ -429,6 +453,8 @@ The `bnd-process-tests` is not executed by default, therefore at least one expli
 | `skip`                  | Skip the goal. _Defaults to `false`._ Override with property `bnd-tests.skip` or `maven.test.skip`.                                                                                                                                                                               |
 | `skipIfEmpty`           | Skip processing if `includeClassesDir` is `true` and the `classesDir` is empty. _Defaults to `false`._ Override with property `bnd.skipIfEmpty`.                                                                                                                                  |
 | `manifestPath`          | Specify the path to store the generated manifest file. _Defaults to `${project.build.testOutputDirectory}/META-INF/MANIFEST.MF`._                                                                                                                                                 |
+| `deleteExistingManifest`          | Whether to delete an existing `META-INF/MANIFEST.MF` inside the `classesDir` directory. Defaults to `false`
```
bjhargrave, line 456:
```suggestion
| `deleteExistingManifest`          | Whether to delete an existing file at `manifestPath`. Defaults to `false`
```

maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
chrisrueger: I added a new option `deleteExistingManifest` (default `false`)


@Marcono1234 with this configuration repeated 
mvn process-classes
mvn process-classes

```
<plugin>
                <version>7.1.0-SNAPSHOT</version>
                <executions>
                    <execution>
                        <goals>
                <goal>bnd-process</goal>
            </goals>
                        <configuration>
                        	<deleteExistingManifest>true</deleteExistingManifest>
                            <bnd><![CDATA[
                                -exportcontents: example.a, example.b, example.c
                                Import-Package: example.c;version="${range;[==,=+)}", *
                            ]]></bnd>
                        </configuration>
                    </execution>
                </executions>
            </plugin>
```

maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
```diff
@@ -259,12 +266,16 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
-				File existingMetaInf = new File(classesDir, "META-INF/MANIFEST.MF");
-				if (existingMetaInf.exists()) {
-					logger.info("Delete existing manifest from previous build: " + existingMetaInf);
-					IO.delete(existingMetaInf);
+				if (deleteExistingManifest) {
+					// delete existing MANIFEST.MF
+					// e.g. from previous build to avoid that this MANIFEST.MF
+					// becpmes input to the bnd process (which could lead to
+					// suprising results in some cases)
+					File existingMetaInf = new File(classesDir, "META-INF/MANIFEST.MF");
```
bjhargrave, line 274:
```suggestion
					File existingMetaInf = getManifestPath();
```

This should be based upon the manifestPath defined for the plugin. Not a hardcoded path.

Since the concern here is repeated runs of the mojo, the manifestPath value is where the manifest is from the previous run.

maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
```diff
@@ -259,12 +266,16 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+					if (existingMetaInf != null && existingMetaInf.exists()) {
```
bjhargrave, line 275: I don't think you need to do these checks. Just call IO.delete. If getManifestPath returns null, we have big problems and should NPE.

bjhargrave:
You also need to remove the merge commit. Only normal commits in a PR.

maven-plugins/bnd-maven-plugin/README.md
```diff
@@ -92,13 +92,37 @@ The `bnd-process` is not executed by default, therefore at least one explicit ex
...
 | `bndfile`               | File path to a bnd file containing bnd instructions for this project. The file path can be either absolute or relative to the project directory. _Defaults to `bnd.bnd`_.                                                                                                         |
 | `bnd`                   | Bnd instructions for this project specified directly in the pom file. This is generally be done using a `<![CDATA[  ]]>` section. If the projects has a `bndfile` configuration property or a file in the default location `bnd.bnd`, then this configuration element is ignored. |
 | `manifestPath`          | Specify the path to store the generated manifest file. _Defaults to `${project.build.outputDirectory}/META-INF/MANIFEST.MF`._                                                                                                                                                     |
+| `deleteExistingManifest`          | Whether to delete an existing file at `manifestPath`. Defaults to `false`
```
bjhargrave, line 124:
```suggestion
| `deleteExistingManifest`          | Whether to delete an existing file at `manifestPath` before execution. _Defaults to `false`._
```

maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
```diff
@@ -258,6 +265,17 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 			// handle projects with no output directory, like
 			// 'test-wrapper-bundle'
 			if (classesDir.isDirectory()) {
+
+				if (deleteExistingManifest) {
```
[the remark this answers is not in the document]
chrisrueger: Right.

maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
```diff
@@ -254,6 +260,16 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 				throw new MojoExecutionException("Sub-bundles not permitted in a maven build");
 			}
 
+			if (deleteExistingManifest) {
+				// delete existing MANIFEST.MF
+				// e.g. from previous build to avoid that this MANIFEST.MF
+				// becpmes input to the bnd process (which could lead to
```
bjhargrave, line 266:
```suggestion
				// becomes input to the bnd process (which could lead to
```

maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
```diff
@@ -254,6 +260,16 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
+				// suprising results in some cases)
+				File existingMetaInf = getManifestPath();
+				logger.info("Delete existing manifest: " + existingMetaInf);
```
bjhargrave, line 269:
```suggestion
				logger.info("Delete existing manifest: {}", existingMetaInf);
```

bjhargrave (changes requested):

bjhargrave:
This all looks good to me. The final thing needed is to squash all this into a single commit. Thanks for the work on this PR!

pkriens:
LGTM

bjhargrave (approved):

maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
```diff
@@ -254,6 +260,16 @@ public void execute() throws MojoExecutionException, MojoFailureException {
...
 			}
 
+			if (deleteExistingManifest) {
+				// delete existing MANIFEST.MF
+				// e.g. from previous build to avoid that this MANIFEST.MF
+				// becomes input to the bnd process (which could lead to
+				// suprising results in some cases)
+				File existingMetaInf = getManifestPath();
+				logger.info("Delete existing manifest: {}", existingMetaInf);
+				IO.delete(existingMetaInf);
```
laeubi, line 270: I think deleting the manifest file is only working when the manifest path points to the default (classes/META-INF/MANIFEST) what is not always the case.

Instead I think the real culprit is some lines below where you create a JAR from the classesDir, this one should be filtered to not include the Manifest-File so `Pattern doNotCopy` might better be specified and remove the unwanted file.

chrisrueger: @laeubi could you create a new Issue based on the comment with a few more details, describing when this could be a problem?

laeubi: I just wanted to mention that this only solves a very specific case in a very specific way.
It might therefore be better to handle it in a generic way (that is when creating the (bnd) Jar from a folder) and without special options / deletion / ...

chrisrueger: Ah ok thanks @laeubi Yes, I agree, but in this case I don't feel confident enough to make such a change, because don't have a specific problem and I am too unfamiliar with the BndMavenPlugin ( I don't use it), that I can foresee the consequences. The only reason I made this PR here is, that I just wanted to help out fixing this specific problem for @Marcono1234 and since it was a new option, I felt it it low risk. Otherwise I wouldn't have touched this. 

So if you have a testcase a PR is welcomed. 

@bjhargrave @rotty3000 any opinions?

Issue #6220, opened 2024-08-17
[bnd-maven-plugin] Generated substitution import does not define version range


Marcono1234:
Version: Maven plugin `biz.aQute.bnd:bnd-maven-plugin` 6.4.0

Hello,
https://github.com/google/gson uses the bnd-maven-plugin to generate the OSGi manifest attributes[^1]. Recently in https://github.com/google/gson/issues/2725 it was reported that the "substitution" `Import-Package` for `com.google.gson.annotations` causes issues when multiple third-party bundles import Gson packages.

In that Gson issue and the linked discussions it was mentioned that the problem might be that Gson's own `Import-Package` (generated by bnd) does not define a version range for `com.google.gson.annotations`[^2]:
```
Import-Package: sun.misc;resolution:=optional,com.google.gson.annotations
```

Do you think as well that this is the cause here, and that the solution is to use `...;version="${range;[==,=+)}"` or similar? But in that case is that something bnd can / should generate on its own?

Disclaimer: I am not very familiar with OSGi, so please let me know if there is any misunderstanding.

[^1]: Maven plugin configuration: https://github.com/google/gson/blob/9615c8d7c895797872d7e601702e690e5ed7a2e8/gson/pom.xml#L125-L157
[^2]: Can for example be seen in the [Gson 2.11.0 JAR](https://repo1.maven.org/maven2/com/google/code/gson/gson/2.11.0/gson-2.11.0.jar), or if you build Gson manually with `mvn package`; the JAR is then under `/gson/target/gson-2.11.1-SNAPSHOT.jar`.

pkriens:
I expect that there is no version available for com.google.gson.annotations when gson is processed? I.e. its JAR contains no version information.

It must then explicitly set in the Import-Package, i.e. don't use a ${range}, just give the version. Best is of course to provide a version for com.google.gson.annotations package in its JAR..

Marcono1234:
There might be a misunderstanding here, `com.google.gson.annotations` is part of the regular Gson JAR. So bnd generates both an `Export-Package` and an `Import-Package` entry for it[^1]. And the `Export-Package` does include the version, but the `Import-Package` does not include any version range. Sorry if I did not make that clear enough.

Therefore I think the version information should be present to bnd, at least in theory. But maybe it is not accessible to that part of bnd for whatever reason?

[...]

PR #6238, opened 2024-08-27
Add default version of export to selfimports

chrisrueger:
This is an thought-experiment / prove of concept / based on discussion for #6220. 
Relates to https://github.com/bndtools/bnd/issues/6220#issuecomment-2302742794

## In this PR

- after `exports` and `imports` have been processed in `Builder.analyze()` 
- we calculate self-imports
- and then call `augmentImports(selfImports, exports)` to `applyVersionPolicy()` to any of those self-imports which 
**a)** do not yet have a version or 
**b)** specify a range-mask with the rang macro (e.g. `Import-Package: example.c;version="${range;[==,=+)}"`)
- `-nodefaultversion true` still works and could disable the behavior

Based on the observation in https://github.com/bndtools/bnd/issues/6220#issuecomment-2299809530 the following will happen for this bnd:

**.bnd:**
```
Bundle-Version: 1.0.1
-exportcontents: example.a, example.b, example.c
Import-Package: example.c;version="${range;[==,=+)}", *
```

**MANIFEST.MF:**
```
Bundle-Version: 1.0.1
Export-Package: example.a;uses:="example.b";version="1.0.1",example.b;version="1.0.1",example.c;version="1.0.1"
Import-Package: example.b;version="[1.0,2)",java.lang,example.c;version="[1.0,1.1)"
```

instead of as it was before:

```
Export-Package: example.a;uses:="example.b";version="1.0.1",example.b;version="1.0.1",example.c;version="1.0.1"
Import-Package: example.b,java.lang,example.c;version="${range;[==,=+)}"
```

## Why or When would that be useful?

As outlined in #6220 3rd-party depdencies like Gson etc. use bnd to create OSGi bundles via the maven or gradle plugins.
But often there is no OSGi-knowledge present hence there are often no version information at the package-level like: 
- packageinfo file in the package directory (recommended)
- package-info.java, bnd’s Version annotation
- Manifest of the exporting JAR

Only a `Bundle-Version` is usually mainted. 

For all those dependencies out there the previous behavior already added the `Bundle-Version` as the last resort to Exports, while self-imports would **not** get any version information. 

But **now** the self-imports would get the version information from the export even it is the "last resort" `Bundle-Version`

## Open Questions

- what about if this be the (new) default behavior?
- does it even make sense?
- what could go wrong if this would be the new default? What could it break? 
- or should there be a new option / instruction to enable this behavior?

Feedback appreciated @pkriens @bjhargrave

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
@@ -458,14 +458,58 @@ public class Builder extends Analyzer {
 	@Override
 	public void analyze() throws Exception {
 		super.analyze();
-		cleanupVersion(getImports(), null, Constants.IMPORT_PACKAGE);
-		cleanupVersion(getExports(), getVersion(), Constants.EXPORT_PACKAGE);
+		Packages exports = getExports();
+		Packages imports = getImports();
+
+		cleanupVersion(imports, null, Constants.IMPORT_PACKAGE);
+		cleanupVersion(exports, getVersion(), Constants.EXPORT_PACKAGE);
+		addExportVersionToSelfImports(imports, exports);
+
 		String version = getProperty(BUNDLE_VERSION);
 		if (version != null) {
 			version = cleanupVersion(version);
 			version = doSnapshot(version);
 			setProperty(BUNDLE_VERSION, version);
 		}
+	}
+
+	/**
+	 * Add version of the export to Selfimports in case the import does not yet
+	 * have a version. This method should be called after exports have been
+	 * processed via {@link #cleanupVersion(Packages, String, String)}
+	 *
+	 * @param imports
+	 * @param exports
+	 * @throws Exception
+	 */
+	private void addExportVersionToSelfImports(Packages imports, Packages exports) throws Exception {
+		Packages selfImports = calcSelfImports(exports, imports);
+
+		// augmentImports() does use applyVersionPolicy internally which is the
+		// main thing we want here
+		augmentImports(selfImports, exports);
+	}
+
+	/**
+	 * Calculates selfImports which are imports which are also exported by the
+	 * same jar
+	 *
+	 * @param exports
+	 * @param imports
+	 * @return
+	 */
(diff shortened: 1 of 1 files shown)


bjhargrave:
This is a bad idea. Because export defaults to using Bundle-Version does not mean it is a valid choice on the import side. There is probably no other bundle which exports the package at the Bundle-Version of the importer. So matching the import at the Bundle-Version will be pointless. Bundle-Version is not semantic and this not suitable for use on imports.

Exporters need to use proper versions. Importing at the Bundle-Version is a bad idea. So I would reject this proposed change.

chrisrueger:
Thanks @bjhargrave for the feedback. 

I am just wondering if there is any other way out of the situation described in https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/3236#issuecomment-2276399714 other than manually fixing in upstream (gson). 

> The problem, as you've noted, is that both gson bundles import just "com.google.gson.annotations" (implied >= 0.0.0) leaving the door open for resolution between different versions of the same library 😐 gson 2.10.1 should import with a range of [2.10.1,2.11.0) and gson 2.11.0 should import with a range of [2.11.0,2.12.0) to be safe.
> 
> com.google.gson_2.10.1.jar
> Import-Package = sun.misc;resolution:=optional,com.google.gson.annotations;version="[2.10.1,2.11.0)"
> 
> com.google.gson_2.11.0.jar
> Import-Package = sun.misc;resolution:=optional,com.google.gson.annotations;version="[2.11.0,2.12.0)"

My interpretation is that gson "expects" that its own `com.google.gson.annotations` is imported, but apparently another bundle of the same library got wired.

bjhargrave:
The solution is always to fix the upstream :-)

In the alternate, if the bundle must not import the package from another bundle, then its bnd instructions should decorate the export with `-noimport:=true`. Then it won't self-import and will just export. When the bundle goes to access the package, it will always use its own copy since it can never import the package.

chrisrueger:
Closing. Continue at comment https://github.com/bndtools/bnd/issues/6220#issuecomment-2312660794

chrisrueger:
@juergen-albert @maho7791 ocx conference car discussion

PR #6253, opened 2024-08-31
NewWorkspace Wizzard: fix switch workspace after finish

### LINKED ISSUE #6252: New bnd workspace: Switch to new workspace after finish not working
`Bndtools	7.1.0.DEV-202408171344`
Eclipse Version: 2024-06 (4.32.0)
Build id: 20240606-1231

1. When creating a new Workspace using the wizzard like this.

image

Expected behavior: 
Eclipse restarts and I am in the new Workspace

Actual behavior:
Eclipse restarts and I am still in the current Workspace I was before (but the new workspace was created... I can manually switch to it)

chrisrueger:
Closes #6252 

Now instead of doing an instant restart (which did not switch to the new workspace) we show the existing Eclipse Switch Workspace dialog but have the new workspace location pre-filled. Since this reuses existing Eclipse feature of OpenWorkspaceAction all the switch workspace logic can be reused.

image

image

Note: It seems the Switch Workspace feature already does not work in a Development instance of Eclipse (when launched from within Eclipse), because I get the following error:

image

Thus I would need to merge this once to master, test and see if it works with the next snapshot build in a real Eclipse instance. 

@pkriens what do you think? It is a slight change from previous UI workflow (showing switch workspace dialog instead of restarting directly), but since the switch did not work, maybe this is an ok compromise?

### DIFF
--- a/bndtools.core/src/bndtools/wizards/newworkspace/Model.java
+++ b/bndtools.core/src/bndtools/wizards/newworkspace/Model.java
@@ -14,7 +14,10 @@ import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.swt.widgets.Display;
+import org.eclipse.ui.IWorkbenchWindow;
 import org.eclipse.ui.PlatformUI;
+import org.eclipse.ui.internal.ide.ChooseWorkspaceData;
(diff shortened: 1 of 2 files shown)


pkriens:
I am ok with this.

chrisrueger:
@pkriens just want to confirm it is working as (I) expected. 
Snapshot: `Bndtools 7.1.0.DEV-202409021143-g564b890`

The "Select workspace" modal dialog appears with the new workspace name prefilled and when I click "Launch" Eclipse restarts and I am in the fresh new workspace.

image

PR #6254, opened 2024-09-01
Fix NPE

chrisrueger:
Related to https://github.com/bndtools/bnd/issues/6199#issuecomment-2323434448

Display.getCurrent() is null already when stop is called. So check for null and only execute if not null. Since it was called when quitting Eclipse, it is probably very late in the process where most things are already shutdown.

This is a similar fix as to the one in https://github.com/bndtools/bnd/pull/6223


chrisrueger:
@pkriens just FYI i merged this because I think it is low risk and similar to https://github.com/bndtools/bnd/pull/6223

PR #6266, opened 2024-09-13
remove usage of groovy nature

peterkir:
Removal for providing oomph setups, without groovy nature


pkriens:
@peterkir can you fix the signing of the commit? Then we can merge & close this?

PR #6270, opened 2024-09-14
Remove imports of exports without a version range

### LINKED ISSUE #6267: Prevent generation of Substitution Packages (Import-Package) without version
## Problem 

In https://github.com/bndtools/bnd/issues/6220 a problem was reported that bnd  has generated an Import-Package header for 

- https://github.com/google/gson/blob/9615c8d7c895797872d7e601702e690e5ed7a2e8/gson/pom.xml#L125-L157

``` 
# .bnd

                Import-Package: sun.misc;resolution:=optional, *

                -exportcontents:\
                    com.google.gson,\
                    com.google.gson.annotations,\
                    com.google.gson.reflect,\
                    com.google.gson.stream
```

created an Import-Package for `com.google.gson.annotations` **without a version range** :

```
# generated Manifest.mf

Export-Package                          com.google.gson.annotations;version="2.11.0"
                                        com.google.gson.reflect;version="2.11.0"
                                        com.google.gson.stream;version="2.11.0"
                                        com.google.gson;uses:="com.google.gson.reflect,com.google.gson.stream";version="2.11.0"
Import-Package                          com.google.gson.annotations
                                        sun.misc;resolution:=optional
```

Note: the version on the `Export-Package`s (`2.11.0`) was taken from the `Bundle-Version` header, which is some kind of last-resort when no version is provided on the exports (not good OSGi practise but very common in the wild for library bundles who "just" want to add OSGi meta data via bnd)

## Expected result

As discussed in the call today with @pkriens it was disucssed that in the case above 

`com.google.gson.annotations` should **NOT** appear in the `Import-Package` because it has no version. 

## Related OSGi spec sections

- https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#i3238802
- https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#framework.module-import.export.same.package

chrisrueger:
Closes #6267

imports (coming from exports) should not be added if the export does not specify a version, because it would lead to imports without a version. this could cause resolver problems because the resolver then has no version constrains and too many wiring options.


chrisrueger:
Another blog post which covers exports/imports: https://blog.osgi.org/2007/04/importance-of-exporting-nd-importing.html

pkriens:
LGTM

chrisrueger:
Thanks @pkriens , I will merge then.

Issue #6273, opened 2024-09-17
Accessing POM via XPath is very slow


reckart:
The Maven integration of the BND tools accesses the Maven POM using XPath which is very slow.

<img width="1651" alt="Screenshot 2024-09-06 at 10 39 56" src="https://github.com/user-attachments/assets/d83895e6-71c4-4ed7-b627-e6497d42caed">

PR #6274, opened 2024-09-17
Issue #6273: Accessing POM via XPath is very slow

reckart:
- For the simple get() method, simply use DOM methods instead of going through XPath

I have tried to profile the issue again, but so far have not been able to encounter the situation where this `get()` method is called again. I even tried setting a breakpoint in the method while using Eclipse "in production" - but so far, the breakpoint did not hit again...

### DIFF
--- a/biz.aQute.repository/src/aQute/maven/provider/POM.java
+++ b/biz.aQute.repository/src/aQute/maven/provider/POM.java
@@ -317,12 +317,15 @@ public class POM implements IPom {
 		return "true".equalsIgnoreCase(other);
 	}
 
-	private String get(Element dependency, String name, String deflt) throws XPathExpressionException {
-		String value = xp.evaluate(name, dependency);
-		if (value == null || value.isEmpty())
-			return Strings.trim(deflt);
+	private String get(Element dependency, String name, String deflt) {
+		var matchingElements = dependency.getElementsByTagName(name);
+		if (matchingElements.getLength() > 0) {
+			var value = matchingElements.item(0).getTextContent();
+			if (value != null && !value.isEmpty())
+				return Strings.trim(replaceMacros(value));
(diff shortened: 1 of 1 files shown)


chrisrueger:
@reckart can you rebase master and push again? This will trigger the rebuild without the windows build ([temporarily](https://github.com/bndtools/bnd/pull/6275))

reckart:
Thing is that I do not really know under which conditions the code that I profiled to be a bottleneck is actually called (see screenshot in issue description). I wanted to check what kind of performance impact my change has - but I have not been able to trigger another situation where I would have been able to profile the same point.

chrisrueger:
Is there a test triggering this code?
Otherwise sounds like caching maybe so that stuff is only done initially? What if you delete/rename your local maven repo / eclipse workspace. Looks like m2e is involved. Not familiar with it

reckart:
I initially had an NPE in the changes and that caused some tests to fail.

chrisrueger:
Sounds like at least no test fails. 
If you are sure it is correct I'm fine with merging if you could have an eye on it and fix any potential issues in case.

 @pkriens what do you think?

pkriens:
LGTM. What's up with the windows check being canceled?

reckart:
So far I couldn't find an issue with the PR. Am I 100% sure that my code usually works - no. That said, I have been running Eclipse with a local build of the plugins installed for some days now and so far didn't hit an issue... 🤷

chrisrueger:
> LGTM. What's up with the windows check being canceled?

This here was before you disabled the windows build in the other PR.

reckart:
I have the feeling that the JAR files in the target folder are no longer updated in my workspace since I have installed the local build. Haven't yet tried tracking it down though...

PR #6277, opened 2024-09-18
[ds] Place properties warning on class; extra warning for =

### LINKED ISSUE #6265: [ds] Tooling support to catch property vs properties confusion
Hi all,

I pray that you are well.

I just wasted a good 15-30m because I created a component, and because I hadn't done it for a while I used the "properties" field rather than the "property" field. When I figured out the problem I remembered that I had done the same thing before.

It would be great if the Bnd DS code generation was able to detect this and issue a warning. A generic solution could be:
* Inspect all filenames specified in the properties annotation field and warn if they do not exist in the finally assembled bundle.
* For any that do not exist, if they are of the form "A=B" then add the suggestion to the warning along the lines of: "Did you mean to use "property" attribute instead?"

 Bonus points if Bndtools could also highlight the offending property in the code directly with the yellow squiggle underline.

kriegfrj:
This PR:

1. makes this warning appear on the source of the class itself to make it more noticeable, rather than on the `bnd.bnd` file;
2. adds an extra hint if it sees that the filename contains an '=', asking "did you mean to use the "property" attribute?"; and
3. replicates this functionality for the `factoryProperties` attribute.

Hopefully this will prevent my future self from wasting more precious brain cycles (which are in ever-increasingly shorter supply).

I did note that Bndtools only seems to display one warning per `@Component` tag (the last one that you specify). I'm not sure if this is Bndtools overwriting or if it is the Bnd reporter mechanism itself. I think it would be preferable if all warnings were displayed.

Fixes #6265


pkriens:
lgtm, thanks

PR #6285, opened 2024-09-23
Template fragments: show Author column to make it clear when you pull in content from non-bndtools / 3rd-party

chrisrueger:
Closes https://github.com/bndtools/bnd/issues/6286

This allows a user to see where a template fragment is coming from. E.g. if the github organisation of the fragment is **not** "bndtools" we show "`Org-name (3rd Party)`"
A double click on the entry opens the Github Repo of the template in the browser (e.g. https://github.com/mnlipp/de.mnl.osgi/tree/HEAD/de.mnl.osgi.bnd.repository/workspace-template).

image

A template definition like this is from an external 3rd-party author.

```
mnlipp/de.mnl.osgi/de.mnl.osgi.bnd.repository/workspace-template#master; \
        name=IndexedMavenRepository; \
        description="Provide an OSGi repository derived from a subset of one or more maven repositories";\
```

**Confirmation popup**

If you have selected at least one 3rd-party template, a confirmation popup shows up. We do this for security reason, because you are downloading content from a potentially untrusted source.

image

But this allows to go with the practise that 3rd-party authors can point to their `master` / `main` branch, which has the practical effect, that users always get the latest template, in case the 3rd-party author makes updates.


chrisrueger:
As discussed in the call with @pkriens and @juergen-albert  I will revert / discard the "`snapshot`" attribute idea.
Instead I will do:

- keep the Author column
- show a green checkmark icon if organization is `bndtools` (Official)
- show a yellow exclamation mark icon if it is a 3rd-party (non bndtools org)
- maybe print a warning / confirmation dialog  if you have checked a template from a 3rd-party so that the user has to acknowledge that again

Based on this we decided to also allow Template authors to point to their master branch (@mnlipp )

I will work on that the next days and update the PR accordingly.

chrisrueger:
@pkriens @mnlipp @juergen-albert I have pushed and  updated the PR description with the latest state, based on our discussion earlier. Also adjusted the docs.

mnlipp:
> @mnlipp are you ok with your repo appearing in the screenshot?

Only just seen that. No problem.

chrisrueger:
@pkriens how should we proceed here? I implemented all of what we discussed and decided in the call and adjusted the PR description accordingly.
Just merging it and give it a try?

pkriens:
Yup! Better to ask for a revert sometimes than to never fail :-)

Merge!

PR #6299, opened 2024-09-27
Resolution view: improve wording of hideSelfImportsFilter Button

chrisrueger:
the wording of the button for  hideSelfImportsFilter was not clear and misleading.
After working on https://github.com/bndtools/bnd/pull/6270 I finally understood what this button actually does.

image

image


chrisrueger:
@pkriens would be great to get that in. Would like to show it at OCX 😃

Issue #6301, opened 2024-09-27
[New Create new Workspace] 


juergen-albert:
![image](https://github.com/user-attachments/assets/8cbe4c98-db87-44b1-8c9c-09624cef09ff)

We should bring back the Create new Workspace and maybe even have 2 links, with one for the new and one for the old mechanism.

![image](https://github.com/user-attachments/assets/b48eb93a-c1b0-4eef-b1cb-d7934529d2e2)

PR #6303, opened 2024-09-27
bring back the no-projects links

chrisrueger:
Closes https://github.com/bndtools/bnd/issues/6301

now the links for
1. Create a new bnd workspace.
2. Bnd OSGi Workspace (Deprecated)

appear again in the Bnd Perspective for an empty workspace if you have no projects. unfortunately I had to shorten the description because it is used instead of the name and was too long. Drawback is, that  the File / New Entries used that as a tooltip, which is now much shorter. But at least both views look ok

image

image

@juergen-albert


chrisrueger:
@juergen-albert since you created the issue. is this what you have in mind?

chrisrueger:
@juergen-albert ping 😄

pkriens:
Taken over by #6351

Issue #6304, opened 2024-09-28
Trying to wrap a non-OSGi jar using bnd, where the jar contains META-INF/services


royteeuwen:
I have a very simple pom.xml where I'm trying to wrap a non-OSGi jar using the bnd-maven-plugin where the only dependency of the pom is the non-osgi jar with exclusion of all dependencies. I've tried the following two approaches:

```
-includeresource: non-osgi-jar-*.jar;lib:=true
```

and 
```
-conditionalpackage: *
```

But for both approaches, it does not generate the `Provide-Capability` headers for `osgi.serviceloader;osgi.serviceloader`

What would be the correct approach so that the `bnd-maven-plugin` detects that the embedded jar contains `META-INF/services` files and generate the correct headers in the osgi-wrapped bundle?

chrisrueger:
This issue https://github.com/bndtools/bnd/pull/5912 and PR https://github.com/bndtools/bnd/issues/5903 sound related. But sounds like the code of the jar need to be under your control.

royteeuwen:
@chrisrueger of course, but that would mean that I have to always manually check when a version update occurs for the wrapped library if it contains ServiceLoader services file. If possible, I'd like to avoid it :)!

chrisrueger:
> @chrisrueger of course, but that would mean that I have to always manually check when a version update occurs for the wrapped library if it contains ServiceLoader services file. If possible, I'd like to avoid it :)!

Yeah i should've have written that about the manual adding 😃, but wasn't sure how familiar you are with bnd so i thought i mention it.

Anyway, the topic is not my expertise. Let's wait if there is anybody else.

royteeuwen:
👍  thanks for checking in any case! So far I have only found a gradle example, where it's at least sort of automated, it checks if any file exists in the META-INF/services and then creates the osgi headers for those, without explicitely naming which ones it are: https://github.com/apache/groovy/blob/master/build-logic/src/main/groovy/org.apache.groovy-library.gradle#L41
 
Hoping to find a (semi) automated manner with maven as well 😄

pkriens:
```
-includeresource @non-osgi-jar.jar
```
This will unroll your jar into the new jar. This will include the manifest but this gets updated. Although I am not sure the capability gets updated? This is normally created via the annotation?

royteeuwen:
[...]

PR #6309, opened 2024-10-03
auto register capabilities for META-INF/services

chrisrueger:
## In this PR

this is an idea based on https://github.com/bndtools/bnd/issues/6304 by @royteeuwen

it leverages the recent addition of PR https://github.com/bndtools/bnd/pull/5912 that allowed to add annotations in comments of files in `META-INF/services`.
See https://github.com/bndtools/bnd/blob/master/docs/_chapters/230-manifest-annotations.md#meta-infservices-annotations

In this PR we basically add a `aQute.bnd.annotation.spi.ServiceProvider` annotation automatically if a service doesn't specify one. this causes bnd to generate `Provide-Capability` manifest headers for files in `META-INF/services` which was the requirement in the #6304 

## example

`bnd.bnd` file and a [groovy-3.0.22.jar](https://mvnrepository.com/artifact/org.codehaus.groovy/groovy/3.0.22) in a the `lib` folder

```
-buildpath: biz.aQute.bnd.annotation;version='7.0'
-includeresource: @lib/groovy-3.0.22.jar
-metainf-services: auto
```

created the following manifest headers

```
Provide-Capability                      osgi.service;objectClass:List<String>="org.codehaus.groovy.transform.ASTTransformation";effective:=active
                                        osgi.serviceloader;osgi.serviceloader="org.codehaus.groovy.transform.ASTTransformation";register:="groovy.grape.GrabAnnotationTransformation"
Require-Capability                      osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
                                        osgi.extender;filter:="(&(osgi.extender=osgi.serviceloader.registrar)(version>=1.0.0)(!(version>=2.0.0)))"
```

Note: `-metainf-services: auto` is a new instruction needed to activate this new behavior (see comment below)

make sure `biz.aQute.bnd.annotation` is on the buildpath which contains the 'aQute.bnd.annotation.spi.ServiceProvider' annotation.

Let's use this as a base for discussion if it makes sense at all, or needs configurability etc.

### DIFF
--- a/docs/_chapters/230-manifest-annotations.md
+++ b/docs/_chapters/230-manifest-annotations.md
@@ -28,9 +28,10 @@ Export-Package: com.acme;version="1.3.4"
 
 ### META-INF/services Annotations
 
-Some developers do not want to rely on the bnd/OSGi annotations. For this reason, it is possible to also apply the annotations to the files in the `META-INF/services`. These Service Loader files have the name of a Service Loader _service_ and contain the names of the implementation classes. One fully qualified name per line.
+Some developers do not want to rely on the additional dependency of bnd/OSGi annotations. For this reason, it is possible to also apply the annotations textually in comments to the files in the `META-INF/services`.
+This trick avoids a compile time dependency. These Service Loader files have the name of a Service Loader _service_ and contain the names of the implementation classes. One fully qualified name per line.
 
-The annotations can be applied in the comments. To make them more readable, it is possible to import the fully qualified name of the annotation.
+To make these annotations in the comments more readable, it is possible to import the fully qualified name of the annotation.
 ```
 META-INF/services/com.example.ServiceType:
 
@@ -42,9 +43,11 @@ META-INF/services/com.example.ServiceType:
 ```
 The processing is identical to the normal class based annotation processing. The `#class` macro will be set to the implementation class. The `#value` will be set in all cases to the service type unless overridden.
 
-This behavior can be controlled with the [-metainf-services](/instructions/metainf-services.html) instruction. Default is `auto` which automatically create `Provide-Capability` headers for services without textual annotations.
+This behavior can be controlled with the [-metainf-services](/instructions/metainf-services.html) instruction. Default is `annotation` which processes the textual annotations above, while the other convenience strategy `auto` automatically creates `Provide-Capability` headers for services without any textual annotations.
 
-Clearly using the class annotation is far superior:
+<hr />
+
+While the above is a compromise, clearly using the real class annotation is far superior:
 
 * Help from the IDE
 * Impossible to make typos
--- a/docs/_chapters/230-manifest-annotations.md
+++ b/docs/_chapters/230-manifest-annotations.md
@@ -42,6 +42,8 @@ META-INF/services/com.example.ServiceType:
 ```
 The processing is identical to the normal class based annotation processing. The `#class` macro will be set to the implementation class. The `#value` will be set in all cases to the service type unless overridden.
 
+This behavior can be controlled with the [-metainf-services](/instructions/metainf-services.html) instruction. You can set `-metainf-services: auto` to also automatically create `Provide-Capability` headers for services without annotations.
+
 Clearly using the class annotation is far superior:
 
 * Help from the IDE
@@ -51,6 +53,12 @@ Clearly using the class annotation is far superior:
 The analysis of these files happens after the analyzer plugins have been run. These plugins can add files if so desired.
 
 Since the annotations & imports happen in the comments, it is not possible to diagnose any errors. If the comment does not match its regular expression, it will be silently ignored.
+
+
+**Note:** Make sure `biz.aQute.bnd.annotation` is on the classpath / buildpath which contains the `aQute.bnd.annotation.spi.ServiceProvider` annotation. Otherwise the annotation won't be processed and you might see a warning in the log like this: 
+
+`Unable to determine whether the meta annotation aQute.bnd.annotation.spi.ServiceProvider applied to type xyz provides bundle annotations as it is not on the project build path. If this annotation does provide bundle annotations then it must be present on the build path in order to be processed`
+
 
 ### @Requirement & @Capability Annotations
 
--- a/docs/_chapters/230-manifest-annotations.md
+++ b/docs/_chapters/230-manifest-annotations.md
@@ -42,7 +42,7 @@ META-INF/services/com.example.ServiceType:
 ```
 The processing is identical to the normal class based annotation processing. The `#class` macro will be set to the implementation class. The `#value` will be set in all cases to the service type unless overridden.
 
-This behavior can be controlled with the [-metainf-services](/instructions/metainf-services.html) instruction. You can set `-metainf-services: auto` to also automatically create `Provide-Capability` headers for services without annotations.
+This behavior can be controlled with the [-metainf-services](/instructions/metainf-services.html) instruction. Default is `auto` which automatically create `Provide-Capability` headers for services without textual annotations.
 
 Clearly using the class annotation is far superior:
 
@@ -53,11 +53,6 @@ Clearly using the class annotation is far superior:
 The analysis of these files happens after the analyzer plugins have been run. These plugins can add files if so desired.
 
 Since the annotations & imports happen in the comments, it is not possible to diagnose any errors. If the comment does not match its regular expression, it will be silently ignored.
-
-
-**Note:** Make sure `biz.aQute.bnd.annotation` is on the classpath / buildpath which contains the `aQute.bnd.annotation.spi.ServiceProvider` annotation. Otherwise the annotation won't be processed and you might see a warning in the log like this: 
-
-`Unable to determine whether the meta annotation aQute.bnd.annotation.spi.ServiceProvider applied to type xyz provides bundle annotations as it is not on the project build path. If this annotation does provide bundle annotations then it must be present on the build path in order to be processed`
 
 
 ### @Requirement & @Capability Annotations
--- a/biz.aQute.bndlib.tests/test/test/annotationheaders/SPIAnnotationsTest.java
+++ b/biz.aQute.bndlib.tests/test/test/annotationheaders/SPIAnnotationsTest.java
@@ -72,6 +72,47 @@ public class SPIAnnotationsTest {
 				"META-INF/services/test.annotationheaders.spi.SPIService;literal='test.annotationheaders.spi.providerE.Provider'");
 			b.setProperty("Provide-Capability",
 				"osgi.serviceloader;osgi.serviceloader='test.annotationheaders.spi.SPIService';register:='test.annotationheaders.spi.providerE.Provider'");
+			b.setProperty(Constants.METAINF_SERVICES, Constants.METAINF_SERVICES_STRATEGY_ANNOTATION);
+			b.build();
+			b.getJar()
+				.getManifest()
+				.write(System.out);
+			assertTrue(b.check());
+
+			Attributes mainAttributes = b.getJar()
+				.getManifest()
+				.getMainAttributes();
+
+			Header req = Header.parseHeader(mainAttributes.getValue(Constants.REQUIRE_CAPABILITY));
+			assertEquals(1, req.size());
+
+			assertEE(req);
+
+			Header cap = Header.parseHeader(mainAttributes.getValue(Constants.PROVIDE_CAPABILITY));
+			assertEquals(1, cap.size());
+
+			Props p = cap.get("osgi.serviceloader");
+			assertNotNull(p);
+			assertNotNull(p.get("osgi.serviceloader"));
+			assertEquals("test.annotationheaders.spi.SPIService", p.get("osgi.serviceloader"));
+			assertNotNull(p.get("register:"));
+			assertEquals("test.annotationheaders.spi.providerE.Provider", p.get("register:"));
+
+			assertServiceMappingFile(b.getJar(), "test.annotationheaders.spi.SPIService",
+				"test.annotationheaders.spi.providerE.Provider");
+		}
+	}
+
+	@Test
+	public void testServiceProvider_existingdescriptorAuto() throws Exception {
+		try (Builder b = new Builder();) {
+			b.addClasspath(IO.getFile("bin_test"));
+			b.setPrivatePackage("test.annotationheaders.spi.providerE");
+			b.setProperty("-includeresource",
+				"META-INF/services/test.annotationheaders.spi.SPIService;literal='test.annotationheaders.spi.providerE.Provider'");
+			b.setProperty("Provide-Capability",
+				"osgi.serviceloader;osgi.serviceloader='test.annotationheaders.spi.SPIService';register:='test.annotationheaders.spi.providerE.Provider'");
+			b.setProperty(Constants.METAINF_SERVICES, Constants.METAINF_SERVICES_STRATEGY_AUTO);
 			b.build();
 			b.getJar()
 				.getManifest()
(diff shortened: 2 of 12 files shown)


pkriens:
I think this is an ok idea and the code is quite suitable for this. 

However, I think the decision can better be done in MetaInfServiceParser. The MetaInfService code is gathering the information of the services directory. It is generally not a good idea then to mix in strategy. I.e. you can now no longer detect if the service implementation was marked or not on the higher level. You never know when you need that. You could also not leverage the project setup to treat it special. Plus it is would be a gratuitous API change.

In MetaInfServiceParser we do policy. We have access to the project and workspace properties so we can make decisions on the actual project setup, unlike MetaInfService who is a worker bee. 

Here we can detect that a given service implementation is not annotated and treat it special. We might add an instruction like `-metainf-services = auto` to the bnd.bnd file to then create the annotation on an empty Implementation if it is set.

chrisrueger:
Thanks for the feedback @pkriens , I will move the code higher up to MetaInfParse. 

> -metainf-services = auto

Wondering what should be possible values and what should be the default? What about:

- `annotation` -> the default if not set, which is current behaviour, which only processes annotations
- `auto` -> as discussed above which processes annotations and also create the annotation on an empty Implementation if it is set.
- `none` (maybe do disable processing of META-INF/services completely. Not sure anybody would want that. )

As I have no experience with this topic just out of curiosity: What would be the effect if "auto" would be the default on libraries out there in the wild?

chrisrueger:
@royteeuwen could you test this PR with your bundle?

royteeuwen:
I tried it on my branch (see referenced commit) but it doesnt produce the headers yet. Anything I can do to debug?

chrisrueger:
> Start using an experimental feature to automatically create the corre…

Hmm...

1.  I see `-includeresource: @*.jar;lib:=true`

When I do this bnd then it seems `biz.aQute.bndlib` is being wrapped (I assume because it is on the `-buildpath`)

If I do it with one specific jar it works:
`-includeresource: @lib/groovy-3.0.22.jar;lib:=true`

royteeuwen:
it would be nice, because as you can see in my referenced branch, I'm wrapping like 20 jars in the same time using pluginManagement, which avoids me from hardcoding the thing. If there is another syntax to achieve the same, I'm completely fine with it of course!

chrisrueger:
> Anything I can do to debug?


Have you built my branch locally too? 
Just wondering where you are pulling in the changes of my PR here?

royteeuwen:
Of course :) 

```
git clone git@github.com:bndtools/bnd.git                                                                                                                                                                        
 gh pr checkout 6309

**Comment by royteeuwen** [comment]
@chrisrueger  I do think its still not using the correct bnd, I'm debugging locally by using `mvnDebug` and then attaching in the project and the lines don't match, trying to figure out why, will get back to you

**Comment by chrisrueger** [comment]
> @chrisrueger I do think its still not using the correct bnd, I'm debugging locally by using `mvnDebug` and then attaching in the project and the lines don't match, trying to figure out why, will get back to you

Just guessing:

Maybe you also need to reference the latest bndlib 7.1.0-SNAPSHOT in the `-buildpath`?

`-buildpath: biz.aQute.bndlib;version='7.1.0-SNAPSHOT'`

**Comment by chrisrueger** [comment]
> I'm wrapping like 20 jars in the same time using pluginManagement, making sure there is always only one dependency (namely the wrappable non-osgi jar), which avoids me from hardcoding the thing. If there is another syntax to achieve the same, I'm completely fine with it of course!

On that one:
If your jar filenames start with a common prefix e.g. `opentele` then you could do
`-includeresource: @opentele*.jar;lib:=true`

This avoids that `bndlib` itself ends up in the resulting jar.

**Comment by royteeuwen** [comment]
OK, I found the issue. You have to have the bnd.annotations jar still on the classpath:

```
<dependency>
                            <artifactId>biz.aQute.bnd.annotation</artifactId>
                            <version>7.1.0-SNAPSHOT</version>
                        </dependency>```

Else you get the following issue in the aQute.bnd.osgi.AnnotationHeaders:

<img width="1271" alt="Screenshot 2024-10-03 at 20 20 10" src="https://github.com/user-attachments/assets/a15c4a30-9072-425d-b757-2af841e96701">

chrisrueger:
> OK, I found the issue. You have to have the bnd.annotations jar still on the classpath:
> 
> ```
>                         <dependency>
>                             <groupId>biz.aQute.bnd</groupId>
>                             <artifactId>biz.aQute.bnd.annotation</artifactId>
>                             <version>7.1.0-SNAPSHOT</version>
>                         </dependency>
> ```
> 
> Else you get the following issue in the aQute.bnd.osgi.AnnotationHeaders:
> 

Yes I got the same issue:

Oh you are right. 
It's `biz.aQute.bnd.annotation` 

it seems `biz.aQute.bndlib` also contains the annotations, that's why it worked for me too. 

I tested with `biz.aQute.bnd.annotation` and it works fine as well and is much better because smaller. 
But I think this dependency is just needed for build / compile time and not at runtime. 

So you get the Provide-Capability headers now? 
Have you tried with a prefix`-includeresource: @opentele*.jar;lib:=true`

royteeuwen:
@chrisrueger well it seems that the `bnd-maven-plugin` doesn't include the annotations by default though, so definitely a remark that will be required in the documentation :)!

Yep, I get the `Provide-Capability` 🎉

I'm adding some extra maven properties and doing it like this:

```
-includeresource: @${opentelemetry.artifactId}-${opentelemetry.java.version}.jar;lib:=true
-metainf-services: auto
```

royteeuwen:
I'll test the actual bundle tomorrow, but seeing as it uses already existing functionality, I guess it should work out ;). Thanks a lot so far!

chrisrueger:
Yeah, and thanks for your feedback. 

> well it seems that the `bnd-maven-plugin` doesn't include the annotations by default though, so definitely a remark that will be required in the documentation :)!

Ok good point. 
Since I am not familiar with the maven plugin, could you maybe suggest a paragraph which could be added to the docs e.g. to the files I touched here?
https://github.com/bndtools/bnd/pull/6309/files#diff-6c05b50a9d4a861349bca27733122a0ea6c002f2d62086c9a857c310b2da2d41

as a code comment would be fine for me, I can work it in. 
I just would like to avoid giving maven recommendations for stuff I am not familiar with.

docs/_chapters/230-manifest-annotations.md
```diff
@@ -52,6 +54,12 @@ The analysis of these files happens after the analyzer plugins have been run. Th
 
 Since the annotations & imports happen in the comments, it is not possible to diagnose any errors. If the comment does not match its regular expression, it will be silently ignored.
 
+
+**Note:** Make sure `biz.aQute.bnd.annotation` is on the classpath / buildpath which contains the `aQute.bnd.annotation.spi.ServiceProvider` annotation. Otherwise the annotation won't be processed and you might see a warning in the log like this: 
+
+`Unable to determine whether the meta annotation aQute.bnd.annotation.spi.ServiceProvider applied to type xyz provides bundle annotations as it is not on the project build path. If this annotation does provide bundle annotations then it must be present on the build path in order to be processed`
```
royteeuwen, line 60: Maybe this warning could be more explicit? I don't see it in my build, it didn't log it. I had to really debug the code to come to it (see screenshot, it doesnt show any output)

Screenshot 2024-10-05 at 10 32 12
tput)

Doing this, would also make it clear for the maven build, seeing as adding something to the build path is done by adding a `<dependency>`

chrisrueger: This is currently logged via Sl4fj at either warning or debug level 
https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/osgi/AnnotationHeaders.java#L336

Do you see it when you enable debug logging in Maven? (not sure, but I think it is the `-X` option).

Or:
Could you enable `-pendantic:true` in the bnd section, to see if it gets logged as WARN?
image
 
Not sure what to do about it. 

@pkriens any thoughts why this was logged as warning or debug? Should we log as error or info?

chrisrueger: Ah ok it seems git blame shows a bit of why:
https://github.com/bndtools/bnd/commit/b1e83609913418b3cd8ade38ab8b260b9dbf302a

royteeuwen: Hmm, so does that mean we could set the log only for the bnd-maven-plugin instead of baseline? Or what would be a good alternative / fix so that we know there is something missing from the buildpath?

chrisrueger: > log only for the bnd-maven-plugin instead of baseline

Yeah something like that... For building to me it sounds like a missing annotation is important to log more visible. 
But I am too unfamiliar with both plugins and the thinking behind it and where to distinguish between those cases.

Maybe bnd-maven-plugin could set the `-pendantic:true` automatically? and then it gets logged as `.warn()`. 
But not sure this makes sense/

Have you tested with `-pendantic:true` to see if the maven build outputs the log message then? @royteeuwen

chrisrueger: I added experimental commit https://github.com/bndtools/bnd/pull/6309/commits/11172b9aa4dbf387340fe67b2e76f11dddea2018 where I set `pendantic:true` in `BndMavenPlugin` only.

Could you please retest by [building bnd and also bnd-maven-plugin](https://github.com/bndtools/bnd/pull/6309#issuecomment-2391967913) and see if you see the log.warn if you do not have the `biz.aQute.bnd.annotation` on the buildpath?

@pkriens @bjhargrave @timothyjward pinging you for feedback. If you have objects, I have no problem reverting this commit. Other suggestions welcome.

royteeuwen: Didn't seem to change anything. If I set the debugger, it states that the builder warnings and errors are null (in `AbstractBndMavenPlugin` but it does show that the builder.annotationHeaders.loggedMissing contains the ServiceProvider:

Screenshot 2024-10-05 at 18 57 18

chrisrueger: Sorry. In the meantime I got `mvnDebug` working too and checked out [your branch](https://github.com/orbinson/opentelemetry-osgi-wrappers/tree/feature/switch-to-bnd) to play around.

I noticed I set the `pendantic` property incorrectly. 
Checkout my new commit https://github.com/bndtools/bnd/pull/6309/commits/ae10e72df508e78dac01619ce12ecf0c18933147 

After that the build output of `mvn package` looks like:

image

and the debugger when into this branch:

image

Could you please try again? @royteeuwen

royteeuwen: Much better :D! Sorry that you had to learn / get `mvn` up and running 🥲 ...

Just going to tag @kwin, any reason for you that the `bnd-maven-plugin` should not have `pedantic:true` by default (meaning that there will be warnings logged). You made the bug for `bnd-baseline-maven-plugin` but it also affected the `bnd-maven-plugin`

royteeuwen:
@chrisrueger  everything seems to work on my side :) i already validated it in a real life OSGi env that uses the opentelemetry jars. What would be required to make this into a release?

chrisrueger:
@royteeuwen great news: I had a little discussion with @pkriens about why bnd.annotations needs to be a dependency on the classpath to use the textual ServiceProvider annotation in services files or via `-metainfo-services: auto` It seems that was a a small flaw in PR 

So he was so kind to work on a fix. https://github.com/bndtools/bnd/pull/6309/commits/b206734514b6c84617a6f75c07e8dcaf8985839b 

Long story short: 

The following dependency is **NOT** required in your pom.xml anymore:

```
<dependency>
    <artifactId>biz.aQute.bnd.annotation</artifactId>
    <version>7.1.0-SNAPSHOT</version>
</dependency> 


I will update the docs again later and then I think this PR is ready to merge.
```

chrisrueger:
@royteeuwen it got even better. We decided to make 'auto' the new default. 
So basically you don't even have to add `-metainf-services: auto` to you `pom.xml`'s `<bnd>` section. 

It is basically like you asked for in #6304 

It would be great if you could give it another testrun too. 

If your feedback is 👍  then @pkriens I think this can be merged.

royteeuwen:
Hmm, I don't know what changed, but a lot of tests are failing when compiling now. Both the `gradle` part as well as the `mvn` part. I saw that you did a rebase, did you pull in changes from master maybe? 

Screenshot 2024-10-07 at 21 15 41
Screenshot 2024-10-07 at 21 12 31
Screenshot 2024-10-07 at 21 12 11

chrisrueger:
> Hmm, I don't know what changed, but a lot of tests are failing when compiling now.

Yeah sorry. I was a bit too fast. It was the change to make 'auto' the default. 
Try to take care of that tomorrow.

royteeuwen:
> > Hmm, I don't know what changed, but a lot of tests are failing when compiling now.
> 
> Yeah sorry. I was a bit too fast. It was the change to make 'auto' the default. Try to take care of that tomorrow.

OK (y) 

I see for example that the Require-Capability in my wrapped bundle changed from
```
Require-Capability: osgi.extender;filter:="(&(osgi.extender=osgi.servi
 celoader.registrar)(version>=1.0.0)(!(version>=2.0.0)))",osgi.ee;filt
 er:="(&(osgi.ee=JavaSE)(version=1.8))"
```
to 
```
Require-Capability: osgi.extender;filter:="(&(osgi.extender=osgi.servi
 celoader.registrar)(version>=1.0.0)(!(version>=2.0.0)))",osgi.ee;filt
 er:="(&(osgi.ee=JavaSE)(version=17))"
```

(notice the Java version)

chrisrueger:
@royteeuwen I reverted for now. so 'annotation' is the default again and you need to have `-metainf-services: auto` in your `<bnd>`

I will investigate the failing tests with 'auto' default in the next days

royteeuwen:
@chrisrueger  I see you already pushed some commits in the meantime and all builds are green except the rebuild jdk17, is that still something you have to look at or should I already retest?

chrisrueger:
> @chrisrueger  I see you already pushed some commits in the meantime and all builds are green except the rebuild jdk17, is that still something you have to look at or should I already retest?

Yeah the rebuild is on my list. Haven't managed to reproduce that locally. 

 If it's not too much hazzle you could retest. But i could also ping you when i know more about the rebuild.

chrisrueger:
> hmm. Maybe making this the default is breaking too much. Hate to do this to you but I guess when you contacted me last night I had a glass of wine and felt more bravado than today.
> 

@pkriens I managed to fix the tests and build.
I squashed all my relevant commits into https://github.com/bndtools/bnd/commit/cfbef6f716160d91f66e2e16c500ce32aca8b684 

The last and maybe important change was to make the new auto-registered annotation `resolution: optional`
image

> Wdyt, shall we make the default no automatic annotations? Sorry for the work :-(

Could you please review this commit above and then let's decide if we change the default or not?

royteeuwen:
> do you see a problem with the `Require-Capability` haviing `resolution: optional` to fix errors in tests like `Unresolved requirements: osgi.extender; (&(osgi.extender=osgi.serviceloader.registrar)(version>=1.0.0)(!(version>=2.0.0)))` ?

Not at all, seems even logical. Using the service loaders should be pluggable, if you don't need them you should be able to disable it :) 

I'll see if I can get a test done tomorrow!

chrisrueger:
@royteeuwen had a review with @pkriens 
A few updates are about to come. So maybe wait with your test.

chrisrueger:
Just sidenote for me how to build locally like build is done via Github Actions (at least that is what I figured)

```
./gradlew :build
./gradlew :gradle-plugins:build
./.github/scripts/rebuild-build.sh ; ./.github/scripts/rebuild-build.sh
```

chrisrueger:
@royteeuwen I think I'm done and would appreaciate your test. 
One little glitch: 
you have to use `-metainf-service: auto` explicitly again. 
Why? After a longer discussion we agreed that we are unsure if this is considered a too big (potentially breaking) change for some projects out there if they start switching to `bnd 7.1.0`, given the effect it might have on bundle resolution. So we decided to play it safe, so that `-metainf-service: annotation` is the default (if not set). This at least requires a concious decision by a developer to annotate the services files. 
If you want 'auto' convenience you have to enable it.

There is still the option to make it the default in a future version of bnd if experience shows this options works. 
At least it is minimal, e.g.

```
# bnd.bnd
-includeresource: @lib/groovy-3.0.22.jar
-metainf-services: auto
```
can be enough.

Note: `biz.aQute.bnd.annotation` is **NOT** required as a depency in you `pom.xml`. bnd automatically knows about the `@ServiceProvider` annotation under the hood.

@pkriens I (think) I implemented everything we talked about today. 
See commit https://github.com/bndtools/bnd/pull/6309/commits/ca16c5070b6639b8d9f1b9dd88185240639f44da and the commits afterwards from today.

Regarding avoiding mixed-services we talked about: 
I splitted the services into two parts: 
- services with annotations
- and services without

https://github.com/bndtools/bnd/pull/6309/commits/1e130b280db0620cf12c02fe9c98ed78f709ee9a#diff-c50541e76ba2146c1d0ee544d6a8df58d45e1a7aaf8f3e1628143d08016e9bb0R58

This might not be the most elegant implementation of splitting, but all I could come up with today. I am open for suggestions.

pkriens:
LGTM, let's merge

royteeuwen:
@chrisrueger just tested it and works as expected, nice! Thanks a lot!

Any idea on when a new release would happen for the bnd plugin? I see the last release was already a year ago, so I'm not sure how often it happens?

pkriens:
We had talked about a release before the OCX conference but that is not happening. We should have a release in the next few months.

royteeuwen:
@pkriens  @chrisrueger  ok sad to hear, I also got a talk at adaptTo conference next week where I wanted to use the wrapped jars in the demo: https://adapt.to/2024/schedule/observability-in-aemaacs-with-opentelemetry

Guess I'll have to try and find a way to release them with the snapshot version of `bnd` 😅

pkriens:
I am trying to find some funding for this work. Since I have no more customers in this area, there is a lack of customer pressure and financial incentive.

PR #6323, opened 2024-10-11
Nice manifest writer

### LINKED ISSUE #6321: [bndlib] Provide a `-nice` option
Felix Bundle plugin has [a `niceManifest` option](https://github.com/apache/felix-dev/blob/736d48ca46f9f7e2f60430333cf2db9df3c355e5/tools/maven-bundle-plugin/src/main/java/org/apache/felix/bundleplugin/BundlePlugin.java#L138) that [formats Manifest](https://github.com/apache/felix-dev/blob/master/tools/maven-bundle-plugin/src/main/java/org/apache/felix/bundleplugin/ManifestWriter.java) 

> Output a nicely formatted manifest that still respects the 72 character line limit.

It would be good if bnd offers a `-nice` option as well, as manifest are sometimes inspected on raw text editors as well, used in debug outputs and so on.
If bndlib offer the code, felix can even migrate to that code to reduce duplication of such efforts.

chrisrueger:
Closes #6321

Felix Bundle plugin has [a niceManifest option](https://github.com/apache/felix-dev/blob/736d48ca46f9f7e2f60430333cf2db9df3c355e5/tools/maven-bundle-plugin/src/main/java/org/apache/felix/bundleplugin/BundlePlugin.java#L138) that [formats Manifest](https://github.com/apache/felix-dev/blob/master/tools/maven-bundle-plugin/src/main/java/org/apache/felix/bundleplugin/ManifestWriter.java)

> Output a nicely formatted manifest that still respects the 72 character line limit.

We adapted this and use it as the new default.

### DIFF
--- a/aQute.libg/src/aQute/lib/manifest/ManifestUtil.java
+++ b/aQute.libg/src/aQute/lib/manifest/ManifestUtil.java
@@ -72,7 +72,16 @@ public final class ManifestUtil {
 
 	private ManifestUtil() {}
 
+	/**
+	 * Writes out the manifest, nicely formatted. Use
+	 * {@link #write(Manifest, OutputStream, boolean)} to control 'nice'
+	 * formatting.
+	 */
 	public static void write(Manifest manifest, OutputStream out) throws IOException {
+		write(manifest, out, true);
+	}
+
+	public static void write(Manifest manifest, OutputStream out, boolean nice) throws IOException {
 		Attributes mainAttributes = manifest.getMainAttributes();
 		Stream<Entry<Name, String>> sortedAttributes = sortedAttributes(mainAttributes);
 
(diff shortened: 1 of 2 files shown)


pkriens:
hmm. I'd like to suggest to do it differently. 

In Analyzer::calcManifest we have the data in Parameters. Why not add an append method that takes a nice option? This will create an attribute in the manifest with newlines. Which is as far as I know allowed ... (might have to check this).

ManifestUtil::write then needs to account for newlines in the attribute value. it should then reset the limit and output an EOL. it should _not_ put the newline itself in the manifest output.

The advantage is that:

* having a formatted output in Parameters is useful in other contexts. 
* The change to ManifestUtil is
  * much smaller,
  * is arguably wrong by not handling newlines today
  * does not require semantic knowledge about the headers, knowledge that Analyzer is the owner of

chrisrueger:
> Why not add an append method that takes a nice option? This will create an attribute in the manifest with newlines. Which is as far as I know allowed ... (might have to check this).

Ok, then let's close this PR and discuss it in the next call maybe. Sounds very core so I am a bit hesitant to take this over.

Issue #6340, opened 2024-10-28
BndtoolsJAREditorTempFiles keeps growing: 1 jar file creates two folders, but only one is removed on closing


chrisrueger:
## Problem

I noticed the following behavior recently in 7.1.0.SNAPSHOT where I am opening a lot of jar files to verify stuff. 

1. I open one .jar file
<img width="508" alt="image" src="https://github.com/user-attachments/assets/7e072192-2afd-4ad6-a38c-4df40e8b12a8">

This creates 2 temp folders

2. now I close the Jar File Viewer window.
This deletes only one of the two temp folders.

<img width="547" alt="image" src="https://github.com/user-attachments/assets/d8dc0c7f-cb49-4379-a2f3-9374e31f41aa">

Expected result:
Both temp folders from the 1st screenshot should have been removed.
Also ending Eclipse does not remove anything. 

I searched other related issues, but not sure if one of it is related. Here is the search result:
https://github.com/bndtools/bnd/issues?q=is%3Aissue+bndtoolsjareditortempfiles+is%3Aclosed

chrisrueger:
Reason seems to be that `JARTreePage.update()` is called two times in different threads. 

And because stuff is happening in a backgroud Job 

`JAREditor.background("Reading zip file", monitor -> {`

see https://github.com/bndtools/bnd/blob/master/bndtools.jareditor/src/bndtools/jareditor/internal/JARTreePage.java#L94 

I was able to fix it by avoiding a background job when the jar is loaded:
so  basically removing the outer

`JAREditor.background("Reading zip file", monitor -> {`

and just do the stuff inside in the main / display thread. 
Now only one folder is created.

Performance wise I do not notice a difference. The jar opens instantly just as before. Maybe even faster, because only one temp folder is created , thus less work. 

Not sure how this is "good" or "bad" in Eclipse. I personally don't know what would be bad to open the jar in the Display thread. I mean, it is like "I want to open that file, and of course I am willing to wait until it is opened".

I will do some more testing and maybe push the PR later.

@pkriens any opinion?

pkriens:
The issue is that some jar files > 50Mb and that will be noticeable. (Or at least it used to be.)

PR #6342, opened 2024-10-28
JarViewer: Fix orphaned temp files (remove background jobs)

chrisrueger:
Closes https://github.com/bndtools/bnd/issues/6340

this fixes all the race conditions where update methods was called multiple times by different threads and created temp folders multiple times. Now only one temp folder per opened jar is created. Also did not notice a performance problem when opening the jar.

Not sure if not using a Background Job is a good or bad to do in Eclipse. But I had the feeling they do more harm than good in this case.


chrisrueger:
As discussed with @pkriens today, removing the Background Job should not be done. He will look into a different approach. Closing this.

PR #6343, opened 2024-10-29
Improve repo tagging

chrisrueger:
As discussed with @pkriens today:

- The Repo Browser does not show 'resolve' tag anymore, for repos which do not have defined a tag

image

- instead only if developer has manually entered a tag in a repo it is shown

image

image

Also fixed a bug I discovered regarding Repository implementations which currently do not implement the `Tagged` interface. Those repositories should still be considered for resolution by default (as if they have the `resolve` tag). This is e.g. the case for any custom Respository plugin in the wild which `implements Repository` but not Tagged (because Tagged interface is just being added in 7.1.0)

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/service/Registry.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/service/Registry.java
@@ -22,7 +22,8 @@ public interface Registry {
 	 * @param c
 	 * @param tags
 	 * @return all plugins matching the given class and any of the given tags.
-	 *         If no tags are given, all plugins are returned without filtering.
+	 *         If a plugin does not support tags or no tags are given, all
+	 *         plugins are returned without filtering.
 	 */
 	default <T> List<T> getPlugins(Class<T> c, String... tags) {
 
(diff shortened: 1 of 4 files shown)


biz.aQute.bndlib/src/aQute/bnd/service/Registry.java
```diff
@@ -31,8 +32,8 @@ default <T> List<T> getPlugins(Class<T> c, String... tags) {
 		}
 
 		return getPlugins(c).stream()
-			.filter(repo -> repo instanceof Tagged taggedRepo && taggedRepo.getTags()
-				.includesAny(tags))
+			.filter(repo -> !(repo instanceof Tagged) || (repo instanceof Tagged taggedRepo && taggedRepo.getTags()
+				.includesAny(tags)))
```
chrisrueger, line 36:
```
// TODO change it to Glob
return getPlugins(c).stream()
			.filter(repo -> (repo instanceof Tagged taggedRepo) 
        ? taggedRepo.getTags()
				.includesAny(tags) : true )
			.collect(Collectors.toList());
	}  
```

chrisrueger:
@pkriens condition improved and Glob support added.

biz.aQute.bndlib/src/aQute/bnd/service/Registry.java
```diff
@@ -22,7 +22,8 @@ public interface Registry {
 	 * @param c
 	 * @param tags
 	 * @return all plugins matching the given class and any of the given tags.
-	 *         If no tags are given, all plugins are returned without filtering.
+	 *         If a plugin does not support tags or no tags are given, all
```
pkriens, line 25: I would say ...
```
All plugins that have a tag that matches at least one of the given tags or have no tags.

pkriens:
LGTM! lets merge this baby!

PR #6355, opened 2024-11-06
Missing package in resolve for the #5829 PR

pkriens:
---
 Signed-off-by: Peter Kriens <Peter.Kriens@aQute.biz>

PR #6356, opened 2024-11-07
Had a dependency on the PDE feature.

pkriens:
---


chrisrueger:
@pkriens @kriegfrj I can confirm this PR fixed the problem reported in https://github.com/bndtools/bnd/pull/5829#issuecomment-2457895318 and https://github.com/bndtools/bnd/pull/6355

kriegfrj:
Thank you @chrisrueger for being brave enough to try it and @pkriens for fixing the "bit rot" :smile:

pkriens:
@kriegfrj well, I did it in a bad way :-) I just copied the PDE code. 

Would be nice if we had a test that validated our p2 distribution repo. I tested this with a bnd p2 repository, should not be that hard.

Issue #6346, opened 2024-10-31
Plugins should not access a task's project at execution time


abstratt:
Plugins should not access a task's project at execution time. This will be deprecated in 8.12, and will be forbidden in a future release. One case:

https://github.com/bndtools/bnd/blob/4550adf924c46d7c5860bbedadc22ea8f946dc11/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/AbstractBndrun.java#L345

```
Invocation of Task.project at execution time has been deprecated. This will fail with an error in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.12-20241030030000+0000/userguide/upgrading_version_7.html#task_project
        ...
	at org.gradle.api.internal.AbstractTask.getProject(AbstractTask.java:238)
	at org.gradle.api.DefaultTask.getProject(DefaultTask.java:59)
	at aQute.bnd.gradle.AbstractBndrun.lambda$bndrunAction$3(AbstractBndrun.java:345)
	at java.base/java.util.concurrent.ConcurrentHashMap.computeIfPresent(ConcurrentHashMap.java:1828)
	at java.base/java.util.Properties.computeIfPresent(Properties.java:1459)
	at aQute.bnd.gradle.AbstractBndrun.bndrunAction(AbstractBndrun.java:345)
        ...
```

Context: https://github.com/gradle/gradle/issues/30860

pkriens:
@bjhargrave any idea how much work this will be?

bjhargrave:
This has long been a known concern and there is a workaround. See https://github.com/bndtools/bnd/tree/master/gradle-plugins#gradle-configuration-cache-support

Certain users of Bnd need access to the open-ended properties in the Gradle Project object.

One can do:

```groovy
tasks.named("jar") {
  bundle {
    properties.put("project.group", provider({project.group}))
  }
}
```

for each Gradle Project property they need to use. But sometimes the list of needed properties is not obvious due to Bnd's support for including other bnd files and macros, etc. So this could be a trail-and-error process which can break when a user modifies a bnd file and not the gradle build file.

A possible solution could be to change:

https://github.com/bndtools/bnd/blob/3f2c1d34ddb2e2b1dc778f55f4abc7fade812daa/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/AbstractBndrun.java#L292

to something like:

```java
[...]

PR #6360, opened 2024-11-08
Added a method to get the Macro references from a Processor

pkriens:
---


pkriens:
Intended for #6346

PR #6362, opened 2024-11-09
Extract code to automatically derive data from maven-project

laeubi:
Currently if one loads the configuration with
`aQute.bnd.maven.lib.configuration.BndConfiguration` then only the
configuration itself is loaded, but actually some implicit configuration later apply (e.g. from the maven project configuration)
that is currently only performed by the `AbstractBndMavenPlugin`.

This extracts this logic from the `AbstractBndMavenPlugin` into the
`BndConfiguration` class for the following benefits:

- making the `AbstractBndMavenPlugin` maintain less code in the execute
method
- allow other users of `BndConfiguration` to easily inherit common
defaults
- separation of concerns, easier to debug and understand

### DIFF
--- a/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/configuration/BndConfiguration.java
+++ b/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/configuration/BndConfiguration.java
@@ -7,18 +7,26 @@ import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
 
+import org.apache.maven.model.Developer;
+import org.apache.maven.model.License;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.model.PluginExecution;
 import org.apache.maven.model.PluginManagement;
 import org.apache.maven.plugin.MojoExecution;
 import org.apache.maven.project.MavenProject;
+import org.codehaus.plexus.util.StringUtils;
 import org.codehaus.plexus.util.xml.Xpp3Dom;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import aQute.bnd.build.Project;
+import aQute.bnd.header.OSGiHeader;
+import aQute.bnd.osgi.Constants;
 import aQute.bnd.osgi.Processor;
+import aQute.bnd.version.MavenVersion;
+import aQute.bnd.version.Version;
 import aQute.lib.io.IO;
+import aQute.lib.strings.Strings;
 import aQute.lib.utf8properties.UTF8Properties;
 
 /**
@@ -26,6 +34,8 @@ import aQute.lib.utf8properties.UTF8Properties;
  * various Mojos.
  */
 public class BndConfiguration {
+	static final String			TSTAMP	= "${tstamp}";
+	static final String			SNAPSHOT	= "SNAPSHOT";
 	private final static Logger	logger	= LoggerFactory.getLogger(BndConfiguration.class);
 
 	private final MavenProject	project;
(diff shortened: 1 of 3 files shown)


laeubi:
The build failure looks like my change in the shared lib is not picked up by the maven build.

laeubi:
@bjhargrave any hint / idea what could be wrong?

bjhargrave:
> @bjhargrave any hint / idea what could be wrong?

I don't know. 

But this PR seems to be poorly designed from an object oriented perspective. There are 2 things unrelated: reading the configuration from the pom/bnd file and setting some defaults, for bundle building, if they have not been set. I don't see why BndConfiguration should have an instance method for setting defaults since default settings never needed the configuration. That seems a poor place to put default settings for the bnd maven plugin and conflates two things.

laeubi:
> There are 2 things unrelated: reading the configuration from the pom/bnd file and setting some defaults, for bundle building, if they have not been set

I think my naming was not really good here, I have now adjusted it to better reflect the meaning. It is actually not any defaults but apply some implicit configuration (and just converted to bnd instructions). Still of course it could also be extracted in a separate class, the main point is that it should not be mangled with the Mojo so it can be used in other context as well.

Sadly this would not resolve the issue that for some reason the invoker tests seem not to "see" the bnd-build jar file with the shared code but some outdated variant :-\

pkriens:
The problem seems to be that the bnd-baseline-maven-plugin has no dependency to biz.aQute.bnd.maven in its pom.xml. Therefore, the latest biz.aQute.bnd.maven it is not copied to the testing repo. With the wonderful maven logic, it will then likely get it from some random repository, which happens to be the previous version ...

Other bnd plugins might have the same problem although the bnd-maven-plugin seems ok.

However, I'd like to see @bjhargrave 's concerns addressed.

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/configuration/BndConfiguration.java
```diff
@@ -47,6 +57,160 @@ public File loadProperties(Processor processor) throws Exception {
 		return loadProjectProperties(processor, project, project, configuration);
 	}
 
+	public void loadImplicitProperties(Processor processor, String outputTimestamp) throws Exception {
```
bjhargrave, line 60: Perhaps a better method name like `setPropertiesDefaults`. The method is not loading anything. It is setting default values for some bnd properties based upon the values of bnd properties and maven project properties.

It is unfortunate that you pass outputTimestamp here. It would be nicer if this value could be found through the plugin configuration since we may need to pass more such values in the future and end up with having to overload this method.

bjhargrave (changes requested):

maven-plugins/bnd-baseline-maven-plugin/pom.xml
```diff
@@ -51,6 +51,10 @@
 			<groupId>${project.groupId}</groupId>
 			<artifactId>biz.aQute.bndlib</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>${project.groupId}</groupId>
+			<artifactId>biz.aQute.bnd.maven</artifactId>
+		</dependency>
```
bjhargrave, line 57: Looks like this same change is needed for bnd-reporter-maven-plugin.

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/configuration/BndConfiguration.java
laeubi: > Perhaps a better method name like setPropertiesDefaults. The method is not loading anything. It is setting default values for some bnd properties based upon the values of bnd properties and maven project properties.

I was just inspired by the existing method name, that in also not necessarily "loads" something but just reads it from the project config and set the values... maybe `inheritPropertiesDefaults` would even be better.

> It is unfortunate that you pass outputTimestamp here. It would be nicer if this value could be found through the plugin configuration since we may need to pass more such values in the future and end up with having to overload this method.

Should be doable.

laeubi:
I have extracted now the adjustments to the poms in a separate PR here:
- https://github.com/bndtools/bnd/pull/6385

once it is merged I will rebase/adjust this change so we only have the changes related to the topic itself.

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/configuration/BndConfiguration.java
laeubi: I have now applied the suggestion to read the property from the configuration.

biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/configuration/BndConfiguration.java
```diff
@@ -111,6 +280,13 @@ private File loadProjectProperties(Processor processor, MavenProject bndProject,
 		return pomFile;
 	}
 
+	private Optional<Xpp3Dom> getConfiguration(MavenProject mavenProject) {
+		if (mavenProject == null) {
+			return Optional.empty();
+		}
+		return getConfiguration(mavenProject.getBuildPlugins()).or(() -> getConfiguration(mavenProject.getParent()));
```
bjhargrave, line 287: I think this here needs to also look in the parentage for properties. The failure stems from not finding the output time stamp in the parent pom of the project. So it is configuration(parentProject)+configuration(thisProject). Also need to allow the descendant projects to override the ancestor projects.

laeubi: Yes I think you are right here, I was hoping this is already resolved by maven but it seems not be the case...

laeubi: The problem was that `project.build.outputTimestamp` is the default value for the mojo, but this is only evaluated when injecting the maven config so it is not part of the config of the mojo itself.

bjhargrave (approved):
LGTM

chrisrueger:
Can I merge this?

@bjhargrave @pkriens

bjhargrave:
> Can I merge this?

I have no objection.

PR #6363, opened 2024-11-10
Add testcase for conditional packages with bnd-process

laeubi:
This adds a testacse where conditional package is used with  bnd-process

### DIFF
--- a/maven-plugins/bnd-maven-plugin/src/it/bnd-process-conditional/README.txt
+++ b/maven-plugins/bnd-maven-plugin/src/it/bnd-process-conditional/README.txt
@@ -1,0 +1,5 @@
+This is an example bundle that used bnd-process together with conditional package.
+
+Points of interest:
+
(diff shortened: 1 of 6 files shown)


pkriens:
I am not sure I understand the purpose? There is no test case with this as far as I can see?

laeubi:
> I am not sure I understand the purpose? There is no test case with this as far as I can see?

Yes there is no test-case yet for this particular use-case and that's why I added one to be sure

1. it currently works
2. it will work in the future if something is changed

laeubi:
This is an [invoker test](https://maven.apache.org/plugins/maven-invoker-plugin/) see the others of the same kind here:

https://github.com/bndtools/bnd/tree/master/maven-plugins/bnd-maven-plugin/src/it

if you run the build you should see the new test somewhere between the lines mentioned when the maven plugins are build.

PR #6383, opened 2024-12-02
Remove fixed bnd workspace problem markers from Problem view again

chrisrueger:
This fixes an annoyance, that Bndtools Workspace Problem Markers in `/cnf/build.bnd` where not removed from Eclipse's Problem-View, although you have fixed a problem. The reason was that model.getErrors() and model.getWarnings() used by org.bndtools.builder.MarkerSupport.setMarkers(Processor, String) where never cleared, so the errors and warnings where always re-added.

@pkriens we just talked about this today

## Example / Testing

1. Make a syntax error in `/cnf/build.bnd `
2. check Eclipse's Problem view that it creates an error marker of type `Bndtools Workspace Problem Marker`

image

3. Now fix the syntax error.

Expected behavior:
- The marker should also be removed from Eclipse's Problem view

Actual behavior before fix:
- the marker stayed there (in fact it was deleted but re-added...but very quickly)


pkriens:
LGTM

PR #6385, opened 2024-12-03
Add explicit dependencies to biz.aQute.bnd.maven

laeubi:
It has shown that the invoker-test in this setup could pull in an outdated artifact if it is not declared as a *direct* dependency.

This adds an explicit dependency to biz.aQute.bnd.maven so we always use the current build version.

This was extracted from
- https://github.com/bndtools/bnd/pull/6362

as it actually do not belong to that change and is a configuration change only that can be merged independently.


bjhargrave (approved):

chrisrueger:
@pkriens since this was already agreed to in https://github.com/bndtools/bnd/pull/6362 and also approved by BJ I took the liberty to merge.

PR #6391, opened 2024-12-06
cnf change: prevent full build although 'Build automatically' is disabled

chrisrueger:
This improves the DX in the following scenario:

When you edit any file inside the `cnf` folder like `/cnf/build.bnd` and you Save the file before this fix, a FULL build was always triggered (by `CnfWatcher.java`), even if `Project / Build Automatically` was disabled. (and such a build can take very long and can become anoying in cases where you need tweak build.bnd) 

Now a build is only triggered if `Build Automatically`  is **enabled**. This at least gives the developer a knob to disable the build-on-change behavior in cases where you work on the file and want to save it a few times along the way (without waiting for a build everytime)

image

A warning is written the Error Log to make it visible somewhere:

image

**Note:** The "Reload Workspace" Button in the Bndtools Explorer is also affected by this change and does not  trigger a FULL build if _Build Automatically_ setting is disabled. This behavior is still handy as a "Reload Workspace without build" could still do some useful work like validation e.g. to add Problem markers.

The tooltip of the "Reload workspace" button now mentions the build behavior:

image

### DIFF
--- a/bndtools.core/src/bndtools/central/sync/WorkspaceChange.java
+++ b/bndtools.core/src/bndtools/central/sync/WorkspaceChange.java
@@ -49,6 +49,12 @@ public class WorkspaceChange {
 			if (cancel || monitor.isCanceled())
 				return Status.CANCEL_STATUS;
 
+			if (!ResourcesPlugin.getWorkspace()
+				.isAutoBuilding()) {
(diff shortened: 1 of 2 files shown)


pkriens:
LGTM

PR #6401, opened 2024-12-11
Fixes a number of errors in the bndrun handling

pkriens:
Some of the changes we made to handle the UI of inherited properties were not correctly
handling all cases. These shortcomings were hidden because we reread the properties
from the file system.

The issue is that the BndEditModel is about the properties in the document while
the resolver handles properties after processing. So we needed to setup a processor
that properly reflected:   

- The correct class hierarchy (there are macros in the Processor subclasses)
- Handled deleted properties
- And provided updated values for properties 

For this I hadded a sub-Processor to the bndrun. This subprocessor contained all the
document values (e.g. -include is in there) and intercepted access to the deleted
keys.

However, we accidentially reread the properties of the bndrun sub processor from file. This
worked fine in the workspace but the LauncUtils lets the Maven handler muck around to
add a repository. Unfortuantely, this also set the base directory of the bndrun to a
temporary directory for unknown reasons. So when the properties were read and an
include was found it could not find it.

Then there was an unrelated problem that I think was also masked. It turns out the UI
was updating the model with requirements even if requirements were not changed or
added. This created a null key in the bndrun that never existed. Since this example
used the `-runrequirements` key in an include file, it hid the original value. And
weridly it caused an NPE later.

So uncomfortable many changes ... I guess we need to test this well and probably release
a 7.2 soon.

Fixes #6393
Fixes #6394
Fixes #6394  

---
 Signed-off-by: github-actions <github-actions@bndtools.org>

PR #6406, opened 2024-12-14
first draft of Effective tab for BndEditor

chrisrueger:
adds a first draft of an readonly "Effective Source view" of a Bnd Editor based on comment https://github.com/bndtools/bnd/pull/6401#issuecomment-2540856340

> It shows you all the effective properties with inheritance and includes.

- supports find text via CMD+F
- add Syntax highlighting (same as Source view)

image

This is a nice additional debugging tool, because it makes things visible to the human eye, which were previously hidden. 

Comment about code: I reused some code of the JarEditor (JarPrintPage.java) and added search-dialog and syntax highlighting. Sorry if this is not state-of-the-art Eclipse JFace code.

### DIFF
--- a/aQute.libg/src/aQute/lib/utf8properties/PropertiesParser.java
+++ b/aQute.libg/src/aQute/lib/utf8properties/PropertiesParser.java
@@ -2,7 +2,6 @@ package aQute.lib.utf8properties;
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.Properties;
 
 import aQute.lib.hex.Hex;
 import aQute.lib.strings.Strings;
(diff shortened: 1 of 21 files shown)


pkriens:
I might take a stab in the next 2 weeks when I am not allowed to sit in my office :-)

chrisrueger:
> I might take a stab in the next 2 weeks when I am not allowed to sit in my office :-)

Thanks. I have pushed some improvements. Hope that might be a good place to takeover. 

- there is a button at the top to toggle between table and source view (button position can be improved, I was just too lazy to mess around with positioning at the moment)
- double clicking on a row opens the editor of the `.bnd / bnd.run` file if it is not in the current file)
- path is only filled if it is from a property file. So if path is empty, the property is coming from somewhere else

image

chrisrueger:
@pkriens thanks for your additions. Did a few simple tests. Looks good. 
Feel free to merge

Issue #6364, opened 2024-11-11
Startlevel in output of bnd-resolver-maven-plugin does not include/incorporate startlevels from runbundles decorator


glimmerveen:
I noticed in a project that any startlevel set using the `-runbundles+` decorator does not show-up in the `-runbundles` outputted by the `bnd-resolver-maven-plugin`. Note that if the bndrun _also_ includes a `-runstartlevel` instruction, then the output of bnd-resolver-maven-plugin will include startlevel attribute, but only with the value(s) following from the `-runstartlevel` instruction, and **not** from the `-runbundles+` decoration.

When a bndrun file that uses a `-runstartlevel` instruction and some specific startlevels assigned using `-runbundles+` decoration is generated into an executable JAR, then the resulting startlevels used within that executable JAR is the expected combination of both.

Is it to be expected that the output of the `bnd-resolver-maven-plugin` with respect to startlevel attributes is aligned with the effective output of the `bnd-exporter-maven-plugin`?

pkriens:
Agree.

Can you provide a PR?

pkriens:
This will be closed if no reply ...

glimmerveen:
> This will be closed if no reply ...

Sorry for not responding to your initial response. I am willing to providing a PR for this, but haven't gotten around to properly dive into this.

pkriens:
ok, then we keep it open ... thanks.

chrisrueger:
Closed via https://github.com/bndtools/bnd/pull/6407

PR #6407, opened 2024-12-14
Apply the -runbundles+ decorator on the computed RunBundles when resolving

glimmerveen:
This PR should address issue #6364 . I applied the same mechanism used within [Project.getBundles](https://github.com/bndtools/bnd/blob/1c97750/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L634-L635) in the RunResolution's updateBundles function, just before the model is updated with the new RunBundles.


biz.aQute.resolve/src/biz/aQute/resolve/RunResolution.java
```diff
@@ -208,7 +210,16 @@ public boolean updateBundles(BndEditModel model) {
 			newer = older;
 		}
 
-		model.setRunBundles(newer);
+		// Apply the -runbundles decorator on the computed RunBundles
```
pkriens, line 213: You have to change the check if the runbundles are identical before on line 197. The decoration might change the runbundles.

maven-plugins/bnd-resolver-maven-plugin/src/it/resolve-with-runstartlevel-and-runbundles-decorator/invoker.properties
```diff
@@ -0,0 +1,7 @@
+invoker.goals=--no-transfer-progress package
```
pkriens, line 1: Hmm. I am not too happy this testing takes place in the maven sub project. The functional change is in bnd so it should be tested in the normal bnd testing. The biz.aQute.resolve.RunResolutionTest in biz.aQute.resolve contains examples.

I know it is more comfortable to stay in your comfort zone but bnd is a shared project. If you change a function in bnd the test must take place as close as possible to the change.

pkriens (changes requested):
Thanks for the fix. 

As said in a comment, I'd really appreciate if you move the tests to the RunResolutionTest to be closer to the original change.

maven-plugins/bnd-resolver-maven-plugin/src/it/resolve-with-runstartlevel-and-runbundles-decorator/invoker.properties
glimmerveen: That makes sense, I will look into adding test case(s) in RunResolutionTest. Do you want to retain these maven sub project test as well for this particular case, or just the tests within biz.aQute.resolve?

biz.aQute.resolve/src/biz/aQute/resolve/RunResolution.java
glimmerveen: I have moved the application of decoration above the older/newer check.

glimmerveen:
I have pushed a commit that should address the feedback. The only question I have is whether the tests in de maven sub module should be retained.

maven-plugins/bnd-resolver-maven-plugin/src/it/resolve-with-runstartlevel-and-runbundles-decorator/invoker.properties
pkriens: I think it is better to test it once in bnd. I consider ANY error that would be detected in a 'driver' (maven, gradle, eclipse, intellij) an error of the highest severity. We rarely have them fortunately. And in this case, the functionality is clearly in bnd, not the driver.

glimmerveen: Ok, I have removed the integration tests from the maven sub module.

pkriens:
thanks!

PR #6421, opened 2025-01-02
conditonalpackage missed classes added by plugins, like Blueprint.

### LINKED ISSUE #6408: conditionalpackage no longer works with Blueprint since 3.4.0
Hello,

I have tried to upgrade to the later bnd version in my project and I have noticed that `conditionalpackage` no longer works when the package in question referred only in Blueprint XML. I have not tested but I guess `conditionalpackage` no longer works with `-plugin`.

First I narrowed down the regression to version 3.4.0 and then I have `git bisect`'ed it to [commit 7a4bc425c8d3e1c2e387c0201e18cba6a0eea435](https://github.com/bndtools/bnd/commit/7a4bc425c8d3e1c2e387c0201e18cba6a0eea435). 

I attach a [sample](https://github.com/user-attachments/files/18141366/bnd-process-conditional-blueprint.tar.gz) which you can use to reproduce the issue. Actually it is a modified version of [maven bnd-process-conditional it test](https://github.com/bndtools/bnd/tree/master/maven-plugins/bnd-maven-plugin/src/it/bnd-process-conditional) so you can easily incorporate it into source tree.

3.3.0:
```
mvn -q clean package -Dbnd.version=3.3.0 && \
  unzip -l target/test-bnd-process-conditional-blueprint-0.0.1-SNAPSHOT.jar | \
  grep -e com/google/common/cache/CacheBuilder.class -e com/google/common/eventbus/EventBus.class
    18643  2024-12-15 12:43   com/google/common/cache/CacheBuilder.class
     4734  2024-12-15 12:43   com/google/common/eventbus/EventBus.class
```

3.4.0
```
mvn -q clean package -Dbnd.version=3.4.0 && \
  unzip -l target/test-bnd-process-conditional-blueprint-0.0.1-SNAPSHOT.jar | \
  grep -e com/google/common/cache/CacheBuilder.class -e com/google/common/eventbus/EventBus.class
```

7.1.0
```
mvn -q clean package -Dbnd.version=7.1.0 && \
  unzip -l target/test-bnd-process-conditional-blueprint-0.0.1-SNAPSHOT.jar | \
  grep -e com/google/common/cache/CacheBuilder.class -e com/google/common/eventbus/EventBus.class
```

Can't confirm 100% but it is probably the case with current master branch as well.

The testcase tests both simple and factory bean but it probably only any of these is enough as it is not a root issue.

Any chance this fix could be merged to 6.x series (the last supporting Java 8) as well?

pkriens:
This PR fixes #6408. The conditional package has been moved _after_ the plugin processing. This felt dangerous but no test failed. We need to keep a close eye on this however.

---
 Signed-off-by: github-actions <github-actions@bndtools.org>

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/AnalyzerTest.java
+++ b/biz.aQute.bndlib.tests/test/test/AnalyzerTest.java
@@ -48,7 +48,6 @@ class T3 extends T2 {}
 public class AnalyzerTest {
 	static File cwd = new File(System.getProperty("user.dir"));
 
-
 	/**
 	 * Verify that the manifest overrides a version in a package info
 	 */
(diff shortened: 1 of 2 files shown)


chrisrueger:
Added the "breakingchange" label just in case. We don't know if anything breaks yet, but since processing order was changed it is likely behaviour changed.

PR #6441, opened 2025-01-23
improve description of -nouses

### LINKED ISSUE #6439: Better explain the implications of -nouses instruction
Because of issue
- https://github.com/bndtools/bnd/issues/6431

I further analyzed the problem and found it to be a missing use clause:

- https://issues.apache.org/jira/browse/LOGGING-198

i then tried to fix the project (what uses maven-bundle plugin / bnd already!) but without success. It took me some time to realize, that with [this commit](https://github.com/apache/commons-parent/commit/740bcf1e180f15931bdc09a27dd2a7f087844671) **all Apache commons projects** effectively disable use clauses on their export package instructions. 

The commit do not mentioned any rationale and I thought it would be good to make the project aware of the issue by linking the documentation that will obviously war that this is dangerous and maybe list cases where it should or should not be used but:

https://bnd.bndtools.org/instructions/nouses.html

only tells 

> Do not calculate the uses directive on package exports or on capabilities.

I find disabling this extremely dangerous and the doc should at least mention the implications of not using uses directives on packages (and capabilities)

chrisrueger:
Closes #6439

### DIFF
--- a/docs/_instructions/nouses.md
+++ b/docs/_instructions/nouses.md
@@ -5,8 +5,13 @@ title: -nouses  BOOLEAN
 summary: Do not calculate uses directives on package exports or on capabilities. 
 ---
 
-Do not calculate the uses directive on package exports or on capabilities.
+Do not calculate the [uses directive](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#i3127019) on package exports or on capabilities, if set to _true_.
+Default: _false_
 
 For example:
 
     -nouses: true
+
+
+**Warning:** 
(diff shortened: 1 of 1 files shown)


chrisrueger:
@laeubi https://github.com/bndtools/bnd/pull/6441 Feel free to propose changes, you are more expert and versed in this topic. Since this is docu we can push it almost instantly and are not bound to a bndtools release.

laeubi (approved):
Sounds reasonable to me, I would probably add a reference to the spec.

chrisrueger:
> I would probably add a reference to the spec.

what do you mean? i added a link to https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#i3127019

laeubi:
> > I would probably add a reference to the spec.
> 
> what do you mean? i added a link to https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#i3127019

uups... I missed that :-D looks all fine to me.

PR #6449, opened 2025-01-29
fix RunRequirements could not be saved

chrisrueger:
The problem was that before this fix the call to super.commit(onSave); has reset the dirty flag from true back to false, which was the reason why commitToModel(onSave); did not work, because it checks for dirty==true

@stbischof could you test the PR?

### DIFF
--- a/bndtools.core/src/bndtools/editor/common/BndEditorPart.java
+++ b/bndtools.core/src/bndtools/editor/common/BndEditorPart.java
@@ -79,8 +79,8 @@ public abstract class BndEditorPart extends SectionPart implements PropertyChang
 	@Override
 	public final void commit(boolean onSave) {
 		committing.compareAndSet(false, true);
-		super.commit(onSave);
 		commitToModel(onSave);
+		super.commit(onSave);
 		committing.compareAndSet(true, false);
 	}
 


stbischof:
I never tested a snapahot of bnd that was selfbuild. Could you help me how to test

chrisrueger:
@stbischof I took the liberty to merge it to master, since the change was small. Snapshot should be available in a few minutes... e.g. 30m.

Please test this. In case something broke, I can revert it.

chrisrueger:
Found it: https://bndtools.org/development.html : "Importing Into Eclipse"
you only need to checkout the branch of the PR and build it
https://docs.github.com/de/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally

But honestly, I need to try it myself too :)

stbischof:
Currently snapahot is not deployed

chrisrueger:
I believe it is. I just got it.

- select Bndtools in Eclipse Installation Details
- press the update button

image

I think it is 
`  Bndtools	7.2.0.DEV-202501291916-g62c11bb	bndtools.main.feature.feature.group	Bndtools`

stbischof:
Thnk you a lot!!!

PR #6455, opened 2025-02-04
Fix missing '.' in filter name

laeubi:
Now that PDE is shipping the jar editor I noticed since a while that a folder/project .BndtoolsJAREditorTempFiles is shown, even though there is a filter to hide it.

Looking a bit closer it seems to me that while it was renamed three years ago to start with a dot the actual filterExpression was not adjusted (only the description/name)

grafik


chrisrueger:
thanks @laeubi Let's very with latest snapshot which should be available in a few minutes after the build.

PR #6456, opened 2025-02-04
Fix MessageReporter#details return null instead of this

laeubi:
`MessageReporter#details` return `null` instead of `this` leading to NPE if one relies on chaining method calls.


chrisrueger:
Good catch. The other methods all return this , except this.

chrisrueger (approved):

chrisrueger:
Merged. 
thanks @laeubi

PR #6457, opened 2025-02-04
Fix of-by-one error in AbstractBndMavenPlugin

laeubi:
According to aQute.service.reporter.Reporter.SetLocation.line(int) javadoc "Line 0 is the top line" but BuildContext#addMessage says "Use 1 (not 0) for the first line" so when converting from bnd location > build context we need to add one to the line.


chrisrueger:
@laeubi merged, since it is a simple change in logging output. 
Could you verify after the build and may post an example output (before / after), just for further reference?

laeubi:
You wont see this in the log but in m2e for example, before nothing was shown in the file itself (only in the problems view) or at the wrong line, after it is shown on first line or "correct" line.

kriegfrj:
Title was "of-by-one", which ironically means that it was off by one (letter). I'm sorry, silly things amuse me. :smile: But I fixed it so that it won't break future searches.

Thank you to all for your continuing hard work on Bnd. I'm sorry I'm not able to contribute much these days.

PR #6463, opened 2025-02-09
fix source lookup for sub-projects via -sub instruction

chrisrueger:
This is a continuation of https://github.com/bndtools/bnd/pull/6165 which fixed it for normal bundle bnd.bnd files. But the previous PR did not consider the -sub instruction feature of bnd where you can define multiple sub-jar per project. The repos inside -includeresource instructions of such sub-bnd.bnd files were not considered yet.

(see also https://bnd.discourse.group/t/generating-multiple-bundles-per-project/84)

## What does it improve?

Let's say you use the -sub instruction to have multiple sub-bundles resulting multiple .jar files generated for your project.

image

In a subbundle you are doing e.g. 

```
-includeresource: \
	${repo;org.apache.poi:poi;latest}; lib:=true,\

```

to include / wrap a dependency from your `repo`

**Before this fix:**
When you were debugging and stepped into one of the files of `org.apache.poi:poi`, Eclipse did not jump into the source code of that file , but instead showed this:

image

(although you are sure that you have the source code because outside of the Debugger you can view the source... but just not when jumping into the code via debugger).

**After this fix:**

Debugging into the same class now opens the source. Just as described in https://github.com/bndtools/bnd/pull/6165

## How to verify / test it:

A way to trigger (and test) this code in BndrunDirectiveSourceContainer is:

- to open the Bnd Dependencies under **Debug Configuration / Source**. You can then observe the added bundles by expanding the runbundles there

image

Basically `runbundles` now shows repoRef entries coming from `-includeresource` which are not there in 7.1.0

### DIFF
--- a/org.bndtools.launch/src/bndtools/launch/sourcelookup/containers/BndrunDirectiveSourceContainer.java
+++ b/org.bndtools.launch/src/bndtools/launch/sourcelookup/containers/BndrunDirectiveSourceContainer.java
@@ -1,8 +1,10 @@
 package bndtools.launch.sourcelookup.containers;
 
+import java.io.IOException;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Objects;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


chrisrueger:
@pkriens I am referring to your comment https://github.com/bndtools/bnd/pull/6165#issuecomment-2225086785 and merge this to try it out with the Snapshot in our real world scenario. I promise to keep an eye on it and fix / revert any issues I find.

chrisrueger:
Already testing the snapshot. Just debugging in our project and it is great that the debugger finally steps into methods of classes coming from sub-bundles :)

PR #6466, opened 2025-02-14
-generate – Allow class path to come from file system

pkriens:
The code from [source code generation](https://bnd.bndtools.org/instructions/generate.html) required the bundle to be part of the repositories, it did not support 'version=file'. This was added. If this throws an exception, we fall back to the old way.
Basic change is that we now use the Project's getBundles to get the class path while we use Workspace.getBundle. The Workspace was used since this could also be used in bndrun files but a bndrun file is also a Project ... The Project.getBundles probably belongs on Workspace ...

---
 Signed-off-by: github-actions <github-actions@bndtools.org>


pkriens:
I closed this because I solved the problem in another way. Instead of referring to a file with `version=file`, I am just wrapping the bad jar into a nice bundle.

However, this is very useful but I just don't have the time to finish this off. 

A key missing feature is also to capture the output of the command, at least stderr, and show it on the errors.

PR #6470, opened 2025-02-16
sub-bundles: show warning markers in sub .bnd file

chrisrueger:
instead of project bnd.bnd

This fixes the problem that previously, when you had sub-subbundles, then Warning / Error Markers where always shown at the parent bnd.bnd in the project, instead of in the sub-bundle's .bnd file (where the warning actually belongs to).

Now it is showing up correctly in the sub-bundle's .bnd file if it is comming from a sub-bundle, or if it is a normal bundle, it shows at the bnd.bnd as it already was the case.

**Before:**

image

image

- actual: warning is shown for project `bnd.bnd`  (see _Resource_ column), but it actually belongs to the sub-bundle
- when I double click the warning, I go to the project `bnd.bnd` file
- the graphical marker in the bndeditor  is in the project `bnd.bnd`

**After:**

image

image

- Correct: warning is shown for the sub-bundle .bnd file (see _Resource_ column)
- when I double click the warning, I come to the correct sub-bundle .bnd file
- also the graphical marker in the bndeditor  is now in the .bnd of the sub-bundle (instead of the project `bnd.bnd`)

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Processor.java
@@ -232,13 +232,23 @@ public class Processor extends Domain implements Reporter, Registry, Constants, 
 	@SuppressWarnings("resource")
 	@Override
 	public SetLocation warning(String string, Object... args) {
-		return current().reporter.warning(string, args);
+		SetLocation warning = current().reporter.warning(string, args);
+		File propertiesFile = getPropertiesFile();
+		if (propertiesFile != null) {
+			warning.file(propertiesFile.getAbsolutePath());
+		}
+		return warning;
 	}
 
 	@SuppressWarnings("resource")
 	@Override
 	public SetLocation error(String string, Object... args) {
-		return current().reporter.error(string, args);
+		SetLocation error = current().reporter.error(string, args);
+		File propertiesFile = getPropertiesFile();
+		if (propertiesFile != null) {
+			error.file(propertiesFile.getAbsolutePath());
+		}
+		return error;
 	}
 
 	/**


chrisrueger:
@pkriens would appreciate some feedback. Tests are passing and the fix does what I expect. But I am not sure if there is a better place for that code, or if there are any side effects I don't see. 

@laeubi pinging you, since you also recently added something related in https://github.com/bndtools/bnd/pull/6458/files#diff-4467b68f13db09a0b59718cfc70696d470c9a0875dfc0c85b7a3fd5ecf312d2cR91

You see any issues? Or is it maybe even helpful for your maven cases?

Just to have more eyes on it.

laeubi (commented):
I think you can even implement this on the Processor level as `getPropertiesFile()` is defined there. Maybe even at `aQute.bnd.osgi.MessageReporter.warning(String, Object...)` what seem to have access to a processor already and then can be overridden in custom reporter implementations if desired.

pkriens:
I think laeubi is right. The Processor is already associated with the proper file. ProjectBuilder should _only_ have functions that are specific to project building.

chrisrueger:
Thanks @laeubi , @pkriens I will move it to Processor. I wasn't sure about that initially, since everything inherits from  Processor and I was afraid it might be too broad. 
Anyway, commit will come in few minutes.

chrisrueger:
Commit pushed. Now the change is in Processor. Tested locally quickly, result is still as I expect it. Let's see if all tests pass.

chrisrueger:
@pkriens Is it ok this way at Processor? Or better at `aQute.bnd.osgi.MessageReporter.warning(String, Object...)` as suggsted by @laeubi

pkriens:
lets see how it works ...

chrisrueger:
Works great in our project. 
Note to others:  A click on the Workspace refresh button is needed, so that problem view is updated for sub-bundles.

image

PR #6471, opened 2025-02-17
Add a ProjectLauncher#getCommand for easier extension

laeubi:
Currently if one wants to extend a Project launcher it is required to completely reimplement the launch() method even if one only wants to add some argument to the commandline. Also things like returncode handling and reporting of results need to be implemented again.

This now adds a new method `ProjectLauncher#getCommand() `that externalize the process of creating a commandline, this also allows to access the final commandline from the outside (e.g. to show it to the user) or further customize it.


pkriens:
nice work. I nowadays always try to have either _control_ (decides flow) or worker (calculates some value) methods, never mix them.

PR #6475, opened 2025-02-18
Add pagefind search to docs

chrisrueger:
adding pagefind based full text search for bnd manual, just as we did recently for bndtools.org

image


chrisrueger:
@pkriens thanks it works :) New search is live. 

I need to fix the close button (Use ESC Button in the meantime to close the modal or just click a result)

PR #6479, opened 2025-02-19
Docs: Improve docs for -library instruction (add examples)

chrisrueger:
Here is a suggested for improving the -library instruction. I have heard it being mentioned a couple of times. But I think the current documentation is missing some examples to illustrate the usefulness (as in "what can I do with it, and when would I use it"?)

Here is what I did:

1. added a introductory sentence
2. added examples
3. moved one section "Creating libraries" from the bottom higher up. To me that improves the reading flow and now is basically: 
- What is it
- examples
- how to create
- how to use

### DIFF
--- a/docs/_instructions/library.md
+++ b/docs/_instructions/library.md
@@ -5,6 +5,69 @@ title: -library library ( ',' library )*
 summary: Apply a bnd library to the workspace, project, or bndrun file
 ---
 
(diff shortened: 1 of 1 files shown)


chrisrueger:
@pkriens  @juergen-albert since you worked on the page and use that feature in the past. Does that make sense what I added?

pkriens:
Just merge it ... :-)

PR #6489, opened 2025-02-24
Fix javadoc of Project#getBundle

### LINKED ISSUE #6482: Project#getBundle does not return null if not found
Project#getBundle says:

https://github.com/bndtools/bnd/blob/5549ecf903ad185898a2abfa93b22d4e37214c2e/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L1249

but actually never return `null` but an error container:

https://github.com/bndtools/bnd/blob/5549ecf903ad185898a2abfa93b22d4e37214c2e/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L1385-L1386

chrisrueger:
Closes #6482

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
@@ -1246,7 +1246,7 @@ public class Project extends Processor {
 	 * @param bsn The bundle symbolic name
 	 * @param range The version range
 	 * @param strategy set to LOWEST or HIGHEST
-	 * @return the file object that points to the bundle or null if not found
+	 * @return the Container that points to the bundle what might be {code}Container.TYPE.ERROR{code} in case it can not be found.
 	 * @throws Exception when something goes wrong
 	 */
 


chrisrueger:
@laeubi ok like that?

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -1246,7 +1246,7 @@ public void release(String name, boolean test) throws Exception {
 	 * @param bsn The bundle symbolic name
 	 * @param range The version range
 	 * @param strategy set to LOWEST or HIGHEST
-	 * @return the file object that points to the bundle or null if not found
+	 * @return the Container that points to the bundle or an error Container (of type Container.TYPE.ERROR) if not found
```
laeubi, line 1249:
```suggestion
	 * @return the Container that points to the bundle what might be {code}Container.TYPE.ERROR{code} in case it can not be found.
```

laeubi (commented):
Thanks for working on this!

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
chrisrueger: thanks

PR #6499, opened 2025-03-07
BndEditor: fix incorrect Effective Source View

### LINKED ISSUE #6498: Effective Source view not showing same value as table view
I noticed that the Effective Table View and the Effective Source View in the BndEditor show different values.
I think this was an oversight after some iterations and introduction of the Provenance. 

Image

The table shows the value `root` which is defined in another .bnd file (Provenance).
But the source view shows the actual value inside the current .bnd file, not respecting the Provenance.

Fix is underway, to basically let the Effective Source View use the same code as the Effective Table View.

chrisrueger:
Closes #6498

Now Effective Table View (which was correct) and Effective Source View (which was wrong) now display the same value, respecting the provenance.


chrisrueger:
@pkriens FYI, I merged, since I think low risk.

PR #6505, opened 2025-03-13
add dev tips for unit tests

chrisrueger:
Adds some recommendations for writing Unit Tests which came up in recent reviews of PRs. 
This might help giving new contributors a link to read up about it.


chrisrueger:
@kriegfrj @bjhargrave  I tried to add the things which you recommended in recent code reviews, so that we have this somewhere written down.

kriegfrj:
> @kriegfrj @bjhargrave I tried to add the things which you recommended in recent code reviews, so that we have this somewhere written down.

This is excellent, thank you. I would also add a comment about asserting directly on the whole collection rather than on individual elements (where possible). We can cut and paste the note I made in the review.

chrisrueger:
> > @kriegfrj @bjhargrave I tried to add the things which you recommended in recent code reviews, so that we have this somewhere written down.
> 
> This is excellent, thank you. I would also add a comment about asserting directly on the whole collection rather than on individual elements (where possible). We can cut and paste the note I made in the review.

Good point. 
added here https://github.com/bndtools/bnd/pull/6505/files#diff-b7b09662c8dd8a47ff2865b2a2a99b5002ff87a27c07ede924f0710cfdfe2908R68

bjhargrave:
There are limits on the applicability of soft assertions. They work well when each assertion does not depend on prior assertions to be successful. Such as making a variety of checks on the contents of a collection. But when later assertions will not work when a previous assertion fails, then the previous assertion should not be soft. So sometimes a test method will need to use a mixture of "hard" assertions and then soft assertions.

kriegfrj:
> There are limits on the applicability of soft assertions. They work well when each assertion does not depend on prior assertions to be successful. Such as making a variety of checks on the contents of a collection. But when later assertions will not work when a previous assertion fails, then the previous assertion should not be soft. So sometimes a test method will need to use a mixture of "hard" assertions and then soft assertions.

The problem with mixing hard assertions is that, if it fails, none of the soft assertion results are reported.

Another way around this that i discovered recently is to use `softly.wasSuccess()`, which returns false if the previous soft assertion failed. You can use this to conditionally execute the dependent assertions.

chrisrueger:
@bjhargrave @kriegfrj Thanks for your additions on the limits.
I tried adding a section "Limits of soft assertions": https://github.com/bndtools/bnd/pull/6505/files#diff-b7b09662c8dd8a47ff2865b2a2a99b5002ff87a27c07ede924f0710cfdfe2908R106

Feel free to come up with better wording, since your are more expert in this.

Issue #5585, opened 2023-03-09
Package imports for java.* & compatibility


laeubi:
See here:

I believe this to be a bug in bnd's heuristics. Imports for `java.*` packages were added (in release 7 I think) to help support modular Java runtimes and not necessary in previous versions. Bnd uses the Java version as a proxy for framework version if no framework package is imported. Java 9 existed long enough ago that it *might* be used with an old framework, hence the heuristic to be "old".

In this case the heuristic is wrong as Multi-Release bundles were added in Core R7, which means that `java.*` packages *should* be imported by bundles using them.

_Originally posted by @timothyjward in https://github.com/osgi/osgi/issues/573#issuecomment-1462241037_

If a multi-release supplemental manifest is computed can always assume that `R7` is used and therefore `java.*` packages can be imported.

pkriens:
We check for >= Java 11. I fail to understand the reasoning since the code explicitly checks for the framework package version. If the import range is pre OSGi 7, we do not import java. 

Any idea why we would check for the EE version? We can run Java 7 code with OSGi 6  on a Java 17 VM? The code level is  not sufficient to determine if we run on JPMS as far as I can see.

I can remove the EE check completely. The framework version check seems to be full proof as far as I can see.

laeubi:
But what if I do not import any framework package? I think the idea is that with Java 11 everything is modularized so package imports for `java.*` are important but thats only a guess, maybe just enable these imports always for Java9+ is simple enought... if I want to build a bundle for an old framework it can be disabled in the bnd-file...

timothyjward:
> Any idea why we would check for the EE version? We can run Java 7 code with OSGi 6 on a Java 17 VM? The code level is not sufficient to determine if we run on JPMS as far as I can see.

The rationale was that “Java 11 is *too new* for any R6 framework to know what it is, so if I want a framework that can run my bundles it will have to be R7.

timothyjward:
I’m not wild about losing the EE check - what happens if there is no framework impl or API on the classpath? I don’t want to have bundles missing these imports when they enable so much value.

laeubi:
As there is an [option to switch this off](https://bnd.bndtools.org/instructions/noimportjava.html) would it not be an option to simply generate such java imports always?
If one is targeting older Frameworks one can simply use that option to prevent that, R7 is out ~ 5 years now and I think anyone upgrading to never BND version has already upgraded to newer frameworks as well.

timothyjward:
[...]

PR #6508, opened 2025-03-14
Reduce pedantic warnings "Imports that lack version ranges" for some JDK packages

chrisrueger:
This reduces warning noise with pedantic:true regarding the warning "Imports that lack version ranges". This warning is not useful for packages from the JDK like java. , javax.net. , javax.security. (and maybe more), because JDK packges have no version. we did not add "javax." because it would be too broad, as there are packages out there not coming from the JDK (e.g. javax.xml.bind)

For example we had lots of warnings like:

```
Imports that lack version ranges: [javax.net.ssl.SSLSessionContext]
Imports that lack version ranges: [javax.net.ssl.SSLSocket]
```

But since those are JDK packages, they do not have version. 

See https://bnd.discourse.group/t/how-to-handle-imports-that-lack-version-ranges-warning-pedantic-true/504

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/model/EETest.java
+++ b/biz.aQute.bndlib.tests/test/test/model/EETest.java
@@ -17,6 +17,8 @@ import org.junit.jupiter.params.provider.ArgumentsSource;
 
 import aQute.bnd.build.model.EE;
 import aQute.bnd.osgi.Clazz;
+import aQute.bnd.osgi.Descriptors;
+import aQute.bnd.osgi.Descriptors.PackageRef;
 import aQute.bnd.version.MavenVersion;
 import aQute.bnd.version.Version;
 
@@ -141,6 +143,20 @@ public class EETest {
 	@DisplayName("Validate Modules exist for each EE beyond Java 8")
 	public void checkEEHasModules(EE ee) throws Exception {
 		assertThat(ee.getModules()).isNotEmpty();
+	}
+
+	@Test
+	public void testEEIsJDKPackage() {
+		// this package appeared in JDK 1.6+
+		// see .properties in aQute.bnd.build.model
+		PackageRef pck = new Descriptors().getPackageRef("javax.xml.transform.stax");
+		assertThat(EE.J2SE_1_2.getPackages().containsKey(pck.getFQN())).isFalse();
+		assertThat(EE.J2SE_1_5.getPackages()
+			.containsKey(pck.getFQN())).isFalse();
+		assertThat(EE.JavaSE_1_6.getPackages()
+			.containsKey(pck.getFQN())).isTrue();
+		assertThat(EE.JavaSE_17.getPackages()
+			.containsKey(pck.getFQN())).isTrue();
 	}
 
 	static class EEsArgumentsProvider implements ArgumentsProvider {
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -55,6 +55,7 @@ import org.slf4j.LoggerFactory;
 
 import aQute.bnd.annotation.Export;
 import aQute.bnd.apiguardian.api.API;
+import aQute.bnd.build.model.EE;
 import aQute.bnd.classindex.ClassIndexerAnalyzer;
 import aQute.bnd.exceptions.ConsumerWithException;
 import aQute.bnd.exceptions.Exceptions;
@@ -2011,6 +2012,7 @@ public class Analyzer extends Processor {
 	void augmentImports(Packages imports, Packages exports) throws Exception {
 		List<PackageRef> noimports = Create.list();
 		Set<PackageRef> provided = findProvidedPackages();
+		EE ee = getEEFromReqsOrHighest();
 
 		for (PackageRef packageRef : imports.keySet()) {
 			String packageName = packageRef.getFQN();
@@ -2114,8 +2116,10 @@ public class Analyzer extends Processor {
 				removeAttributes(importAttributes);
 
 				String result = importAttributes.get(Constants.VERSION_ATTRIBUTE);
-				if (result == null || !Verifier.isVersionRange(result))
+				if ((result == null || !Verifier.isVersionRange(result))
+					&& complainAboutMissingVersionRange(packageRef, ee)) {
 					noimports.add(packageRef);
+				}
 			} finally {
 				unsetProperty(CURRENT_PACKAGE);
 				unsetProperty(CURRENT_BUNDLESYMBOLICNAME);
@@ -2126,6 +2130,46 @@ public class Analyzer extends Processor {
 		if (isPedantic() && noimports.size() != 0) {
 			warning("Imports that lack version ranges: %s", noimports);
 		}
+	}
+
+	private EE getEEFromReqsOrHighest() {
+		SortedSet<EE> eesFromRequirement = EE.getEEsFromRequirement(get(Constants.REQUIRE_CAPABILITY));
+		if (!eesFromRequirement.isEmpty()) {
+			// return lowest (minimum required EE)
+			return eesFromRequirement.first();
+		}
+		return EE.parse(getHighestEE().getEE());
+	}
+
+	/**
+	 * If <code>false</code> then this means this package probably does not
+	 * provide a version and we are fine and do not complain about it (e.g. JDK
+	 * packages). <code>true</code> means, we will complain about a missing or
+	 * invalid version-range.
+	 *
+	 * @param pck
+	 * @return <code>true</code> if should complain about missing version
+	 *         <code>false</code> otherwise.
+	 */
+	private boolean complainAboutMissingVersionRange(PackageRef pck, EE ee) {
+
+		if (pck.isJava() || isJDK(pck, ee)) {
+			return false;
+		}
+		return true;
+	}
+
+	/**
+	 * @return <code>true</code> if this package is part of the JDK for the
+	 *         given ee.
+	 */
+	private boolean isJDK(PackageRef pck, EE ee) {
+		if (ee == null) {
+			return false;
+		}
+
+		return ee.getPackages()
+			.containsKey(pck.fqn);
 	}
 
 	Pair<Packages, Parameters> divideRegularAndDynamicImports() {
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Descriptors.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Descriptors.java
@@ -120,6 +120,9 @@ public class Descriptors {
 			return binaryName;
 		}
 
+		/**
+		 * @return <code>true</code> if this is a "java." package.
+		 */
 		public boolean isJava() {
 			return java;
 		}
--- a/biz.aQute.bndlib.tests/test/test/BuilderTest.java
+++ b/biz.aQute.bndlib.tests/test/test/BuilderTest.java
@@ -9,6 +9,8 @@ import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
+import static org.osgi.framework.namespace.ExecutionEnvironmentNamespace.CAPABILITY_VERSION_ATTRIBUTE;
+import static org.osgi.framework.namespace.ExecutionEnvironmentNamespace.EXECUTION_ENVIRONMENT_NAMESPACE;
 
 import java.io.File;
 import java.io.FileInputStream;
@@ -51,6 +53,7 @@ import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.ArgumentsProvider;
 import org.junit.jupiter.params.provider.ArgumentsSource;
 
+import aQute.bnd.build.model.EE;
 import aQute.bnd.header.Attrs;
 import aQute.bnd.header.OSGiHeader;
 import aQute.bnd.header.Parameters;
@@ -65,8 +68,11 @@ import aQute.bnd.osgi.Packages;
 import aQute.bnd.osgi.Processor;
 import aQute.bnd.osgi.Resource;
 import aQute.bnd.osgi.Verifier;
+import aQute.bnd.osgi.resource.FilterBuilder;
+import aQute.bnd.osgi.resource.RequirementBuilder;
 import aQute.bnd.test.jupiter.InjectTemporaryDirectory;
 import aQute.bnd.version.Version;
+import aQute.bnd.version.VersionRange;
 import aQute.lib.collections.SortedList;
 import aQute.lib.hex.Hex;
 import aQute.lib.io.FileTree;
@@ -75,8 +81,8 @@ import aQute.lib.strings.Strings;
 import aQute.lib.zip.ZipUtil;
 import aQute.service.reporter.Report.Location;
 
-@ExtendWith(SoftAssertionsExtension.class)
 @SuppressWarnings("resource")
+@ExtendWith(SoftAssertionsExtension.class)
 public class BuilderTest {
 
 	@InjectSoftAssertions
@@ -727,7 +733,6 @@ public class BuilderTest {
 			b.setProperty("Require-Bundle", "osgi.core");
 			b.setProperty("-noimportjava", "true");
 			b.build();
-			assertTrue(b.check("Imports that lack version ranges: \\[javax.swing\\]"));
 			assertThat(b.getImports()
 				.keySet()).containsExactlyInAnyOrder(b.getPackageRef("javax.swing"));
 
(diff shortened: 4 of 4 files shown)


kriegfrj:
I think we might be able to do something better here.

> This reduces warning noise with pedantic:true regarding the warning "Imports that lack version ranges". This warning is not useful for packages from the JDK like java. , javax.net. , javax.security. (and maybe more), because JDK packges have no version. we did not add "javax." because it would be too broad, as there are packages out there not coming from the JDK (e.g. javax.xml.bind)
> 

I had a high-level idea for a different approach to this that I wanted to float: I'm not an expert on contracts (see https://docs.osgi.org/reference/portable-java-contracts.html), but to me this seems identical to problem to what contracts are designed to solve.

A contract is a set of packages where the version number applies to the entire set (ie, the "contract") rather than to the individual packages. My understanding is that Bnd doesn't (or at least shouldn't) complain about missing versions on packages that are known to be part of a contract, because the version is effectively set by the contract version that those packages are a part of. Instead, it creates a `Req-Cap` for the contract with the proper contract version.

Is there perhaps some way we can leverage this functionality to solve the same problem here, by treating the set of packages provided by the JDK as a contract? Bnd already does the hard work to calculate the minimum osgi.ee level required for a bundle and insert this as a `Req-Cap` header. This requirement could be internally mapped to a set of available packages and treated as a contract. Then (assuming that we internally define in bndlib somewhere all of the correct contract definitions) this will automatically be handled by the core contract handling code - which would include not warning about missing version ranges on imports.

I can foresee situations where there are multiple solutions to a missing version range, where packages have moved out of the JDK and "into the wild" (eg, for `javax.xml.bind`). I think that this could be handled by clearly separating the code that 1. calculates the minimum EE requirement, and 2. calculates the missing imports based on contract, which always uses the highest allowed contract version based on the bundle's EE req-cap. That way you can resolve conflicts by manually specifying an upper bound on your bundle's EE requirement if necessary. Eg, if I want to use an unversioned `javax.xml.bind` import, I can specify the maximum supported version of my bundle is EE JavaSE 1.8. Bnd will then use the 1.8 contract definition to see that `javax.xml.bind` is part of the Java 8 contract and so it will not generate the warning. I note that this problem (ie, packages moving in-and-out of the contract) is not specific to the JDK but also applies to contracts in general.

The advantage of this approach is that it provides a natural evolution path as packages move in-and-out of the JDK across different JDK versions.

Thoughts? Or is this overkill?

chrisrueger:
> I had a high-level idea for a different approach to this that I wanted to float: I'm not an expert on contracts (see https://docs.osgi.org/reference/portable-java-contracts.html), but to me this seems identical to problem to what contracts are designed to solve.

I agree, this indeed sounds like it could help.

> My understanding is that Bnd doesn't (or at least shouldn't) complain about missing versions on packages that are known to be part of a contract

Yes I think bnd does already do that. A little bit further down in that code snippet the warning is produced which this PR is about. But because of the code snipped below it would be skipped. 

https://github.com/bndtools/bnd/blob/eab701184c3f07ff153668d93aa8ab878a9afd81/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L2063-L2067


> Then (assuming that we internally define in bndlib somewhere all of the correct contract definitions)

Is there already "something" or is that something still to be implemented. 
Do I understand you correctly that we need to maintain mappings for all packages per Java-Version somewhere?

> Or is this overkill?

Since we are talking about a "pedantic warning" it might be overkill 🤣  But in theory I agree that it looks like that contracts could solve this problem. The question if this is overkill or not may be:

- where do the definitions come from defining which packages belongs to which java version. Maybe you could give an example how that could look, so I see if I have the same understanding as you
- could this be useful for something else, other than suppressing a pedantic warning

chrisrueger:
@kriegfrj Your hint with contracts was great. I think I figured out a solution which at least works for us.

Based on 

- https://bnd.bndtools.org/instructions/define-contract.html
- https://bnd.bndtools.org/instructions/contract.html

I have put that into our `build.bnd`

```
-contract: *

-define-contract:\
  osgi.contract;\
    osgi.contract=JavaPackages;version:Version=17;\
    uses:="javax.net,javax.net.ssl,javax.crypto"
```

and the warnings we care about are gone. 

Note: I discovered that the example on https://bnd.bndtools.org/instructions/define-contract.html is incorrect. The version-attribute `version="3.0"` is at the wrong place, leading to an exception during build. It has to be as in my example above, which is what is used in the Testcases e.g. `test.ContractTest.testDefinedContract()`

I will fix that in this PR and rollback by current changes.

kriegfrj:
> > I had a high-level idea for a different approach to this that I wanted to float: I'm not an expert on contracts (see https://docs.osgi.org/reference/portable-java-contracts.html), but to me this seems identical to problem to what contracts are designed to solve.
> 
> I agree, this indeed sounds like it could help.
> 
> > My understanding is that Bnd doesn't (or at least shouldn't) complain about missing versions on packages that are known to be part of a contract
> 
> Yes I think bnd does already do that. A little bit further down in that code snippet the warning is produced which this PR is about. But because of the code snipped below it would be skipped.
> 
> https://github.com/bndtools/bnd/blob/eab701184c3f07ff153668d93aa8ab878a9afd81/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L2063-L2068
> 
> > Then (assuming that we internally define in bndlib somewhere all of the correct contract definitions)
> 
> Is there already "something" or is that something still to be implemented. Do I understand you correctly that we need to maintain mappings for all packages per Java-Version somewhere?
>

There is something, but also something still to be implemented.

Using contracts, the contract's package membership is defined by the `uses` clause of the `Provide-Capability` header in the provider bundle. In this case, the provider "bundle" is the JDK/system bundle, so the equivalent of the `Provide-Capability` header would need to be synthesized by Bnd during analysis.

> > Or is this overkill?
> 
> Since we are talking about a "pedantic warning" it might be overkill 🤣 But in theory I agree that it looks like that contracts could solve this problem. The question if this is overkill or not may be:
> 
> * where do the definitions come from defining which packages belongs to which java version. Maybe you could give an example how that could look, so I see if I have the same understanding as you

The definitions would need to be defined and maintained somewhere. The sensible place to do it is probably as static definitions here somewhere:

https://github.com/bndtools/bnd/blob/eab701184c3f07ff153668d93aa8ab878a9afd81/biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java#L119-L128

> * could this be useful for something else, other than suppressing a pedantic warning

Maybe?

I have had a think about this. Contracts are potentially a different class of things to JDK package lists. So I think implementing it as a contract might be overkill. But maybe it would be ok to do something similar by enhancing this bit:

https://github.com/bndtools/bnd/blob/eab701184c3f07ff153668d93aa8ab878a9afd81/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L2025-L2032

I think this is already doing what you are trying to do - the only problem is that the `isJava()` test seems to be catching "java.*" only. I believe that we would want to treat all JDK packages the same way - I think that this is the root problem here? I also note that `isJava()` is reused a lot in other places of `Analyzer`, so maybe fixing it there will fix other problems?

To implement the kind of functionality that I was proposing, you'd also need `isJava()` to take an EE as a parameter because the answer may depend on which EE you're analyzing against. But `PackageRef` is public API so this ouldn't be a simple change. Perhaps we can add a method `isJDK()` that takes an EE as a parameter and then replace usage of `isJava()` in `Analyzer` with `isJDK()` where appropriate. Or perhaps this `isJDK()` method can simply live as a private method in `Analyzer` itself (effectively your `complainAboutMissingVersionRange()` method, where it has access to `ees`.

chrisrueger:
Thanks @kriegfrj for all the pointers. I need to digest this.

> the only problem is that the `isJava()` test seems to be catching "java.*" only

When I started with my PR `PackageRef.isJava()` was exactly the place where I hooked into first. The problem was that then tests failed, which currently expect a javax.net package, e.g.

https://github.com/bndtools/bnd/blob/041ba15a03cf635e5e377cd14d2b9319c4257dec/biz.aQute.bndlib.tests/test/test/fragments/FragmentTest.java#L32-L34

So I refrained from that I idea, because I was afraid it is too general and could break other stuff. So I ended up where my current change is. 

> The definitions would need to be defined and maintained somewhere.

Could you outline, how you imagine this definition? Similar to `-define-contract`?

I wonder, what if we make a mistake? What would be the consequences if e.g. one package in our definition is incorrectly assigned to a specific java version. What could go wrong, if it is at a central place as `Analyzer` or `PackageRef.isJava()`?

The other question is: How would be start creating those definitions? Would we literally go through each JDK and generate a list of packages of the rt.jar ? (or all modules in JDK 9+)?

chrisrueger:
I just did a small experiment to get a list of all JDK packages of my JDK 17. 
@kriegfrj even experimentally: could this theoretically be a way to obtain those packages? 

```
java -version
openjdk version "17.0.9" 2023-10-17
OpenJDK Runtime Environment Temurin-17.0.9+9 (build 17.0.9+9)
OpenJDK 64-Bit Server VM Temurin-17.0.9+9 (build 17.0.9+9, mixed mode)
```

Little shell script:

```bash
#!/bin/bash

# List all modules
modules=$(java --list-modules | cut -d '@' -f1)

# Loop through each module and describe it
for module in $modules; do
    echo "============================="
    echo "Module: $module"
    echo "============================="
    java --describe-module $module
    echo
done
```

Execute it: 
```sh
./list_jdk_modules.sh | grep export | awk '/^exports / {print $2}' > list_jdk_modules.txt
```

Result with a list of all JDK packages.

[list_jdk_modules.txt](https://github.com/user-attachments/files/19301869/list_jdk_modules.txt)

kriegfrj:
> We could also think about adding this as a general hint to the documentation.
> But for this we might discuss what exactly to write. For example the Version=17 in my example above for me is just a placeholder... it just means "we are using JDK17, because I know this". But I could have placed anything there. Maybe there is a better way to use that.

Correct. I was thinking the "better way" is to use the `Require-Capability` in your bundle that specifies the `osgi.ee` for your bundle. This is normally calculated by `Analyzer` by default based on the maximum class file version of the classes in the bundle, but can also be set explicitly in `bnd.bnd`.

In the context of `augmentImports()`, we can look for an explicit EE requirement, and if none is found (and if `-noee` has not been set) we can use `getHighestEE()` to determine the calculated highest EE that Bnd would use to synthesize the EE requirement. Once we have the EE requirement we can determine the list of packages to check against.

> Thanks @kriegfrj for all the pointers. I need to digest this.
>

Me too!

> When I started with my PR `PackageRef.isJava()` was exactly the place where I hooked into first (experimentally and hardcoded). The problem was, that then tests failed, which currently expect a `javax.net` package, e.g.
> 
> https://github.com/bndtools/bnd/blob/041ba15a03cf635e5e377cd14d2b9319c4257dec/biz.aQute.bndlib.tests/test/test/fragments/FragmentTest.java#L32-L34
>
> So I refrained from that I idea, because I was afraid it is too general and could break other stuff. So I ended up where my current change is.
>

Yes, good point. I remember now that "java.*" packages were treated differently - in earlier versions of OSGi (<R7) you couldn't include `java.*` packages in your `Import-Package` statement, but you *could* include other packages from the JDK that weren't in the `java.*` namespace (such as `javax.net`). Bnd was changed (#4241, #5585) to only generate `java.*` imports if you're using OSGi >= R7 and it appears it uses this `isJava()` method to implement this conditional functionality. So we have to leave it as it is and implement the "isJDK()" or "isEE()" (or equivalent) overload.

> Could you outline, how you imagine this definition? Similar to `-define-contract`?
>
> I wonder, what if we make a mistake? What would be the consequences if e.g. one package in our definition is incorrectly assigned to a specific java version. What could go wrong, if it is at a central place as `Analyzer` or `PackageRef.isJava()`?
>

Re: if we made a mistake - I think that we can rule out using `isJava()` per above. This in itself significantly reduces the potential scope for breaking things.

Also, I think that is unlikely that we would make a mistake with package definitions - the package definitions in any given Java release are very static. But if it were to happen, the worst that should happen (I think) is that a package import fails to have an automatic version range calculated when it should, or has one calculated when it shouldn't.  Both of these can be fixed with some manual intervention.

> The other question is: How would we start creating those definitions? Would we literally go through each JDK and generate a list of packages of the rt.jar ? (or all modules in JDK 9+)?

Well, I started thinking about answers to this question and went looking for the best place to put such code... then I realized that the hard work has already largely been done! (Which is not really a surprise to me... @bjhargrave was all over this stuff. :smile:)

https://github.com/bndtools/bnd/blob/520d236bf9ff808262347b6c2dce55220bd5ffa2/biz.aQute.bndlib/src/aQute/bnd/build/model/EE.java#L222-L227

The package list defintions are in the `.properties` files in this package. It looks like we haven't updated the package definition lists since Java 17, but it shouldn't be too hard to catch up for the missing versions. The good news is that they don't change.

There also appears to be a little bit of functionality duplication between `aQute.bnd.osgi.Clazz.JAVA` and `aQute.bnd.build.model.EE`, and I'm not quite sure what the separation of duties is between these two classes or if there are any cross-dependency issues we're trying to avoid.  Perhaps @pkriens or @bjhargrave can enlighten us on that.

kriegfrj:
> I just did a small experiment to get a list of all JDK packages of my JDK 17. @kriegfrj could this theoretically be a way to obtain those packages?

Well done! I'm sure that something like this could be used to generate the missing `.properties` files (see my previous comment).

chrisrueger:
> Well done! I'm sure that something like this could be used to generate the missing `.properties` files (see my previous comment).

> then I realized that the hard work has already largely been done! (Which is not really a surprise to me... @bjhargrave was all over this stuff. 😄)

It seems, BJ has developed a  tool for the `.properties` files https://github.com/bjhargrave/java-platform-packages/blob/master/src/main/java/io/hargrave/java/packages/CalculateJavaPlatformPackages.java
 i discovered mentioned in this commit https://github.com/bndtools/bnd/commit/be103d1d61592e45c07b45b334451e50e9394c2d

And then there seems to be 

https://github.com/bndtools/bnd/blob/be103d1d61592e45c07b45b334451e50e9394c2d/biz.aQute.bndlib.tests/compilerversions/compile.sh

which might also be related for different JDKs. 


> Re: if we made a mistake - I think that we can rule out using `isJava()` per above. This in itself significantly reduces the potential scope for breaking things.

Right, that should reduce it. 

I will play around with all the infos. Thanks @kriegfrj  :)

chrisrueger:
I have build a little script around your [CalculateJavaPlatformPackages.java](https://github.com/bjhargrave/java-platform-packages/blob/master/src/main/java/io/hargrave/java/packages/CalculateJavaPlatformPackages.java)

See https://github.com/chrisrueger/jdk-packages

Question @bjhargrave : 
Would that make sense for your repo or even to include it directly into bnd ?

For example it allows you to create the .properties files like this:

```
# Download JDK 17 (from https://adoptium.net/)
./download_jdk.sh 17

# create .properties for bnd
./list_jdk_packages_java.sh 17 > "JavaSE_17.properties"
```

Just for simplicity I have created an own GH repo, but I can also contribute it somewhere else. Whatever makes most sense.

chrisrueger:
> Well, I started thinking about answers to this question and went looking for the best place to put such code... then I realized that the hard work has already largely been done! (Which is not really a surprise to me... @bjhargrave was all over this stuff. 😄)
> 
> https://github.com/bndtools/bnd/blob/520d236bf9ff808262347b6c2dce55220bd5ffa2/biz.aQute.bndlib/src/aQute/bnd/build/model/EE.java#L222-L227
> 
> The package list defintions are in the `.properties` files in this package. It looks like we haven't updated the package definition lists since Java 17, but it shouldn't be too hard to catch up for the missing versions. The good news is that they don't change.

@kriegfrj I have tried using EE.getPackages() and adding the `.isJDK(EE)` idea in https://github.com/bndtools/bnd/pull/6508/commits/646f651dbebb1f558462a45bca60d02a4a7c93e0

Is this what you had in mind?

chrisrueger:
> You just need docs on how to make the properties files which you sort of found in a commit message. Maybe a wiki page would be better :-)

I would have expected that too somehow :) I will try to write up something.
What about putting a README.md directly next to the `.properties` files describing how to create them? and link that Readme.md also somewhere in our CONTRIBUTION.md and/or DEV_README.md ? 

The goal of "my CLI" was to basically have a way to run your script with different JDKs - as an easy way to create new properties for new JDKs - and a helper for the next person needing to generate files for new JDKs for bnd.

bjhargrave:
> What about putting a README.md directly next to the `.properties` files describing how to create them

That seems a reasonable idea (just make sure the readme is not slurped into the bnd jar along with the properties files).

chrisrueger:
@kriegfrj I think I have a state now which could be considered for review. I tested with our application and all the warnings `Imports that lack version ranges` referencing JDK packages (we are on JDK17) are gone.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -2011,6 +2012,7 @@ private void fixupOldStyleVersions(Attrs attrs) {
 	void augmentImports(Packages imports, Packages exports) throws Exception {
 		List<PackageRef> noimports = Create.list();
 		Set<PackageRef> provided = findProvidedPackages();
+		EE ee = EE.parse(getHighestEE().getEE());
```
kriegfrj, line 2015: I think that instead of simply using `getHighestEE()`, it should first look in `requirements.containsKey(ExecutionEnvironmentNamespace.EXECUTION_ENVIRONMENT_NAMESPACE)` to see if an EE requirement has been set manually - if yes, then we should check against the minimum EE specified in that requirement; if no, *then* use `getHighestEE()`.

biz.aQute.bndlib/src/aQute/bnd/osgi/Descriptors.java
```diff
@@ -120,10 +121,26 @@ public String getPath() {
...
+		 */
 		public boolean isJava() {
 			return java;
 		}
 
+		/**
+		 * @return <code>true</code> if this package is part of the JDK for the
+		 *         given ee.
+		 */
+		public boolean isJDK(EE ee) {
```
kriegfrj, line 135: I'm having second thoughts about whether or not to make a method here or simply inline the check in `Analyzer`. Adding it here makes it part of public API which adds to the maintenance overhead. Further,  it's also coupling this API to another package which is not otherwise coupled to this package.

It might be safer to inline it as we can always add it to public API later if there is a demand, but it's hard to remove it from public API once it is there.

biz.aQute.bndlib.tests/test/test/model/EETest.java
```diff
@@ -143,6 +145,17 @@ public void checkEEHasModules(EE ee) throws Exception {
 		assertThat(ee.getModules()).isNotEmpty();
 	}
 
+	@Test
+	public void testEEIsJDKPackage() {
```
kriegfrj, line 149: Should also have an equivalent test for this in `Analyzer` to check (eg) `javax.xml.transform.stax` against JavaSE-1.5 (which should complain about missing version ranges) and JavaSE-1.6 (which shouldn't).

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
chrisrueger: Hmm I am having some trouble with this. 
I have the feeling you are refferering to the `requirements` somewhere around here

https://github.com/bndtools/bnd/blob/3b0b28f2468869ee385ae355d8495f8ee9ae2932/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L1138

and here

https://github.com/bndtools/bnd/blob/3b0b28f2468869ee385ae355d8495f8ee9ae2932/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L1159

but as far as I see it this is executed after `augmentImports(imports, exports)`

So I have the feeling I am misunderstanding. 
Any other pointers? @kriegfrj

biz.aQute.bndlib/src/aQute/bnd/osgi/Descriptors.java
chrisrueger: I will move as private in `Analyzer`.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
chrisrueger: @kriegfrj I think in the meantime I came up with something. Let me know if this is what you had in mind.
see https://github.com/bndtools/bnd/pull/6508/commits/2cd573d423e3fff52238f9bcbd06bfc033cc629e

biz.aQute.bndlib/src/aQute/bnd/osgi/Descriptors.java
chrisrueger: done in https://github.com/bndtools/bnd/pull/6508/commits/2cd573d423e3fff52238f9bcbd06bfc033cc629e

biz.aQute.bndlib.tests/test/test/model/EETest.java
chrisrueger: done in https://github.com/bndtools/bnd/pull/6508/commits/2cd573d423e3fff52238f9bcbd06bfc033cc629e

chrisrueger:
@kriegfrj I think I got something in https://github.com/bndtools/bnd/commit/2cd573d423e3fff52238f9bcbd06bfc033cc629e

chrisrueger:
@kriegfrj I squashed all commits now. Could you have a look if the current state is ok now, so we could give this a try in the wild?

kriegfrj:
> @kriegfrj I squashed all commits now. Could you have a look if the current state is ok now, so we could give this a try in the wild?

Let's give it a try - we still have a couple months at least until 7.2.0 is released so that's plenty of time to fix it if it causes issues.

Thank you for all your hard work and patience. :smile:

kriegfrj:
CI workflow failure seems to have been due to a transient network issue. Simply re-running the failed CI workflow succeeded.

PR #6513, opened 2025-03-17
Add feature for ECF remote services

scottslewis:
New bndtools.ecf for ECF remote services support in bndtools

### DIFF
--- a/dist/bnd.bnd
+++ b/dist/bnd.bnd
@@ -38,7 +38,6 @@
             bndtools.jareditor, \
             bndtools.m2e, \
             bndtools.pde, \
-            bndtools.ecf, \
             bndtools.release, \
             org.bndtools.applaunch, \
             org.bndtools.headless.build.manager, \
--- a/dist/bnd.bnd
+++ b/dist/bnd.bnd
@@ -38,6 +38,7 @@
             bndtools.jareditor, \
             bndtools.m2e, \
             bndtools.pde, \
+            bndtools.ecf, \
             bndtools.release, \
             org.bndtools.applaunch, \
             org.bndtools.headless.build.manager, \
--- a/org.bndtools.p2/bnd.bnd
+++ b/org.bndtools.p2/bnd.bnd
@@ -13,6 +13,7 @@ timestamp: ${tstamp}
     bndtools.jareditor, \
     bndtools.m2e, \
     bndtools.pde, \
+    bndtools.ecf, \
     bndtools.release, \
     org.bndtools.headless.build.manager, \
     org.bndtools.headless.build.plugin.ant, \
--- a/org.bndtools.p2/bnd.bnd
+++ b/org.bndtools.p2/bnd.bnd
@@ -13,7 +13,6 @@ timestamp: ${tstamp}
     bndtools.jareditor, \
     bndtools.m2e, \
     bndtools.pde, \
-    bndtools.ecf, \
     bndtools.release, \
     org.bndtools.headless.build.manager, \
     org.bndtools.headless.build.plugin.ant, \
--- a/cnf/ext/repositories.bnd
+++ b/cnf/ext/repositories.bnd
@@ -37,6 +37,12 @@ ossrh:                  https://oss.sonatype.org/content/repositories/snapshots
         index="${.}/baseline.mvn";\
         readOnly=true
 
+-plugin.4.ECFRSSDK: \
+ 	aQute.bnd.repository.p2.provider.P2Repository; \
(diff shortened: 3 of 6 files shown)


scottslewis:
The builds seem to be failing  because of this error.

The new bndtools.ecf feature includes several existing ecf features from the ECF repo (added to cnf/ext/repositories.bnd).  I suspect that for some reason that the feature build is somehow not able to access these repo-provided features to include them, but I don't know how things are added to the target platform for the p2 feature creation task

> Task :biz.aQute.repository:test
D:\a\bnd\bnd\org.bndtools.p2\bnd.bnd:0: error: Missing dependson bndtools.ecf

D:\a\bnd\bnd\org.bndtools.p2\bnd.bnd:0: error: ${repo} macro refers to an artifact bndtools.ecf-latest (HIGHEST) that has an error: bndtools.ecf;version=latest Not found in [bnd-cache [D:/a/bnd/bnd/cnf/cache/7.2.0/bnd-cache   r/w=false], MavenBndRepository [localRepo=D:\a\bnd\bnd\dist\m2, storage=Maven Central, inited=true, redeploy=false], MavenBndRepository [localRepo=D:\a\bnd\bnd\dist\m2, storage=Local, inited=true, redeploy=false], MavenBndRepository [localRepo=D:\a\bnd\bnd\dist\m2, storage=Release, inited=true, redeploy=false], MavenBndRepository [localRepo=D:\a\bnd\bnd\dist\m2, storage=JFrog, inited=true, redeploy=false], Eclipse 4.25 (2022-09) [D:\a\bnd\bnd\cnf\cache\stable\Eclipse-4.25 r/w=false], P2Repository [ECF Remote Services SDK], MavenBndRepository [localRepo=D:\a\bnd\bnd\dist\m2, storage=Baseline, inited=true, redeploy=false]]
> Task :org.bndtools.p2:jar FAILED

dist/bnd.bnd
```diff
@@ -38,6 +38,7 @@
             bndtools.jareditor, \
             bndtools.m2e, \
             bndtools.pde, \
+            bndtools.ecf, \
```
kriegfrj, line 41: I had a look at the code - you haven't actually created a `bndtools.ecf` project in the workspace, have you? Only the feature?

The build is failing because `-dependson` doesn't list Eclipse features - it contains the list of projects in the Bnd workspace that the feature/distribution depends on. There is a `bndtools.pde` and a `bndtools.m2e` project, that is why they are included in `-dependson` - but if the feature you're adding doesn't actually contain a corresponding bundle called `bndtools.ecf`, then you won't need to include those in the `-dependson`. As far as I can see there is no Bnd project in this commit called `bndtools.ecf`.

I suspect if you remove this (and its partner-in-crime below), it will work.

org.bndtools.p2/bnd.bnd
```diff
@@ -13,6 +13,7 @@ timestamp: ${tstamp}
     bndtools.jareditor, \
     bndtools.m2e, \
     bndtools.pde, \
+    bndtools.ecf, \
```
kriegfrj, line 16: Ditto above comment.

kriegfrj (commented):
I think this is your problem:

scottslewis:
@kriegfrj thanks...now all checks builds are succeeding.   I don't know how to verify the shape of the built p2 repo...i.e. if the check builds actually include the new ECF feature in the right category, so I'll ask someone @kriegfrj to check that and merge if it looks ok.  Thanks.

kriegfrj:
> @kriegfrj thanks...now all checks builds are succeeding. I don't know how to verify the shape of the built p2 repo...i.e. if the check builds actually include the new ECF feature in the right category, so I'll ask someone @kriegfrj to check that and merge if it looks ok. Thanks.

I'm not an expert on this either, but I think you can add the built P2 repository to your Eclipse installation by adding a repo ("Install new software...") and specifying the path to the built p2 directory. Then you can see how it looks in Eclipse, which will effectively be verifying the shape of the repo.

scottslewis:
> > @kriegfrj thanks...now all checks builds are succeeding. I don't know how to verify the shape of the built p2 repo...i.e. if the check builds actually include the new ECF feature in the right category, so I'll ask someone @kriegfrj to check that and merge if it looks ok. Thanks.
> 
> I'm not an expert on this either, but I think you can add the built P2 repository to your Eclipse installation by adding a repo ("Install new software...") and specifying the path to the built p2 directory.   

Yes, but I don't know the url for the check builds above.

kriegfrj:
Thanks!

For future reference, this probably should have been squashed before merging, and also a "Fixes #6513" note in the commit message so that merging would also close the corresponding issue.

PR #6518, opened 2025-03-18
add README how to generate JDK .properties

chrisrueger:
Based on https://github.com/bndtools/bnd/pull/6508#issuecomment-2733296245 

This adds a readme how to create the `.properties ` files for different JAVA / JDK versions in https://github.com/bndtools/bnd/tree/master/biz.aQute.bndlib/src/aQute/bnd/build/model

### DIFF
--- a/DEV_README.md
+++ b/DEV_README.md
@@ -1,5 +1,4 @@
-Bndtools Development: Tips and Tricks
-=====================================
+# Bndtools Development: Tips and Tricks
 
 Here are some tips for developing in bnd / bndtools.
 
@@ -136,7 +135,7 @@ public class MyTest {
 ```
 
 
-## Method level:
+### Method level:
 
 ```java
 import aQute.bnd.test.jupiter.InjectTemporaryDirectory;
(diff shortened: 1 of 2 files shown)


chrisrueger:
@bjhargrave in https://github.com/bndtools/bnd/pull/6508#issuecomment-2733329482 you said

>  (just make sure the readme is not slurped into the bnd jar along with the properties files).

Any pointer of how to exclude it? (i checked [-includeresource](https://bnd.bndtools.org/instructions/includeresource.html) but no luck... excluding was mainly for unrolling jars)

I noticed that the [readme.md ](https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/readme.md) in the root is also in the final `.jar` and even pre-processed (curly braces - seems on purpose)

https://github.com/bndtools/bnd/blob/592c71e0a41879141a2cd09901f46676ab321f01/biz.aQute.bndlib/bnd.bnd#L53

Any specific reason why you want to exclude the readme?

chrisrueger:
Another idea if excluding them is difficult:
I could make the README smaller (basically a one liner) and point to https://github.com/bndtools/bnd/blob/master/DEV_README.md

That way the README in the jar is very small and I move the larger portion to DEV_README.md

PR #6528, opened 2025-03-21
parallelize build of -sub bundles

chrisrueger:
I could speed up our wrapper project which contains currently 12 sub-bundles from `19s` down to `11s`. 
This project mainly wraps 3rd party non-osgi jars to make them OSGi bundles for our application. 

Feedback welcome whether there are side-effects I don't see. 
Are sub-bundle builds independent from each-other?

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
@@ -1520,7 +1520,11 @@ public class Builder extends Analyzer {
 		List<Builder> builders = getSubBuilders();
 
 		for (Builder builder : builders) {
-			startBuild(builder);
+			try {
+				startBuild(builder);
+			} catch (Exception e) {
+				builder.exception(e, "Exception Start Building %s", builder.getBsn());
+			}
 		}
 
 		Collection<Jar> result = new ConcurrentLinkedQueue<>();
@@ -1538,12 +1542,17 @@ public class Builder extends Analyzer {
 				} catch (Exception e) {
 					builder.exception(e, "Exception Building %s", builder.getBsn());
 				}
-				if (builder != this)
-					getInfo(builder, builder.getBsn() + ": ");
 			});
 
 		for (Builder builder : doneBuilders) {
-			doneBuild(builder);
+			try {
+				doneBuild(builder);
+			} catch (Exception e) {
+				builder.exception(e, "Exception Done Building %s", builder.getBsn());
+			}
+
+			if (builder != this)
+				getInfo(builder, builder.getBsn() + ": ");
 		}
 
 		return result.toArray(new Jar[0]);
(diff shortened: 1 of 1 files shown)


pkriens:
The builder works independently from the rest. However, `startBuild` and `doneBuild` integrate the result in the project and this is not thread safe.

To be safe, I would call them outside the parallel body.

chrisrueger:
Thanks Peter.

> To be safe, I would call them outside the parallel body.

You mean like this? @pkriens 

```
builders = getSubBuilders();

		for (Builder builder : builders) {
			startBuild(builder);
		}

		builders.parallelStream()
			.forEach(builder -> {
				try {

					Jar jar = builder.build();
					jar.setName(builder.getBsn());
					result.add(jar);

				} catch (Exception e) {
					builder.exception(e, "Exception Building %s", builder.getBsn());
				}
				if (builder != this)
					getInfo(builder, builder.getBsn() + ": ");
			});

		for (Builder builder : builders) {
			doneBuild(builder);
		}
```

chrisrueger:
Update 4: I tried making the start and done methods synchronized 73bc5fb0583aa029f41937b5e292ae3e6252f315

pkriens:
> Update 4: I tried making the start and done methods synchronized https://github.com/bndtools/bnd/commit/73bc5fb0583aa029f41937b5e292ae3e6252f315

We tried to remove all synchronization from all bnd methods except in the repositories. bnd is not thread safe. So I think this is the wrong direction ... 

If starting and done method requires strict ordering, then you cannot parallelize it.

chrisrueger:
> We tried to remove all synchronization from all bnd methods except in the repositories. bnd is not thread safe. So I think this is the wrong direction ...

Oh ok I see. I will revert. 

> If starting and done method requires strict ordering, then you cannot parallelize it.

How can I find out, if this is the case?

https://github.com/bndtools/bnd/blob/144b558b96d9d40a2adbbf93136c779335dcbdda/biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java#L846-L863

https://github.com/bndtools/bnd/blob/144b558b96d9d40a2adbbf93136c779335dcbdda/biz.aQute.bndlib/src/aQute/bnd/build/ProjectBuilder.java#L866-L892

chrisrueger:
I am wondering if 

- a) there is some condition where a parallel build of sub-bundles is less problematic? 

e.g. our wrapper bundle has

```
# bnd.bnd
-resourceonly: true
-sub: *.bnd
```

Would `-resourceonly: true` be less risky to build in parallel?

- b) or we could add a new instruction (e.g. `-parallel: sub`)
I found [-noparallel](https://bnd.bndtools.org/instructions/noparallel.html) but that is only evaluate by gradle.

chrisrueger:
> We tried to remove all synchronization from all bnd methods except in the repositories. bnd is not thread safe. So I think this is the wrong direction ...
> 
> If starting and done method requires strict ordering, then you cannot parallelize it.

I think I will close this for now. I would have liked to get the speed benefit, but I don't want to risk breaking things.

PR #6529, opened 2025-03-21
Jdk24 preparation

chrisrueger:
- Adding files to support JDK 20 to 24
- generated Java .properties files based docs adding https://github.com/bndtools/bnd/pull/6518


chrisrueger:
@bjhargrave could you maybe have a look? I tried having a look at recent PRs from you, but not sure what I missed. 
@kriegfrj FYI since related to recent #6508

I will squash commits at the end.

biz.aQute.bndlib/src/aQute/bnd/build/model/JavaSE_18.properties
bjhargrave, line 1: You only need to add a JavaSE_*.properties file when there is a difference from the prior version. See https://github.com/bndtools/bnd/commit/8739a86eff35761d683b6783bdca0e523a5065e8

So you may be able to remove some of these if they are identical to the previous version's properties file.

bjhargrave (commented):
I think you also should update the CI build to test with JDK 17 (which is the minimum version required), all subsequent LTS versions (21) and also the latest released version (24).

You should also investigate any changes to the JVMS since JDK 17 which may need updates to the biz.aQute.bnd.util/src/aQute/bnd/classfile package.

biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java
```diff
@@ -159,6 +159,8 @@ public Map<String, Set<String>> getProfiles() throws IOException {
 		Java_22,
 		Java_23,
 		Java_24,
+		Java_25,
+		Java_26,
```
bjhargrave, line 163: You might want to go big here and add the next 4-6 versions. You can always go back later with properties files, if needed, once each version is released.

Adding 2 versions is only one year and people may use a Bnd version for a long time.

biz.aQute.bndlib/src/aQute/bnd/build/model/EE.java
```diff
@@ -72,6 +72,8 @@ public enum EE {
 	JavaSE_22(22),
 	JavaSE_23(23),
 	JavaSE_24(24),
+	JavaSE_25(25),
+	JavaSE_26(26),
```
bjhargrave, line 76: Go big and add more versions.

biz.aQute.bndlib/src/aQute/bnd/build/model/EE.java
```diff
@@ -302,7 +304,7 @@ public static SortedSet<EE> getEEsFromRequirement(String requirement) {
 	final static EE[] classFileVersionsMinus44 = {
 		UNKNOWN, JRE_1_1, J2SE_1_2, J2SE_1_3, J2SE_1_4, J2SE_1_5, JavaSE_1_6, JavaSE_1_7, JavaSE_1_8, JavaSE_9,
 		JavaSE_10, JavaSE_11, JavaSE_12, JavaSE_13, JavaSE_14, JavaSE_15, JavaSE_16, JavaSE_17, JavaSE_18, JavaSE_19,
-		JavaSE_20, JavaSE_21, JavaSE_22, JavaSE_23, JavaSE_24
+		JavaSE_20, JavaSE_21, JavaSE_22, JavaSE_23, JavaSE_24, JavaSE_25, JavaSE_26
```
bjhargrave, line 307: Go big and add more versions.

.github/workflows/cibuild.yml
```diff
@@ -45,7 +45,7 @@ jobs:
         os:
         - 'ubuntu-latest'
         java:
-        - '20'
+        - '21'
```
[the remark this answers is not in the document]
chrisrueger: I didn't add JDK24 yet, because JDK24 currently needs a newer gradle version which is not yet released. I tried with [Gradle 8.14 nightly 8.14-20250329033919+0000](https://gradle.org/release-nightly/) but there seems to be a [groovy problem still not supporting JDK 24.](https://docs.gradle.org/release-nightly/release-notes.html#support-for-building-projects-with-java-24) 

So I may have to wait a little more.

chrisrueger:
> You should also investigate any changes to the JVMS since JDK 17 which may need updates to the biz.aQute.bnd.util/src/aQute/bnd/classfile package.

I checked https://docs.oracle.com/en/java/javase/24/language/java-language-changes-summary.html

To me it looks like SealedClasses permanent in Java 17 where the last major JVM change relevant<sup>*</sup> for bnd. 

_<sup>*</sup> with 'relevant' I mean mainly things which affect types and packages which are needed to calculate manifest._

**What is your opinion on Preview features?** When would we consider supporting them? @bjhargrave 

I first thought about [Module Import Declarations](https://docs.oracle.com/en/java/javase/24/language/module-import-declarations.html#GUID-E1581E3A-2B0F-4F3B-8FF1-6EBEC8A210DB) but I don't think we need to handle them specifically since the byte code still contains `import package` declarations as usual (in other words they are expanded under the hood).

stbischof:
i am just here to say thank you @chrisrueger

.github/workflows/cibuild.yml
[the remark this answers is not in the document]
bjhargrave: Well then use 23 until a version of gradle is available which supports 24.

bjhargrave:
> **What is your opinion on Preview features?** When would we consider supporting them?

Your link is to the JLS. I am talking about the JVMS which specifies the class file format: https://docs.oracle.com/javase/specs/jvms/se24/html/jvms-4.html. But it does appear there are no class file format changes since 17.

biz.aQute.bnd.util/src/aQute/bnd/classfile/ClassFile.java
```diff
@@ -15,7 +15,7 @@
 import aQute.lib.io.IO;
 
 public class ClassFile extends ElementInfo {
-	public static final int		MAJOR_VERSION	= 61;
+	public static final int		MAJOR_VERSION	= 68;
```
bjhargrave, line 18: No. Only change this when there is a class file change you are supporting added by that major version. So leave as 61 (Java 17) for now.

chrisrueger: oh ok, sorry. reverted.

chrisrueger:
@bjhargrave question to PR https://github.com/bndtools/bnd/pull/6536 next door where I am experimenting with upgrading to use latest **Eclipse 2025-03**.

I get `java.lang.foreign` not found, which is a package from [JavaSE_21.properties](https://raw.githubusercontent.com/bndtools/bnd/aa849528419ef5d565de0664117c2133d5eda9ce/biz.aQute.bndlib/src/aQute/bnd/build/model/JavaSE_21.properties)

```
Resolution failed. Summary:
      ⇒ Bundle: org.eclipse.sdk cannot be resolved
              ⇒ Require-Bundle: org.eclipse.help.base; version=[4.0.0,5.0.0) cannot be resolved
                      ⇒ Require-Bundle: org.apache.lucene.core; version=[10.0.0,11.0.0) cannot be resolved
                              ⇒ because Import-Package requirement for: java.lang.foreign could not be provided by any available bundle or dependency
```

i have `-runee: JavaSE-21`

Would that error go away once this PR here is inside bnd? Are those package `.properties` files considered for resolution?
My interpretation would be yes because of 

https://github.com/bndtools/bnd/blob/ff9e429062a65188ccf83396c5b709747e4e9de4/biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java#L172-L179 

but wanted to confirm.

BTW: Eclipse 2024-12 is running fine. This error starts with 2025-03

biz.aQute.bndlib/src/aQute/bnd/osgi/Clazz.java
```diff
@@ -159,6 +159,12 @@ public Map<String, Set<String>> getProfiles() throws IOException {
...
+		Java_27,
+		Java_28,
+		Java_29,
+		Java_30,
```
laeubi, line 167: Why not even to Java_50 ... ?

chrisrueger: No special reason... just because it was suggested here https://github.com/bndtools/bnd/pull/6529#discussion_r2021408271

bjhargrave: > Why not even to Java_50 ... ?

Making predictions is hard. Especially about the future. :-)

PR #6536, opened 2025-03-24
WIP Upgrade bndtools to higher eclipse

chrisrueger:
Trying with 2024-12


chrisrueger:
Current status:

- I can launch the bnd eclipse instance with Eclipse 2024-12
- but still tests failing... mainly because changed dependencies e.g. which were part of of Eclipse 2022-09, but are not in 2024-12 anymore.

image

chrisrueger:
@kriegfrj You seem to be very familiar with all the JUnit Testcases stuff in bnd / bndtools. 
Bndtools can be started with the Eclipse 2024-12 just fine, but I cannot sort out how to  get the testcases running. 
I think I got lost in all the JUnit versions since a lot of libs got updated or were removed in the newer Eclipse. So there is something messed up with different bundle (incompatible) versions loaded and classloading issues and what not. 

Some tests are working, but I got there via trial and error and I always feel I'm not doing it right. 

So I could need some support there. 
Maybe you see something in the errors https://github.com/bndtools/bnd/actions/runs/14252255522/job/39947564487?pr=6536 and have some pointers or thoughts. 

Thanks

cnf/ext/repositories.bnd
```diff
@@ -4,43 +4,47 @@ ossrh:                  https://oss.sonatype.org/content/repositories/snapshots
...
         noupdateOnRelease=true
 
 -plugin.1.Eclipse:\
-    aQute.bnd.repository.osgi.OSGiRepository;\
-        name="Eclipse 4.25 (2022-09)";\
-        locations="https://bndtools.jfrog.io/bndtools/bnd-build/eclipse/4.25/index.xml.gz";\
-        poll.time=-1;\
-        cache="${workspace}/cnf/cache/stable/Eclipse-4.25"
+	aQute.bnd.repository.p2.provider.P2Repository; \
+	 	url = https://download.eclipse.org/releases/2024-12/; \
```
laeubi, line 32: Instead of larger upgrades, I would recommend to first go from 

2022-09 > 2022-12

using smaller increments (that can then already be merged) usually works better for plugins quite tightly integrated (like Bndtools)

kriegfrj:
> @kriegfrj You seem to be very familiar with all the JUnit Testcases stuff in bnd / bndtools. Bndtools can be started with the Eclipse 2024-12 just fine, but I cannot sort out how to get the testcases running. I think I got lost in all the JUnit versions since a lot of libs got updated or were removed in the newer Eclipse. So there is something messed up with different bundle (incompatible) versions loaded and classloading issues and what not.
> 
> Some tests are working, but I got there via trial and error and I always feel I'm not doing it right.
> 
> So I could need some support there. Maybe you see something in the errors https://github.com/bndtools/bnd/actions/runs/14252255522/job/39947564487?pr=6536 and have some pointers or thoughts.
> 
> Thanks

Sorry, I have a lot going on and haven't been paying close attention to this recently.

I'm a bit rusty on the details, but I remember that these tests were a bit of a pain because they run inside an Eclipse instance which might have different JUnit on its classpath than the one you are trying to test against. Then the launchpad would include a version of JUnit in the framework-under-test that it launched, which might be different to both of those.

The error:
```
 # org.junit.platform.launcher.TestExecutionListener: org.junit.platform.launcher.listeners.UniqueIdTrackingListener not a subtype
    java.util.ServiceConfigurationError: org.junit.platform.launcher.TestExecutionListener: org.junit.platform.launcher.listeners.UniqueIdTrackingListener not a subtype
```

...seems to suggest that this is in that category. It seems that UniqueIdTrackingListener being run inside the launchpad framework is linked to a different version of the TestExecutionListener class.

I'll see if I can have a look in the next day or so.

chrisrueger:
Closing and retrying after https://github.com/bndtools/bnd/pull/7000 which should solve the JUnit problems. But we will start over fresh and just use this as a template.

PR #6537, opened 2025-03-24
Fix keep duplicate -runbundles

### LINKED ISSUE #6526: [resolver] Bundle version being removed from resolution saved to .bndrun
I have come across what looks like a regression.

I am working on iDempiere, and it has quite a few bundles that are not, shall we say, "best practice". However, I have managed to have my resolution working recently and it has been stable for some time.

Today I have tried re-resolving, and I think it's the first time since recently upgrading to HEAD. I have come across a strange issue.

I ran my resolve as usual. The resolver determined (correctly) that two versions of FastInfoset are required:

Image

However, when I click "update", the lower of the two versions is omitted from the `-runbundles` that is written to the `.bndrun`:

Image

I removed the `-runstartlevel: orderby...` just to see if that was the cause and it didn't seem to fix the problem.

I am currently running Bndtools 7.2.0.DEV-202502261413-ga418a9b.

chrisrueger:
Closes #6526

this fixes the problem that in a RunResultion which suggested the same bundle but in different versions only one bundle (the highest) was kept. Now both suggestions are kept.

The problem was that Parameters.toParameter() used parameters.put() instead of parameters.add(). .put() removes duplicate keys, .add() keeps duplicates by using the duplicate markers.


chrisrueger:
@kriegfrj @pkriens Please have a look at the fix. The problem was basically that `HeaderClause.toParameters()` removed duplicates, because it used `.put()` instead of `.add()`.

I would like to get it into master / snapshot, to get some more real word testing.

biz.aQute.resolve/test/biz/aQute/resolve/RunResolutionTest.java
```diff
@@ -517,24 +518,40 @@ public void testPrintHumanReadableDifference() throws Exception {
...
+			.toString()).isEqualTo("org.apache.felix.gogo.runtime;version='[0.10.0,0.10.1)';startlevel=110");
+		softly.assertThat(runBundles.get(2)
+			.toString()).isEqualTo("osgi.enroute.junit.wrapper;version='[4.12.0,4.12.1)';startlevel=120");
+
+		softly.assertThat(runBundles.get(3)
 			.getName()).isEqualTo("test.simple");
-		assertThat(runBundles.get(1)
+		softly.assertThat(runBundles.get(3)
 			.getAttribs()).containsEntry(Constants.RUNBUNDLES_STARTLEVEL_ATTRIBUTE, "90");
+		softly.assertThat(runBundles.get(3)
```
kriegfrj, line 542: I think this test is flaky as the resolver will not always produce the same order by result.

Maybe converting runbundles into a `List<String>` and then assert on it using `containsExactlyInAnyOrder()`?

biz.aQute.resolve/test/biz/aQute/resolve/RunResolutionTest.java
```diff
@@ -517,24 +518,40 @@ public void testPrintHumanReadableDifference() throws Exception {
...
+			.toString()).isEqualTo("test.simple;version=snapshot;startlevel=90");
```
kriegfrj, line 543: This assertion is almost a duplicate of the one on line 540/541.

pkriens:
i am ok with this

biz.aQute.resolve/test/biz/aQute/resolve/RunResolutionTest.java
chrisrueger: fixed by recent commit.

biz.aQute.resolve/test/biz/aQute/resolve/RunResolutionTest.java
```diff
@@ -517,24 +518,34 @@ public void testPrintHumanReadableDifference() throws Exception {
...
-		assertThat(runBundles.get(0)
-			.getAttribs()).containsEntry(Constants.RUNBUNDLES_STARTLEVEL_ATTRIBUTE, "100");
-		assertThat(runBundles.get(1)
-			.getName()).isEqualTo("test.simple");
-		assertThat(runBundles.get(1)
-			.getAttribs()).containsEntry(Constants.RUNBUNDLES_STARTLEVEL_ATTRIBUTE, "90");
+		softly.assertThat(runBundles).hasSize(4);
+		softly.assertThat(runBundles.stream()
+			.map(rb -> rb.toString())
+			.toList())
```
kriegfrj, line 533: I *think* `toList()` is redundant - you can assert directly on the stream.

kriegfrj (changes requested):
I think I'm OK to give this a go. We're likely to find any potential bugs quicker "in the wild"

biz.aQute.resolve/test/biz/aQute/resolve/RunResolutionTest.java
```diff
@@ -517,24 +518,34 @@ public void testPrintHumanReadableDifference() throws Exception {
...
-		assertThat(runBundles).hasSize(2);
-		assertThat(runBundles.get(0)
-			.getName()).isEqualTo("osgi.enroute.junit.wrapper");
-		assertThat(runBundles.get(0)
-			.getAttribs()).containsEntry(Constants.RUNBUNDLES_STARTLEVEL_ATTRIBUTE, "100");
-		assertThat(runBundles.get(1)
-			.getName()).isEqualTo("test.simple");
-		assertThat(runBundles.get(1)
-			.getAttribs()).containsEntry(Constants.RUNBUNDLES_STARTLEVEL_ATTRIBUTE, "90");
+		softly.assertThat(runBundles).hasSize(4);
```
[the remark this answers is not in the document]
chrisrueger: Ups sorry, somehow missed that comment. I will push another PR with it.

Issue #6524, opened 2025-03-19
[bndtools] Effective view tooltips to have scrollbar


kriegfrj:
I am really liking the Effective view that is now available in the Bnd Editor.

I like the fact that I can hover over it to get the full property value setting as a tooltip.

One thing I just noticed: for property definitions that are very long (eg, `-runbundles`), the tooltip can be so large as to not fit on the screen. There are no scrollbars available to scroll to see all of the data (eg, the Eclipse built-in Javadoc tooltip has a scrollbar for longer Javadoc definitions).

Alternatively (or perhaps additionally) the ability to copy the values into the clipboard so you can paste them somewhere and examine them would be useful.

chrisrueger:
@kriegfrj have a look at https://github.com/bndtools/bnd/pull/6552

PR #6552, opened 2025-04-07
Effective: show textbox tooltips and checkbox

chrisrueger:
Closes https://github.com/bndtools/bnd/issues/6524

**Tweaks for the Effective Tab in BndEditor**

- now the tooltips have scrollbars and allow to click into and copy the content to clipboard
- also I added a Checkbox "Show expanded values" to toggle between raw value and expanded value. for both table and source view

https://github.com/user-attachments/assets/1e162cf7-42d2-4dda-9d72-b675a805da87


chrisrueger:
@pkriens FYI want to test it in the wild.

PR #6554, opened 2025-04-08
RepoView: improve performance & fix memory leak

chrisrueger:
When working on the recent Eclipse P2 SimRel Repo stuff I noticed that my Eclipse instance was getting really slow over time. Especially when I did something in the Repositories view. I had to search for dependencies the whole day. 

So I investigated a bit and found out that `RepositoryTreeContentProvider.java` maintains an ever growing cache where it stores each filter term per repo.

So for each filter term I use for search it stores an entry with a large list in a map. 
For example if I would type `javax.xml` very slow, it would put each filter term in the cache:
- j
- ja
- jav
- java
- javax
- javax.
- javax.x

image

This cache is only cleared when you add / remove a repo... but that does not happen so often. 

## This PR

- fix memory leak using LRUCache, limiting the cache to 10 entries, removing the oldest. I experimented a bit and I feel it is ok and usable. 
- save / restore tree expansion state before filtering and when filter is cleared again
  - this can speed up performance e.g. if a large P2 repository was collapsed before filtering, so it appears collapsed again after filtering -> avoiding unnecessary redrawing of hundreds of nodes

Overall this makes the Repositories view feel more responsive. 
And when I have the heap usage open, I see much smaller spikes when typing search filter terms.

### DIFF
--- a/bndtools.core/src/bndtools/model/repo/RepositoryTreeContentProvider.java
+++ b/bndtools.core/src/bndtools/model/repo/RepositoryTreeContentProvider.java
@@ -7,6 +7,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


chrisrueger:
@pkriens I will test it again and roll back in case something goes wrong. I hope it can only get better.

PR #6564, opened 2025-04-17
HeaderClause: fix wrong quoting of directives

### LINKED ISSUE #6563: Bndeditor: Customize imports: Incorrect quotes when adding version to an optional import
## Steps to reproduce 

1. mark an import as optional and save
2. then add a version, then save again. 

Image

**before:**
`com.netflix.*;resolution:=optional`

**after:**
`com.netflix.*;'resolution:'=optional;version='1.0.0`

**expected correct would be:** 
`com.netflix.*;resolution:=optional;version='1.0.0`

If I remove the version again in the UI then the string is still broken:

`com.netflix.*;'resolution:'=optional,\`

expected:
`com.netflix.*;resolution:=optional,\`

same for ranges:

**before:**
`com.netflix.*;resolution:=optional`

add range [1.0.0,2.0.0)

**after:**
`com.netflix.*;'resolution:'=optional;version='[1.0.0,2.0.0)',\`

**expected correct would be:** 
`com.netflix.*;resolution:=optional;version='[1.0.0,2.0.0)',\`

UPDATE:
I think the version is not the problem. The **directive** `resolution:=` is the problem. this gets converted incorrectly.

chrisrueger:
Closes #6563

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/ResourceTest.java
+++ b/biz.aQute.bndlib.tests/test/test/ResourceTest.java
@@ -30,10 +30,13 @@ import org.osgi.resource.Namespace;
 import org.osgi.resource.Requirement;
 import org.osgi.resource.Resource;
 
+import aQute.bnd.annotation.Resolution;
+import aQute.bnd.build.model.clauses.ImportPattern;
 import aQute.bnd.build.model.clauses.VersionedClause;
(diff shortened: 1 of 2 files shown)


chrisrueger:
@pkriens Since it is bndlib, I would appreciate a quick review. Thanks

PR #6571, opened 2025-04-21
Effective view: show merged properties

### LINKED ISSUE #6525: [bndtools] Effective view to show merged properties?
As noted earlier, I like the effective view.

One thing I think we should consider is showing in the Effective view the results of the property merging. So if I had:

a.bnd:
```
-runrequires.a: bundle.a
```

b.bnd:
```
-include: a.bnd
-runrequires.b: bundle.b
```

... then the effective view would be:

```
-runrequires: bundle.a, bundle.b
```
The Provenance whould then reveal where in a.bnd and b.bnd those definitions were from. If it might be useful, it could also name the original pre-merged property name alongside the file name.

chrisrueger:
Closes #6525

- add a new option to show the properties as merged properties (group the properties by stem)
- works in table and source view
- refactored BndSourceEffectivePage slightly. I removed the functions from ColSpec and introduced record PropertyRow which is populated in getTableData()
- added: hovering over 'key' column now also shows the Syntax.HELP text like the sourceViewer does

image

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
@@ -1601,7 +1601,7 @@ public class BndEditModel {
 			return (Converter) converters.get(header);
 	}
 
-	private static String getStem(String header) {
+	public static String getStem(String header) {
 		int n = header.indexOf('.');
 		if (n < 0)
 			return header;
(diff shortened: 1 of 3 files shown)


chrisrueger:
Let's play around with it in the wild and do finetuning in another PR. 
Let me know if you like changes or find bugs. 

@kriegfrj @pkriens

PR #6581, opened 2025-04-30
addSources context menu for BndPomRepo

chrisrueger:
Relates to https://github.com/bndtools/bnd/issues/6578#issuecomment-2840463367

This adds the "Add Sources" feature - known from MavenBndRepository - to download source code for a Repo Revision 
 also to BndPomRepository.

image

It shares the same code with a minor refactoring.

Issue #6578, opened 2025-04-28
Remove OSGi-OPT folder from biz.aQute.launcher  and biz.aQute.launcher.pre


stbischof:

kriegfrj:
Hi Stefan, what is is the rationale for this request? I don't see a problem with it but it would be nice to understand the use case. Is it just to make the bundle smaller?

stbischof:
Yes, it is just to make it smaller.

Als the initial usecase of osgi-opt was to have te sources in place, in a time in history here the setup did not work with sources published to mavencentral.

Since this work now fine ther is no casero Store this files.

chrisrueger:
I don't know about the history of including the source code in OSGI-OPT, but given that the sources also appear on Maven central it seems redundant.

https://repo1.maven.org/maven2/biz/aQute/bnd/biz.aQute.launcher/7.1.0/

<img width="666" alt="Image" src="https://github.com/user-attachments/assets/573bee29-b535-4c79-83d8-850557813755" />

kriegfrj:
One use case for having the source in OSGI-OPT is that it helps the source lookup in Bndtools. I am a bit hazy on the details but I remember that the Bndtools source lookup would look for sources in OSGI-OPT of a bundle. However, there are other ways to make source lookup work. For example, the MvnBndRepository plugin, if you right-click on a bundle and select "download sources", it will add the sources into the bundle in OSGI-OPT in your local Maven cache (that's when the jars get renamed from `bundlename.jar` to `+bundlename.jar`).

I believe that generally speaking there is probably not a need for these in the bundles themselves. And the launcher bundles in particular could benefit from being smaller for use in embedded environments.

stbischof:
Yes meaned the mecanism with MvnBndRepository.

I hought this works for all repos.
I wasreall sueprised that the launcher is nearly 400kb.

chrisrueger:
> I wasreall sueprised that the launcher is nearly 400kb.

I experimentally used `-sources: false` for the launcher and the included pre.launched and got the final jar down to 280kb.



> Yes meaned the mecanism with MvnBndRepository.
> 
> I hought this works for all repos.

Just tested this. It seems unfortunatelly it is repo-dependent. Didn't work for BndPomRepository. I was surprised too. 

I will try if this can be added so that both MavenBndRepo and BndPomRepo support "Add sources" (right click context menue)

chrisrueger:
[...]

PR #6584, opened 2025-05-02
Launcher: don't include source code in OSGI-OPT

chrisrueger:
- Related to #6578 where it was requested to not include source code in the jars, to keep them smaller

The source code is available as separate jars on Maven Central anyway. 

https://repo1.maven.org/maven2/biz/aQute/bnd/biz.aQute.launcher/7.1.0/

Image

I also added  https://github.com/bndtools/bnd/pull/6581 , which adds the "Add sources" context menu for in Repo browser for BndPomRepo. This then makes it easier to use the source code artifacts on maven central if needed.

### DIFF
--- a/biz.aQute.launcher/bnd.bnd
+++ b/biz.aQute.launcher/bnd.bnd
@@ -32,3 +32,6 @@
 -builderignore: testresources
 
 -baseline: *
+
+# Do not include source code in OSGI-OPT  
+-sources: false
--- a/biz.aQute.launcher/bnd/pre.bnd
+++ b/biz.aQute.launcher/bnd/pre.bnd
@@ -3,3 +3,6 @@
 	aQute.launcher.pre
 
 -baseline:
+
+# Do not include source code in OSGI-OPT  
+-sources: false


bjhargrave:
> The source code is available as separate jars on Maven Central anyway.

Maven central is not always reachable from the machine wanting the source code. 

This seems a lot of fuss over a few 100K.

stbischof:
But what is the usecase where a developer has no connection to maven or a local proxy and can't see the sources of all its dependencies he uses. And then he wants exactly the bnd.launcher.jar.

chrisrueger:
> And then he wants exactly the bnd.launcher.jar.

Ok, I should have mentioned that my change is not only for the launcher, but _all_ bundles of bnd. 
I could narrow it to only the launcher if needed (and requested in #6578 ) but I saw no point in doing it for some only.
But let me know if launcher (+ pre-launcher) is enough. 

I guess it is relevant for people working with embedded devices where kilobytes might matter. So @stbischof , it is up to you to make a strong case for this.

stbischof:
So , if i am the only one with an case i will Switch the Launcher and everything is fine.

laeubi:
> In low-memory devices this might matter, to reduce the footprint to a minimum.

Actually shipping the source with an open-source bundle is good. There are many other ways than maven central one can get a bundle, e.g. with a third party product.

Also I want to mention [what the spec says](https://docs.osgi.org/specification/osgi.core/8.0.0/framework.module.html#d0e1760):

> Can contain optional documentation in the OSGI-OPT directory of the JAR file or one of its sub-directories. Any information in this directory is optional. For example, the OSGI-OPT directory is useful to store the source code of a bundle. **Management systems may remove this information** to save storage space in the OSGi framework. 

So anyone that obtains this bundle and really don't want/need it for storage concerns **can safely remove this**. This can even be done on-the-fly while downloading things using ZipInput/OutputStream combination.

chrisrueger:
> Actually shipping the source with an open-source bundle is good. There are many other ways than maven central one can get a bundle, e.g. with a third party product.

Hmm ok, there seem to be contradicting opinions on this topic. 
I'm a little surprised at the otherwise minimalist OSGi community (only pull in what is really necessary). 

> So anyone that obtains this bundle and really don't want/need it for storage concerns **can safely remove this**. This can even be done on-the-fly while downloading things using ZipInput/OutputStream combination.

On that note: Sounds interesting. My question would be if that should be done with the bnd-way of wrapping it into a new jar with a new name , or if the original jar should be touched. 
But I wonder how this impacts the md5/shaxxx checksums. 

So I guess the current bnd-way to remove the OSGI-OPT folder would be wrapping the jar like this:

```
# central.mvn
biz.aQute.bnd:biz.aQute.launcher:7.1.0
```
in your wrapping bundle:
```
#bnd.bnd
-includeresource: @${repo;biz.aQute.bnd:biz.aQute.launcher;latest}!/!OSGI-OPT/*
```

See https://bnd.bndtools.org/instructions/includeresource.html#inline-form

I wonder if it would be practical to have feature in bnd to do this automatically directly in the repo definition:
e.g. using some kind of special `nosources` qualifier or something like this.

```
biz.aQute.bnd:biz.aQute.launcher:7.1.0:nosources
```

laeubi:
> But I wonder how this impacts the md5/shaxxx checksums.

Checksums are only relevant for verify the download integrity, so *after* you downloaded it you can modify it in any way without a problem.

 > I wonder if it would be practical to have feature in bnd to do this automatically directly in the repo definition

Yes this likely is in the sense of the specification, an alternative would be e.g. when creating an executable jar, a product or ...

laeubi:
You mentioned above:

> I guess it is relevant for people working with embedded devices where kilobytes might matter. 
> In low-memory devices this might matter, to reduce the footprint to a minimum.

@stbischof did not mentioned the use-case so safe a few kilobytes at all.

In any case in the scenario of a "low-memory devices" (even though it won't usually account to the memory usage) or "embedded devices" I would except none of them are using a bndrun but some kind of carefully crafted package of executable code. This "deployment artifact" can easily have sources removed using whatever technique is sufficient (e.g. any zip utility).

chrisrueger:
Ok I will close this now. There seems no consensus and I personally don't have the problem. Feel free to take over and reopen.

Issue #6585, opened 2025-05-02
New Workspace Wizard (Template Fragments): Update Current workspace: imported projects do not show up


chrisrueger:
I just discovered a surprising behavior of the New Workspace wizard with the new Fragments.

- I start my bnd dev setup with the workspace `/Users/christoph/git/bnd/bndtools.test/workspace`
- there I "Import existing bnd workspace" which is at a different location

So Eclipse workspace and bnd workspace is different. 

Now I want to add template Fragment e.g. the new ["Servlet" template](https://github.com/bndtools/workspace-templates/pull/3) I am working on. This template not only adds something to /cnf/ext, but also imports a new project into the workspace. 

## Problem

After the wizard is finished I do not see the file added to `cnf/ext/` (a `.pmvn` file), nor did I see the new project in my workspace.

## Analysis

**Actual behavior:**

I found out that `bndtools.wizards.newworkspace.Model.location` points to the **Eclipse workspace**, instead of my **bnd workspace**.
The import just created a cnf/ext/ folder in my Eclipse workspace and put the file there.

<img width="355" alt="Image" src="https://github.com/user-attachments/assets/a5c41d87-2f6d-45c0-a9aa-ecfde26d6fec" />

**Expected behavior:**

I expected the file to be added to my bnd workspace instead.

The same happened for the project being added from the template. 
It was copied to the Eclipse workspace instead of my bnd workspace. 

I will work on a PR to change this. 
Not exactly sure how the logic should be , but if the current workspace is a bnd workspace, then the bnd workspace should be used. 

It might be related to https://github.com/bndtools/bnd/issues/6300 and https://github.com/bndtools/bnd/issues/6302#issuecomment-2380115998

PR #6586, opened 2025-05-02
Template fragments: use bnd workspace if possible

chrisrueger:
Closes https://github.com/bndtools/bnd/issues/6585

This fixes 2 Problems:

- if you select "Update current workspace" in the  NewWorkspaceWizard then the bnd workspace is used instead of the Eclipse workspace, if it is a real bnd workspace (contains a cnf project)

- also fixes the problem that new projects (added by a template fragment) did not show up. This is solved by Refreshing the bnd workspace (see Model. syncWorkspace()) on finish of the Wizzard.


chrisrueger:
@pkriens would be nice if you could have a look. @peterkir too, since we both talked about using the Template Fragments more and more

chrisrueger:
I will keep an eye on it when testing in the wild and fix or revert.

PR #6589, opened 2025-05-04
Experimental: BndRunAnalyzer context menu

chrisrueger:
- adds a right-click context menu for .bndrun files "Analze Bndrun"
- currently it outputs: unreferenced Bundles (Bundle Symbolic Names over all repositories, which are not in -runbundles, -buildpath, -testpath, or ${repo;...} references e.g. in -includeresources)

The result is a text dialog where you can see and copy paste it, like this:

image

## Why?

We have a historically grown `central.mvn` (MavenBndRepository) and I wanted to find out orphaned entries, which could be removed. 

Feedback welcome.


chrisrueger:
Seems to work good so far. I used it to remove a lot of orphaned dependencies in the `central.mvn` file of our project.
I manually checked each dependency outputted by the report, ran all tests. No false positive so far.

Issue #6593, opened 2025-05-10
includeresource with lib:=true copies jar with sources into the final jar in Eclipse


chrisrueger:
# Problem

I am including a dependency like this, with `lib:=true`, so that the dependecy ends up in the final jar and added to the Bundle-Classpath:

```
-includeresource: \
  ${repo;org.springframework:spring-tx;latest}; lib:=true,\
```

For this dependency I have also Added the source code in Eclipse (via right click in the Repo-Viewer).

Then when looking at the my final jar in the `generated` folder I noticed that the jar is included with a little + prefix. 

<img width="320" alt="Image" src="https://github.com/user-attachments/assets/1968a47d-58e7-4bb2-b9e0-d86e792d1b59" />

I read somewhere that this is the mechanism of bnd of making the downloaded source code available, by basically creating a copy of the .jar in the local .m2 maven repo and marking it with a + prefix. 

This is what is in my local maven repo. 
<img width="417" alt="Image" src="https://github.com/user-attachments/assets/1c6e58e5-9c3c-46dd-ab25-5991998457dc" />

I was a bit surprised by this behavior, because:

- it makes the local jar bigger than needed
- it makes the build slower , if I have the source code of lots of dependencies

## Expected behavior:

- I would expect that `-includeresource` would take the dependency jar without the +-prefix. 

The is mainly a problem for local development and not on our CI pipeline. But for local development it is relevant as it slows down the build more than necessary. 

Another possible solution would be to change bndtools' Source lookup and "Add sources" mechanism to not modify the dependy, but just use the `dep-sources.jar` in the .m2 repo.

PR #6595, opened 2025-05-10
rework Add sources & calculateSourceAttachmentPath

chrisrueger:
Closes https://github.com/bndtools/bnd/issues/6593

This PR does mainly 2 things:

- it removes a "hack" in `MavenBndRepository.get()` which was using the +-prefixed jar file (e.g. +somedependency-6.1.17.jar) which is created by "Add sources" in the repo browser. This had negative side effects because those +jars were put into the build output e.g. by `-includeresources`. 
- It improves the Eclipse Source attachment lookup to properly use bnds integrated mechanism to get a -sources.jar if it exists via the BSN_SOURCE_SUFFIX (bsn + ".source") mechanism

Note: I did not change the "Add sources" behavior. The +-jar with included sourcecode under OSGI-OPT is still created. I initially had it and it is reflected in the discussions below, but I reverted that. 

## What does it solve?

- for the initial issue https://github.com/bndtools/bnd/issues/6593 it solves the problem that other jars put inside my jar via `-includeresource` no longer contain the source code just because I used "Add sources" in the IDE
- this also makes the Eclipse IDE created jars more reproducible because they now should be the same as in a CI build

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
@@ -582,9 +582,6 @@ public class MavenBndRepository extends BaseRepository implements RepositoryPlug
 		}
 
 		File f = storage.toLocalFile(archive);
-		File withSources = new File(f.getParentFile(), "+" + f.getName());
-		if (withSources.isFile() && withSources.lastModified() > f.lastModified())
-			f = withSources;
 
 		if (listeners.length == 0) {
 			return f;
--- a/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerSourceManager.java
+++ b/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerSourceManager.java
@@ -1,5 +1,10 @@
 package org.bndtools.builder.classpath;
 
+import static aQute.bnd.osgi.Constants.BSN_SOURCE_SUFFIX;
+import static aQute.bnd.osgi.Constants.VERSION_ATTR_LATEST;
+import static aQute.bnd.service.Strategy.EXACT;
+import static aQute.bnd.service.Strategy.HIGHEST;
+
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
@@ -11,6 +16,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Properties;
+import java.util.SortedSet;
 import java.util.jar.JarInputStream;
 import java.util.jar.Manifest;
 
@@ -34,6 +40,7 @@ import aQute.bnd.build.WorkspaceRepository;
 import aQute.bnd.header.Attrs;
 import aQute.bnd.osgi.Domain;
 import aQute.bnd.service.RepositoryPlugin;
+import aQute.bnd.service.Strategy;
 import aQute.bnd.version.Version;
 import aQute.lib.io.IO;
 import bndtools.central.Central;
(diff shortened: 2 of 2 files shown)


bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
```diff
@@ -488,7 +488,42 @@ private void addProjectEntry(IPath path, List<IAccessRule> accessRules, List<ICl
...
+		 */
+		private IPath trySourcesJar(File file) {
+
+			if (file.getName()
+				.endsWith(".jar")) {
+
+				try {
+					File withSources = new File(file.getParentFile(), file.getName()
+						.substring(0, file.getName()
+							.indexOf(".jar"))
```
reckart, line 516: Should probably be a `lastIndexOf`. How about using a method like Apache Commons IO `FilenameUtils.getExtension()`?

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
```diff
@@ -488,7 +488,42 @@ private void addProjectEntry(IPath path, List<IAccessRule> accessRules, List<ICl
...
+		 * Maven repos
+		 *
+		 * @param file
+		 * @return the bsn-sources.jar file if found, otherwise
+		 *         <code>null</code>
+		 */
+		private IPath trySourcesJar(File file) {
+
+			if (file.getName()
+				.endsWith(".jar")) {
```
reckart, line 511: Should maybe also allow for `.JAR`. File names are not case sensitive on all platforms.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
```diff
@@ -488,7 +488,42 @@ private void addProjectEntry(IPath path, List<IAccessRule> accessRules, List<ICl
...
+					File withSources = new File(file.getParentFile(), file.getName()
+						.substring(0, file.getName()
+							.indexOf(".jar"))
+						+ "-sources.jar");
+
+					if (withSources.isFile()) {
+						return fileToPath(withSources);
+					}
+				} catch (Exception e) {
+					// ignore
```
reckart, line 523: A trace-level logging might be helpful here for possible debugging cases and would remove the need to actually fire up a debugger.

biz.aQute.repository/src/aQute/bnd/repository/maven/util/RepoActionsUtil.java
```diff
@@ -21,42 +22,17 @@ public static void addSources(final Archive archive, Function<Archive, Promise<F
...
+
+				Archive a = archive.getOther(JAR_EXTENSION, SOURCES_CLASSIFIER);
+				lookup.apply(a);
+
+			});
+			return;
 		}
 
-		map.put("-Add Sources", null);
+		// map.put("-Add Sources", null);
```
reckart, line 35: Is this commented out line still necessary? If yes, a comment explaining its use would be helpful.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
chrisrueger: Thanks. I'll try to improve it. FilenameUtils is currently not a dependency at the moment and I don't want to introduce one just for this.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
chrisrueger: Good point.

biz.aQute.repository/src/aQute/bnd/repository/maven/util/RepoActionsUtil.java
chrisrueger: I'll remove that too.

bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java
chrisrueger: I added support for case-insensitive file extension. Although I read that the convention is all lowercase, also enforced by maven central, but unfortunatelly it does not seem to be specified. So I assume there could be uppercase .JAR files out there in enterprise maven repos.

chrisrueger:
Anybody from the longtime committers with knowledge about the history of `+filename.jar` files? @pkriens @bjhargrave @kriegfrj Any objections?

bjhargrave:
> Anybody from the longtime committers with knowledge about the history of `+filename.jar` files?

As far as I can recall, they are just source-enriched (source code added to OSGI-OPT) versions of the original jar so Bnd did not have to manage -sources.jar files.

bjhargrave:
This PR seems to be undoing the original goal of a single jar with classes and source.

chrisrueger:
> As far as I can recall, they are just source-enriched (source code added to OSGI-OPT) versions of the original jar so Bnd did not have to manage -sources.jar files.

This is my impression too.

> This PR seems to be undoing the original goal of a single jar with classes and source.

Correct. 

Now bndtools can use the `-sources.jar` which is downloaded by the MavenBndRepository and BndPomRepository anyway, and does not create a copy of the original .jar + sourcecode anymore .

Is this a problem? What was the reason in the past to not use the  `-sources.jar`? Was it only for the Eclipse-Source-Lookup?

kriegfrj:
> > As far as I can recall, they are just source-enriched (source code added to OSGI-OPT) versions of the original jar so Bnd did not have to manage -sources.jar files.
> 
> This is my impression too.
> 
> > This PR seems to be undoing the original goal of a single jar with classes and source.
> 
> Correct.
> 
> Now bndtools can use the `-sources.jar` which is downloaded by the MavenBndRepository and BndPomRepository anyway, and does not create a copy of the original .jar + sourcecode anymore .
> 
> Is this a problem? What was the reason in the past to not use the `-sources.jar`? Was it only for the Eclipse-Source-Lookup?
> 

I don't have a time for a deep-dive, but my initial concern is to make sure that the source lookup will continue to work after this change. It could be that the `BndContainerInitializer` might need some re-working to attach the sources to the `IClasspathEntry`s that it creates, which (from my very old and fading memory) it doesn't currently do - it relies on them being present in the binary jar in `OSGI-OPT`. You might also need a hook to refresh this when you download sources from the MavenBndRepository context menu (perhaps an option of the MavenBndRepository config to automatically download sources alongside the binaries, ideally do it asychronously).

Also need to see how this change will interplay with the P2Repository.

reckart:
Maybe it would be a good idea to use the legacy JAR-with-source as the default and if there are no sources in the JAR, then check if there is a Maven-style `-sources` JAR which could be used?

chrisrueger:
> Maybe it would be a good idea to use the legacy JAR-with-source as the default and if there are no sources in the JAR, then check if there is a Maven-style `-sources` JAR which could be used?

That is exactly how it now works. I forgot to mention that I left the existing lookup of OSGI-OPT/src in place. So this is done first. Only if no sources are found inside the jar, I do the new lookup in this PR. @reckart @kriegfrj

chrisrueger:
> * this also makes the Eclipse IDE created jars more reproducible because they now should be the same as in a CI build

I just want to highlight this **reproducibility** fact which might appear less important at first glance, but in fact it is IMO

I was bitten by that again. 
I just noticed that a local gradle build output executable jar (using bnd gradle plugin) was much larger than the build on our CI server. 
After unpacking some the executable jar and some jars in there I noticed that it was because my local maven repository (`~/.m2`)  contained the `+` jar files because of `-includeresource`. This happened because ~/.m2 contained such files, because I did "Add sources" in Eclipse for some of those jars which got included. 

image

In other words: a reproducible build is difficult with this situation, because it now heavily depends on the environment and whether or not there are `+` jar files in the local maven repo or not. 

So I hope that this supports the case for this PR.

chrisrueger:
For all involved here in the discussion:
I decided to merge this, after talking to @pkriens and approaching another strategy. So most of the discussions above is a bit outdated, since the change is much smaller than I had initially.

I have used the current code for a day with no issues so far and I will continue monitoring it. If I discover problems I will try to address them quickly. 

Please try out the next SNAPSHOT and let me know of any issues.

PR #6597, opened 2025-05-12
Fix duplicates in Open Type dialog (write empty.index file with correct version)

### LINKED ISSUE #6596: Duplicates in Open Type Dialog (maybe reason identified: empty.index not understood anymore by eclipse
While debugging to understand the [long term issue](https://github.com/bndtools/bnd/issues/5250#issuecomment-1136317708) of duplicates in the Open Type Dialog I may have found a reason. 

## Problem

In Bndtools when using the Open Type Dialog and you search for something, then I almost always get duplicate search results.
1. one from the actual source of my project
2. one from the project's generated .jar file in the `generated` folder

Image

The 1st one is what I want. I don't want the second one, because if I open it, I land in the .class file and not the .java file. 

## Debugging

I debugged because I discovered this code:

https://github.com/bndtools/bnd/blob/0b6dca338ee5dceba4f1b839632c12a857597259/bndtools.builder/src/org/bndtools/builder/classpath/BndContainerInitializer.java#L409-L415

which defines an `empty.index` (this EMPTY_INDEX is a URL reference to: `platform:/plugin/bndtools.builder/org/bndtools/builder/classpath/empty.index`) which is set for the `.jar` file in the generated folder, with exactly the purpose that Eclipse should not show  the generated jar's content in the Open Type dialog.

But I seems Eclipse cannot understand this index file anymore, and rebuilds the index of the jar 
-> which is probably the reason why the duplicates show up. 

Image

Image

It seems it breaks here where it is expecting a version string 1.131 inside the `empty.index` file (at least in Eclipse 2022-09).

Image

But `empty.index` contains 1.127:

Image

if I change empty.index to contain the expected 1.131 then it works and the Open Type Dialog no longer shows duplicate entries.

Image

**But:** in upcoming Eclipse 2025-06 it is even 1.134

Image

## Conclusion

It seems this expected Signature version string may change with each Eclipse version. 
So the question is how we could provide an `empty.index` which always works. 

Or we need to find an alternative way to achieve something similar.

chrisrueger:
Closes #6596

This removes the duplicate search results (types from the .jar files the `/generated/` folder)
It does it by 

- writing an `empty.index` file (as before) but now dynamically with correct index-signature-version of the current Eclipse Version

In other words: it still leverages the idea of the `empty.index` file. But this mechanism was broken, because the Version in that file changes with Eclipse versions. Now we create a correct `empty.index` ourselves (instead of a hard-coded one) and write it in the workspace `.metadata`

image

This file is created dynamically and always contains the eclipse index version signature of your current eclipse instance. 

image

## Important

Make sure you manually remove the `OpenTypeHistory.xml`  in e.g. `/MyEclipseWorkspace/.metadata/.plugins/org.eclipse.jdt.ui/OpenTypeHistory.xml` when testing this. 

It took me ages to realize that Open Type dialog was showing me (wrong) results from the history, even though this fix is in place. In other words: If you still see results from the `.jar` files `/generated/` folder, make sure you don't see results from the history. Btw: it is also possible to [delete those history entries in the UI manually](https://stackoverflow.com/a/30503240).


chrisrueger:
I'll give this a go to see how it works in the wild. @pkriens

PR #6600, opened 2025-05-13
create per-project empty.index

chrisrueger:
Another attempt to fix https://github.com/bndtools/bnd/issues/6596

Successor of https://github.com/bndtools/bnd/pull/6599

During debugging I found why it is still not working. I found the index is build parallel in multiple threads and if there is a single shared empty.index file the threads have a problem with this (`org.eclipse.jdt.internal.core.search.indexing.AddJarFileToIndex.execute(IProgressMonitor)`), close the empty.index and start an (potentially in memory) reindexing of the actual .jar (which is what we want to prevent). So giving each project its own empty-projectname.index we don't run into race conditions of a shared empty.index


chrisrueger:
Yeah, I think I finally got it. No more duplicates in Open Type dialog 🥳 

Remember to do the following cleanup, because Eclipse caches stuff in various places. I needed to do:

- empty the folder `/eclipse-workspace/.metadata/.plugins/bndtools.builder`
- remove all .index and .txt files in `/eclipse-workspace/.metadata/.plugins/org.eclipse.jdt.core`
- remove the `/eclipse-workspace/.metadata/.plugins/org.eclipse.jdt.ui/OpenTypeHistory.xml`
- restart Eclipse

PR #6603, opened 2025-05-15
Update to latest guava 33.4.8

chrisrueger:
@stbischof pointed out 3 CVEs in https://mvnrepository.com/artifact/biz.aQute.bnd/biz.aQute.bnd.reporter/7.1.0 because of an old Guava version. 

This is an attempt upgrade to the latest version. 

This seems possible.
The only case where Guava 19.0 is still needed is for `biz.aQute.bndlib.comm.tests` which uses no-longer maintained `littleshoot proxy` (https://mvnrepository.com/artifact/org.littleshoot/littleproxy)


stbischof:
owesome. thx for the fix

PR #6607, opened 2025-05-17
Move org.bndtools.build.api to api bundle

laeubi:
Today I surprisingly noticed that `org.bndtools.build.api` is part of [bndtools.core](https://github.com/bndtools/bnd/tree/master/bndtools.core) while it would better fit into [bndtools.api](https://github.com/bndtools/bnd/tree/master/bndtools.api). This currently ties e.g. [org.bndtools.launch](https://github.com/bndtools/bnd/tree/master/org.bndtools.launch) to bndtools.core while it is mostly independent from that.

This simply moves the package to the API bundle.


laeubi:
There are basically two issues:

1. The package name is rather odd and do not match other names
2. It contains some default implementation that uses an extension point

I have adjusted the code but it would actually be better to:

1. Align package name
2. Only put API in the new package keep implementation in `bndtool.core`

this would be a "breaking" change in bndtools even though I'm not sure how much this public API is really use _outside_ bndtools ...

laeubi:
I now created an alternative approach here:

- https://github.com/bndtools/bnd/pull/6608

PR #6608, opened 2025-05-17
Move the BuildListener to the API

laeubi:
Currently there is a need to directly reference bndtools.core to participate in build updates.

This now moves the build listener to the API to avoid this situation.


synesty-cr:
I have no objections. 
Anybody else?

chrisrueger:
I have no objections. 
Anybody else?

Issue #6609, opened 2025-05-18
Show Setups in the CONTRIBUTING.md file


laeubi:
Here we have some setups to build an bnd-ide (even though it claims to be 7.1 it works for 7.2-SNAPSHOT of course)

https://bndtools.org/bndtools.p2.repo/

It would be actually more usefull to place it in [CONTRIBUTING.md](https://github.com/bndtools/bnd/blob/master/CONTRIBUTING.md) where it can be found much easier, e.g. Eclipse do it like this:

https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#creating-an-eclipse-development-environment

chrisrueger:
Closed via #6610

We can reopen for more finetuning in the future.

PR #6610, opened 2025-05-18
move dev-setup from bndtools website to CONTRIBUTING

chrisrueger:
Related to #6609 
This adds the content of the current page https://bndtools.org/development.html to github CONTRIBUTING.md because this is a better place.

I will forward the old URL to the CONTRIBUTING.md

Advantage:
- Developer setup instructions in a single place
- no jumping between websites


chrisrueger:
@laeubi I did not include the Oomph Links directly yet, just still link to https://bndtools.org/bndtools.p2.repo/ because this page is still in the making. We should coordinate with @peterkir which link we could add directly as a shortcut to the CONTRIBUTING.md

But at least now there is a central place and I moved the section for the development setups at the top. 
I think we can build on this.

PR #6612, opened 2025-05-20
allow excluding repositories from Project -buildpath via tags

chrisrueger:
It is related to a forum post some time ago which was the base for [Repo-Tagging](https://bnd.bndtools.org/chapters/870-plugins.html#tagging-of-repository-plugins):

- https://bnd.discourse.group/t/baseline-repo-exclude-from-the-build-path/391/2

e.g. only consider "resolve" repos for Project. getRepositories()

This fixes a cornerish-case problem with compilation errors when you temporarily add a new Repository to  your workspace which is purely for lookup/research/browsing purposes and should neither be considered for resolution nor to the buildpath or Eclipse Classpath. It should just "be there" but do nothing.

It fixes a problem I had when I temporarily added an Eclipse Repo for the research purpose above and suddenly got compile errors because the new repo contained a 2.x version of slf4j.simple while the code expected to get an 1.7.x via 

`-buildpath: slf4j.simple;version=latest`

But maybe this is a practical feature, not only for this corner case.

## Example

Let's say I temporarily want to add this repo for research & lookup purposes.

```
-plugin.1.EclipseMilestone:\
	aQute.bnd.repository.p2.provider.P2Repository; \
	 	url = https://download.eclipse.org/eclipse/updates/4.36-I-builds/I20250515-1800/; \
	 	name = "Eclipse-4_36";\
	 	tags = "noResolve"
```

adding 
```
tags = "noResolve"
```

allows me to exclude this repository from any `-buildpath`. 

image

This has two positive effects:
- also the Eclipse classpath ignores this repo (because currently bnd adds -buildpath of all projects to build the Eclipse classpath)
- the build works in case it got broken by the new repository (see slf4j.simple case above)

## TODO

This is experimental and only for discussion. 
What are drawbacks of this approach? Is it useful? Or should I just not do this or use a narrower version-range instead of 'latest' (`-buildpath: slf4j.simple;version=latest`)

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
@@ -105,7 +105,6 @@ import aQute.bnd.service.action.NamedAction;
 import aQute.bnd.service.export.Exporter;
 import aQute.bnd.service.release.ReleaseBracketingPlugin;
 import aQute.bnd.service.specifications.RunSpecification;
-import aQute.bnd.service.tags.Tags;
 import aQute.bnd.stream.MapStream;
 import aQute.bnd.version.Version;
 import aQute.bnd.version.VersionRange;
@@ -638,6 +637,8 @@ public class Project extends Processor {
 			decorator.decorate(bundles, true);
 		}
 
+		List<RepositoryPlugin> repos = getRepositories(repoTagsBySource(source));
+
 		List<Container> result = new ArrayList<>();
 		try {
 			for (Entry<String, Attrs> entry : bundles.entrySet()) {
@@ -650,12 +651,12 @@ public class Project extends Processor {
 				boolean triedGetBundle = false;
 
 				if (bsn.indexOf('*') >= 0) {
-					return getBundlesWildcard(bsn, versionRange, strategyx, attrs);
+					return getBundlesWildcard(bsn, versionRange, strategyx, attrs, repos);
 				}
 
 				if (versionRange != null) {
 					if (versionRange.equals(VERSION_ATTR_LATEST) || versionRange.equals(VERSION_ATTR_SNAPSHOT)) {
-						found = getBundle(bsn, versionRange, strategyx, attrs);
+						found = getBundle(bsn, versionRange, strategyx, attrs, repos);
 						triedGetBundle = true;
 					}
 				}
@@ -694,7 +695,7 @@ public class Project extends Processor {
 							found = new Container(this, bsn, "file", Container.TYPE.EXTERNAL, f, error, attrs, null);
 						}
 					} else if (!triedGetBundle) {
-						found = getBundle(bsn, versionRange, strategyx, attrs);
+						found = getBundle(bsn, versionRange, strategyx, attrs, repos);
 					}
 				}
 
@@ -728,6 +729,25 @@ public class Project extends Processor {
 		return result;
 	}
 
+	private String[] repoTagsBySource(String source) {
+
+		if (source == null) {
+			return null;
+		}
+
+		if (Constants.BUILDPATH.equals(source)) {
+			return new String[] {
+				Constants.REPOTAGS_COMPILE
+			};
+		} else if (Constants.TESTPATH.equals(source)) {
+			return new String[] {
+				Constants.REPOTAGS_COMPILE, Constants.REPOTAGS_TEST
+			};
+		}
+		// TODO how to determine tag "debug"
+		return null;
+	}
+
 	/**
 	 * Just calls a new method with a default parm.
 	 *
@@ -750,6 +770,13 @@ public class Project extends Processor {
 	 */
 	public List<Container> getBundlesWildcard(String bsnPattern, String range, Strategy strategyx,
 		Map<String, String> attrs) throws Exception {
+
+		List<RepositoryPlugin> plugins = getRepositories();
+		return getBundlesWildcard(bsnPattern, range, strategyx, attrs, plugins);
+	}
+
+	private List<Container> getBundlesWildcard(String bsnPattern, String range, Strategy strategyx,
+		Map<String, String> attrs, List<RepositoryPlugin> repos) throws Exception {
 
 		if (VERSION_ATTR_SNAPSHOT.equals(range) || VERSION_ATTR_PROJECT.equals(range))
 			return Collections.singletonList(new Container(this, bsnPattern, range, TYPE.ERROR, null,
@@ -774,8 +801,7 @@ public class Project extends Processor {
 
 		SortedMap<String, Pair<Version, RepositoryPlugin>> providerMap = new TreeMap<>();
 
-		List<RepositoryPlugin> plugins = getRepositories();
-		for (RepositoryPlugin plugin : plugins) {
+		for (RepositoryPlugin plugin : repos) {
 
 			if (repoFilter != null && !repoFilter.match(plugin))
 				continue;
@@ -1256,6 +1282,14 @@ public class Project extends Processor {
 	public Container getBundle(String bsn, String range, Strategy strategy, Map<String, String> attrs)
 		throws Exception {
 
+		List<RepositoryPlugin> plugins = getRepositories();
+		return getBundle(bsn, range, strategy, attrs, plugins);
+
+	}
+
+	private Container getBundle(String bsn, String range, Strategy strategy, Map<String, String> attrs,
+		List<RepositoryPlugin> repos) throws Exception {
+
 		if (range == null)
 			range = "0";
 		if (attrs == null) {
@@ -1281,7 +1315,7 @@ public class Project extends Processor {
 		useStrategy = overrideStrategy(attrs, useStrategy);
 		RepoFilter repoFilter = parseRepoFilter(attrs);
 
-		List<RepositoryPlugin> plugins = getRepositories();
+
 
 		if (useStrategy == Strategy.EXACT) {
 			if (!Verifier.isVersion(range))
@@ -1291,7 +1325,8 @@ public class Project extends Processor {
 			// For an exact range we just iterate over the repos
 			// and return the first we find.
 			Version version = new Version(range);
-			for (RepositoryPlugin plugin : plugins) {
+
+			for (RepositoryPlugin plugin : repos) {
 				DownloadBlocker blocker = new DownloadBlocker(this);
 				File result = plugin.get(bsn, version, attrs, blocker);
 				if (result != null)
@@ -1307,7 +1342,8 @@ public class Project extends Processor {
 			// multiple repos we take the first
 
 			SortedMap<Version, RepositoryPlugin> versions = new TreeMap<>();
-			for (RepositoryPlugin plugin : plugins) {
+
+			for (RepositoryPlugin plugin : repos) {
 
 				if (repoFilter != null && !repoFilter.match(plugin))
 					continue;
@@ -1385,8 +1421,7 @@ public class Project extends Processor {
 		//
 
 		return new Container(this, bsn, range, Container.TYPE.ERROR, null,
-			bsn + ";version=" + range + " Not found in " + plugins, attrs, null);
-
+			bsn + ";version=" + range + " Not found in " + repos, attrs, null);
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


biz.aQute.bndlib/src/aQute/bnd/build/Project.java
```diff
@@ -1395,7 +1398,19 @@ public Container getBundle(String bsn, String range, Strategy strategy, Map<Stri
...
 	public List<RepositoryPlugin> getRepositories() {
-		return workspace.getRepositories();
+		// if (tags == null || tags.includesAny(Constants.REPOTAGS_RESOLVE)) {
+
+		return workspace.getRepositories()
+			.stream()
+			.filter(repo -> {
+				Tags tags = repo.getTags();
+
+				if (tags == null || tags.includesAny(Constants.REPOTAGS_RESOLVE)) {
```
laeubi, line 1408: Should it not be NO_RESOLVE ?

laeubi:
> It should just "be there" but do nothing.

I think it would be more useful to allow adding repositories to individual (bndrun) files, the whole problem seem to stem from the fact that repositories need to be added on the workspace level.

biz.aQute.bndlib/src/aQute/bnd/build/Project.java
chrisrueger: I think it is correct as it is. My thinking was to use the same logic which is used for Resolving 

https://github.com/bndtools/bnd/commit/2688cb56f84db9e1619a7af6a4c2316fb39cafd6#diff-096900d1f971d5ff7606d028e316560c8e89c25031bb473a493b776642f7f347R34 called by 

https://github.com/chrisrueger/bnd/blob/180a01a327d1ca3b001206ace8d024dd3994cd7a/biz.aQute.resolve/src/biz/aQute/resolve/BndrunResolveContext.java#L410

In other words: The repositories a Project should "care" about are the same repositories which are considered for resolving by `BndrunResolveContext.java`

chrisrueger:
> > It should just "be there" but do nothing.
> 
> I think it would be more useful to allow adding repositories to individual (bndrun) files, the whole problem seem to stem from the fact that repositories need to be added on the workspace level.

Maybe that would solve a different problem. 

The change I added is on the Project and the project gets a list of Repositories to do its work for `-buildpath` and what not. The project itself and bndrun are at this point not related (bndrun comes into play later).

So the current discrepancy (in my opinion) is that:
- `-buildpath` considers ALL Repos in the Workspace
- .bndrun resolving can look at only a subset (because of the tagging and considering repos with either no tags (Backwards compatibility or the tag "resolve")

My idea is to have both use the same logic.

laeubi:
The classpath is controlled completely by the user and nothing is resolved (as far as I know) it actually _look up_ the bundles, therefore only using resolve repositories might be confusing. So I would argue that if you don't want a higher version then simply adjust the version range in your classpath instead of try to "ignore" a source of bundles.

Bndruns are working differently there is a _resolve phase_ (either explicitly or implicitly) where only the metadata of respositories is used to collect all transitive required things. These are then used in the _execution phase_ where the actual bundles are fetched (from the repositories).

chrisrueger:
> The classpath is controlled completely by the user and nothing is resolved (as far as I know) it actually _look up_ the bundles,

What does "controlled by the user" mean? How can I control it? 

Let me try again: 

The effect I saw after adding the additional Repository was that:
- a command line build failed
- and Eclipse showed also compile errors

Because the new repo made a higher version available which was chosen by the -buildpath because of the `latest` (classifier). So far, so correct.

I debugged a bit and saw that the ProjectBuilder.java is calling `Project.getBuildPath()`. 
image

Also BndContainerInitializer (the Eclipse part) does it.

image


The current behavior is absolutly correct ( am not saying that it is wrong). And you are right, I could solve it by not using "latest" and use a stricter version range. 

BUT: I would like to "control" it in a way, that I can exclude a Repository from `-buildpath` (as also requested in the [Forum](https://bnd.discourse.group/t/baseline-repo-exclude-from-the-build-path/391/2)) and also from resolver stuff.
I just want to have the control-knob at a different place. 

Basically the same way I can control it in .bndrun files.
In .bndrun I can specify either:
- a list of repos via `-runrepo`
- or provide no `-runrepo` which causes to use all Repositories without tags or the tag "resolve".

If the confusion is about the tag name "resolve" although I am talking about `-buildpath` then yes, we could also think about introducing a "buildpath" repo-tag which is only used in the context to control it for buildpath.

chrisrueger:
Thanks @pkriens , and welcome back :)

> I think when we worked on the tags it was implied that we would also apply them to the buildpath.

I just realized that recently when I re-read the forum post in the context of this PR here. I completely forgot about it when we worked on tags. I was so focused on .bndrun at that time. 

> I think a special tag for this would be much better than reusing the resolve tag. I.e. the resolve tag is about runtime dependencies and the buildpath _should_ be about API compile time dependencies.

Ok makes sense and basically supports my last sentence above ("_we could also think about introducing a "buildpath" repo-tag which is only used in the context to control it for buildpath._")

Ok, I will add a new tag. 
Is `buildpath` for the tagname ok? What about `-testpath`? 
Or would `compile` be better?

chrisrueger:
@pkriens 
If we introduce a new tag for compilation e.g. `compile` and we already have `resolve` I think we might have  created a 7.1.0 backwards compatibility problem situation.

Example:

- If somebody has already tagged a repo with something e.g. `resolve` (we have done it)

Now if we filter for the existence of tag with

`if (tags == null || tags.includesAny(Constants.REPOTAGS_COMPILE))`

then the repo which already has a tag, but definitely does NOT have the new 'compile' tag. 
So suddenly builds would fail in 7.2.0

I wonder how the filter condition should look like?
Do we maybe need to the opposite? a `nocompile` tag?

Maybe this is what @laeubi was thinking with his comment:

> Should it not be NO_RESOLVE ?

chrisrueger:
I added back the positive tag check "compile" (in addition to negative "nocompile" precedence ) in e467b0b
I solved my [backwards compatibility dilemma](https://github.com/bndtools/bnd/pull/6612#issuecomment-2898763421) by now also providing TRANSITION phase:

- For one release cycle in 7.2.0 we silently promote every repository that has `resolve` but not `compile` into the new `compile` when the caller asks for `compile`.

- Until version 7.1.x having the tag list `resolve` implied visibility in
both the resolve and the compile phase.

- Starting with 7.2.0 that implication is deprecated.
  - 7.2.0 still honours it for backward compatibility and emits a warning.

- 7.3 will no longer honour it.

peterkir:
I have many discussions about compile time and runtime dependencies with my colleaques and would appreciate having an more explicit way of making it visible on a repository base.
For the current tag usage I fould the following documentation https://bnd.bndtools.org/chapters/870-plugins.html#tagging-of-repository-plugins
Could you extend this with and example for using a repo only for compile and another only for runtime dependencies.
I was not able to quickly grab the description above, especially as it is also mixing the intent/api and the implementation with consideration of the existing historical implementation.

chrisrueger:
> Could you extend this with and example for using a repo only for compile and another only for runtime dependencies.

So with this PR here it would be possible to specify:

- RepoA:
  - no tags (e.g. existing repo)

Meaning: the backwards-compatibility default repo is consider for all phases (resolve and compile)

- RepoB:
  - tags='compile'

Meaning: repo is only considered for `-buildpath`, but resolution

- RepoC:
  - tags='resolve'

Meaning (actual) repo is only considered for resolution, but not -buildpath
But: in 7.2.0 I added a transition hack so that this a repo with 'resolve' is considered also for 'compile'. This is to avoid breaking current builds. It will be removed in 7.3.0 where the actual meaning would apply. This needs to be communicated in release notes.

- RepoD:
  - tags='resolve,compile'

Meaning: repo is only considered for both resolution and -buildpath

Yesterday I added also no+[tag] negative tags.

- RepoE:
  - tags='resolve,nocompile'

Meaning (actual) repo is only considered for resolution, but not -buildpath.
This is e.g. needed in the 7.2.0 Transition phase mentioned above to exclude a repo from -buildpath but keep it in resolution

I tried added that also in the tests https://github.com/bndtools/bnd/pull/6612/commits/977955e4bd5602357da2caaab313a0dd64eb9a21

Let's discuss

chrisrueger:
Summary from copilot about my current state:


```java
Tags activeTags = Tags.parse(getProperty(Constants.ACTIVETAGS), 
                              Tags.of(Constants.REPOTAGS_RESOLVE));

return data.repositories.get()
    .stream()
    .filter(repo -> {
        Tags repotags = repo.getTags();
        
        if (repotags == null || !activeTags.includesAny(tags) || repotags.includesAny(tags)) {
            return true;
        }
        return false;
    })
    .toList();
```

## Corrected Understanding of ACTIVETAGS

You're right — `ACTIVETAGS` is an **expert feature for backward compatibility control** that determines **which tag categories are active in the workspace**. Let me re-interpret this:

### The Logic (Reinterpreted):

```
activeTags = [resolve]  (default from -activetags or default to "resolve")

For EACH repository, include it if:
  1. Repository has NO tags (repotags == null)  → include (legacy)
  2. OR active tags DON'T include requested tags (!activeTags.includesAny(tags))
     → include (expert override: if activeTags doesn't cover this phase, accept all)
  3. OR repo's tags include requested tags (repotags.includesAny(tags))
     → include (repo explicitly tagged for this phase)
```

### Use Case Example:

**Scenario**: An expert user runs on an older JVM where `compile` tag doesn't make sense, or they want to disable tagging entirely:

```properties
# cnf/build.bnd
-activetags: resolve

# This way, even if you request "compile" tags,
# since "compile" is NOT in activeTags,
# condition 2 triggers: !activeTags.includesAny(["compile"]) = true
# So ALL repositories are returned (no filtering happens)
```

### Another Scenario - Multi-phase Activation:

```properties
# cnf/build.bnd
-activetags: resolve,compile,test

# Now all three phases can be filtered independently
# -buildpath can request ["compile"]
# -testpath can request ["compile", "test"]
# Both will filter properly
```

---

## Revised Assessment

**With this interpretation, the logic makes much more sense:**

✅ **`activeTags` acts as a "feature toggle" for the entire tagging system**
- If a caller requests tags that aren't in `activeTags`, the filter short-circuits and returns all repos
- This preserves backward compatibility: users who don't set `-activetags` still get the default `"resolve"` behavior
- Expert users can enable additional phases like `compile` and `test`

✅ **The condition structure now makes sense:**
```java
!activeTags.includesAny(tags) || repotags.includesAny(tags)
```
Means: **"Include if the phase isn't active OR if the repo explicitly tags itself for this phase"**

---

## Remaining Issues (But Less Critical Now)

However, there are still **actual bugs** in the implementation:

### Bug 1: Negative Tags Never Actually Checked
The tests expect this to work:
```java
Tags.of("resolve", "nocompile").includesAny("compile")  // should return false
```

But `Tags.includesAny()` **never checks for the "no" prefix**. The test passes because `"nocompile"` doesn't glob-match `"compile"`, but this is accidental, not intentional:

```java
public boolean includesAny(String... tags) {
    if (isEmpty()) {
        return true;  // OK
    }
    
    for (String tagGlob : tags) {
        if (matchesAny(new Glob(tagGlob))) {  // "nocompile" doesn't match glob "compile"
            return true;
        }
    }
    return false;  // So returns false - but by accident!
}
```

**Problem**: If someone uses glob patterns like `"comp*"`, then `"nocompile"` **would** accidentally match! This is dangerous.

### Bug 2: Documentation Claims Negative Tag Logic
The PR's documentation in `Tags.java` **documents** negative tag semantics:
```java
* 2. **Negative tag overrides** – if the entity contains a literal tag
*    {@code "no" + tag} it is <b>excluded</b> from that phase
```

But the code **doesn't implement it**. This is a documentation-code mismatch.

---

## Recommendation

With your clarification on `activeTags`:

1. ✅ The `activeTags` logic is **actually correct** (backward compatibility feature)
2. ⚠️ **BUT** negative tags need proper implementation to match the documented behavior
3. ✅ The test `testTags()` works, but only by accident (glob non-matching)

**Before merge**, the PR should:
- Properly implement negative tag checking (don't rely on glob accidents)
- Or remove the negative tag documentation until it's actually implemented
- Update the Constants.java comment to explain `-activetags` is for expert use

Does this revised understanding align with what you intended?

chrisrueger:
This PR will be superceeded by the new approach suggested in https://github.com/bndtools/bnd/issues/7288

PR #6621, opened 2025-06-02
Eclipse: Support source attachments for sub-bundles + ${repo} refs

chrisrueger:
This makes the Eclipse Source Attachments work also for sub-bundles. This is especially useful for  pure wrapper bundles which contains only sub-bundles to wrap non-osgi jars and all projects in the workspace only have a build-path reference to the wrapper bundle, but not to the wrapped actual jars (which have the `-sources.jar` in `~/.m2` repo).

Also improved how the version is used, but also using `Container.getVersion()` in case `Container.getAttributes().get(Constants.VERSION_ATTRIBUTE)` is `null`. This makes it work for cases where a ${repo} reference specifies an explicit version (not latest) e.g.` -includeresource: ${repo;org.apache.lucene:lucene-analyzers-common;[4.0.0,4.0.1)};lib:=true`

## Example of a wrapper bundle

```
# bnd.bnd

-resourceonly: true
-sub: *.bnd
```

And there are a bundle of subbundles wrapping the actual wrapped non-osgi jars:

image

**Before:**

image

Using Type Search Dialog (Cmd + Shift + T) for a class inside a wrapper sub-bundles showed this:

image

**After:**

image

As you see the **Bnd Bundle Path** now also contains references to the actual contained dependencies. 
And doing the same  Type Search Dialog (Cmd + Shift + T) I go to the class which is referenced in the project and correctly source attached:

image

image

## TODO

- [ ] Update description on https://bndtools.org/manual/packageexplorer.html#icons

image

e.g. we should mention that also `${repo}` references are taken into account

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/RepoCollector.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/RepoCollector.java
@@ -8,6 +8,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 
+import aQute.bnd.build.Container.TYPE;
 import aQute.bnd.header.Attrs;
 import aQute.bnd.header.Parameters;
 import aQute.bnd.osgi.Constants;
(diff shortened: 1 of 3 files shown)


chrisrueger:
One small issue i notice when using that locally for testing:

- Cmd + Shift + T -> works, it opens source file in my Wrapper bundle
- Cmd + Left Click directly in the source code -> fails, opens the .class file in the reference the Wrapper .jar of the current project , which obviously does not have source attached, because the source attachments live in the Wrapper Bundle's "Bnd Bundle Path" container.

chrisrueger:
I was working with it from local build over the last few days and discovered no issues. It feels good that **Open Type Dialog** (Cmd + Shift + T) is now opening source code for classes which did not have source code attached before , although it was there. This was especially the case in pure wrapper bundles which use sub-bundles.

Thus I merged to master and hope to get more real world feedback from others.

UPDATE: I investigated a bit into this issue. I did not find a solution. I think eclipse behaves this way and it is also not really wrong. It just opens the class in the jar which is a direct dependency of the current project (bundle) you are in which in my case is the jar of a wrapper sub-bundle. And obviously this .jar has no direct source attachment, because it is a conglomerate of multiple dependencies, which currently live only as a Bnd Build Path entry of the wrapper bundle. 

And I cannot add multiple source attachement to a single jar. So maybe this is not really solvable. Anyway this PR is useful on its own already and closes a gap.

PR #6629, opened 2025-06-13
try to make test predictable / less flaky

chrisrueger:
for some reason the windows build failed because the order of runbundles was different. so I try sortbynameversion instead of leastdependenciesfirst which makes the sorting always predictable

This is about the failing windows build e.g. here https://github.com/bndtools/bnd/actions/runs/15640349278/job/44073573245

```
RunResolutionTest > testStartLevelDecoration(SoftAssertions) FAILED
    org.assertj.core.error.AssertJMultipleFailuresError: 
    Multiple Failures (2 failures)
    -- failure 1 --
    Expecting actual:
      ["org.apache.felix.gogo.runtime;version='[0.10.0,0.10.1)';startlevel=100",
        "org.apache.felix.gogo.runtime;version='[0.12.0,0.12.1)';startlevel=110",
        "osgi.enroute.junit.wrapper;version='[4.12.0,4.12.1)';startlevel=120",
        "test.simple;version=snapshot;startlevel=90"]
    to contain exactly in any order:

107 tests completed, 3 failed
      ["org.apache.felix.gogo.runtime;version='[0.12.0,0.12.1)';startlevel=100",
        "org.apache.felix.gogo.runtime;version='[0.10.0,0.10.1)';startlevel=110",
        "osgi.enroute.junit.wrapper;version='[4.12.0,4.12.1)';startlevel=120",
        "test.simple;version=snapshot;startlevel=90"]
    elements not found:
      ["org.apache.felix.gogo.runtime;version='[0.12.0,0.12.1)';startlevel=100",
        "org.apache.felix.gogo.runtime;version='[0.10.0,0.10.1)';startlevel=110"]
    and elements not expected:
      ["org.apache.felix.gogo.runtime;version='[0.10.0,0.10.1)';startlevel=100",
        "org.apache.felix.gogo.runtime;version='[0.12.0,0.12.1)';startlevel=110"]
    at RunResolutionTest.testStartLevelDecoration(RunResolutionTest.java:532)
    -- failure 2 --
    expected: "org.apache.felix.gogo.runtime;version="[0.12.0,0.12.1)";startlevel=100,org.apache.felix.gogo.runtime;version="[0.10.0,0.10.1)";startlevel=110,osgi.enroute.junit.wrapper;version="[4.12.0,4.12.1)";startlevel=120,test.simple;version=snapshot;startlevel=90"
[...]

PR #6630, opened 2025-06-14
just testing if HashMap is the problem

chrisrueger:
and the source of the failing test under windows RunResolutionTest.testStartLevelDecoration(SoftAssertions)

This is just a test for an assumption in https://github.com/bndtools/bnd/pull/6629#issuecomment-2972312884


chrisrueger:
Ok doesn't have any effect. Windows build still fails. Continueing in https://github.com/bndtools/bnd/pull/6629#issuecomment-2972312884

PR #6634, opened 2025-06-18
Use PDE Views (RepoBrowser/Resolution)

chrisrueger:
- Eclipse 4.36 (2025-06) deployes org.eclipse.pde:org.eclipse.pde.bnd.ui:1.2.200 to maven central and includes the views (Repositories View / Resolutions View ) which were migrated from bndtools as part of the bndtools-PDE-colaboration.
- this is the first step now to include the Eclipde PDE views into the bndtools Perspective.
- the existing views from bndtools are still there, but not in the perspective anymore (but can be used independently via Eclipse / Window / Show View. They are marked with "EOL" (End of life) in the part title
- in future releases we can remove them from bndtools completely as development will happen in Eclipse PDE for those parts

### DIFF
--- a/bndtools.core/_plugin.xml
+++ b/bndtools.core/_plugin.xml
@@ -222,13 +222,13 @@
 		</view>
 		<view category="bndtools.viewCategory"
 			class="bndtools.views.resolution.ResolutionView"
-			icon="${icons.capreq}" id="bndtools.impExpView" name="Resolution"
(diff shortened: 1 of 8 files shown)


chrisrueger:
Closing for now because using the PDE views directly was so easy because:
- required JDK21
- required newer SWT stuff because of SVG icon support

In the end I cound't easily get it working. 
I may open another PR where I mark the Bnd views as _deprecated_ and add some hints that the views in current Eclipse versions can / should be used and that the old views might be removed in some future.

PR #6641, opened 2025-06-26
Set TCCL on tester thread when testing

kriegfrj:
This makes sure that the tests work properly even when they are being launched by a different version of junit-platform-launcher than what is being installed in the system-under-test. This fixes #6640.

The first commit also sets up tests for the tester so that it will test against both the lowest version (ie, the compile version) of junit-platform-launcher and against the highest. This will enable us to stay abreast of changes in JUnit Platform by incrementing the "high" version as new versions of JUnit come out, which will allow us to proactively find and fix problems like #6614 rather than reactively.

PR #6649, opened 2025-06-30
[tester] BundleEngine to report failing test engine

### LINKED ISSUE #6646: [tester] BundleEngine silently ignores engine failures
This is a spin-off of #6614 

When working on #6614, I noted a related issue: when `BundleEngine` hands off the discovery request to the sub-engines, it surrounds them with a try/catch that swallows the exception with only minimal logging:

https://github.com/bndtools/bnd/blob/09fae58c0b4e5a42639d84fb2748698d59215f5e/biz.aQute.tester.junit-platform/src/aQute/tester/bundle/engine/discovery/BundleSelectorResolver.java#L594-L602

It would be better if the caught exception were reported by the `BundleEngine` as a `StaticFailureDescriptor` so that the test launching/reporting framework can report the underlying exception to the end user. This would help with debugging. In the case of #6614 it would have made the root cause more obvious instead of the user having to infer it.

kriegfrj:
`BundleEngine` will now report in the test output when a test engine's discovery method throws an exception. See a screenshot for an example of how this looks:

image

Fixes #6646


chrisrueger:
@kriegfrj really appreciate your efforts regarding the testing / junit infrastructure. I hope it makes upgrades to a higher Eclipse version easier. I believe those issues you addressed are exactly the ones I am facing.

kriegfrj:
> @kriegfrj really appreciate your efforts regarding the testing / junit infrastructure.

You're welcome!

> I hope it makes upgrades to a higher Eclipse version easier. I believe those issues you addressed are exactly the ones I am facing.

I believe so too - actually, I think that #6641 will fix most of the issues you were having.

Once I get the next (and hopefully final) PR in place, I'd encourage you to have another go at bumping the baseline version of Eclipse again.

Issue #6642, opened 2025-06-27
Simplify export from bndrun to a java application under folder


scottslewis:
Currently, the way to export a bndrun to a functional java application is:

1) Open the bndrun editor

2) Click on Export button in upper right of bndrun editor

3) Choose  Executable JAR exporter from list (only exporter in my installation

4) On Executable JAR wizard page choose Folder rather than Jar to export

5) Set a location on the file system to export to and click Finish

I suggest that this be made more simple for devs by either:

a) Adding a new property to the bndrun file...e.g. -export-directory=build that when present in the bndrun file automatically clears, exports to <project dir>/build/ folder in a single Export (no UI shown...simply done).

b) Add a new wizard that provides a default of project-dir/build/ (e.g.) and allows the user change/reset it.

c) Some other UI

For a or b, I'm willing to implement this feature.   It will be very simple to do and could be ready in a few days.   I personally prefer a and have already implemented and tested this, but before submitting pr would like to get input from project committers.

chrisrueger:
Sounds useful.  Although I admit I never used the Export button in the .bndrun Editor in Eclipse. We usually only use the gradle export task. 

Few thoughts to  **a)**

So the logic would be:
- if `.bndrun` contains new property e.g. `-exportpath-hint=build` then the following dialog would be skipped completely?

![Image](https://github.com/user-attachments/assets/78a31a10-11a0-47ef-b269-804fb4a87f88)

- What about a small success dialog showing where the exported file has been placed, so that users know "something" has happened. Maybe with a hint, why it happened (e.g. "because `-exportpath-hint=build` in your .bndrun file... remove this property to manually choose the location" ... or something like that. 

- should the path of this property be relative to the workspace, or support an absolute path too? (I somehow have the feeling that it should be relative to the workspace... maybe also for security reasons)

- should this new property only be used in Eclipse UI? Or should / would other parts (bnd cli, gradle, maven...) also use it (as a default if not set explicitly)? 
 
- sounds like it should be mentioned at https://bnd.bndtools.org/chapters/300-launching.html#packaging and also in `aQute.bnd.help.Syntax.java` so that some help-tooltip appears in .bndrun source editor when hovering over the new property

kriegfrj:
> Currently, the way to export a bndrun to a functional java application is:
> 
> 1. Open the bndrun editor
> 2. Click on Export button in upper right of bndrun editor
> 3. Choose  Executable JAR exporter from list (only exporter in my installation
[...]

PR #6656, opened 2025-07-01
Set bndeditor as default for bnd extensions

chrisrueger:
based on discussion https://github.com/bndtools/bnd/issues/6642#issuecomment-3022090707

This is an attempt to fix the problem that newer Eclipse version took over the .bnd extension and opened in an Eclipse editor instead of the bndtools' BndEditor.


chrisrueger:
@peterkir as discussed. I will merge if it builds and then we can install the snapshot to see if it has any effect.

PR #6657, opened 2025-07-02
Custom tooltip with help button in BndEditor source view

chrisrueger:
The goal is to add a link with with a help button to the tooltip shown when you hover over a word in BndEditor.
The help button should link to e.g. a URL in /docs / bnd manual 

Originated from https://github.com/bndtools/bnd/issues/6642#issuecomment-3017862729 

> ... where is the information in the tooltip coming from?
> We could add the link to the documentation there or grep the content from the docs on demand?

Current WIP preview:

image

image

## Auto-detect help url

I added an auto-detection / heuristic for constructing the URL to the manual (because it is a rather predictable pattern). This makes it easier to provide useful help urls for most instructions and headers without manually specifying it. 

For example: 

image

This is an icon pointing to https://bnd.bndtools.org/instructions/export.html although this URL is nowhere specified, but automatically constructed.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/help/Syntax.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/help/Syntax.java
@@ -33,6 +33,7 @@ public class Syntax implements Constants {
 	final Pattern							pattern;
 	final String							values;
 	final Syntax[]							children;
+	final String							helpurl;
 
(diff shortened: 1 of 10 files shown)


chrisrueger:
Feedback appreciated @peterkir @pkriens

PR #6663, opened 2025-07-08
Finetune docs with Copilot

chrisrueger:
This is an experiment to use Github Copilot Agent Mode to improve / enrich some of the "old" `.md` pages for headers, instructions and commands with prose. 
Most of those "old" files had a relevant Java Sourcecode snippet in them. I tried to ask Copilot (Model GPT 4.1) to turn that into text. 

Thus, I focus only on files with source code in them (a few exceptions (mostly accidental) are there, where copilot rewrote some already prose content, which I left, because I thought it might be an improvement)

## Task

Not sure how we go about this. I probably need all of the long time bnd developer's human intelligence to help going through the produced AI text and verify correctness.

After this is done, the last step is to re-generate the documentation which will append all those files under the `_ext` folder to the generated files. I didn't do that already, because this would inflate this PR which would make it harder to see the woods from the trees.


chrisrueger:
@peterkir please wait for https://github.com/bndtools/bnd/pull/6667/files before starting to review. I will #6667  first and then rebase this PR here. 

UPDATE: I rebased. We can continue.

chrisrueger:
It was agreed in the bnd call (@pkriens @peterkir @juergen-albert) that we merge this to master, since it seems better than everything we have so far. 
I added a disclaimer `TODO Needs review - AI Generated content` to each affected file. This allows us to later find those files and go through them one by one and review and correct it.

PR #6674, opened 2025-07-13
WIP bnd build -watch flag for continuous build

chrisrueger:
The goal is to have an IDE-independent live coding workflow known from the javascript / npm world e.g. "vue-cli serve" which starts a dev-webserver and allows for hot code reloading when changes are detected. It should basically do the same what Eclipse "Build automatically" is doing, but without Eclipse. So I can use any text editor VSCode, vim etc.

A workflow could look like:

```
cd mybndworkspace
bnd build --watch & cd mywebapp; bnd run launch.bndrun
```

Ideally even simpler with a new command:

```
cd mybndworkspace
bnd serve
```

This is basically. a combination of the three existing commands:

``
bnd compile
bnd build
bnd run
```


chrisrueger:
Requires https://github.com/bndtools/bnd/pull/6675

chrisrueger:
Closing because this branch moved to bnd main repo in https://github.com/bndtools/bnd/pull/6768

PR #6675, opened 2025-07-14
[CLI] Change bnd build to also compile

chrisrueger:
because `bnd compile` alone just compiles .java files to class files but will fail for workspace bundles which require another bundle (because compile expects the built .jar  of that dependent bundle on the classpath, but the .jar is not there unless bnd build has created the jar file). So kind of a henn-egg problem.

My use case was this:

I'm trying to build my workspace with it (instead of gradle) but somehow. 

I first tried 

```
cd mybndworkspace
```

```
bnd clean
bnd build
```

That has produced `/generated/xxx.jar` files per bundle but the jar files did not contain the .class files of the bundle. 
After thinking about it, because ```build``` expects the classes to be _compiled_ first.

So I went ahead and tried

```
bnd clean
bnd compile
```

Unfortunatelly compile failed as soon as it arived at a bundle which required a previously compiled bundle - because it needs the .jar from it. 

But the jar is created by `bnd build`. 

Henn egg ?

Somehow compile and build per bundle should be done together.

## Conclusion

This PR basically does `compile` + `build` together when you do a 

```
bnd build
```

I could also create another command e.g. `bnd compilebuild`. Not sure what is better.

## Question:

`bnd compile` alone is not useful at the moment to compile a full workspace. 
It only works for very small self contained bundles. I wonder what the idea was behind that.

Or am I doing something wrong?

### DIFF
--- a/biz.aQute.bnd/src/aQute/bnd/main/bnd.java
+++ b/biz.aQute.bnd/src/aQute/bnd/main/bnd.java
@@ -984,9 +984,15 @@ public class bnd extends Processor {
 	@Description("Build a project. This will create the jars defined in the bnd.bnd and sub-builders.")
 	public void _build(final buildoptions opts) throws Exception {
 
-		perProject(opts, p -> p.build(opts.test()));
+		perProject(opts, p -> {
+			p.getGenerate()
+				.generate(opts.force());
+			p.compile(opts.test());
+			p.build(opts.test());
+		});
 	}
 
+	@Description("Compile a project or the workspace. DEPRECATED: This command will be removed in bnd 8.0. Use 'bnd build' for compile and build.")
 	interface CompileOptions extends ProjectWorkspaceOptions {
 
 		@Description("Compile for test")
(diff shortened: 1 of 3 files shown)


chrisrueger:
@pkriens thoughts?

kriegfrj:
I'm sorry, but I've only rarely used the bnd command line tool, and never for building/compiling. So I don't think I have much useful guidance to give.

chrisrueger:
Thanks @kriegfrj , no problem. I am just exploring the CLI currently. It is quite powerful.

I noticed that it theoretically should be able to build a full bnd workspace. But currently it doesn't without this fix. 
And building and running turns out to be really useful to a) be independent of Eclipse, Gradle or Maven and b) that can be a good base for using other IDEs / Texteditors like VSCode. This is what I am currently experimenting with in https://github.com/bndtools/bnd/pull/6674

chrisrueger:
Bnd Call Meeting Notes:
We agreed on:
- modify (fix) the bnd build command to do `generate` + `compile` + `build`
- deprecate compile (because compile alone is not useful because it depends on build, as it expects built .jar files). 

- [x]  TODO  mark `bnd compile` it for removal in 8.0 and in 7.2. output a deprecation warning.

PR #6677, opened 2025-07-14
[CLI] rename createOption.Manifest() to skipmanifest to fix random exception in Commandline.getOptions and swallowed option

chrisrueger:
the new bnd manual generator was randomly throwing a exception, because the `bnd.createOptions` contained two 'manifest' options, one lowercased `manifest()` one Capitalized `Manifest()` and violates the rule in the comment below that the first 3 characters must not be the same (`3+ throw an error) `

https://github.com/bndtools/bnd/blob/6e0b150b13324e425e4d1deedaedbd42bd7f6bed/aQute.libg/src/aQute/lib/getopt/CommandLine.java#L401-L409

Depending on the mood of the jvm in which it returns `Class.getMethods()` sometimes it will get Manifest() first and sometimes the lowercase version. In the map of options only the lowercase letter is used (-m) but with the method of the uppercase method. Clearly random and messed up. 

image

image

Since the user only sees the lowercase `-m` version, it also means that the other one is currently swallowed and not usable.

I renamed it to `skipmanifest`, because I think this createOptions violates what the exception is telling us and is also confusingly named IMO.
And `skipmanifest` seems not wrong.

```
Errors
000: Exception: java.lang.Error: Only commands with lower case first char are acceptable (Manifest)
        at aQute.lib.getopt.CommandLine.getOptions(CommandLine.java:408)
        at aQute.lib.getopt.CommandLine.generateDocumentationForCommand(CommandLine.java:206)
        at aQute.bnd.main.ManualGenerator.generateCommandsFiles(ManualGenerator.java:90)
        at aQute.bnd.main.bnd._generatemanual(bnd.java:1286)
```

Now the command help `bnd help create` reads:

```
NAME
  create                      - Equivalent jar command c[v0mf] command (supports
                                the jar tool's syntax). Will wrap the bundle
                                unless --wrapnot is specified

SYNOPSIS
   create [options] ...


OPTIONS

   [ -b, --bsn <string> ]     - Bundle Symbolic Name for wrap
   [ -c, --cdir <string> ]    - directory
   [ -f, --file <string> ]    - Jar file (f option)
   [ -F, --force ]            - Force write event if there are errors
   [ -m, --manifest <string> ] - Use manifest (m option)
   [ -n, --nocompression ]    - No compression (0 option)
   [ -p, --properties <string> ] - Properties for wrapping
   [ -s, --skipmanifest ]     - No manifest.
   [ -v, --verbose ]          - Verbose (v option)
   [ -V, --version <version> ] - Bundle Version for wrap
   [ -w, --wrap ]             - Wrap
```

### DIFF
--- a/docs/_commands/create.md
+++ b/docs/_commands/create.md
@@ -15,9 +15,10 @@ note: AUTO-GENERATED FILE - DO NOT EDIT. You can add manual content via same fil
 - `[ -c --cdir <string> ]` directory
 - `[ -f --file <string> ]` Jar file (f option)
 - `[ -F --force ]` Force write event if there are errors
-- `[ -m --manifest ]` No manifest (M option)
+- `[ -m --manifest <string> ]` Use manifest (m option)
 - `[ -n --nocompression ]` No compression (0 option)
 - `[ -p --properties <string> ]` Properties for wrapping
+- `[ -s --skipmanifest ]` No manifest.
 - `[ -v --verbose ]` Verbose (v option)
 - `[ -V --version <version> ]` Bundle Version for wrap
 - `[ -w --wrap ]` Wrap
--- a/docs/_heads/target_dir.md
+++ b/docs/_heads/target_dir.md
@@ -11,3 +11,6 @@ note: AUTO-GENERATED FILE - DO NOT EDIT. You can add manual content via same fil
 
 - Pattern: `.*`
 
+<!-- Manual content from: ext/target_dir.md --><br /><br />
+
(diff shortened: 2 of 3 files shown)


chrisrueger:
Alternative to maybe remain 50% backwards compatible:
`mnomanifest` -> also leads to a capitable `-M` option (disadvantage the long form `--mnomanifest` looks strange)

```
 [ -m, --manifest <string> ] - Use manifest (m option)
 [ -M, --mnomanifest ]      - No manifest (M option).
 ```

PR #6700, opened 2025-07-24
Clarify major version 0 with respect to baselining

### LINKED ISSUE #6691: baseline: Clarify special handling of version 0.x for bundle and package version
Currently in https://bnd.bndtools.org/chapters/180-baselining.html it just says:

> the baselining is started with one exception: the bundle version must be more than 1.0.0. If the version is less no baselining is possible, the purpose is to allow the primordial baseline to be established without errors.

This is a bit fuzzy and not quite correct. According to https://github.com/bndtools/bnd/blob/28ee79bf1ee0a68d12c599cceafca679ac1c7958/biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java#L319-L327 it rather seems that baselining will never emit any errors for bundle versions nor package versions (if any of those start with 0.x). However still baselining issues will be emitted (just with a lower severity, i.e. never fail the build).

Also this seems to conflict a bit with the default import range for 0.x versions which is for consumers `[0.x - 1.0)` i.e. all consumers are only compatible with unstable API but implicitly no longer compatible with stable API starting with 1.x (even if that one is actually fully backwards compatible but just stable). Compare with https://semver.org/#spec-item-4.

I am wondering about the best way to make an unstable API (0.x) stable (1.x) without unnecessarily break all consumers of the unstable API as the signature has actually not changes compared to the last unstable version...

kwin:
Add some more links to Maven plugin and semantic versioning specs. This closes #6691

### DIFF
--- a/docs/_chapters/180-baselining.md
+++ b/docs/_chapters/180-baselining.md
@@ -4,15 +4,15 @@ layout: default
 requires: 2.2.0
 ---
 
-Baselining compares a bundle with another bundle, the _baseline_, to find mistakes in the semantic versioning. For example, there is a binary incompatible change in the new bundle but the version has not been bumped. Baselining can be run from the command line (see `bnd help baseline`) or it can be run as part of a project build. 
+Baselining compares a bundle with another bundle, the _baseline_, to find mistakes in the [semantic versioning](https://docs.osgi.org/whitepaper/semantic-versioning/). For example, there is a binary incompatible change in the new bundle but the version has not been bumped. Baselining can be run from the command line (see `bnd help baseline`) or it can be run as part of a project build. 
 
 APIs are compared for backward compatibility using the semantic versioning rules defined in this chapter. Baselining is aware of the `@ConsumerType` and `@ProviderType` rules. Proper versions are calculated and suggested.
 
 ## Setting Up a Project for Baselining
 
-During the build, bnd will check the `-baseline` instruction at the end of the build when the JAR is ready. This instruction is a selector on the symbolic name of the building bundle. If it matches, the baselining is started with one exception: the bundle version must be more than 1.0.0. If the version is less no baselining is possible, the purpose is to allow the primordial baseline to be established without errors.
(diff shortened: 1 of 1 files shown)


chrisrueger:
@kwin Do you want this to be merged already, or wait for https://github.com/bndtools/bnd/pull/6702 ?

kwin:
This is orthogonal and can be merged already.

docs/_chapters/180-baselining.md
```diff
@@ -4,15 +4,15 @@ layout: default
...
 
 ## Setting Up a Project for Baselining
 
-During the build, bnd will check the `-baseline` instruction at the end of the build when the JAR is ready. This instruction is a selector on the symbolic name of the building bundle. If it matches, the baselining is started with one exception: the bundle version must be more than 1.0.0. If the version is less no baselining is possible, the purpose is to allow the primordial baseline to be established without errors.
+During the build, bnd will check the [`-baseline` instruction](../_instructions/baseline.html) at the end of the build when the JAR is ready. This instruction is a selector on the symbolic name of the building bundle. If it matches, the baselining is started with one exception: the bundle/package version must be 1.0.0 or above. If the version is less (i.e. major version being `0`) no baselining is possible, the purpose is to allow the [primordial baseline to be established without errors](https://semver.org/#spec-item-4).
```
laeubi, line 13: I think referring to https://semver.org/#spec-item-4 is wrong here, above the correct specification (https://docs.osgi.org/whitepaper/semantic-versioning/)) is mentioned.

I also think the statement that only a version 1 and above is possible to baseline is wrong. Actually any version (except the default what is `0.0.0`) can be used for baseline compare a bundle, especially the version `0.0.1`

kwin: https://github.com/bndtools/bnd/issues/6701#issuecomment-3139008625

chrisrueger: I think the reason for the `version >= 1` comes from this

https://github.com/bndtools/bnd/blob/28ee79bf1ee0a68d12c599cceafca679ac1c7958/biz.aQute.bndlib/src/aQute/bnd/differ/Baseline.java#L319-L327 as mentioned in the issue #6691

Issue #2452, opened 2018-05-18
bnd-baseline-maven-plugin: Skip baselining if major version is 0?


io7m:
Hello!

See [item 4](https://semver.org/#spec-item-4) of the semantic versioning spec:

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

It would be really great if the baseline plugin followed the _should_ in this part of the spec: If the current major version of the bundle is `0`, the baseline check should not be performed (or should at least not treat failures as an error). If I'm developing version`0.0.1` of a package, and I then try to move up to `0.0.2`, I don't want to be told that I've made a breaking change and must now release `1.0.0`! :smile:

One option is to set `bnd.baseline.skip` in the POM until `1.0.0` is released. The problem with this is that it's easy to forget to do so - it would be very unpleasant to release `1.1.0` only to find that it really should have been `2.0.0` and that the plugin was accidentally left disabled.

bjhargrave:
I think for baselining, both the bundle being baselined and the baseline bundle both must have version >= 1.0. This fix will affect more than maven plugin (also gradle plugin and possible bndlib).

io7m:
Yep, I agree. Requiring both to be >= 1.0 makes sense.

bjhargrave:
@pkriens @rotty3000 @timothyjward What are your thoughts on this? Basically the change would cause baselining to report no changes when either input (older, newer) has a version < 1.0.

That is, if older < 1.0, then it is not a suitable baseline since "real" API does not start until 1.0. If newer < 1.0, then it is not real API and thus there is no point in baselining it.

rotty3000:
It sounds reasonable. Will there be an override? Because this sounds like a good _default policy_ rather than a hard and fast rule (kind of like import range policy, though I'm not quite sure how you'd express it.)

bjhargrave:
Well, if we are to follow semver, then major version 0 is a "playground" outside the space of semantic versioning. So our baseliner should treat it as such. So we can still make a full diff of something with major version 0, any changes detected must not be baselining failures.

bjhargrave:
I also think this fits well with the idea in #2422 which also suggests using major version 0 for wrapped jars whose semantic versioning status is broken or unknown.

rotty3000:
Got it. I'm good with that.

io7m:
Thanks for implementing this. I'm looking forward to 4.1!

timothyjward:
[...]

PR #6702, opened 2025-07-24
Prevent generating import-package version ranges for unstable API

### LINKED ISSUE #6701: Don't generate "import-package" version restrictions for preliminary APIs (i.e. major version being 0)
By default bnd generates version restrictions for `import-package` statements according to the [import version policy](https://bnd.bndtools.org/chapters/170-versioning.html#import-version-policy). However APIs having a major version `0` are [unstable by definition](https://semver.org/#spec-item-4). As in general for unstable API it is impossible to predict which future versions may be compatible or not bnd should in that special case not automatically create version restrictions. However a warning during the build should be emitted that one depends on an unstable version.

This was triggered from #6691.

kwin:
This closes #6701

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
@@ -2070,6 +2070,8 @@ public class Analyzer extends Processor {
 					continue;
 				}
 
+				// flag indicating if importing an unstable API (major version 0)
+				boolean isUnstable = false;
 				if (exportVersion == null) {
 					// TODO Should check if the source is from a bundle.
 					if (importRange != null) {
@@ -2096,9 +2098,16 @@ public class Analyzer extends Processor {
 					}
 					exportVersion = cleanupVersion(exportVersion);
 
-					importRange = applyVersionPolicy(exportVersion, importRange, provider);
+					isUnstable = isUnstableVersion(exportVersion);
+					importRange = applyVersionPolicy(exportVersion, importRange, provider, isUnstable);
 					if (Strings.nonNullOrTrimmedEmpty(importRange)) {
 						importAttributes.put(VERSION_ATTRIBUTE, importRange);
+					} else {
+						if (isUnstable) {
+							warning(
+								"Depending on unstable API %s in version %s, therefore not generating any version range restriction",
+								packageName, exportVersion);
+						}
 					}
 				}
 
@@ -2117,7 +2126,7 @@ public class Analyzer extends Processor {
 
 				String result = importAttributes.get(Constants.VERSION_ATTRIBUTE);
 				if ((result == null || !Verifier.isVersionRange(result))
-					&& complainAboutMissingVersionRange(packageRef, ee)) {
+					&& complainAboutMissingVersionRange(packageRef, ee) && !isUnstable) {
 					noimports.add(packageRef);
 				}
 			} finally {
@@ -2128,8 +2137,22 @@ public class Analyzer extends Processor {
 		}
 
 		if (isPedantic() && noimports.size() != 0) {
-			warning("Imports that lack version ranges: %s", noimports);
+			warning("Imports that lack version ranges due to not being found in any bundle on the -buildpath: %s",
+				noimports);
 		}
+	}
+
+	/**
+	 * Checks if the given version is an unstable version (i.e. has major
+	 * version 0).
+	 *
+	 * @param version
+	 * @return {@code true} if the given version can be parsed and has major
+	 *         version 0, {@code false} otherwise.
+	 */
+	private static boolean isUnstableVersion(String version) {
+		// check major version of exportVersion
+		return Version.isVersion(version) && new Version(version).getMajor() == 0;
 	}
 
 	private EE getEEFromReqsOrHighest() {
@@ -2192,14 +2215,16 @@ public class Analyzer extends Processor {
 		return new Pair<>(regularImports, dynamicImports);
 	}
 
-	String applyVersionPolicy(String exportVersion, String importRange, boolean provider) {
+	String applyVersionPolicy(String exportVersion, String importRange, boolean provider, boolean isUnstable) {
 		try {
 			setProperty(CURRENT_VERSION, exportVersion);
 			if (importRange != null) {
 				importRange = cleanupVersion(importRange);
 				importRange = getReplacer().process(importRange);
 			} else {
-				importRange = getVersionPolicy(provider);
+				if (!isUnstable) {
+					importRange = getVersionPolicy(provider);
+				}
 			}
 		} finally {
 			unsetProperty(CURRENT_VERSION);
--- a/biz.aQute.bndlib.tests/test/test/BuilderTest.java
+++ b/biz.aQute.bndlib.tests/test/test/BuilderTest.java
@@ -601,7 +601,8 @@ public class BuilderTest {
 		p3.addClasspath(new File("jar/osgi.jar"));
 		p3.setExportPackage("test.activator.inherits;version=0.0.0");
 		p3.build();
-		assertTrue(p3.check("Import package test.activator not found in any bundle on the -buildpath."));
+		assertTrue(p3.check("Import package test.activator not found in any bundle on the -buildpath.",
+			"Depending on unstable API test.activator.inherits in version 0.0.0"));
 
 		p3.getJar()
 			.getManifest()
@@ -3733,7 +3734,8 @@ public class BuilderTest {
 			b.setPedantic(true);
 			b.setProperty("Import-Package", "foo.bar");
 			b.build();
-			softly.assertThat(b.check("Imports that lack version ranges: \\[foo.bar\\]",
+			softly.assertThat(b.check(
+				"Imports that lack version ranges due to no being found in any bundle on the -buildpath: \\[foo.bar\\]",
 				"The JAR is empty: The instructions for the JAR named biz.aQute.bndlib.tests did not cause any content to be included, this is likely wrong"))
 				.isTrue();
 			softly.assertThat(b.getImports()
@@ -3768,7 +3770,8 @@ public class BuilderTest {
 				.toString());
 			b.setProperty("Import-Package", "javax.xml.transform.stax");
 			b.build();
-			// there should be NO warning: Imports that lack version ranges:
+			// there should be NO warning: Imports that lack version ranges due
+			// to no being found in any bundle on the -buildpath:
 			// [javax.xml.transform.stax]
 			softly.assertThat(b.check(
 				"The JAR is empty: The instructions for the JAR named biz.aQute.bndlib.tests did not cause any content to be included, this is likely wrong"))
@@ -3808,9 +3811,11 @@ public class BuilderTest {
 				.toString());
 			b.setProperty("Import-Package", "javax.xml.transform.stax");
 			b.build();
-			// there should be NO warning: Imports that lack version ranges:
+			// there should be NO warning: Imports that lack version ranges due
+			// to no being found in any bundle on the -buildpath:
 			// [javax.xml.transform.stax]
-			softly.assertThat(b.check("Imports that lack version ranges: \\[javax.xml.transform.stax\\]",
+			softly.assertThat(b.check(
+				"Imports that lack version ranges due to no being found in any bundle on the -buildpath: \\[javax.xml.transform.stax\\]",
 				"The JAR is empty: The instructions for the JAR named biz.aQute.bndlib.tests did not cause any content to be included, this is likely wrong"))
 				.isTrue();
 			softly.assertThat(b.getImports()
--- a/docs/_chapters/170-versioning.md
+++ b/docs/_chapters/170-versioning.md
@@ -159,6 +159,8 @@ The resulting manifest would look like:
         Import-Package:  org.osgi.service.event; version="[1.1,1.2)", ...
         ...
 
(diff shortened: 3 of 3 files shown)


chrisrueger:
Looks good to me. The only thought I have is wether  this warning should be always emitted (like you have it now) or only with [-pedantic: true](https://bnd.bndtools.org/instructions/pedantic.html).

Furthermore: Where could this behavior be described in the docs. Maybe somewhere [here](https://bnd.bndtools.org/chapters/170-versioning.html#import-version-policy)?

kwin:
There is already some functionality which prints warnings about missing versions on import-package headers: https://github.com/bndtools/bnd/blob/e517467ee73e0cc800a6d4426e8e2f7224e0b6e0/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L2130-L2132 which is given out with `-pedantic: true`. I am gonna rework my patch to not emit any warning and just rely on the existing logic for the warning. However this warning does not indicate why the version range is missing....

kwin:
After having a 2nd thought I would rather prefer to always emit a warn when importing unstable API (not only with pedantic). This is different from the case where you explicitly import a package without a version when bnd cannot find any matching dependency in the classpath. I will just look into preventing the duplicate warning when depending on unstable API. WDYT @chrisrueger

chrisrueger:
I have no strong feelings about this just wanted to bring it up for discussion. I'm fine with the warning always. Maybe let's wait for some more feedback. I mentioned this pr in the bnd slack and @peterkir and @reckart had some comments.

reckart:
How about leaving the range as-is and instead you could add package-level versioning to your exports? That way, if a consumer of your artifact upgrades from 0.x to 1.x, they may still get package-level 0.x compatibility - wouldn't that resolve the issue as well?

Alternatively, you could explicitly export using `*;version=0.0.0` while still in a phase where you do not want to have ranges being generated.

kwin:
I would argue that version ranges are not useful for unstable API (not complying with semantic versioning) in almost all of the cases. Rather generating a version range for unstable APIs should be a manual thing (as this would be special and cannot be derived from semantic versioning rules), but the default should be no restriction.

biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java
```diff
@@ -2117,7 +2126,7 @@ void augmentImports(Packages imports, Packages exports) throws Exception {
 
 				String result = importAttributes.get(Constants.VERSION_ATTRIBUTE);
 				if ((result == null || !Verifier.isVersionRange(result))
-					&& complainAboutMissingVersionRange(packageRef, ee)) {
+					&& complainAboutMissingVersionRange(packageRef, ee) && !isUnstable) {
```
chrisrueger, line 2129: what about passing `isUnstable` to the `complainAboutMissingVersionRange` method? so this method is the one deciding wether to complain or not. Maybe that would be the better place, as it also considers JDK-packages - JDK packages do not have a version which is fine.

kwin: I would like to expose a different explanation for unstable APIs being referenced and for non-versioned or unknown dependencies being referenced.

peterkir:
> I would argue that version ranges are not useful for unstable API (not complying with semantic versioning) in almost all of the cases. Rather generating a version range for unstable APIs should be a manual thing (as this would be special and cannot be derived from semantic versioning rules), but the default should be no restriction.

Why is it not complying with with semantic versioning.
As long as I don't have a public/official consumer, I always use 0.x versions and have a proper tooling with bnd.
I'm starting my APIs with 0.1.0 and use already baselining and API tooling.
This would **break existing workspaces** :disappointed:
So from my point of view I **disapprove this PR**.
You can use 'Export-Package: *:version=' or 'Export-Package: *:version=0.0.0'
If you do not want use the a concrete versioning.

kwin:
This does not change export versions. It will merely no longer generate import version ranges for unstable API. A version range `(0.x, 1.0]` is not really useful as unstable API may break with every non major version increment but must not necessarily break when promoting to 1.x.

> I'm starting my APIs with 0.1.0 and use already baselining

I am really keen to know how that works for you as baselining checks against 0.x packages will never emit violations…

laeubi:
I don't think that there is any mean for "unstable API" in the OSGi specification. Exported packages per se has to comply with the [OSGi semantic versioning rules](https://docs.osgi.org/whitepaper/semantic-versioning/), so there is no exception where it is "allowed" to break it. If you want something like make your consumers aware that you do not plan to follow these rules,  I would recommend to use [@API Guardian](https://github.com/apiguardian-team/apiguardian) (what is supported by bnd already if I remember correctly).

So I think the assumption

> It will merely no longer generate import version ranges for unstable API. A version range (0.x, 1.0] is not really useful as unstable API may break with every non major version increment but must not necessarily break when promoting to 1.x.

is simply wrong and does not match the OSGi specification and what bnd does here is generally valid. The main problem with an export without a version actually is that one _can not define a lower bound_ but this does not mean an upper bound is of no use!

Eclipse makes some use of this by do not version the API of _packages_ but only on the _bundle_ level in wich case one then need to require the bundle with a suitable range (what is a much stronger requirement).

chrisrueger:
I digged a bit in git to find out where the bnd treatment of `versions <= 1` comes. I found:

- https://github.com/bndtools/bnd/issues/2452
- https://github.com/bndtools/bnd/commit/1cc175bcc344834f3e4dbea3f0f5000c4db0e27f

laeubi:
> This does not change export versions. It will merely no longer generate import version ranges for unstable API. A version range (0.x, 1.0] is not really useful **as unstable API may break with every non major version increment** but must not necessarily break when promoting to 1.x.

If you are really concerned for this I think the solution is not to use _no version range_ but instead bnd must generate a _provider range_ that would be `(0.x, 0.x+1]`.

peterkir:
> I digged a bit in git to find out where the bnd treatment of `versions <= 1` comes. I found:
> 
> * [bnd-baseline-maven-plugin: Skip baselining if major version is 0? #2452](https://github.com/bndtools/bnd/issues/2452)
> * [1cc175b](https://github.com/bndtools/bnd/commit/1cc175bcc344834f3e4dbea3f0f5000c4db0e27f)

@kwin You are right with baselining. I was refering to the release process of bundles, which is working for 0.x bundle versions. Even though from my point of view this behaviour should not spill into import/export and consumer/provider functionality. From my point of view it would be nice to optionally activate 0.x baselining.
Maybe I'm stubborn on this, but coming from linux development I hesitate to put the mature 1.x version when starting with something. I find it very funny to see versions like guava (which is currently 33) or even worse chrome 138. This is ridiculous from my point of view.
And as discussed previously there are options available to tweak provider exports and consumer imports with bnd in almost any flavour. I'm against merging and would propose closing this PR.

kwin:
I didn't see all this discussion coming, but here we are now. Still I have problems with the way how you argue.

My point is:
- baselining and import version ranges should consider the same semantic versioning rules
- baselining treats 0.x specially meaning it will ignore backwards incompatible changes without major version increment
- import version range calculation does the opposite: i.e. it will assume 0.x sticks to the same rules as every other major version

We should fix this inconsistency one way or another.

peterkir:
Total agreement and the inconsistency should be fixed.
I vote for fixing baselining to also work on 0.x packages.

chrisrueger:
> I vote for fixing baselining to also work on 0.x packages.

Should this be configurable? (keeping the current behavior as default?)

We should digest the original issue and thinking which led to the current state (where <= 1.0 is excluded as "[playground](https://github.com/bndtools/bnd/issues/2452#issuecomment-390255971)" and [here](https://github.com/bndtools/bnd/issues/2452#issuecomment-390255971) ).

https://github.com/bndtools/bnd/issues/2452#issuecomment-390247811

> [@pkriens](https://github.com/pkriens) [@rotty3000](https://github.com/rotty3000) [@timothyjward](https://github.com/timothyjward) What are your thoughts on this? Basically the change would cause baselining to report no changes when either input (older, newer) has a version < 1.0.
> 
> That is, if older < 1.0, then it is not a suitable baseline since "real" API does not start until 1.0. If newer < 1.0, then it is not real API and thus there is no point in baselining it.

There was also a question whether this should be [configurable](https://github.com/bndtools/bnd/issues/2452#issuecomment-390251844)

laeubi:
From my point of view there is nothing like "no real API"... if something is accessible people will use it and no matter what you will break someones workflow:

grafik

So I would propose to simply support baseline in any version, if one thing its not needed, simply skip baseline validation seems the logical consequence until the project feels "ready" to do so and hopefully there consumers are aware of that fact.

kwin:
For me 0.x is a good way to express "this API should be considered experimental/unstable" without needing to use additional annotations/documentation.

laeubi:
> For me 0.x is a good way to express "this API should be considered experimental/unstable" without needing to use additional annotations/documentation.

No one prevents you from doing that but then you don't need to enable baseline, what is all about stable API and ensuring proper use of semantic versioning rules (according to OSGi).

chrisrueger:
> I vote for fixing baselining to also work on 0.x packages.

> So I would propose to simply support baseline in any version,

@peterkir @laeubi @kwin I have created https://github.com/bndtools/bnd/issues/6771  for this. 
Let's continue the discussion there maybe and let me know if the issue description needs correction. 
I will probably not implement the feature, because I am not familiar with the feature, but of course I can support with everything else.

kwin:
> > For me 0.x is a good way to express "this API should be considered experimental/unstable" without needing to use additional annotations/documentation.
> 
> No one prevents you from doing that but then you don't need to enable baseline, what is all about stable API and ensuring proper use of semantic versioning rules (according to OSGi).

Again this issue is not at all about baselining but just how the default import-package version ranges are gonna be calculated.

laeubi:
> Again this issue is not at all about baselining but just how the default import-package version ranges are gonna be calculated.

The calculation is correct, so maybe this should be simply closed as `WON'T FIX` then if you no longer care about baselining?

kwin:
I do care about consumers of unstable versions (indicated via 0.x). As long as they get incorrect version ranges for unstable API version this issue is valid. @chrisrueger Could we at least optionally enable this behaviour then?

laeubi:
>  As long as they get incorrect version ranges for unstable API version this issue is valid

The version ranges here are not incorrect, that is the main problem here, you are stating things as you interpret them, but this is not how it is specified. As mentioned above one might argue that the version range already is to broad if we assume < 1.x is unstable (what is not specified anywhere at OSGi), but any broader range is simply invalid and I won't add any option to allow such bad behavior globally.

Apart from that consumer of course already can do that (by explicitly adding an open upper bound), so there is literally no need for having any additional configuration option here.

kwin:
The main problem with unstable APIs (0.x) is that just promoting them to stable (i.e. 1.x) breaks all consumers with default import package version ranges.

rombert:
> The main problem with unstable APIs (0.x) is that just promoting them to stable (i.e. 1.x) breaks all consumers with default import package version ranges.

Right, this is the precise situation I am in with https://github.com/apache/sling-org-apache-sling-auth-oauth-client . I started it off at 0.x package exports because the APIs were in flux and did not want to release a 'stable' version with 3.x exports.

After a few iterations we're done and the APIs are final at 0.x and we are discussing what to do because bumping to 1.x would break consumers.

laeubi:
@rombert @kwin please reread what you writing here, so you claim having an unstable API but don't want to break consumers because it becomes stable now... that completely contradictory because you have breaking consumers all the time before with your unstable API.

Furthermore you now want them to having used a open range what would even break them anytime soon after your stable API change.

So matter of fact is, once the API becomes stable, it is the first time consumers can rely on not being broken and need one update to the stable API version (1.0.0) in your case.

chrisrueger:
> I do care about consumers of unstable versions (indicated via 0.x). As long as they get incorrect version ranges for unstable API version this issue is valid. @chrisrueger Could we at least optionally enable this behaviour then?

@kwin Just wanted to get back to this.

I think nothing we are discussing here would help your existing consumers, since they already have their ranges in their jar's manifest baked in and would need to be regenerated / rebuilt anyway. I think this was also concluded in [here](https://lists.apache.org/thread/xtdr34srcl9d9oszfozw4phn0g715ghn).

I suggest we close this PR. 

For the long run, maybe we could come up with a suggestion for projects on how to communicate the versioning policy for consumers of 0.x versions. You might tell them to use or more open version range, e.g. :

> Dear alpha/beta users of our library, we release our lib with version 0.x in the beginning. Please use a more open import range like `[0,2)` if you are consuming our OSGi library to ensure no hickups when we release 1.0. During the initial 0.x releases expect lots of instabilities and breaking changes, since this is pre 1.0, but it will get better once we a closer to 1.0 and we promise to not break when going to 1.0

But I think this is a policy thing individual to each project and not something we should generalize here.

Sorry for being mostly absent in this discussion. I realized due to my lack of experience with this, that I could not contribute much and after reading the other comments I think I underestimated this PR. 

Update: Forgot to answer the question for configurability. After the discussion above I have the feeling there is pushback agains the idea. ~~But in case it would be configurable, could you outline how you would like to configure such an option? Maybe in a new issue?~~ 
Update2: Also thought about it: I also don't think we should have an option to create more open ranges beyond the next major version. It would create libraries which then exist in the world which have this open range (or unconstraint range) which can cause problems, as a bundle now could be wired to much higher incompatible versions in the future (see https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/3236#issuecomment-2276399714 as an example).

chrisrueger:
Closing this as the consensus so far has been to not do this.

PR #6712, opened 2025-07-27
docs: update to ruby 3.4.5

chrisrueger:
this is an attempt to make dependabot security upgrades run again. Most likely the docs build and dependencies have been to far in the past that updates don't work


chrisrueger:
This has fixed all current dependabot alerts
image

PR #6726, opened 2025-07-30
Add dependencies in sub-bundle pom.xml

### LINKED ISSUE #6725: Sub-bundle pom.xml does not contain dependencies with -pom: true
## Problem

I have set `-pom: true` in `cnf/build.bnd` so that all built `.jar`s of each project contains a `META-INF/maven/groupid/artifactid/pom.xml`  which also have a populated `<dependencies>` section with all dependencies from the buildpath. 
This works fine for normal bundles (although `${repo}` references are not considered... other problem).

BUT it sub-bundles (-sub: true) the pom.xml exists but does not contain a `<dependencies>` section. 

I will provide a PR which ensures that sub-bundle pom.xml also contain a populated `<dependencies>` section.

chrisrueger:
Closes #6725


chrisrueger:
Note: I discovered an effect of https://github.com/bndtools/bnd/pull/2851 which always removes the existing `META-INF/maven` folders from the jar before putting the generated `pom.xml` in case of `-pom: true`

Something to keep in mind. I stumbeled over this behavior.

PR #6727, opened 2025-07-30
Experimental: Add all pom.xml files of all runbundles to exported / executable jar

chrisrueger:
This adds all pom.xml of all runbundle jars to the the executable jar.

Example:

```
executable.jar
  /META-INF/maven/foo/bar/pom.xml
  /META-INF/maven/foo/baz/pom.xml
  /jar/foo.bar.jar (groupId: foo, artifactId: bar)
  /jar/foo.baz.jar (groupId: foo, artifactId: baz)
```

Depends on https://github.com/bndtools/bnd/pull/6726 to work properly


chrisrueger:
I discard this idea, because it makes the executable jar much bigger than needed. 
I fixed problem with sub-bundles in https://github.com/bndtools/bnd/pull/6726 which now adds more meta data to sub-bundle jars which is helpful for SBOM tools like Syft .

Issue #5156, opened 2022-03-03
org.bndtools.p2: Update the build for the p2 repository to replace md5 checksum


bjhargrave:
The latest Eclipse now complains that our p2 repository only generates md5 checksums. We use the p2 repository builder from Eclipse 3.5.2 (circa 2009). Recent p2 updates now generate sha256 checksums.

So we need to modernize the p2 repository building in the `org.bndtools.p2` project. I suspect we will need to move to use the Maven Tycho support. For example, https://www.eclipse.org/tycho/sitedocs/tycho-p2/tycho-p2-repository-plugin/assemble-repository-mojo.html.

```
eclipse.buildId=4.22.0.I20211124-1800
java.version=17.0.2
java.vendor=Azul Systems, Inc.
BootLoader constants: OS=macosx, ARCH=x86_64, WS=cocoa, NL=en_US
Framework arguments:  -product org.eclipse.epp.package.java.product -keyring /Users/hargrave/.eclipse_keyring
Command-line arguments:  -os macosx -ws cocoa -arch x86_64 -product org.eclipse.epp.package.java.product -keyring /Users/hargrave/.eclipse_keyring

org.eclipse.equinox.p2.repository
Warning
Thu Mar 03 09:22:46 EST 2022
The digest algorithms (md5) used to verify osgi.bundle,biz.aQute.bndlib,6.3.0.202203030031-SNAPSHOT have severely compromised security. Please report this concern to the artifact provider.
```

bjhargrave:
We also need to support the PGP signing recently added to eclipse. See <https://gitlab.eclipse.org/eclipse-wg/ide-wg/community/-/issues/11#downstream-product-considerations>.

pkriens:
@maho7791 has a p2 exporter. he will talk to Jürgen and make a plan of action.

pkriens:
Could you look at the P2 PR? I've created a new P2 exporter and that would solve this issue. Need someone to engage before I want to use it ourselves

pkriens:
I plan to include the P2 exporter in release 7 but not use it yet. For 7.1.0-SNAPSHOT I will try to use it for the release

chrisrueger:
Reopening

PR #6738, opened 2025-08-04
enable sonatype deployments

peterkir:
enable releae to sonatype / maven-central via 
https://central.sonatype.com/api/v1/publisher/upload
as documented here 
https://central.sonatype.org/publish/publish-portal-api/#uploading-a-deployment-bundle

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
@@ -12,6 +12,9 @@ import java.io.OutputStream;
 import java.io.UnsupportedEncodingException;
 import java.net.URI;
 import java.net.URLDecoder;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -125,6 +128,7 @@ public class MavenBndRepository extends BaseRepository implements RepositoryPlug
 	private final AtomicReference<Throwable>	open				= new AtomicReference<>();
 	Optional<Workspace>							workspace;
 	private AtomicBoolean						polling				= new AtomicBoolean(false);
+	private Path								sonatypeDir;
 
 	/**
 	 * Put result
@@ -278,8 +282,7 @@ public class MavenBndRepository extends BaseRepository implements RepositoryPlug
 			String clazz = extra.clazz;
 			File file = new File(path);
 			if (!file.isFile())
-				reporter.error("-release-maven archive contains a path to a file that does not exist: %s",
-					file);
+				reporter.error("-release-maven archive contains a path to a file that does not exist: %s", file);
 			else {
 				try (Resource r = new FileResource(file)) {
 					Resource what;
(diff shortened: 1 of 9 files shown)


stbischof:
@maho7791 
@juergen-albert

biz.aQute.repository/src/aQute/maven/provider/Releaser.java
```diff
@@ -58,8 +82,23 @@ public void close() throws IOException {
...
+					Iterator<Archive> iterator = upload.iterator();
+					uploadAll(iterator);
 					updateMetadata();
+					List<MavenBackingRepository> releaseRepositories = home.getReleaseRepositories();
+					String isSonatypePublish = context.getProperty("sonatypePublish", "false");
+					if (!releaseRepositories.isEmpty()
+						&& Boolean.parseBoolean(isSonatypePublish)) {
+						MavenBackingRepository mbr = releaseRepositories.get(0);
+						if (mbr.toString()
+							.contains("/sonatype_")) {
```
chrisrueger, line 94: Maybe something to improve in the future, as relying on `.toString()` might be fragile.

chrisrueger (approved):
Looks good. Just a minor comment regarding `.tostring()`. 
There might be other possible cleanups here and there but still something we can discuss in the future.

chrisrueger:
related https://github.com/bndtools/bnd/issues/5156 org.bndtools.p2: Update the build for the p2 repository to replace md5 checksum


and 

> We also need to support the PGP signing recently added to eclipse. See https://gitlab.eclipse.org/eclipse-wg/ide-wg/community/-/issues/11#downstream-product-considerations.

source: https://github.com/bndtools/bnd/issues/5156#issuecomment-1155143451

biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
```diff
@@ -703,6 +717,12 @@ synchronized boolean init() {
 		}
 	}
 
+	private boolean isSonatypeCentralPortal(String releaseUrl) {
+		if (releaseUrl == null)
+			return false;
+		return releaseUrl.contains("central.sonatype.com/api/v1/publisher/upload/");
```
maho7791, line 723: I am not sure, if its a good idea to depend this feature on a fixed URL. If we would update our Sonatype Nexus to the latest version, we would also get the new release process. But then the repository would only work for this URL but not ours. Wouldn't it be better to use a flag on the repository configuration?

peterkir: I'm not familiar with nexus installations (is it available as OSS installation). My assumption was that only sonatype is using this new api for deployments.
Are you using a nexus server? Do you have more references on nexus?

maho7791: Sonatype Nexus is a product, like artifactory is. We run the previous version on our infrastructure, the Eclipse Foundation runs an own instance (also still an older version). The new version will use the same API like for maven central. But the URL's are then different.

juergen-albert: @peterkir can you make this fixed URL overwritable by configuration property. Having this as the default seems okay to me, but it would be nice to be able to overwrite it.

juergen-albert:
@peterkir Have you been able to work on what we discussed a while back?

PR #6739, opened 2025-08-05
speedup CDIAnnotations plugin

chrisrueger:
This improves the check if a bundle-classpath entry (BCP entry) contains a META-INF/beans.xml

in my usecase (11MB hibernate-core.jar) calculating the `Map<String, Discover> discoverPerBCPEntry` took `> 800ms`, because `Jar.getResources(resource, beansResourceFilter)` used a `ZipResourceSpliterator`. It turned out that this is slower than using a JarResource directly which got it down to e.g. 57 ms


chrisrueger:
See https://bnd.discourse.group/t/performance-improve-build-performance-by-narrowing-annotation-scanning-to-specific-packages/535 for benefits of this PR

PR #6371, opened 2024-11-15
Fix Bnd Launcher with Java 24

chrisrueger:
Related to #6370
hope to fix a Java 24 build problem, where the Launcher isn't working with Java 24.

@pkriens is this something which should still go to 7.1.0 ?

And also: I added Java 23 temporarily to the build (Java 24 was not available in the Github actions... probably because it is still early access @timothyjward ). But maybe we already see something with JDK23. 

@scordio Are you able to test this fix in your build with Java 24 to see if it fixes your problem?

PR #6746, opened 2025-08-08
allow usage of lower biz.aQute.launcher versions

### LINKED ISSUE #6737: Feature Request: Select Executable JAR Launcher Based on `-runee` Compatibility
**Is your feature request related to a problem? Please describe.**

The Bndtools 7.1 release is built on Java 17. This enables developers to use the latest features of Bndtools with modern Eclipse versions. For local development, it's possible to configure Eclipse to run Bndtools with a Java 17+ execution environment while ensuring project compliance with Java 1.8.

The problem arises during the "Export Executable Jar" process for a `.bndrun` file. Bndtools 7.1 currently embeds its own `biz.aQute.launcher:7.1.0`, which is compiled with Java 17 class file format. This makes the resulting executable JAR incompatible with older target runtimes like Java 1.8.

When such a JAR is executed on a JRE 1.8, it fails with an `UnsupportedClassVersionError`:

```
Exception in thread "main" java.lang.UnsupportedClassVersionError: biz/aQute/launcher/Launcher has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

Attempts to work around this by explicitly adding a Java 1.8-compatible launcher (e.g., `biz.aQute.launcher:6.4.1`) to the `-runpath` are not respected by the export dialog; the bundled `biz.aQute.launcher:7.1.0` is always used.

This forces projects needing to produce Java 1.8 compatible executables to not being benefitted from recent advancements.

**Describe the solution you'd like**

We propose enhancing the "Export Executable Jar" functionality to make it aware of the target execution environment specified in the `.bndrun` file (or derived from workspace).

1.  When exporting an executable JAR from a `.bndrun` file, Bndtools should inspect the `-runee` instruction (e.g., `-runee: JavaSE-1.8`).
2.  Bndtools should then use the specified OSGi execution environment capability (e.g., `osgi.ee; filter:="(&(osgi.ee=JavaSE)(version=1.8))"`) to search the configured repositories for a compatible version of the `biz.aQute.launcher` bundle.
3.  If a compatible launcher is found (for instance, `biz.aQute.launcher:6.4.1` for Java 1.8), that version should be resolved and packaged into the executable JAR. The JAR's `Main-Class` attribute should be set to `Embedded Launcher` class stored in the prelauncher jar which is located inside of this resolved launcher.
4.  If the `-runee` instruction specifies a modern environment (e.g., Java 17 or higher), the current behavior of using the bundled `biz.aQute.launcher:7.1.0` should be maintained.
5.  If no launcher matching the `-runee` capability can be found in the configured repositories, the export process should fail with an informative error message, such as: "Could not find a compatible launcher for the target execution environment 'JavaSE-1.8' in the configured repositories."

This solution is non-invasive and leverages Bnd's existing resolver and repository capabilities. It allows developers to use the latest Bndtools for development while still producing executables that are compatible with specified legacy runtimes.

**Describe alternatives you've considered**

1.  **Sticking with Bndtools 6.4.1 and older Eclipse:** This is the current, undesirable workaround. It blocks access to new features, performance improvements, and security fixes in both Bndtools and the Eclipse IDE.
2.  **Manually specifying `biz.aQute.launcher:6.4.1` in `-runpath`:** This was attempted, but the export process ignores this instruction and always prefers the bundled launcher.

This proposed feature provides a robust, forward-looking solution that aligns the export behavior with the project's declared runtime requirements.

peterkir:
resolves #6737
enables the capability of launching with the lowest available version of 'biz.aQute.launcher' inside the workspace repositories - if no lower version is available the existing behaviour is persisted and the version from the workspace/release is used

### DIFF
--- a/.github/workflows/cibuild.yml
+++ b/.github/workflows/cibuild.yml
@@ -92,10 +92,17 @@ jobs:
     #      ~/.bnd/urlcache/
     #    key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/cnf/**/*.bnd', '**/cnf/**/*.mvn', '**/cnf/**/*.maven', '**/cnf/ext/*.*') }}
 
-    - name: Build
+    - name: Build (default)
       id: build
+      if: ${{ matrix.java != '24' }}
       run: |
         ${{ format(matrix.runner, './.github/scripts/ci-build.sh') }}
+
+    - name: Build Java 24 with stacktrace output
+      id: buildJDK24
+      if: ${{ matrix.java == '24' }}
+      run: |
+        ${{ format(matrix.runner, './.github/scripts/ci-build.sh --stacktrace') }}
 
     - name: Configure settings.xml for Publish
       if: ${{ matrix.canonical }}
--- a/.project
+++ b/.project
@@ -1,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>_root</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+	</buildSpec>
+	<natures>
+	</natures>
+	<filteredResources>
+		<filter>
+			<id>1753880320173</id>
+			<name></name>
+			<type>5</type>
+			<matcher>
+				<id>org.eclipse.ui.ide.multiFilter</id>
+				<arguments>1.0-name-matches-false-true-.*</arguments>
+			</matcher>
+		</filter>
+		<filter>
+			<id>1753880320175</id>
+			<name></name>
+			<type>9</type>
+			<matcher>
+				<id>org.eclipse.ui.ide.multiFilter</id>
+				<arguments>1.0-name-matches-false-true-(\.github|\.mvn)</arguments>
+			</matcher>
+		</filter>
+	</filteredResources>
+</projectDescription>
--- a/biz.aQute.bndall.tests/test/biz/aQute/launcher/AlsoLauncherTest.java
+++ b/biz.aQute.bndall.tests/test/biz/aQute/launcher/AlsoLauncherTest.java
@@ -36,6 +36,7 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledIf;
+import org.junit.jupiter.api.condition.EnabledForJreRange;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.ValueSource;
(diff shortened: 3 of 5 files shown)


chrisrueger:
Thanks for the fix. Unfortunatelly test fails. 
Can you check one thing:

The getHandler() method you modified is called by two callers, `getProjectLauncher()` and `getProjectTester()`.
Should `getProjectTester()` also be affected by the same logic?

image

peterkir:
I was not able to reproduce the test error with Java17. I will give it a try with Java24

peterkir:
I see. I have to check how to update the testdata.
`[/error/java.io.FileNotFoundException: Download biz.aQute.launcher-3.0.0.201508170645 into repo01]`

chrisrueger:
> I was not able to reproduce the test error with Java17. I will give it a try with Java24

Ah ok, maybe it is related to this https://github.com/bndtools/bnd/pull/6371 where an older laucher does not work with Java24 anymore, because of removed Security Manager

peterkir:
I removed the old launcher from the testdata, cause there is always a launcher bundle inside the workspace.
That does the trick locally.

peterkir:
I think there is an more serious issue with Java 24 which we have to look into
``` 
# version biz.aQute.launcher.pre;version=7.1.0.202411251545, org.eclipse.osgi; singleton:=true;version=3.18.200.v20221116-1324, biz.aQute.launcher;version=7.1.0.202411251545
    java.lang.UnsupportedOperationException: Setting a system-wide Policy object is not supported
    at java.base/java.security.Policy.setPolicy(Policy.java:114)
    at aQute.launcher.Launcher.activate(Launcher.java:493)
    at aQute.launcher.Launcher.launch(Launcher.java:403)
    at aQute.launcher.Launcher.run(Launcher.java:185)
    at aQute.launcher.Launcher.main(Launcher.java:161)
    at aQute.launcher.pre.EmbeddedLauncher.executeWithRunPath(EmbeddedLauncher.java:169)
    at aQute.launcher.pre.EmbeddedLauncher.findAndExecute(EmbeddedLauncher.java:134)
    at aQute.launcher.pre.EmbeddedLauncher.main(EmbeddedLauncher.java:51)
    ! Unexpected error in the run body: Setting a system-wide Policy object is not supported
    java.lang.UnsupportedOperationException: Setting a system-wide Policy object is not supported
    at java.base/java.security.Policy.setPolicy(Policy.java:114)
    at aQute.launcher.Launcher.activate(Launcher.java:493)
    at aQute.launcher.Launcher.launch(Launcher.java:403)
    at aQute.launcher.Launcher.run(Launcher.java:185)
    at aQute.launcher.Launcher.main(Launcher.java:161)
    at aQute.launcher.pre.EmbeddedLauncher.executeWithRunPath(EmbeddedLauncher.java:169)
    at aQute.launcher.pre.EmbeddedLauncher.findAndExecute(EmbeddedLauncher.java:134)
    at aQute.launcher.pre.EmbeddedLauncher.main(EmbeddedLauncher.java:51)
    # stopped system bundle due to leaving run body
    ```
    Could we exclude 24 for the moment?

**Comment by chrisrueger** [comment]
I think this Java24 exception  is expected, because 7.1. has the problem which is fixed by https://github.com/bndtools/bnd/pull/6371 which is not released yet. 

> Could we exclude 24 for the moment?

I think we have exclude this in JDK >=24 to and it is the only choice. 

You can try the `@EnabledForJreRange(maxVersion = 23)` annotation (should work now since @kriegfrj updated JUnit recently to 5.12.2). 

If that doesn't work you can try the workaround I used in `biz.aQute.launcher.AlsoLauncherTest.testOlderLauncherOnRunpath()`

```
        @Test
	@DisabledIf("isJava24OrAbove")
	public void sometest() throws Exception {..}

         /**
	 * Workaround for Jupiter < 5.12.x which does not allow to
	 * use @DisableForJRERange with int values for min max
	 */
	@SuppressWarnings("unused")
	private static boolean isJava24OrAbove() {
		return Runtime.version()
			.feature() >= 24;
	}

```

Issue #6730, opened 2025-07-31
Wrong Minimal required maven version for 7.0.0


cpfeiffer:
Same as #5801, see https://github.com/bndtools/bnd/wiki/Changes-in-7.1.0#backward-compatibility

Is this being generated from somewhere?

chrisrueger:
I can only guess but maybe it is coming from here?

https://github.com/bndtools/bnd/blob/7480096852568e4da8cb6688cba3640eb0c2be5e/maven-plugins/bnd-plugin-parent/pom.xml#L24

cpfeiffer:
It _should_ come from there, but apparently it doesn't (because the changelog says Maven 3.3.9).

chrisrueger:
I guess the problem is that something was forgotten to keep in "sync" 

https://github.com/bndtools/bnd/blob/7480096852568e4da8cb6688cba3640eb0c2be5e/maven-plugins/bnd-plugin-parent/pom.xml#L23-L24

https://github.com/bndtools/bnd/blob/7480096852568e4da8cb6688cba3640eb0c2be5e/cnf/ext/maven-plugin.bnd#L1-L4

chrisrueger:
I think I found the reason why the versions are not in sync. A dependabot update. 

https://github.com/bndtools/bnd/commit/87892faa3799ec1d7523a774bbbcfa9380f96297

Any idea how to prevent dependabot from touching this specific section in the pom.xml ?

chrisrueger:
@cpfeiffer I added https://github.com/bndtools/bnd/pull/6731 
Feedback welcome, but especially testing if it has the desired outcome.

cpfeiffer:
> Any idea how to prevent dependabot from touching this specific section in the pom.xml ?

I'm not familiar with configuring dependabot, but it looks like you could add an ignore-section to the `dependabot.yaml`:
- https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/controlling-dependencies-updated#ignoring-specific-dependencies
- https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#ignore-

```
ignore:
  - dependency-name: "org.apache.maven:maven-*"
    # Keep our maven.target.version at 3.3.9
    versions: [ ">3.3.9" ]
```

I'm not sure if `maven-*` would be correct, or if there are other dependencies named maven-something but not using `maven.target.version`.

cpfeiffer:
[...]

PR #6748, opened 2025-08-11
fix for CVE-2021-26291

peterkir:
resolves https://github.com/bndtools/bnd/security/dependabot/20


chrisrueger:
@peterkir I think this PR would introduce #6730 again.

This issue was the reason I set that fixed to maven 3.3.9 because 3.8.1 was by accident via dependabot. 

If we go to 3.8.1 then this probably should go to the [release notes](https://github.com/bndtools/bnd/wiki/Changes-in-7.2.0).

PR #6763, opened 2025-08-15
MavenBndRepo: Fix for Already closed error

chrisrueger:
now we log.warn the Already closed Exception of MavenBndRepo and just return false (instead of throwing exception), as the method does in other cases too.

I don't really know why the exception was thrown in the first place. The thing is that .init() is constantly called by polling (see .startPoll()) or other Workspace refresh actions which have references to the repo.


chrisrueger:
@gg-dim could you test this?

chrisrueger:
I merged please test with snapshot available in a few minutes. @gg-dim  

Marking @kriegfrj too because you mentioned this error on another issue

Issue #6480, opened 2025-02-20
Make the usage of RepositoryPlugin <-> Repository more flexible in bnd (run)s


laeubi:
One thing I noticed quite often now is that when using Bndruns programmatically (not sure if the same problem might occur elsewhere) to perform a resolve operation one sometimes get very strange `could not resolve <<INITIAL>>...` errors. 

After some digging one finds that this is because there are two kind of "repositories":

1. "bnd" repositories (like [FileRepo](https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/lib/deployer/FileRepo.java)) that are implemented as `RepositoryPlugin` and can be used to resolve e.g. bundle classpath and alike
2. "osgi" repository, that is required to perform a resolve operation (e.g. [FileSetRepository](https://github.com/bndtools/bnd/blob/master/biz.aQute.repository/src/aQute/bnd/repository/fileset/FileSetRepository.java)) that are usually also `RepositoryPlugin`s

One example is `biz.aQute.resolve.BndrunResolveContext.getAllRepos()` where it only will get all that are OSGi Repository but this will of course not cover "plain" ones. This is very confusing because the usual bnd operations work, but the resolve suddenly fails.

I could think of two ways to resolve this:

1. Make bndrun / maybe others "smarter" ...
2. Have some kind of way to define a "converter" that is when I call `aQute.bnd.service.Registry.getPlugins(Repository.class)` automatically wraps `RepositoryPlugin`s that do not implement that interface in a `Repository` thats uses the repository plugin as a backend.

One example is the "implicit" repository `CachedFileRepo` that come with each Workspace by default, one can use that for all operations in bnd except for resolve... for this purpose one has to explicitly add for example `biz.aQute.tester.junit-platform` even though it is part of the implict repository.

chrisrueger:
I stumbled on the two kinds of repositories too when I worked on the tags. But I didn't pay it much attention. But maybe somebody could explain the intention / reasoning behind this. 

> One example is the "implicit" repository `CachedFileRepo` that come with each Workspace by default, one can use that for all operations in bnd except for resolve

> 1. Make bndrun / maybe others "smarter" ...

I just found this:

https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/BndrunContainer.java#L216

Do you mean by "making smarter" to do something similar?

laeubi:
[...]

PR #6764, opened 2025-08-16
RepoBrowser: show the bnd-cache repo

chrisrueger:
this makes the previously hidden [CachedFileRepo](https://github.com/bndtools/bnd/blob/fda9509b81a0e61b90c4600d68ead2cf6b43d9ec/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java#L640) visible in the repo-browser. It helps devs to know what's going on behind the scenes e.g. because this bnd-cache contains the ProjectLauncher which is automatically added to exportable .jars. But so far it has been unclear where this launcher is magically coming from. So this is an attempt to make this repo visible.

image

Btw: I found there is the instruction https://bnd.bndtools.org/instructions/nobuildincache.html which can disable this cache completely (@amitjoy might be interesting in context of Launcher too for you)

TODO 

- [ ] what are other places where this should be shown... e.g. what about the .bndrun editor for `-runrepos`

## Docs

I found the following places where this bnd-cache is mentioned. 

- https://bnd.bndtools.org/chapters/150-build.html#the-cnfcache-directory
- https://bnd.bndtools.org/chapters/150-build.html#overriding-the-plugins

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java
@@ -638,11 +638,25 @@ public class Workspace extends Processor {
 	}
 
 	class CachedFileRepo extends FileRepo {
+
 		final Lock	lock	= new ReentrantLock();
 		boolean		inited;
 
 		CachedFileRepo() {
 			super(BND_CACHE_REPONAME, getCache(BND_CACHE_REPONAME), false);
+		}
+
+		@Override
+		public String tooltip(Object... target) throws Exception {
+
+			if (target == null || target.length == 0) {
+				String statustext = isTrue(getProperty(NOBUILDINCACHE)) ? "Disabled by -nobuildincache"
+					: "Enabled - Use -nobuildincache to disable this cache";
+				return String.format("%s (%s)%n%s", getName(), statustext, root);
+			}
+
+			return super.tooltip(target);
+
 		}
 
 		@Override
@@ -708,6 +722,7 @@ public class Workspace extends Processor {
 				}
 			}
 		}
+
 	}
 
 	public void syncCache() throws Exception {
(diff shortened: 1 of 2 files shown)


chrisrueger:
@peterkir @amitjoy a first start about what we talked about in yesterday's call

peterkir:
Yes this would be really helpful. I think it should be visible everywhere the repos are shown.
Maybe we can tweak the "tooltip/hover" and mention the switch `-nobuildincache` for disabling it completely.

chrisrueger:
> Maybe we can tweak the "tooltip/hover" and mention the switch `-nobuildincache` for disabling it completely.

I modified the tooltip like this:
image

> If disabled then it reads "Disabled by -nobuildincache

> Yes this would be really helpful. I think it should be visible everywhere the repos are shown.

The other places I found are all + buttons in the .bndrun editor:

image

or

image

I am a bit hesitant to show it there because I am not sure that the cached repo is availble for run-bundles and resolving? 
The reason is that e.g. in the RepoBrowser the repositories are fetched via `RepositoryPlugin.class`like this:

```
List<RepositoryPlugin> plugins = localWorkspace.getPlugins(RepositoryPlugin.class);
```

which includes the `CachedFileRepo` (which is the bnd-cache)


but for resolving we use

```
List<RepositoryPlugin> plugins = localWorkspace.getPlugins(Repository.class);
```

which does NOT return `CachedFileRepo`.  (might be related to https://github.com/bndtools/bnd/issues/6480 )

So I think it could be misleading to show the `bnd-cache` in other places than the RepoBrowser.

I think maybe for a start the Repo-Browser is enough and an improvement over the current situation by making the `bnd-cache` visible. We could build on that in the future.

What do you think? @peterkir

chrisrueger:
mentioned the `bnd-cache` repo on this page https://bndtools.org/manual/repositories-view.html so that people can discover it by clicking on the Help button of the repo browser.

PR #6781, opened 2025-08-25
add warn marker for Non-empty, non-contained pck

chrisrueger:
Adds an additional WARNING marker in Eclipse Problems View for packages which get the little red icon , indicating that the package is non-empty but is neither in -privatepackage or Export-Package

The goal is that this kind of problem gets noticed easier, because in a large workspace with deeply nested packages (and maybe package presentation "hierarchical") it is hard to notice. 

## Example

For a package like this 

image

you now also get a warning like this :

image

**Bndtools Path Problem Marker**
```
Non-empty, non-contained package: my.package.foo (Add package to private or exported packages in bnd.bnd)
```

Suggestions for better wording welcome. The current message is based on an existing comment in the code.


chrisrueger:
Merged and will test it in the real world with our stuff.

PR #6797, opened 2025-09-04
Switch to the correct BuildContext API

gnodet:
This avoids rebuilding the jar when nothing has changed, and in projects with subproject, avoids downstream project recompilations.

### DIFF
--- a/maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
+++ b/maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
@@ -68,7 +68,7 @@ import org.apache.maven.settings.Settings;
 import org.apache.maven.shared.mapping.MappingUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.sonatype.plexus.build.incremental.BuildContext;
+import org.codehaus.plexus.build.BuildContext;
 
 /**
  * Abstract base class for all bnd-maven-plugin mojos.


cstamas (approved):

maven-plugins/bnd-plugin-parent/pom.xml
```diff
@@ -196,6 +196,11 @@
 				<version>${osgi.annotation.version}</version>
 				<scope>provided</scope>
 			</dependency>
+			<dependency>
+				<groupId>org.codehaus.plexus</groupId>
+				<artifactId>plexus-utils</artifactId>
+				<version>3.6.0</version>
```
bjhargrave: Is the scope of this not provided? Like other artifacts which come built in to maven?

maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
```diff
@@ -457,8 +458,12 @@ private void attachArtifactToProject(Jar bndJar) throws Exception {
 
 			Files.createDirectories(artifactFile.toPath()
 				.getParent());
-			try (OutputStream os = buildContext.newFileOutputStream(artifactFile)) {
+			try (CachingOutputStream os = new CachingOutputStream(artifactFile)) {
```
bjhargrave: Seems like this capability should be part of the build context rather than requiring all plugins to do something like this :-)

maven-plugins/bnd-plugin-parent/pom.xml
cstamas: plexus-utils is not exported by Maven core (was "injected" by Maven 2) to allow plugins to bring their own versions. Declaring plexus-utils as "provided" is wrong, as it is not provided for plugins.

maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
cstamas: This build context (the one you use)? https://github.com/sonatype/sisu-build-api

cstamas: No, my mistake, sorry, the used one is this https://github.com/codehaus-plexus/plexus-build-api

maven-plugins/bnd-plugin-parent/pom.xml
bjhargrave: Just checking!

maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
bjhargrave: We use this one:

https://github.com/bndtools/bnd/blob/a644f2d34e976de1ddf811b20f5e789f54fbb02c/maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java#L169-L170

It is injected by maven.

gnodet: I think we're on the wrong track.  The default build context should already cache the stream, see https://github.com/codehaus-plexus/plexus-build-api/blob/903ed6c3e9dc767da4f9ac6cf5a22f0324400bbf/src/main/java/org/codehaus/plexus/build/DefaultBuildContext.java#L96-L101

However, multiple successive runs of bad generate different output, for example:
```
< Export-Package: org.eclipse.aether;uses:="org.eclipse.aether.artifact,
<  org.eclipse.aether.collection,org.eclipse.aether.deployment,org.eclip
<  se.aether.graph,org.eclipse.aether.installation,org.eclipse.aether.me
<  tadata,org.eclipse.aether.repository,org.eclipse.aether.resolution,or
<  g.eclipse.aether.scope,org.eclipse.aether.transfer";version="2.0.11",
---
> Export-Package: org.eclipse.aether;version="2.0.11";uses:="org.eclipse
>  .aether.artifact,org.eclipse.aether.collection,org.eclipse.aether.dep
>  loyment,org.eclipse.aether.graph,org.eclipse.aether.installation,org.
>  eclipse.aether.metadata,org.eclipse.aether.repository,org.eclipse.aet
>  her.resolution,org.eclipse.aether.scope,org.eclipse.aether.transfer",
```

is there a way to have a consistent manifest written ?

gnodet: The issue seems a bit weird, as running `mvn clean install` seems to always create the same manifest.  However, a subsequent `mvn install` (with no clean) generates a slightly different one.

cstamas: Nope, I was right in my original comment: the 15 years old (like my son) is being used, as this commit was not yet released: 6f5823e92e070d6088c7406a6b0e2dfa1542fe61

gnodet: > Nope, I was right in my original comment: the 15 years old (like my son) is being used, as this commit was not yet released: [6f5823e](https://github.com/bndtools/bnd/commit/6f5823e92e070d6088c7406a6b0e2dfa1542fe61)

This should fix the need for the current change in the PR, but not the fact that `bnd` is not idempotent.

cstamas: Yes, lack of build idempotency nullifies all this (in this PR, or the functionality in alive build context, that is not present in abandoned).

chrisrueger: > lack of build idempotency

Not sure if I interpret "idempotency" correctly, but if you are referring to reproducible build outputs maybe have a look at https://bnd.bndtools.org/instructions/reproducible.html  and https://bnd.bndtools.org/instructions/noextraheaders.html (also see https://github.com/bndtools/bnd/blob/master/maven-plugins/bnd-maven-plugin/README.md#reproducible-builds )

 This is mainly about timestamps.. but I will have a look if it is used for something else. 

Question: Whic bnd version are you using here?

chrisrueger: @gnodet @cstamas I just remembered a past issue which might be related to your problem (not sure):
https://github.com/bndtools/bnd/issues/6221#issuecomment-2296567709 and the PR https://github.com/bndtools/bnd/pull/6222

I believe the #6222 should be in bnd 7.1.0 https://github.com/bndtools/bnd/releases/tag/7.1.0

Docs: https://github.com/bndtools/bnd/blob/master/maven-plugins/bnd-maven-plugin/README.md#configuration-parameters-1 

see the `deleteExistingManifest` option

gnodet: Sure, but that looks like a work-around.  The input looks the same, just the order of clause seems different.  Given that has no effect in OSGi, a fix order should be used somehow I think.

chrisrueger: @gnodet 

I had a quick look and maaaybe a suspicion:


I noticed that in your debug output [above](https://github.com/bndtools/bnd/pull/6797/files#r2322459119) the `uses:` directive comes first in example A, and then at the end in Example B. 
So this means the attributes (Attrs) of the Export-Package parameter (or any) are ordered differently... potentially in insertion-order.

https://github.com/bndtools/bnd/blob/a644f2d34e976de1ddf811b20f5e789f54fbb02c/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L2465-L2469

And given the comment in 

https://github.com/bndtools/bnd/blob/a644f2d34e976de1ddf811b20f5e789f54fbb02c/biz.aQute.bndlib/src/aQute/bnd/osgi/Analyzer.java#L2426

It looks like this could be your case where "someone else" is your previous build.

Attrs.java 

https://github.com/bndtools/bnd/blob/a644f2d34e976de1ddf811b20f5e789f54fbb02c/biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java#L86-L88 is initialized with a LinkedHashMap which indicates insertion order.

Maybe we should think about using a Treemap instead so that attributes are consistently ordered by key.

Thoughts @bjhargrave @pkriens ?

gnodet:
I repurposed this PR to switch to the correct library for BuildContext.
That one does indeed support using caching streams.

maven-plugins/bnd-maven-plugin/src/main/java/aQute/bnd/maven/plugin/AbstractBndMavenPlugin.java
bjhargrave: Or fix where emitted. Isn't that already done?

PR #6798, opened 2025-09-04
Fix directive ordering inconsistency in OSGi headers

gnodet:
When bnd processes existing manifests without cleaning, the order of
attributes and directives within OSGi headers (like Export-Package)
could vary between builds. This happened because:

1. When reading an existing manifest, attributes/directives are inserted
   in the order they appear in the file
2. When generating a fresh manifest, they are inserted in the order
   processed by the code
3. The Attrs.append() method preserved insertion order using LinkedHashMap

This fix ensures consistent ordering by sorting attributes and directives
in the Jar.writeManifest() method:
- Attributes (no ':' suffix) come first, sorted alphabetically
- Directives (with ':' suffix) come second, sorted alphabetically

This resolves the issue where Export-Package headers would show
different ordering of 'version' and 'uses:' directives between
clean and incremental builds.

Fixes the inconsistent manifest generation reported in the issue where
the same content appeared in different orders:
- Before: uses:="...";version="2.0.11"
- After:  version="2.0.11";uses:="..."

Now both scenarios produce: version="2.0.11";uses:="..."

Signed-off-by: Guillaume Nodet <gnodet@gmail.com>

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
@@ -782,12 +782,13 @@ public class Jar implements Closeable {
 	private static Manifest clean(Manifest org) {
 		Manifest result = new Manifest();
 		Attributes mainAttributes = result.getMainAttributes();
+		Collator collator = attrCollator(); // Create once and reuse
 		for (Map.Entry<?, ?> entry : org.getMainAttributes()
 			.entrySet()) {
 			String nice = clean((String) entry.getValue());
 			Object key = entry.getKey();
 			if (Constants.OSGI_SYNTAX_HEADERS.contains(key.toString())) {
-				nice = reorderClause(nice);
+				nice = reorderClause(nice, collator);
 			}
 			mainAttributes.put(key, nice);
 		}
@@ -810,14 +811,14 @@ public class Jar implements Closeable {
 		return result;
 	}
 
-    private static final Collator ATTR_COLLATOR;
-    static {
-        ATTR_COLLATOR = Collator.getInstance(Locale.ROOT);
-        ATTR_COLLATOR.setDecomposition(Collator.CANONICAL_DECOMPOSITION);
-        ATTR_COLLATOR.setStrength(Collator.SECONDARY); // case-insensitive
+    private static Collator attrCollator() {
+        Collator collator = Collator.getInstance(Locale.ROOT);
+        collator.setDecomposition(Collator.CANONICAL_DECOMPOSITION);
+        collator.setStrength(Collator.SECONDARY); // case-insensitive
+        return collator;
     }
 
-    private static int compareAttrKeys(String k1, String k2) {
+    private static int compareAttrKeys(String k1, String k2, Collator collator) {
         boolean isDirective1 = Attrs.isDirective(k1);
         boolean isDirective2 = Attrs.isDirective(k2);
         // Attributes come before directives
@@ -825,17 +826,17 @@ public class Jar implements Closeable {
             return isDirective1 ? 1 : -1;
         }
         // Within same type, sort using collator
-        return ATTR_COLLATOR.compare(k1, k2);
+        return collator.compare(k1, k2);
     }
 
-    private static String reorderClause(String s) {
+    private static String reorderClause(String s, Collator collator) {
         Parameters header = OSGiHeader.parseHeader(s);
         for (Map.Entry<String, Attrs> entry : header.entrySet()) {
             Attrs newAttrs = new Attrs();
             Attrs oldAttrs = entry.getValue();
             oldAttrs.keySet()
                 .stream()
-                .sorted(Jar::compareAttrKeys)
+                .sorted((k1, k2) -> compareAttrKeys(k1, k2, collator))
                 .forEachOrdered(key -> newAttrs.put(key, oldAttrs.getType(key), oldAttrs.get(key)));
             entry.setValue(newAttrs);
         }
--- a/biz.aQute.bndlib.tests/test/test/JarTest.java
+++ b/biz.aQute.bndlib.tests/test/test/JarTest.java
@@ -16,6 +16,8 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.lang.reflect.Method;
 import java.nio.charset.StandardCharsets;
+import java.text.Collator;
+import java.util.Locale;
 import java.util.jar.Attributes;
 import java.util.jar.JarInputStream;
 import java.util.jar.Manifest;
@@ -463,12 +465,17 @@ public class JarTest {
 	@Test
 	public void testReorderClauseMethod() throws Exception {
 		// Test the reorderClause method directly using reflection
-		Method reorderClauseMethod = Jar.class.getDeclaredMethod("reorderClause", String.class);
+		Method reorderClauseMethod = Jar.class.getDeclaredMethod("reorderClause", String.class, Collator.class);
 		reorderClauseMethod.setAccessible(true);
+
+		// Create a collator for testing
+		Collator collator = Collator.getInstance(Locale.ROOT);
+		collator.setDecomposition(Collator.CANONICAL_DECOMPOSITION);
+		collator.setStrength(Collator.SECONDARY); // case-insensitive
 
 		// Test case 1: Mixed attributes and directives
 		String input1 = "com.example.api;uses:=\"com.example.internal\";version=\"1.0.0\";mandatory:=\"version\";provider=acme";
-		String output1 = (String) reorderClauseMethod.invoke(null, input1);
+		String output1 = (String) reorderClauseMethod.invoke(null, input1, collator);
 
 		// Verify that all components are present
 		assertTrue(output1.contains("com.example.api"), "Package name should be preserved");
@@ -487,7 +494,7 @@ public class JarTest {
 
 		// Test case 2: Simple case with one attribute and one directive
 		String input2 = "com.example;directive:=value;attribute=value";
-		String output2 = (String) reorderClauseMethod.invoke(null, input2);
+		String output2 = (String) reorderClauseMethod.invoke(null, input2, collator);
 
 		assertTrue(output2.contains("com.example"), "Package name should be preserved");
 		assertTrue(output2.contains("directive:=value"), "directive should be present");
@@ -495,7 +502,7 @@ public class JarTest {
 
 		// Test case 3: Only attributes
 		String input3 = "com.example;version=\"1.0.0\";provider=acme";
-		String output3 = (String) reorderClauseMethod.invoke(null, input3);
+		String output3 = (String) reorderClauseMethod.invoke(null, input3, collator);
 
 		assertTrue(output3.contains("com.example"), "Package name should be preserved");
 		assertTrue(output3.contains("version=\"1.0.0\""), "version should be present");
@@ -503,7 +510,7 @@ public class JarTest {
 
 		// Test case 4: Only directives
 		String input4 = "com.example;uses:=\"com.other\";mandatory:=\"version\"";
-		String output4 = (String) reorderClauseMethod.invoke(null, input4);
+		String output4 = (String) reorderClauseMethod.invoke(null, input4, collator);
 
 		assertTrue(output4.contains("com.example"), "Package name should be preserved");
 		assertTrue(output4.contains("uses:="), "uses directive should be present");


gnodet:
With #6797, it seems the API behaves correctly and does cause the jar to be modified at the end.

bjhargrave:
Which bnd-maven-plugin goal are you using? See `deleteExistingManifest` config parameter for when using `bid-process` goal. https://github.com/bndtools/bnd/blob/master/maven-plugins/bnd-maven-plugin/README.md#bnd-process-goal

biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
```diff
@@ -399,7 +399,25 @@ public String toString() {
 
 	public void append(StringBuilder sb) {
 		String del = "";
-		for (Map.Entry<String, String> e : entrySet()) {
+		// Sort entries to ensure consistent ordering of attributes and directives
```
bjhargrave: If we are going to got this route, then perhaps Attrs.keySet and Attrs.entrySet should be overridden to return items in the desired order. Then everyone benefits and we don't need multiple sorting algorithms which can get out of sync.

bjhargrave (commented):
However, this change can present issues for some since it will change the manifests which otherwise would not change. So this change would need to be noted in the release notes as potential breaking change.

gnodet:
> Which bnd-maven-plugin goal are you using? See `deleteExistingManifest` config parameter for when using `bid-process` goal. https://github.com/bndtools/bnd/blob/master/maven-plugins/bnd-maven-plugin/README.md#bnd-process-goal

Yes, though it looks to me as a workaround for not fixing the actual problem.

biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
chrisrueger: Would using a Treemap in the constructor and  the current logic in a comparator solve this?

e.g.

```
        Comparator<String> keyComparator = (k1, k2) -> {
            boolean isDirective1 = k1.endsWith(":");
            boolean isDirective2 = k2.endsWith(":");

            // Attributes come before directives
            if (isDirective1 != isDirective2) {
                return isDirective1 ? 1 : -1;
            }

            // Within same type (attribute or directive), sort alphabetically ignoring case
            return k1.compareToIgnoreCase(k2);
        };

        Map<String, String> map = new TreeMap<>(keyComparator);
```

Not sure what drawbacks the Treemap impl would have.

biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
```diff
@@ -84,7 +86,29 @@ private Attrs(Map<String, String> map, Map<String, Type> types) {
...
+	}
+
+	/**
+	 * Create a TreeMap with custom comparator that ensures consistent ordering:
+	 * - Attributes (no ':' suffix) come first, sorted alphabetically
+	 * - Directives (with ':' suffix) come second, sorted alphabetically
+	 */
+	private static Map<String, String> createSortedMap() {
+		Comparator<String> keyComparator = (k1, k2) -> {
+			boolean isDirective1 = k1.endsWith(":");
```
chrisrueger: Small suggestion: there are static helpers `Attrs.isAttribute(String)` and `Attrs.isDirective(String)` . Just for centralizing those specific checks.

docs/_chapters/160-jars.md
```diff
@@ -48,6 +48,19 @@ You can also make an ant file but this is not described here (volunteer?). The o
 ## Manifest
 Applying this recipe gives the following manifest in a JAR named `org.websocket.jar`:
 
+### Manifest Header Ordering (bnd 7.2.0+)
```
bjhargrave: You don't need to mention the bnd versions in the docs. The docs are generated and available for prior bnd versions. 

e.g. https://bnd.bndtools.org/releases/7.0.0/

@chrisrueger did the team fail to publish the 7.1.0 doc pages? https://github.com/bndtools/bnd/wiki/Release-Process#adding-bnd-release-documentation

chrisrueger: @bjhargrave seems so, though I wasn't involved in the release.
Do I understand correctly that this is something which can still be done based on the current [docs folder](https://github.com/bndtools/bnd/tree/next/docs) in `next` branch, following the [instructions](https://github.com/bndtools/bnd/wiki/Release-Process#adding-bnd-release-documentation)  ?

Update: Created https://github.com/bndtools/bnd/pull/6801
Update2: 7.1.0 docs are here now https://bnd.bndtools.org/releases/7.1.0/

gnodet:
I've changed and rewritten the PR to provide a simpler non breaking change.

chrisrueger:
Thanks. I ran your new `AttrsTest.testDirectiveOrderingConsistency` on `master` branch and I can confirm that it fails there, proving this PR fixes it. 
I also checked out the PR, built it and used it on our product. No problems.

Let's wait for some more feedback @bjhargrave @pkriens

biz.aQute.bndlib.tests/test/test/AttrsTest.java
```diff
@@ -65,4 +65,44 @@ public void testVersion() {
...
+
+		// Add in the order that would appear in a fresh manifest generation
+		merged.put("uses:", "org.eclipse.aether.artifact,org.eclipse.aether.collection");
+		merged.put("version", "2.0.11");
+
+		String mergedString = merged.toString();
+		System.out.println("Merged:   " + mergedString);
+
+		// The order should remain consistent after merging with itself
+		assertEquals(originalString, mergedString,
```
bjhargrave: I don't see that this test case actually merges 2 Attrs objects which the comments seem to indicate is being done. 

This test relies only on the last-in/last-out nature of the change to put.

biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
```diff
@@ -349,7 +351,7 @@ public Type getType(String key) {
 	public void putAll(Attrs attrs) {
 		types.keySet()
 			.removeAll(attrs.map.keySet());
-		map.putAll(attrs.map);
+		attrs.map.forEach(this::put);
```
bjhargrave: Add a comment here about why we don't use Map.putAll here when it seems like the obvious choice. This will help future maintainers from not breaking the code.

biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
```diff
@@ -265,6 +265,7 @@ public String put(String key, Type type, String value) {
 		} else {
 			types.put(key, type);
 		}
+		map.remove(key);
```
bjhargrave: I don't see how this forces a consistent ordering. It just forces puts to the end of the linkedhashmap. Last-in/last-out. If this is what we want, then a comment is needed to explain.

biz.aQute.bndlib/src/aQute/bnd/header/Attrs.java
```diff
@@ -273,6 +274,7 @@ public String put(String key, String value) {
 		if (key == null)
 			return null;
 
+		map.remove(key);
```
bjhargrave: I don't see how this forces a consistent ordering. It just forces puts to the end of the linkedhashmap. Last-in/last-out. If this is what we want, then a comment is needed to explain.

biz.aQute.bndlib.tests/test/test/AttrsTest.java
```diff
@@ -65,4 +65,44 @@ public void testVersion() {
...
+		String originalString = original.toString();
+		System.out.println("Original: " + originalString);
+
+		// When the manifest is loaded, some attributes are already populated
+		Attrs merged = new Attrs();
+		merged.put("version", "2.0.11");
+
+		// Add in the order that would appear in a fresh manifest generation
+		merged.put("uses:", "org.eclipse.aether.artifact,org.eclipse.aether.collection");
+		merged.put("version", "2.0.11");
```
bjhargrave:
```suggestion
		merged.mergeWith(original, true);
```

This would be more inline with the test case comments.

But this fix/test case is too simple. 

If you have an Attrs (from an existing Manifest) with a,b,c and you want to merge in a,b; you end up with c,a,b

bjhargrave (changes requested):
I am not excited by this change. It is always putting at the end which doesn't provide consistency. It may fix the narrow case at hand, but this class should handle more than merging manifest attributes. It is Bnd API and should be more precise.

gnodet:
> I am not excited by this change. It is always putting at the end which doesn't provide consistency. It may fix the narrow case at hand, but this class should handle more than merging manifest attributes. It is Bnd API and should be more precise.

You're right, I'll give it another try asap.

gnodet:
I pushed a minimalist version which solves the very specific problem.  The downside is that it may leave other similar holes.

biz.aQute.bndlib/src/aQute/bnd/osgi/Builder.java
```diff
@@ -547,8 +547,10 @@ public void cleanupVersion(Packages packages, String defaultVersion, String what
 						warning("No export version for exported package %s", entry.getKey());
 				}
 			}
-			if (v != null)
+			if (v != null) {
+				attributes.remove(Constants.VERSION_ATTRIBUTE);
```
bjhargrave: This seems odd to always move version to the end of the iteration order even in cases when `v.equals(cleanupVersion(v))`.

This seems too much of a hack and would definitely need comments here to explain why.

I think I would rather see something more complete like a definitive ordering or something similar.

There is already a work around for this in maven for the bnd-process goal. I would rather you use that than add this to everyone else's builds.

gnodet: > This seems odd to always move version to the end of the iteration order even in cases when `v.equals(cleanupVersion(v))`.
> 
> This seems too much of a hack and would definitely need comments here to explain why.
> 
> I think I would rather see something more complete like a definitive ordering or something similar.
> 
> There is already a work around for this in maven for the bnd-process goal. I would rather you use that than add this to everyone else's builds.

Deleting the file is definitely not a work around.
My goal is that subsequents `mvn install` don't cause the output jar to be changed _at all_.  The use of the caching mechanism avoids the writing of the same manifest to cause an actual modification on the file.  Deleting the file would obviously not work.
It may be possible to _ignore_ or _not read_ the manifest file, which would work, but _deleting_ it is a bad idea.

chrisrueger: @gnodet What about the Treemap approach in  `Attrs.java` you had in one of your recent iterations? What was the reason you decided against it? IMO  `Attrs.java` would be the most central place to get consistent ordering and every caller would benefit.

gnodet: Whatever you guys prefer.  But that's a breaking change.  If you're fine with that, I can get back to using a `TreeMap`.

chrisrueger: I think you described the potential "breaking" behavior that pretty good in one of you your docs changes (If I remember correctly, I saw a sentence that mostly tools would be affected which compare previous and currentl Manifest files). We can mention it in the release notes. 

In my opinion it is only "kinda" breaking because technically the manifest contains the same information - some attributes are just presented in a different order.
You have to adjust the failing bnd tests accordingly to pass the build.

gnodet: @bjhargrave you also agree with going back to using the `TreeMap` ?

bjhargrave: > Deleting the file is definitely not a work around.
> My goal is that subsequents `mvn install` don't cause the output jar to be changed _at all_.

Deleting a file before writing a replacement should only be a "change" if the file contents change. If they are the same (e.g. same SHA), then what has changed? Just the timestamp on the file? I don't know how sophisticated the build context is for determining file changes.

> you also agree with going back to using the `TreeMap` ?

Having a definitive ordering is a good thing. I think the answer is not to change Attrs but to fix ManifestUtil which is what sorts the manifest attributes as part of writing the manifest file out.

https://github.com/bndtools/bnd/blob/297bacb92707861108d77b2fc7d423da6bac75bb/aQute.libg/src/aQute/lib/manifest/ManifestUtil.java#L154-L159

If we just change Attrs, ManifestUtil could alter the ordering when the manifest is written out obviating the ordering in Attrs. 

So please focus on ManifestUtil as the way to define the definitive order of the written manifest.

gnodet: > > Deleting the file is definitely not a work around.
> > My goal is that subsequents `mvn install` don't cause the output jar to be changed _at all_.
> 
> Deleting a file before writing a replacement should only be a "change" if the file contents change. If they are the same (e.g. same SHA), then what has changed? Just the timestamp on the file? I don't know how sophisticated the build context is for determining file changes.

Yes, the timestamp is used to avoid rebuilding the downstream jar.

> 
> > you also agree with going back to using the `TreeMap` ?
> 
> Having a definitive ordering is a good thing. I think the answer is not to change Attrs but to fix ManifestUtil which is what sorts the manifest attributes as part of writing the manifest file out.
> 
> https://github.com/bndtools/bnd/blob/297bacb92707861108d77b2fc7d423da6bac75bb/aQute.libg/src/aQute/lib/manifest/ManifestUtil.java#L154-L159
> 
> If we just change Attrs, ManifestUtil could alter the ordering when the manifest is written out obviating the ordering in Attrs.
> 
> So please focus on ManifestUtil as the way to define the definitive order of the written manifest.

gnodet: `ManifestUtil` seems completely agnostic about the content, it sort manifest entries, but currently does not alter them.  Furthermore, it cannot access `Constants.OSGI_SYNTAX_HEADERS`.
As an alternative, I modified the `Jar.writeManifest`, which already performs some cleanup on the manifest before writing.  I also ensure `ManifestResource` does delegate to the same method.  Those are the only two callers of `ManifestUtil.write` I could see (and now, there's only one).

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -794,12 +798,35 @@ private static Manifest clean(Manifest org) {
 			for (Map.Entry<?, ?> entry : org.getAttributes(name)
 				.entrySet()) {
 				String nice = clean((String) entry.getValue());
-				attrs.put(entry.getKey(), nice);
+				Object key = entry.getKey();
```
bjhargrave: This looks in the wrong place. This loop is processing the individual (`Name:`) sections of the manifest. OSGi headers are all in the main section (`getMainAttributes`).

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -794,12 +798,35 @@ private static Manifest clean(Manifest org) {
...
+				attrs.put(key, nice);
 			}
 		}
 		return result;
 	}
 
+	private static String reorderClause(String s) {
+		Parameters header = OSGiHeader.parseHeader(s);
+		for (Map.Entry<String, Attrs> entry : header.entrySet()) {
+			TreeMap<String, String> map = new TreeMap<>((k1, k2) -> {
```
bjhargrave: It seems wasteful to create a TreeMap just to sort. Also the plain TreeMap here will lose the typing information in Attrs.

Create a case-insensitive collator above the outer loop:

```java
Collator collator = Collator.getInstance(Locale.ROOT);
collator.setDecomposition(Collator.CANONICAL_DECOMPOSITION);
collator.setStrength(Collator.SECONDARY); // case-insensitive
```

Sort the keys and put the type and value in the new Attrs.

```java
Attrs newAttrs = Attrs();
Attrs oldAttrs = entry.getValue()
oldAttrs.keySet()
    .stream()
    .sorted(collator::compare)
    .forEachOrdered(key -> newAttrs.put(key, oldAttrs.getType(key), oldAttrs.get(key));
```

gnodet:
@chrisrueger this PR requires the `breakingchange` label back.

chrisrueger:
> @chrisrueger this PR requires the `breakingchange` label back.

@gnodet  Thanks for the reminder, I was a few days off and now added back the label.

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -779,7 +785,11 @@ private static Manifest clean(Manifest org) {
 		for (Map.Entry<?, ?> entry : org.getMainAttributes()
 			.entrySet()) {
 			String nice = clean((String) entry.getValue());
-			mainAttributes.put(entry.getKey(), nice);
+			Object key = entry.getKey();
+			if (Constants.OSGI_SYNTAX_HEADERS.contains(key.toString())) {
```
chrisrueger, line 790: Maybe it's obvious and I just don't see it, but what was the reason to only do `reorderClause()` for `OSGI_SYNTAX_HEADERS`? Reason I ask is: Couldn't this inconsistency also happen for other headers?

chrisrueger:
I would wait on @bjhargrave final approval, since I was wrong all too often in my comments :)

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
bjhargrave: Bnd cannot know the syntax of headers which OSGi does not specify. So we should not attempt to parse and reorder any header in the manifest since we may mess it up.

biz.aQute.bndlib/src/aQute/bnd/osgi/Jar.java
```diff
@@ -800,6 +810,38 @@ private static Manifest clean(Manifest org) {
 		return result;
 	}
 
+    private static final Collator ATTR_COLLATOR;
```
bjhargrave: Collator is not thread-safe. So we cannot have a single static instance which could be used concurrently. We will need to create a new, local collator instance for each invocation of the clean method.

gnodet: Fixed

chrisrueger:
Thanks @gnodet for this fix and @bjhargrave for support. I added a link for breaking changes PRs in https://github.com/bndtools/bnd/wiki/Changes-in-7.2.0   ([example](https://github.com/bndtools/bnd/pulls?q=is%3Apr+is%3Amerged+updated%3A2024-11-13..2025-12-29+-label%3Adependencies+label%3Abreakingchange+sort%3Aupdated-desc+))

chrisrueger:
@gnodet bnd 7.2.0 is on the way (RC3). Would be great if you could test it with regards to your contribution
https://github.com/bndtools/bnd/issues/6974

Thanks a lot

PR #6777, opened 2025-08-22
use P2Exporter for org.bndtools.p2

chrisrueger:
Closes #5156 

first draft of using https://bnd.bndtools.org/plugins/p2export.html for bnd itself the goal is to replace the veery old way of building the p2 artifact and use bnd itself.

Importantly this solves the issue to use SHA-256 for checksum in addition to md5 to solve the Eclipse warning like "The digest algorithms (md5) used to verify ...jar have severely compromised security. Please report this concern to the artifact provider."

PR #6808, opened 2025-09-12
use P2Exporter for org.bndtools.p2

### LINKED ISSUE #5156: org.bndtools.p2: Update the build for the p2 repository to replace md5 checksum

chrisrueger:
Closes #5156

- use bnd's own P2Export https://bnd.bndtools.org/plugins/p2export.html for bnd itself the goal is to replace the veery old way of building the p2 artifact and use bnd itself.

Importantly this solves the issue to use SHA-256 for checksum in addition to md5  to solve the Eclipse warning like "The digest algorithms (md5) used to verify ...jar have severely compromised security. Please report this concern to the artifact provider." (#5156 )

Note: This PR is identical to (successor of) https://github.com/bndtools/bnd/pull/6777 but this time created inside bndtools organization. This is needed so that we can use gpg variables for the P2 Signing


org.bndtools.p2/bndtools.ecf.feature.bndrun
```diff
@@ -0,0 +1,20 @@
...
+    feature;name=org.eclipse.ecf.core.feature;version="[0.0.0,1000)",\
+    feature;name=bndtools.main.feature,\
+    feature;name=org.eclipse.ecf.discovery.feature;version="[0.0.0,1000)",\
+    feature;name=org.eclipse.ecf.discovery.jmdns.feature;version="[0.0.0,1000)",\
+    feature;name=org.eclipse.ecf.osgi.services.feature;version="[0.0.0,1000)",\
+    feature;name=org.eclipse.ecf.provider.generic.feature;version="[0.0.0,1000)",\
+    feature;name=org.eclipse.ecf.provider.generic.remoteservice.feature;version="[0.0.0,1000)",\
+    feature;name=org.eclipse.ecf.remoteservice.feature;version="[0.0.0,1000)",\
+    feature;name=org.eclipse.ecf.remoteservice.sdk.bndtools.feature;version="[0.0.0,1000)",\
+    feature;name=org.eclipse.ecf.sharedobject.feature;version="[0.0.0,1000)"
```
chrisrueger: Hi @scottslewis , could you please go through those features and suggest more sensible version ranges? My current hack of `version="[0.0.0,1000)` is too broad and might introduce trouble as unrestricted version ranges typically tend to do. I just did it to mimic the existing behavior you had in https://github.com/bndtools/bnd/blob/master/org.bndtools.p2/features/bndtools.ecf/feature.xml#L511

I think it would be good to limit it to a major version range something like `[3,4)`(just an example) or something like that.

Or what do you think?

chrisrueger:
I will keep an eye on it and revert if there is something wrong. I will focus on the setup of P2 PGP Signing in a separte PR.

PR #6824, opened 2025-09-21
include files in default package for -includepackage *;from:=project

### LINKED ISSUE #6821: Resource at root of source folder is excluded from bundle
I have a Bnd workspace where bundles have these properties, among others:

```
tool-modifier:          ${if;${is;${driver};gradle};/gradle}
target-dir:             target
src:                    src
bin:                    ${target-dir}${tool-modifier}/classes
```

There is no Maven-style separate source folder for resources.

I have a bundle with these files:

```
src/
   about_files/
      … some resources
   com/
      ... more package levels and Java source files
   logback.xml
```

When I build the bundle with Bnd 7.1.0, `logback.xml` is missing.

Experimenting, I find that all files directly under `src/` are missing from the bundle, but files in subdirectories are not missing.

This worked OK in Bnd 6.4.0.

For now, I'm working around it with `-includeresource: src/logback.xml` but this shouldn't be required.

chrisrueger:
Closes #6821

Fixes a problem (regression of #5522 ?) that files in default package where not included in the built jar

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/BuilderTest.java
+++ b/biz.aQute.bndlib.tests/test/test/BuilderTest.java
@@ -172,6 +172,24 @@ public class BuilderTest {
 	}
 
 	/**
+	 * This tests that files in default project e.g.
+	 * /src/file-in-default-package.txt are also included with
+	 * -includepackage=Constants.ALL_FROM_PROJECT
+	 */
+	@Test
+	public void testIncludeWholeProject2() throws Exception {
+		try (Builder b = new Builder()) {
+			Jar jar = new Jar(IO.getFile("jar/osgi.jar"));
+			jar.putResource(Constants.PROJECT_MARKER, new EmbeddedResource("", 0L));
+			jar.putResource("file-in-default-package.txt", new EmbeddedResource("", 0L));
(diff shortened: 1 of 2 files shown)


chrisrueger:
Mentioning @juergen-albert @stbischof @bjhargrave @pkriens  since you were active in #5407 (which was the issue closed by #5522 )

bjhargrave:
No one using bnd should be using the default package for anything. OSGi and default package do not work together.

chrisrueger:
> No one using bnd should be using the default package for anything. OSGi and default package do not work together.

Thanks. I guess we then call that a breaking change from 6.4.x to 7.x

@ejjcase

ejjcase:
> No one using bnd should be using the default package for anything. OSGi and default package do not work together.

My project is a fragment whose host is `ch.qos.logback.classic`, and that is where Logback expects to find the file. Why on earth shouldn’t I do this in bnd?

Or what if the resource was an Eclipse `plugin.xml`?

It’s not like I want to export the default package.

I find it highly counterintuitive and annoying that adding the bnd plugin to a Gradle project now breaks something that Just Works in Gradle without bnd.

PR #6826, opened 2025-09-22
P2Export optimizations

chrisrueger:
- add proper manifest in p2 bundle
- remove the hack
- add p2.index file
- put pgp.publicKeys once in <repository> instead of redundantly in each artifact. makes the xml file smaller


chrisrueger:
I am going to merge because the rebuild error will correct itself with the next PR

PR #6847, opened 2025-10-01
new -nosubstitution instruction to disable package substitution globally (comparable to -noimport:=true)

### LINKED ISSUE #6844: Add support to disable substiution packages
https://bnd.bndtools.org/chapters/170-versioning.html#substitution describes how bnd currently handle substitution packages. Often one do not want to use substitution packages at all and then needs to disable them on each export.

It would be more convenient if substitution package handling must be explicitly enabled (or can be completely disabled).

e.g. one could have an instruction `-substitution: true/false` that allows to completely disable it (=`false`) or enable it like before (=`true`)

chrisrueger:
Closes #6844

This is about https://bnd.bndtools.org/chapters/170-versioning.html#substitution

Setting this new instruction to `-nosubstitution:true` disables package substitution globally (**default** is `false`). That means, that bnd does not calculate `Import-Package` references for any package in exports.

This is equivalent to `Export-Package: *;-noimport:=true`  or `-exportcontents: *;-noimport:=true`

Note: This is an optional / opt-in feature. Nothing changes if you do not use it. 

## Why is it useful?

This new instruction is useful for library authors who **never** want to have any exported packages in Import-Package ([example](https://github.com/cucumber/cucumber-parent/pull/61/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8R257))

## Example

```
-nosubstitution:true
Export-Package: *
Import-Package: *
```

or 

```
-nosubstitution:true
-exportcontents: *
Import-Package: *
```

This ensures that no `Import-Package` is calculated for any package in `Export-Package`

The advantage is that `-nosubstitution:true` is a global switch, thus a developer cannot forget to add `-noimport:=true` in case of a more fine-grained `Export-Package` declaration with multiple packages.

```
-nosubstitution:true
Export-Package: a;-noimport:=true,\
                              b,\ # here -noimport:=true,\ was forgotten
                              c # here -noimport:=true,\ was forgotten
Import-Package: *
```

In the latter example even though `-noimport:=true` directive was forgotten, the `-nosubstitution:true` takes care of it. 

@laeubi could give additional examples on use cases

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/BuilderTest.java
+++ b/biz.aQute.bndlib.tests/test/test/BuilderTest.java
@@ -1719,6 +1719,102 @@ public class BuilderTest {
 	}
 
 	@Test
+	public void testSubstitutionEnabled() throws Exception {
+
+		try (Builder b = new Builder()) {
+			b.setProperty(Constants.NOSUBSTITUTION, "false");
+			b.setProperty("Export-Package", "org.osgi.util.measurement, org.osgi.service.http;");
+			b.setProperty("Private-Package", "org.osgi.framework, test.refer");
+			b.addClasspath(IO.getFile("jar/osgi.jar"));
+			b.addClasspath(new File("bin_test"));
+			Jar jar = b.build();
+			assertTrue(b.check());
+
+			Manifest m = jar.getManifest();
+			String imports = m.getMainAttributes()
+				.getValue("Import-Package");
+			String exports = m.getMainAttributes()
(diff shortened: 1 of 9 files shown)


laeubi (approved):
Looks good to me!

chrisrueger:
Renamed it to `-nosubstitution: true` after feedback.

chrisrueger:
Going to merge since no objections in the recent weeks.

Issue #6651, opened 2025-06-30
[junit-platform] Tester incompatible with JUnit Platform 1.13.x


kriegfrj:
The adoption of this PR in JUnit Platform 1.13 has broken `BundleEngine`:

junit-team/junit-framework#4281

For historical reasons, the BundleEngine implementation has relied on an API that was flagged as "Internal". It seems that this dependency has come back to bite us (see also #6614). If you try to run tests with Jupiter 5.13, it will fail to load because the NamespacedHierarchicalStore is not supplied to the sub-engines.

On the bright side, it could be that a lot of the jiggery-pokery that is present in BundleEngine is no longer necessary. Much of said jiggery-pokery was to work around the fact that JUnit-Platform was not classloader aware and would naively resolve ClassSelectors and MethodSelectors using the default or system classloader. A lot of these issues have since been fixed, so it could be that a lot of what is being done in `BundleEngine` is no longer necessary and the appropriate methods in core JUnit-Platform can be used directly.

At some point soon it would be worth looking at this again to see if we can re-implement BundleEngine in a way that makes use of public APIs (and therefore hopefully avoid future nasty surprises like this one).

kriegfrj:
I have had an initial look at this, and unfortunately it doesn't look like the solution will be trivial. Or perhaps (as is so often the case in this industry) the solution will turn out to be trivial but finding said solution has already shown to be a non-trivial exercise.

I need to get back to my "real work" so I think I'm going to have to leave this one for now. Hopefully the abillity for us to upgrade at least to 5.12.2 is a good improvement in the meantime.

PR #6854, opened 2025-10-04
Upgrade to Eclipse 2023-12 (e.g. for Java 21 Execution Environment support)

chrisrueger:
This upgrades the Eclipse version used by bndtools' dev instance from Eclipse 2022-12 to 2023-12. 
For me mainly because I want to test a Java 21 project in the dev instance when contributing to bndtools and 2023-12 is the first version to support that Execution-Environment. This makes this process a lot easier. And the other reason is just to keep constantly upgrading the Eclipse instance.

In this PR my main issues I had with JUnit again to get all the testcases running.


chrisrueger:
@kriegfrj Could you have a look please? Especially because one thing I needed to do was to "downgrade"

```
junit.jupiter.version=5.12.2
junit.platform.version=1.12.2
```
to 
```
junit.jupiter.version=5.10.1
junit.platform.version=1.10.1
```

in [junit.bnd](https://github.com/bndtools/bnd/pull/6854/files#diff-098e0861cd307b703acf27acb89560edf8bea267d2add2be2e758526215aabf0R21-R24)

This was the final piece in my puzzle to make all tests pass again. 

If I did not do it I had the problem that somehow a mix of 5.12.2 and 5.10.1 was pulled which caused test case failures. But since you recently just bumped to 5.12.1 I wanted to ask you.

kriegfrj:
Hi @chrisrueger ! Thanks again for your hard work on this.

There are two versions of JUnit used by the tester project:

1. A minimum supported version of JUnit 5 (`junit.platform.tester.version` & `junit.jupiter.tester.version`)
2 The maximum supported version of JUnit 5, which is also the version used by the Bnd workspace's own tests (`junit.platform.version` & `junit.jupiter.version`). 

The minimum version is used as the compilation version for `biz.aQute.tester.junit-platform`. both are used in two sets of tests - one set that tests the tester against the highest version, and one that tests it against the lowest version.

It should not be necessary for you to change the highest version (if anything, we should be trying to bump this version higher to the latest release version of JUnit, if we weren't blocked by #6651). It should be sufficient to bump the _minimum_ supported version to the minimum required by Eclipse 2023-12. In other words, bump `junit.platform.tester.version` and `junit.jupiter.tester.version` to 1.10.1/5.10.1 but leave `junit.platform.version` and `junit.jupiter.version` at 1.12.2/5.12.2. 

If this does not work for you, let me know and I will try and take a look.

chrisrueger:
thanks @kriegfrj 

I reverted the version back to 1.12.2/5.12.2 as you suggested.
I also did a few other modifications to fix the build again. 

If you check out this PR you can run:

`./gradlew :biz.aQute.tester.test:test`

Your help is really appreciated.

UPDATE: I modified this comment as I had a couple of false positive tries and my initial comment is not valid anymore. So I removed most of it.

gradle-plugins/biz.aQute.bnd.gradle/build.gradle.kts
```diff
@@ -77,7 +77,7 @@ dependencies {
 	implementation("biz.aQute.bnd:biz.aQute.resolve:${version}")
 	runtimeOnly("biz.aQute.bnd:biz.aQute.bnd.embedded-repo:${version}")
 	// keep in sync with cnf/junit.bnd e.g. 'junit.jupiter.version'
-	testImplementation(enforcedPlatform("org.junit:junit-bom:5.12.2"))
+	testImplementation(enforcedPlatform("org.junit:junit-bom:5.10.1"))
```
kriegfrj: If this change is necessary, then we need to change the comment: `eg 'junit.jupiter.tester.version'`

However, this again feels a bit wrong. Upgrading the Eclipse baseline shouldn't require us to downgrade JUnit.

chrisrueger: Thanks. I will have a look. Maybe something I forgot to revert

chrisrueger: @kriegfrj just looked. Could it be that you reviewed an outdated commit? it is already at org.junit:junit-bom:5.12.2

kriegfrj: Yes, you are correct - sorry - I didn't realise I was looking at an old commit.

cnf/ext/junit.bnd
```diff
@@ -37,9 +37,9 @@ junit: ${junit-osgi},\
 	org.osgi.framework;version='[1.8,1.9)'
 
 junit-osgi: \
-	junit-platform-commons;version=latest,\
-	junit-platform-engine;version=latest,\
-	junit-platform-launcher;version=latest,\
+	junit-platform-commons;version='${range;[===,==+);${junit.platform.version}}',\
+	junit-platform-engine;version='${range;[===,==+);${junit.platform.version}}',\
+	junit-platform-launcher;version='${range;[===,==+);${junit.platform.version}}',\
```
kriegfrj, line 42: I think this change is fairly harmless, but I'm curious what your thought process was, "latest" wasn't good enough?

biz.aQute.tester.test/bundleenginetest.bndrun
```diff
@@ -1,5 +1,6 @@
 -include: bundleenginetest-noengines.bndrun
 
 -runbundles.engines: \
-	junit-jupiter-engine;version='[5.9.1,5.9.2)',\
-	junit-vintage-engine;version='[5.9.1,5.9.2)'
+	junit-jupiter-engine;version='[5.10.1,5.10.2)',\
+	junit-vintage-engine;version='[5.10.1,5.10.2)',\
+	junit-platform-engine;version='${range;[===,==+);${junit.platform.tester.version}}'
```
kriegfrj: This last line one should not be necessary. The launchpad setup causes the framework-under-test to link to the junit-platform-engine that is used on the classpath that launches the framework (hence the `-runsystempackages` declarations in `bundleenginetest-noengines.bndrun`).

kriegfrj (commented):
Looks good! A couple of suggestions where I think yo

cnf/ext/junit.bnd
chrisrueger: I can't remember anymore. I think on my way to the current state I was experimenting. I also remember that in my recent attempts to Upgrade to a higher Eclipse 2024-XX I had trouble with "latest" because at some point the JUnit Versions the Eclipse repo brings in, suddeny become "latest" (while before they were lower than our own in Maven Repo). 
So I thought using an explicit range based on the variables in `junit.bnd` is more predictable.

biz.aQute.tester.test/bundleenginetest.bndrun
chrisrueger: Thanks, I will try building without it.

kriegfrj: I forgot to tell you that I tried building without it and it worked.

cnf/ext/junit.bnd
kriegfrj: Yes, if you upgraded to a version of Eclipse that pulls in a JUnit Platform 1.13 or higher, then `latest` would try and use that, which would cause issues due to #6651.

biz.aQute.tester.test/bundleenginetest.bndrun
chrisrueger: @kriegfrj you are right. I removed it and still works.

cnf/ext/junit.bnd
chrisrueger: So, do we leave it as is with the range? or should I revert to `latest`? Any other remarks?

As soon as we agree, then I will squash all commits.

kriegfrj: Let's leave the range in.

chrisrueger:
I squashed all commits. Ready to go, if nothing else comes up 😄

kriegfrj:
Great work!

chrisrueger:
Thanks for the support and testen 👍

PR #6868, opened 2025-10-16
Do not concat Bundle-Activator when already defined

### LINKED ISSUE #6867: Bundle-Activator errors when dependencies contain Activators
Creating a fat jar with Gradle and the bnd plugin fails with the message:

```
error: The Bundle-Activator header only supports a single type. The following types were found: org.example.Activator,org.apache.logging.log4j.util.Activator. This usually happens when a macro resolves to multiple types
```

A similar issue has already been reported in https://github.com/bndtools/bnd/issues/5257 but is still broken in the current release of bnd (7.1.0).

I have attached a sample project that creates a fat jar by combining the log4j library with a custom Bundle-Activator. The issue can be reproduced by simply calling 

```gradlew  jar```

This yields the above error message because bnd finds a Bundle-Activator in log4j although one is explicitly specified in the bnd instructions.

```
tasks.named("jar", Jar) {
    from(configurations.shadow.elements.map {it.collect{zipTree(it)}})
    bundle {
        bnd(
                'Bundle-Activator': 'org.example.Activator',
                'Import-Package': '*',
                'Export-Package': 'org.example.*',
                '-removeheaders': 'Private-Package,Require-Capability',
                '-nodefaultversion': true,
                '-nouses': true,
                '-noimport': true,
                '-fixupmessages': '*wrong directory*' // apparently bnd cannot deal with multi-release JARs
        )
    }
}
```

[bnd-example.zip](https://github.com/user-attachments/files/22925817/bnd-example.zip)

chrisrueger:
Closes #6867

fixes a problem that `Verifier.verifyActivator()` throws an error "The Bundle-Activator header only supports a single type. The following types were found" when the bnd instructions already define a Bundle-Activator but a processed jar also contributed a Bundle-Activator (e.g. via `@Header(name = Constants.BUNDLE_ACTIVATOR, value = "${@class}")` annotation). 

see [org.apache.logging.log4j.util.Activator](https://github.com/apache/logging-log4j2/blob/2.x/log4j-api/src/main/java/org/apache/logging/log4j/util/Activator.java#L49)

image

Before this fix the Bundle-Activator header was concatenated to the existing `Bundle-Activator` header in the outer bnd of the bundle and thus resulted in multiple values (separated by comma). But for Bundle-Activator this does not make sense as it supports only a single value, as there can only be one single Bundle-Activator in the manifest - which is correctly complained about by `Verifier.verifyActivator()`.

This PR fixes the root cause where the multiple-values are created, so that `Verifier.verifyActivator()` does not complain anymore.

This for example happened with `-includeresource` of a Bundle (unrolling or lib:=true) or other ways where bnd scans classes of jar files.

I also identified some other Bundle headers in which should also only contain a single value and not be concatenated.

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/BuilderTest.java
+++ b/biz.aQute.bndlib.tests/test/test/BuilderTest.java
@@ -3187,6 +3187,86 @@ public class BuilderTest {
 		}
 	}
 
+	/**
+	 * Check that the defined activator is found, and not concatenated by the
+	 * Bundle-Activator annotation in TypeInVersionedPackage
+	 *
+	 * @throws Exception
+	 */
+	@Test
+	public void testBundleActivatorShouldNotBeConcatenated() throws Exception {
+		Builder bmaker = new Builder();
+		try {
+			bmaker.setProperty("Bundle-Activator", "test.activator.Activator");
+			bmaker.setProperty("build", "xyz"); // for @Version annotation
+			bmaker.setProperty("Private-Package", "test.*");
+			bmaker.setProperty("-bundleannotations",
+				"test.annotationheaders.attrs.std.activator.TypeInVersionedPackage");
+			// bmaker.setProperty("-dsannotations", "!*");
+			// bmaker.setProperty("-metatypeannotations", "!*");
+			bmaker.setClasspath(new File[] {
+				new File("bin_test")
+			});
+			bmaker.setProperty("-fixupmessages.export",
+				"The annotation aQute.bnd.annotation.Export applied to package test.versionpolicy.api is deprecated and will be removed in a future release. The org.osgi.annotation.bundle.Export should be used instead");
+			bmaker.setProperty("-fixupmessages.directive",
+				"Unknown directive foobar: in Export-Package, allowed directives are uses:,mandatory:,include:,exclude:,-import:, and 'x-*'");
+
+			Jar jar = bmaker.build();
+			report("testBundleActivatorShouldNotBeConcatenated", bmaker, jar);
+
+			Attributes main = jar.getManifest()
+				.getMainAttributes();
+			assertEquals("test.activator.Activator", main.getValue(Constants.BUNDLE_ACTIVATOR));
+			assertFalse(
+				bmaker
+					.check(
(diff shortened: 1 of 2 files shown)


chrisrueger:
Any thoughts @pkriens @peterkir ?

Do you think there are other headers too, which should have only a single value, if already defined? Or is Bundle-Activator the only header like this?

Update: Added a few other headers which only allow a single value

chrisrueger:
Note: A related issue which touches  a similar topic https://github.com/bndtools/bnd/issues/5363 (and non-merged PR https://github.com/bndtools/bnd/pull/5389 )

From there it also becomes apparent that `Require-Capability` & `Provide-Capability` are the main targets of that method. 
That means, that the current fix is at least **not** wrong.

stbischof:
plans to merge, what blocks?

chrisrueger:
> what blocks?

The lack of feedback. And the controversy in the issue discussion. I think the PR fixes the problem reported in the #6868 . If it fixes it in the best way - I don't know. 

I will try to bring it up in the next bnd zoom call

stbischof:
+1

juergen-albert:
The Mechanism should also make sure, that any bundle activator header is not just a singlton, but that the property in the bnd file wins, so I can overwrite whatever included classfiles may define here.

chrisrueger:
> that the property in the bnd file wins, so I can overwrite whatever included classfiles may define here.

That is the exact purpose of that PR. But only for the properties in the new Set `BUNDLE_HEADERS_SINGLE_VALUE` I defined (which includes `Bundle-Activator`).

image

https://github.com/bndtools/bnd/pull/6868/files#diff-be27a5a501f05ec3c1bc10a14a8bed0592a76cd02407998f98792675cd2b60c5R153-R156

Is that what you mean? @juergen-albert

chrisrueger:
I renamed the private variable to `AnnotationHeaders.BUNDLE_HEADERS_ANALYZER_WINS` and added some explanation. I hope this makes the intention clearer. Since it's all private people can come up with better wording in the future.

stbischof:
Here is an example where i would use this 

https://github.com/eclipse-osgi-technology/command/blob/main/all/src/main/java/org/eclipse/osgi/technology/command/all/Activator.java#L26

chrisrueger:
I discussed with @peterkir that this is good enough to be merged. It shouldn't break anything, and fixes the issue.

PR #6859, opened 2025-10-06
more lenient osgi.ee=JavaSE for unknown JDKs

chrisrueger:
Closes #6858

**Experimental:**

For a yet to bnd unknown JDK (e.g. JDK-10000) Instead of

```
Require-Capability: osgi.ee;filter:="(osgi.ee=UNKNOWN)"
```

we now create

Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=9955))"

for a  .class file with u2 major_version 10000 (9955 == 10000 - 45)

Issue #6858, opened 2025-10-06
Can the osgi.ee=JavaSE be more lenient on unknown JDK versions?


ben-manes:
As an OSGi novice, I was very confused when my tests failed on JDK-25 and the Pax Exam team kindly helped me [resolve](https://github.com/ops4j/org.ops4j.pax.exam/issues/1131#issuecomment-3370313321) that setting the `Require-Capability` explicitly. Since this is just the Java version can that be added dynamically instead of `osgi.ee;filter:="(osgi.ee=UNKNOWN)"`? Then the bnd plugin is future compatible. While I release on JDK-11, the tests failed due to cross-building on the latest JDK.

```ini
# Java 24
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=24))"

# Java 25
Require-Capability: osgi.ee;filter:="(osgi.ee=UNKNOWN)"
```

I manually forced the correct value using,
```kts
// Workaround until the bnd plugin supports the latest JDK
val javaVersion = java.toolchain.languageVersion.get()
if (javaVersion.canCompileOrRun(25)) {
  bnd(mapOf("Require-Capability" to
    """osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=${javaVersion}))""""))
}
```

chrisrueger:
I think this will be fixed in upcoming bnd 7.2.0 which added support for JDK24 and JDK25.

- https://github.com/bndtools/bnd/pull/6529/files#diff-85fbc89f0c6509380298a0e20b8a5520338a927f87b7f85ff882a678c6cbf35e
- https://github.com/bndtools/bnd/pull/6672

ben-manes:
Great, but won’t it reoccur in 26, etc? Is there a reason to require a parallel release? That change adds up to 30 which is nice, but seems like it should be dynamic?

chrisrueger:
Well good question. I remember the consensus was to increase the JAVA constants up to 30. I just tried to follow the existing pattern. 

@bjhargrave @pkriens any comments?

chrisrueger:
I created an experimental PR to adress this https://github.com/bndtools/bnd/pull/6859

But I am not sure yet, if this "dynamic" is a good idea.
[...]

PR #6891, opened 2025-10-27
#6858 add enums up to JDK50

chrisrueger:
"hack" (alternative to https://github.com/bndtools/bnd/pull/6859 ) to postpone the problem of #6858 a bit longer until we agree on a more generic solution. Also update the `EE.MAX_SUPPORTED_RELEASE` variable which was forgotten previously


chrisrueger:
@laeubi  fine with this "predicting the future" as a quick hack to buy us some more time?

PR #6895, opened 2025-10-28
effective view: fix incorrectly merged props and add support for decorated instructions

### LINKED ISSUE #6893: In Bndrun editor "Effective" tab, version property is mistakenly shown as merged
My project has some properties called `version.foo` which control the versions of various dependencies in the repositories.

```
version.eclipse.base: 2.64.0-dev.646364
version.gecko: 1.2.1-dev.652527
version.shared: 1.3.1
```
etc.

I also have a plain `version` property. `version` comes from `gradle.properties` which is included by a bnd file in `cnf/ext`, and the other version properties are all defined in a different bnd file in `cnf/ext`.

In the "Effective" tab of a bndrun editor for a bndrun that inherits all of these, when I check "Show as merged properties", all these values are shown as merged:

```
version: 2.64.0-local,2.64.0-dev.646364,1.2.1-dev.652587,1.3.1,...
```
Bnd itself correctly does not merge these properties. If it did, I could rename them.

It's not a major problem since it's only in the UI, but it is confusing to read when it doesn't match Bnd's actual behaviour.

chrisrueger:
Closes #6893

- only show properties as "merged" which are in Constants.MERGED_HEADERS

## Merged Properties and instructions

Now MERGED_HEADERS and instructions (properties starting with `-` as in `-buildpath`) are properly merged. 
E.g.

For 

```
version: 1.2.3
version.eclipse.base: 2.64.0-dev.646364
version.gecko: 1.2.1-dev.652527
version.shared: 1.3.1
```

image

the single 'version' is now not merged anymore. (because it is NOT contained in `Constants.MERGED_HEADERS`)

While 

```
-buildpath: foo
-buildpath.a: bar
```

is still merged

image

because it is contained in `Constants.MERGED_HEADERS`

https://github.com/bndtools/bnd/blob/cbc7b063f023d0b17eec2397912355c0a4ad119c/biz.aQute.bndlib/src/aQute/bnd/osgi/Constants.java#L611

## Decorated instructions

This PR also handles [decorated instructions](https://bnd.bndtools.org/chapters/820-instructions.html#decorated-instructions) and displays them properly merged. 

e.g.

```
-foo       a, b, c;skip=true, d
-foo+      b;skip=true,c;skip=false
-foo+.d    d;skip=true
```

is now displayed as 

```
-foo: a,b;skip=true,c;skip=false,d;skip=true
```

image

Note: I also fixed the hover feature in source and effective view which had the same issue

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
@@ -1602,11 +1602,20 @@ public class BndEditModel {
 	}
 
 	public static String getStem(String header) {
+		String stem;
 		int n = header.indexOf('.');
 		if (n < 0)
-			return header;
+			stem = header;
+		else
+			stem = header.substring(0, n);
 
-		return header.substring(0, n);
+		// remove any trailing +, ++ (decorated instructions)
+		int m = stem.indexOf('+');
+		if (m < 0)
+			return stem;
+		else
+			return stem.substring(0, m);
+
 	}
 
 	@SuppressWarnings("unchecked")
--- a/biz.aQute.bndlib/src/aQute/bnd/help/Syntax.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/help/Syntax.java
@@ -1134,4 +1134,8 @@ public class Syntax implements Constants {
 		return ret.toLowerCase()
 			.replace("-", "_");
 	}
+
+	public static boolean isInstruction(String key) {
+		return key.startsWith("-");
+	}
 }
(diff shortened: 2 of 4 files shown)


chrisrueger:
@ejjcase Is that what you had in mind ?

chrisrueger:
Found another problem  the [decorated instructions](https://bnd.bndtools.org/chapters/820-instructions.html#decorated-instructions) are currently not properly handled by the Effective view

ejjcase:
> @ejjcase Is that what you had in mind ?

@chrisrueger I think so, thank you.

chrisrueger:
@ejjcase let's give it a try. new snapshot should be available in about 30minutes

ejjcase:
@chrisrueger The fix works for me, thank you!

chrisrueger:
@ejjcase just mentioning that I pushed another quickfix this morning... in case you got a NPE while hovering in the source view, get the latest snapshot again.

ejjcase:
I don't think I checked until after you pushed your quickfix. I didn't see
an NPE today.

Issue #6894, opened 2025-10-28
Can't combine literal backslash with macro expansion


ejjcase:
I understand that `\` is the escape character in the Bnd macro language, and that the way to write a literal like `${foo}` is `\${foo}`. But I can't seem to write a macro that expands to a property value following a literal `\` character.

Source:
```
a: 42
a0: ${a}
a1: \${a}
a2: \\${a}
a3: \\\${a}
a4: \\\\${a}
a5: \\\\\${a}
a6: \\\\\\${a}
a7: \\\\\\\${a}
a8: \\\\\\\\${a}
a9: \\\\\\\\\${a}
```
Expanded values in the "Effective" tab:
```
a: 42
a0: 42
a1: 42
a2: ${a}
a3: ${a}
a4: \${a}
a5: \${a}
a6: \\${a}
a7: \\${a}
a8: \\\${a}
a9: \\\${a}
```
I can't find a number of `\` characters that expands to a literal  `\` without also escaping the `$`. It's like the same escape character is sometimes used twice.

chrisrueger:
That sounds like a question for @pkriens

ejjcase:
I suppose I could use `${separator}` for the backslash (though only on Windows, and I can't work out how to define my own property with the same value because of how backslashes are treated at the ends of lines).

But that breaks again if I put the expression inside an `if` or similar.

chrisrueger:
Could you maybe outline what you try to accomplish? Also I remember @juergen-albert often had some useful macro magic tricks in some forum discussions.
Other that that there is [${pathseparator}](https://bnd.bndtools.org/macros/pathseparator.html) which seem to handle different platforms.

ejjcase:
@chrisrueger First, I think you mean `${separator}` because `${pathseparator}` gives `;` or `:`.

You're right that I was trying to construct a Windows file path. I eventually realised I could just use `/` and Bnd and Eclipse would understand it just fine, even on Windows. So I don't really need any macro magic, but I reported this anyway because it strikes me as a symptom of how something in the macro processor is broken.

The same syntax works in Kotlin strings, so I tried it in Kotlin:
```kotlin
    val a = 42
    println("${a}")
    println("\${a}")
    println("\\${a}")
    println("\\\${a}")
    println("\\\\${a}")
    println("\\\\\${a}")
    println("\\\\\\${a}")
    println("\\\\\\\${a}")
    println("\\\\\\\\${a}")
    println("\\\\\\\\\${a}")
```
[...]

PR #6904, opened 2025-10-29
Macro: fix for dollar escaping

chrisrueger:
For now this is a base for discussion of my comment https://github.com/bndtools/bnd/issues/6894#issuecomment-3463694757

It fixes a parsing problem for escaped dollar characters in macros.
But the problem is that - while this works for the testcase - this fix is never used in real-life because java Properties parsing drops backslashes in the first place (see spec https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Properties.html#load(java.io.Reader)) 

> The method does not treat a backslash character, \, before a non-valid escape character as an error; the backslash is silently dropped. For example, in a Java string the sequence "\z" would cause a compile time error. In contrast, this method silently drops the backslash. Therefore, this method treats the two character sequence "\b" as equivalent to the single character 'b'.

Since I have no idea how to handle this I keep this PR as draft.

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/MacroTest.java
+++ b/biz.aQute.bndlib.tests/test/test/MacroTest.java
@@ -12,6 +12,7 @@ import static org.junit.jupiter.api.condition.OS.WINDOWS;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.StringReader;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


chrisrueger:
Closes because of https://github.com/bndtools/bnd/issues/6894#issuecomment-3464106362 
alternative is https://github.com/bndtools/bnd/pull/6905

PR #6905, opened 2025-10-30
source editor: warning on single backslashes

chrisrueger:
For https://github.com/bndtools/bnd/issues/6894
Single backslashes are silently dropped by Java Properties parsing (https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Properties.html). 

This causes surprising behavior since bnd allows that they can be entered in the source editor. This is an attempt to show red markers and a warning for such illegal single backslashes. the only valid single backslash is still the line-continuation (for multiline support)

Now it looks like this for a single backslash (or odd number of consecutive backslashes)

image

image

Backslashes followed by a space or tab are still invalid as before

image

image

We also modified the PropertiesParser and show a warning in Eclipse.

image

Disclaimer: I used AI in the process to help with the parsing.

### DIFF
--- a/docs/_chapters/125-tour-features.md
+++ b/docs/_chapters/125-tour-features.md
@@ -71,6 +71,28 @@ Though lines can be as long as you want, it is better for your brains to keep th
 
 Since bnd files are property files, you cannot repeat a property. Later properties will completely overwrite earlier ones and there is no order between properties.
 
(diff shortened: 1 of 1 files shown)


chrisrueger:
I would appreciate a review: @pkriens @peterkir @bjhargrave @laeubi 

I know this is pretty core, thus I kept it in separate commits and I would encourage not squashing in this case, which allows us easier roll-back. The commits are separated by the idea:

- Eclipse UI red marking of text in bnd source editor (`BndScanner.java`)
- Warning message: `PropertiesParser.java`
- docs

Note: I think the failing builds are some kind of github or maven deps hickup. I am restarting them one by one now and they seem to work.

PR #6911, opened 2025-10-31
Enhance template fragment wizard

peterkir:
- Change default behavior to apply fragments in existing eclipse workspace
- increased wizard size and some text for clearer user experience
- added tooltip


chrisrueger (commented):
Tested briefly. Looks good and works. 
DCO reminder 😄

PR #6914, opened 2025-11-04
Don't use the Project or Task at execution time

ejjcase:
Work in progress.

One thing that strikes me is that if we're going to read the bndfile at task configuration time, the entire chain of included bndfiles will need to be inputs. This was already a bug (not sure if it was reported) but will become more important now.

#6346

### DIFF
--- a/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
+++ b/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
@@ -154,7 +154,8 @@ public class BundleTaskExtension {
 	 * Properties of the Gradle project and task that are referenced in the
 	 * bndfile or in {@link #getProperties()}.
 	 * <p>
-	 * The following properties are set by the builder and are also available:
+	 * The following properties are set by the builder at runtime and are also available:
+	 * </p>
 	 * <dl>
 	 * <dt>{@code project.dir}</dt>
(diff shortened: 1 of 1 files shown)


ejjcase:
Current test results:
```
> Task :gradle-plugins:biz.aQute.bnd.gradle:test

TestBundlePlugin > Configuration Cache Support Test FAILED
    Condition not satisfied:

    result.task(":bundle").outcome == UP_TO_DATE
    |      |               |       |
    |      :bundle=SUCCESS SUCCESS false
    <org.gradle.testkit.runner.internal.FeatureCheckBuildResult@6fa2448b delegateBuildResult=org.gradle.testkit.runner.internal.DefaultBuildResult@5c53f292 outputFeatureCheck=org.gradle.testkit.runner.internal.feature.BuildResultOutputFeatureCheck@46fb0c33>
        at aQute.bnd.gradle.TestBundlePlugin.Configuration Cache Support Test(TestBundlePlugin.groovy:412)
```
The `bundle` task is not up to date on the second run because `project.buildpath` now includes `builderplugin4/build/libs/builderplugin4-1.0.0.jar`, which must have been created during the first run.

Since calculating the buildpath doesn't use anything we can't access at runtime, maybe I'll go back to setting it at runtime.

ejjcase:
> Since calculating the buildpath doesn't use anything we can't access at runtime, maybe I'll go back to setting it at runtime.

Done. Now all tests are passing for me locally.

bjhargrave:
Closing as #6915 was merged to fix the issue.

Thanks for taking a look into this!

PR #6915, opened 2025-11-04
Avoid using Gradle Project object at execution time

bjhargrave:
We use the new getMacroReferences method to determine which properties
are referenced in the bnd instruction so that we can populate any
reference project properties in the provider.

Fixes https://github.com/bndtools/bnd/issues/6346

### DIFF
--- a/.github/scripts/rebuild-build.sh
+++ b/.github/scripts/rebuild-build.sh
@@ -2,8 +2,5 @@
 set -ev
 ./gradlew --no-daemon --version
 ./gradlew --no-daemon -Dmaven.repo.local=dist/m2 :buildscriptDependencies :publish
-# Temporarily ignore warnings like "Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0."
-# so we use --warning-mode=all instead of --warning-mode=fail
-#./gradlew --no-daemon -Dmaven.repo.local=dist/m2 --warning-mode=fail :gradle-plugins:build
-./gradlew --no-daemon -Dmaven.repo.local=dist/m2 --warning-mode=all :gradle-plugins:build
+./gradlew --no-daemon -Dmaven.repo.local=dist/m2 --warning-mode=fail :gradle-plugins:build
 ./gradlew --no-daemon -Dmaven.repo.local=dist/m2 :gradle-plugins:publish
--- a/.github/scripts/rebuild-test.sh
+++ b/.github/scripts/rebuild-test.sh
@@ -1,7 +1,4 @@
 #!/usr/bin/env bash
 set -ev
 ./gradlew --no-daemon --version
-# Temporarily ignore warnings like "Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0."
-# so we use --warning-mode=all instead of --warning-mode=fail
-#./gradlew --no-daemon -Dmaven.repo.local=dist/m2 -Pbnd_snapshots=./dist/bundles --warning-mode=fail :buildscriptDependencies :build :publish "$@"
-./gradlew --no-daemon -Dmaven.repo.local=dist/m2 -Pbnd_snapshots=./dist/bundles --warning-mode=all :buildscriptDependencies :build :publish "$@"
+./gradlew --no-daemon -Dmaven.repo.local=dist/m2 -Pbnd_snapshots=./dist/bundles --warning-mode=fail :buildscriptDependencies :build :publish "$@"


gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
```diff
@@ -242,6 +241,73 @@ public BundleTaskExtension(org.gradle.api.tasks.bundling.Jar task) {
 			.property("default Bundle-Version", getDefaultBundleVersion());
 	}
 
+	private Map<String, String> getGradleProjectProperties() throws Exception {
+		Properties gradleProperties = new BeanProperties();
+		gradleProperties.put("project", getTask().getProject());
```
ejjcase, line 246: Should the task also be in here?

bjhargrave: No, because under current behavior, the task is always added. Doing here would mean task is not set if any properties are set in the build script since this code is called in a convention. (I thought about doing this but then realized it would change behavior.)

chrisrueger:
Thanks @bjhargrave . Now we have the unprecedented situation of having two PRs for the issue :) 

Not sure how much I can contribute to a review here.

I tried running it locally with:

```
./gradlew --no-daemon -Dmaven.repo.local=dist/m2 --continue :build "$@"  -x testOSGi
./gradlew --no-daemon -Dmaven.repo.local=dist/m2 --continue :gradle-plugins:build 
```

Both work fine.

I just notice the warning:


and when running with `--warning-mode all`

```
./gradlew --no-daemon -Dmaven.repo.local=dist/m2 --continue --warning-mode all :gradle-plugins:build
```

I get this.

```
./gradlew --no-daemon -Dmaven.repo.local=dist/m2 --continue --warning-mode all :gradle-plugins:build

To honour the JVM settings for this build a single-use Daemon process will be forked. For more on this, please refer to https://docs.gradle.org/8.14/userguide/gradle_daemon.html#sec:disabling_the_daemon in the Gradle documentation.
Daemon will be stopped at the end of the build 
Settings file '/Users/christoph/git/bnd/settings.gradle': line 26
Mutating configuration container for settings file 'settings.gradle' using create(String, Closure) has been deprecated. This will fail with an error in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.14/userguide/upgrading_version_8.html#mutating_buildscript_configurations
        at settings_5ah3375x6topgpc3o7srn3c7h$_run_closure2.doCall$original(/Users/christoph/git/bnd/settings.gradle:26)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
        at settings_5ah3375x6topgpc3o7srn3c7h.run(/Users/christoph/git/bnd/settings.gradle:14)
        (Run with --stacktrace to get the full stack trace of this deprecation warning.)
```

When I try to build your PR with Gradle Wrapper 9.1.0 (`gradle wrapper --gradle-version 9.1.0 `) I get :



```
FAILURE: Build failed with an exception.

* Where:
Settings file '/Users/christoph/git/bnd/settings.gradle' line: 26

* What went wrong:
> Cannot mutate configuration container for settings file 'settings.gradle' using create(String, Closure). Configurations cannot be added or removed from the buildscript configuration container.
```


Do you think this is also something which should / could be adressed here?


Additional question: How do you develop / debug this usually?

chrisrueger:
What about re-using the testcase https://github.com/bndtools/bnd/pull/6914/files#diff-5f39dd0673d31c529b14ac8f10d09cad7dab3eb8504591438532df95d61cf911 by @ejjcase ?

bjhargrave:
> Do you think this is also something which should / could be adressed here?

I have updated the PR to address all the gradle warnings. This include settings.gradle in the main build and some new warnings in the gradle test case which were not noticed while warning mode was all. I also restored the warning mode to fail.

Once the PR is merged, we can then look to update the gradle version in the build to 9+.

bjhargrave:
> What about re-using the testcase https://github.com/bndtools/bnd/pull/6914/files#diff-5f39dd0673d31c529b14ac8f10d09cad7dab3eb8504591438532df95d61cf911 by @ejjcase ?

There are already tests cases covering user defined properties. So I don't think this change is needed.

bjhargrave:
> Additional question: How do you develop / debug this usually?

Usually I don't :-)

I have not had Eclipse installed on my system for a few years. So for this effort, I just built from the command line to put the latest jars in my local m2 repo. Then I used IntelliJ on the gradle-plugin directory (same concept for the maven-plugins directory if you want to work on the maven plugins). This time I did not need to debug (beyond some temporary logging statements). There is a way to debug the gradle plugins but I can no longer find that information in my head :-/

PR #6917, opened 2025-11-04
gradle: Support gradle 9 in the plugins

bjhargrave:
This means dropping support for gradle 7 since the support requires the use of deprecated methods removed in gradle 9.

https://github.com/gradle/gradle/issues/34505 causes the 
`TestBundlePlugin."Kotlin Bnd Builder Plugin Test"` to fail when building with Gradle 9.

We will also need to merge this PR before upgrading the gradle version to 9.3 (version targeted for the fix of the above issue) since we will need to remove the use of deprecated methods before we update gradle.

### DIFF
--- a/gradle-plugins/README.md
+++ b/gradle-plugins/README.md
@@ -5,10 +5,8 @@ A typical Gradle build is a non-Bnd workspace build.
 A Bnd Workspace build uses the information specified in the Bnd Workspace's `cnf/build.bnd` file and each project's `bnd.bnd` file to configure the Gradle projects and tasks.
 
 The [`biz.aQute.bnd.gradle`][2] jar contains the Bnd Gradle Plugins.
-These plugins require Java 17 and
-at least Gradle 7.3 for Java 17,
-at least Gradle 7.5 for Java 18,
-and at least Gradle 7.6 for Java 19.
+These plugins require Java 17 and at least Gradle 8.0.
+More recent Java versions will require more recent Gradle versions.
 
 This README represents the capabilities and features of the Bnd Gradle Plugins in the branch containing this README.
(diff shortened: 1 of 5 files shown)


chrisrueger:
Thanks again @bjhargrave 

> https://github.com/gradle/gradle/issues/34505 causes the
> TestBundlePlugin."Kotlin Bnd Builder Plugin Test" to fail when building with Gradle 9.

I tried building your branch with Gradle 9.2.0 and both were successful. I did not hit the problem you described. 

```
./gradlew wrapper --gradle-version=9.2.0 && ./gradlew wrapper
./gradlew --no-daemon -Dmaven.repo.local=dist/m2 --continue :build "$@" -x test  -x testOSGi
./gradlew --no-daemon -Dmaven.repo.local=dist/m2 --continue  --warning-mode all :gradle-plugins:build
```

both BUILD SUCCESSFUL. Ok I skipped the tests for the main build, to build faster. but the gradle-plugins:build had all tests pass.

I got a warning though:

```
The archives configuration has been deprecated for artifact declaration. This will fail with an error in Gradle 10. Add artifacts as direct task dependencies of the 'assemble' task instead of declaring them in the archives configuration. Consult the upgrading guide for further information: https://docs.gradle.org/9.2.0/userguide/upgrading_version_9.html#sec:archives-configuration
```

Is this the problem? 
Or if not what did you do to trigger it? Or do I misunderstand?

bjhargrave:
> Is this the problem?

It is not what I saw. 

The archives configuration warning is what is fixed by this PR. So on rebuild, it is gone.

So let's merge this PR and then rebase your PR on it and update it for Gradle 9.2.

chrisrueger (approved):

Issue #6920, opened 2025-11-06
Add special handling for proxy interfaces when import package is calculated


laeubi:
Today I encountered an interesting problem, see here for a full analysis:

- https://github.com/laeubi/equinox/pull/130

The problem here is that when calculating packages with bnd the package `javax.servlet.descriptor` seems unused in the code but is actually required at runtime due to 

```
return (ServletContext) Proxy.newProxyInstance(
    curClassLoader, 
    new Class[] { ServletContext.class }, 
    new AdaptorInvocationHandler()
);
```

and ServletContext has a getter `JspConfigDescriptor getJspConfigDescriptor()`

As bnd already has detects the case where `classForName` witha fqdn is used, it would be great if the proxy case can also be automatically detected so it could work like this:

- In the bytecode a call to `Proxy.newProxyInstance` is detected
- all types in the  new Class[] { SomeInterface.class, ... maybe more ... } are now considered as if it where implemented, that means all types this interface references in parameters and return types need to be imported as well

PR #6921, opened 2025-11-06
ClassParser: Implement Proxy.newProxyInstance detection for import packages

laeubi:
Add detection for Proxy.newProxyInstance calls to automatically include packages referenced by proxy interface methods. This works similarly to the existing Class.forName detection.

Fix https://github.com/bndtools/bnd/issues/6920


chrisrueger (commented):
Thanks, looks useful and all tests pass. 
One tiny thing: could you mention that in the Syntax entry for NOCLASSFORNAME https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib/src/aQute/bnd/help/Syntax.java#L618

This sentence will then also end up in the docs and in syntax highlighting in bnd editor.

laeubi:
Done!

chrisrueger (approved):

chrisrueger:
UPDATE: The test below is not good. Please ignore this comment. See further down. 

I just asked AI if there is a risk of false positive detection. It came up with that. Do you think this is valid?

Could you maybe have a look at the following :

```
package test.proxy;
import java.lang.reflect.Proxy;

public class FalsePositiveProxyExample {
    public void tricky() {
        // 1️⃣ Create a Class[] array, but NOT for proxy
        Class<?>[] arr = new Class<?>[1];
        arr[0] = Object.class;  // ldc(Object.class), aastore
        doSomething(arr);

        // 2️⃣ Later, completely unrelated call to Proxy.newProxyInstance,
        //     but the Class[] is built elsewhere, not inline.
        Proxy.newProxyInstance(
            getClass().getClassLoader(),
            getInterfaces(),      // comes from a variable, not the inline array above
            (proxy, method, args) -> null
        );
    }

    private Class<?>[] getInterfaces() {
        // Constructed in a different method — analyzer cannot see contents
        return new Class<?>[] { Runnable.class };
    }

    private void doSomething(Class<?>[] a) {
        // no-op
    }
}
```

```
@Test
	public void testProxyFalsePositiveNotDetected() throws Exception {
		// Enable class-for-name and proxy detection
		a.setProperty("-noclassforname", "false");

		// Load the compiled class file that could cause a false positive
		Clazz c = new Clazz(a, "test/proxy", null);
		c.parseClassFile(getClass().getResourceAsStream("proxy/FalsePositiveProxyExample.class"));

		// When a Proxy.newProxyInstance call occurs later but the Class[] array
		// was created earlier for a different purpose, we should NOT treat
		// those earlier ldc/aastore class constants (e.g., Object.class)
		// as proxy interfaces.

		assertThat(c.getReferred())
			// Ensure we did NOT incorrectly add Object.class as a proxy
			// interface
			.doesNotContain(a.getPackageRef("java.lang.Object"))
			// Optional: if the analyzer can’t detect the true interface array
			// built in getInterfaces(), it’s fine; just ensure no bogus one is
			// added
			.doesNotContain(a.getPackageRef("java.lang"))
			// For documentation: Proxy.newProxyInstance is still recognized,
			// but its interface array (built in a separate method) isn't
			// analyzed
			.contains(a.getPackageRef("java.lang.reflect"));
	}
```

laeubi:
Test case added and detection improved (by the AI)

chrisrueger:
@laeubi thanks. I added my testcase from my comment [above](https://github.com/bndtools/bnd/pull/6921#issuecomment-3500933005) and this fails. 

```
java.lang.AssertionError: 
Expecting
  [test.proxy, java.lang, java.lang.reflect, java.lang.invoke]
not to contain
  [java.lang]
but found
  [java.lang]

	at test.ClassParserTest.testProxyFalsePositiveNotDetected(ClassParserTest.java:577)
```

in https://github.com/bndtools/bnd/pull/6921/files#diff-85fbc89f0c6509380298a0e20b8a5520338a927f87b7f85ff882a678c6cbf35eR1301 eclipse shows me a potential NPE for `proxyInterfaces.add(lastReference);`

One more question: In first commit you also submitted `.class` files, and in the new commit you don't. Are they needed? I deleted them locally just to see, and the tests still work. So maybe we can remove them? I guess the classes are compiled by the normal compilation?

laeubi:
> I added my testcase from my comment https://github.com/bndtools/bnd/pull/6921#issuecomment-3500933005 and this fails.

I think `java.lang` is really a bad choice as such import is likely coming from elsewhere (especially `import java.lang.reflect.Proxy;`

> One more question: In first commit you also submitted .class files, and in the new commit you don't. 

The git ignores of bnd repository are often not really ignoring files created in test, I even get a dirty workingtree when running bndtools :-\

So I guees its possibly not needed.

chrisrueger:
Ok I confirm that my AI false-positive test was not good. 
I rolled back to master, and bnd would detect that normally, even without your new code. 

Still working on it.

chrisrueger:
In agreement with @laeubi I have added another FalsePositive test `ClassParserTest#testProxyFalsePositiveNotDetected` using his concrete ServletContext.class example mentioned in the description of #6920

peterkir:
Without reviewing the code completely ;-)
Will there also be an instruction option similar to https://bnd.bndtools.org/instructions/noclassforname.html and corresponding documentation site?

chrisrueger:
Currently -noclassforname is reused. Do you want a separate instruction e.g. -noproxysomesthing ? @peterkir

chrisrueger:
> Will there also be an instruction option similar to https://bnd.bndtools.org/instructions/noclassforname.html and corresponding documentation site?

I added `-noproxyinterfaces:true` to disable the new processing of `Proxy.newProxyInstance` usage - similar to how `-noclassforname` works. It will be generated under https://bnd.bndtools.org/instructions/noproxyinterfaces.html  after merging.

chrisrueger:
Thanks @laeubi for the contribution.

PR #6925, opened 2025-11-06
gradle: Compile Kotlin code to work on Gradle 8

bjhargrave:
Update the gradle plugins to use Kotlin 2.0 as the language level to support Gradle 8. Gradle 9 has updated to Kotlin 2.2.

We also update the build to Gradle 9.2 and use Java 25 (replacing 24) for testing.


chrisrueger:
Thanks a lot

PR #6950, opened 2025-11-20
remove superfluous dummy

peterkir:


chrisrueger:
Thanks @peterkir for finding it. I will examine the jar afterwards and if it looks good we can merge. 
As discussed, not sure anymore why I left it in the first place.

Issue #6960, opened 2025-11-24
Support registering JUnit tester runnable as a service to be run by custom code


ejjcase:
In `biz.aQute.tester.junit-platform` I can choose to run the tests on their own thread (the default behaviour) or register them as a service. If they are registered as a service, the service always has the attribute `main.thread=true`, which means `aQute.launcher.Launcher` picks it up and runs it on the main thread.

I would like the option to register the service without `main.thread=true` so that I can run it at a different time.

Motivation: I am migrating an old RCP application to Bnd. It has many tests (thousands) that rely on the workbench being up and running, and/or which need to run in the Eclipse UI thread. So I would like to hook into Eclipse's support for workbench tests by obtaining the tester `Runnable` and then doing this:

```java
PlatformUI.getTestableObject().setTestHarness(() -> Display.getDefault().asyncExec(runnable));
```

Simply omitting `main.thread=true` from the service registration is enough to make this work, because then I can find the service by filtering on `(&(objectclass=java.lang.Runnable)(service.description=JUnit tester))`.

PR #6961, opened 2025-11-24
Support tester.customservice property.

ejjcase:
For https://github.com/bndtools/bnd/issues/6960.

I've written it so that if you set `tester.customservice`, you get a service with `junit.tester=true`. Also, if you set `tester.customservice` you don't need to also set `launch.services`.

Test forthcoming.


chrisrueger:
Since @ejjcase asked for feedback in slack, maybe @kriegfrj has some opinion.

PR #6992, opened 2025-12-08
bugfixing sonatype publisher url

peterkir:


biz.aQute.repository/test/aQute/bnd/repository/maven/provider/SonatypeDeploymentTest.java
```diff
@@ -236,4 +285,52 @@ private void dropDeployment(HttpClient client, String deploymentId) throws Excep
...
+	}
+
+	private static boolean ensureEnvVar(String varName) {
+		boolean found = false;
+		String value = System.getenv(varName);
+		if (value == null || value.isEmpty()) {
+			System.err.println("Mandatory environment variable '" + varName + "' is not set or empty.");
+		}
+		return found;
+	}
```
chrisrueger, line 336: @peterkir these method seem to be incorrect. 

- ensureEnvVar always returns false
- hasRequiredEnvironment always returns true

Maybe try to change it like this

```suggestion
	private static boolean hasRequiredEnvironment() {
		String[] vars = {
			"SONATYPE_BEARER", "GPG_KEY_ID", "GPG_PASSPHRASE"
		};

		boolean allFound = true;

		for (String var : vars) {
			if (!ensureEnvVar(var)) {
				allFound = false;
			}
		}

		return allFound;
	}

	private static boolean ensureEnvVar(String varName) {
		String value = System.getenv(varName);

		if (value == null || value.isEmpty()) {
			System.err.println("Mandatory environment variable '" + varName + "' is not set or empty.");
			return false;
		}

		return true;
	}
```

PR #7000, opened 2025-12-14
Fix Junit Platform 1.13+ Support (up to JUnit Jupiter and Platform to 5.14.2/1.14.2)

chrisrueger:
Closes #6651
Copied from fork PR https://github.com/chrisrueger/bnd/pull/1 which was created via Copilot.

> ## JUnit Platform 1.13+ Support - Fix NamespacedHierarchicalStore Error
> 
> ### Overview
> Addressing the "No NamespacedHierarchicalStore" error when using JUnit Platform 1.13+ with the BundleEngine. The issue occurs because JUnit Platform 1.13+ changed how execution context is propagated to child test engines.
> 
> ### Implementation Status
> 
> - [x] Understand the problem: ExecutionRequest context not properly propagated in JUnit Platform 1.13+
> - [x] Modify `BundleEngine.execute()` to pass the full ExecutionRequest to child execution methods
> - [x] Update `BundleDescriptor.executeChild()` signature to accept ExecutionRequest instead of just listener and params
> - [x] Implement proper context propagation using ExecutionRequestFactory with reflection
> - [x] Update version constraint in `biz.aQute.tester.junit-platform/bnd.bnd` to allow JUnit Platform 1.13+
> - [x] Address all code review feedback
> - [x] Fix test setup to include ExecutionRequestFactory class
> - [x] Update ExecutionRequestFactory to handle JUnit Platform 1.13.4 constructor signature changes
> - [x] Update to JUnit Jupiter 5.13.4 / Platform 1.13.4
> - [x] Fix test classes for JUnit 5.13+ @BeforeAll static requirement
> 
> ### Changes Made
> 
> 1. **BundleEngine.java** - Pass ExecutionRequest through to child execution methods
> 2. **BundleDescriptor.java** - Use ExecutionRequestFactory to create child requests with proper context
> 3. **ExecutionRequestFactory.java** - Factory with reflection-based context propagation
>    - Detects both 4-param (pre-1.13) and 5-param (1.13+) constructor signatures
>    - Handles OutputDirectoryProvider parameter added in 1.13+
>    - Properly propagates NamespacedHierarchicalStore to child engines
> 4. **bnd.bnd** - Updated Import-Package to allow JUnit Platform 1.13+
> 5. **BundleEngineTest.java** - Updated test setup to include ExecutionRequestFactory class
> 6. **cnf/ext/junit.bnd** - Updated to junit.jupiter.version=5.13.4 and junit.platform.version=1.13.4
[...]

PR #7024, opened 2025-12-27
Add optional imports for Guava / Error Prone 

### LINKED ISSUE #7022: Make some dependencies optional to avoid problems for consumers after 7.2.0 upgrade
Eclipse has problems after upgrading to 7.2.0 because e.g. our `org.bndtools.templating` now imports some packages because of our upgrade of Guava

See

I think the latest release of the BND tool set is kind of messed up.

There is this failure:

> 17:18:26  [ERROR] Cannot resolve dependencies of project org.eclipse.pde:org.eclipse.pde.bnd.ui:eclipse-plugin:1.2.400-SNAPSHOT
> 17:18:26  [ERROR]  with context {osgi.os=linux, osgi.ws=gtk, org.eclipse.update.install.features=true, org.eclipse.jdt.buildtime=true, osgi.arch=x86_64, org.eclipse.update.install.sources=true}
> 17:18:26  [ERROR]   Software being installed: org.eclipse.pde.bnd.ui 1.2.400.qualifier
> 17:18:26  [ERROR]   Missing requirement: org.bndtools.templating 7.2.0.202512261929 requires 'java.package; com.google.common.util.concurrent.internal 0.0.0' but it could not be found
> 17:18:26  [ERROR]   Cannot satisfy dependency: org.eclipse.pde.bnd.ui 1.2.400.qualifier depends on: java.package; org.bndtools.templating [2.0.0,3.0.0): See log for details

Looking at the new version of that bundle, I think this import should probably be `Private-Package` like the other `com.google.common` packages:

image

Looking at

https://repo.maven.apache.org/maven2/org/bndtools/org.bndtools.templating/7.2.0/org.bndtools.templating-7.2.0.pom

the dependency listed as `provided`:

> <dependency>
> <groupId>com.google.guava</groupId>
> <artifactId>guava</artifactId>
> <version>33.4.8-jre</version>
> <scope>provided</scope>
> </dependency>

I don't think we want to drag guava into the Platform/SDK.

---

I don't think that's the only think that' appears messed up....

When I try to build these updates with Orbit, that build fails like this:

> [ERROR]   Missing requirement: biz.aQute.tester.junit-platform 7.2.0.202512261929 requires 'java.package; org.junit.platform.commons [1.10.1,1.13.0)' but it could not be found
> [ERROR]   Cannot satisfy dependency: org.eclipse.orbit.maven.osgi.all.feature.group 4.39.0.v20251227-1000 depends on: org.eclipse.equinox.p2.iu; biz.aQute.tester.junit-platform [7.2.0.202512261929,7.2.0.202512261929]
> [ERROR]   Cannot satisfy dependency: org.eclipse.orbit.maven.osgi.site raw:1.0.0.'SNAPSHOT'/format(n[.n=0;[.n=0;[-S]]]):1.0.0-SNAPSHOT depends on: org.eclipse.equinox.p2.iu; org.eclipse.orbit.maven.osgi.all.feature.group 0.0.0: See log for details
> [ERROR] -> [Help 1]

The current version of the `org.junit.platform.commons` package is `1.14.1`:

image

The older version of that bundle has a broader range for the dependency:

image

So this too appear inappropriate to me.

---

I can't be sure there are other problems we've not noticed yet...

_Originally posted by @merks in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3579#issuecomment-3693797403_
            

## Tasks

Make following optional:

- com.google.errorprone.annotations
- com.google.errorprone.annotations.*

Before:

```
 com.google.common.util.concurrent.internal
 com.google.errorprone.annotations.concurrent
 com.google.errorprone.annotations
```

Goal:

```
com.google.common.util.concurrent.internal;resolution:=optional
com.google.errorprone.annotations.concurrent;resolution:=optional
com.google.errorprone.annotations;resolution:=optional
```

chrisrueger:
Closes #7022

## Guava

- Removed `org.jspecify.jspecify` and  `com.google.errorprone.annotations` 
- following packages are now private-package (instead of Import-Package)

```
com.google.common.util.concurrent.internal.*
```

Hopefully makes life easier for https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3579#issuecomment-3693797403

Full  MANIFEST.MF (Print view from JAR viewer):

```
[MANIFEST]

Automatic-Module-Name                   org.bndtools.templating
Bundle-ActivationPolicy                 lazy
Bundle-Copyright                        Copyright (c) Neil Bartlett (2009, 2025) and others. All Rights Reserved.
Bundle-Developers                       bjhargrave;email="bj@hargrave.dev";name="BJ Hargrave";organization=IBM;organizationUrl="https://developer.ibm.com";roles=developer;timezone="America/New_York";url="https://github.com/bjhargrave"
                                        chrisrueger;email="chrisrueger@gmail.com";name="Christoph Rueger";organization="Synesty GmbH";organizationUrl="https://synesty.com/";roles=developer;timezone="Europe/Berlin"
                                        peterkir;email="peter@klib.io";name="Peter Kirschner";organization="Kirschners GmbH";organizationUrl="https://peterkir.github.io/";roles=developer;timezone="Europe/Berlin";url="https://peterkir.github.io"
                                        pkriens;email="Peter.Kriens@aQute.biz";name="Peter Kriens";organization=Bndtools;organizationUrl="https://github.com/bndtools";roles="architect,developer";timezone=1
                                        rotty3000;email="raymond.auge@liferay.com";name="Ray Augé";organization="Liferay Inc.";organizationUrl="https://www.liferay.com";roles=developer;timezone="America/New_York";url="https://rotty3000.github.io"
Bundle-DocURL                           https://bndtools.org/
Bundle-License                          (Apache-2.0 OR EPL-2.0);description="This program and the accompanying materials are made available under the terms of the Apache License, Version 2.0, or the Eclipse Public License 2.0.";link="https://opensource.org/licenses/Apache-2.0,https://opensource.org/licenses/EPL-2.0"
Bundle-ManifestVersion                  2
Bundle-Name                             org.bndtools.templating
Bundle-SCM                              connection=scm:git:https://github.com/bndtools/bnd.git
                                        developerConnection=scm:git:git@github.com:bndtools/bnd.git
                                        tag=7.3.0-SNAPSHOT
                                        url=https://github.com/bndtools/bnd
Bundle-SymbolicName                     org.bndtools.templating
Bundle-Vendor                           Bndtools https://bndtools.org/
Bundle-Version                          7.3.0.202512290835-SNAPSHOT
Export-Package                          org.bndtools.templating.util;version="1.0.0";uses:="org.osgi.service.metatype"
                                        org.bndtools.templating;version="2.0.0";uses:="aQute.service.reporter,org.eclipse.core.runtime,org.osgi.framework,org.osgi.service.metatype,org.osgi.util.promise"
Git-Descriptor                          7.3.0.DEV-96-gf5db14361
Git-SHA                                 f5db14361114564f2162531fbcfb160e2aac43fe
Import-Package                          aQute.bnd.exceptions;version="[3.0,4)"
                                        aQute.bnd.osgi;version="[7.6,8)"
                                        aQute.service.reporter;version="[1.3,2)"
                                        android.os;resolution:=optional
                                        com.google.appengine.api.utils;resolution:=optional
                                        com.google.appengine.api;resolution:=optional
                                        com.google.apphosting.api;resolution:=optional
                                        javax.crypto
                                        javax.crypto.spec
                                        javax.swing
                                        javax.swing.border
                                        javax.swing.event
                                        javax.swing.text
                                        javax.swing.tree
                                        org.eclipse.core.runtime;bundle-symbolic-name="org.eclipse.core.runtime";bundle-version="[3.30,4)"
                                        org.osgi.framework;version="[1.8,2)"
                                        org.osgi.service.metatype;version="[1.3,2)"
                                        org.osgi.util.promise;version="[1.3,2)"
                                        sun.misc;resolution:=optional
Manifest-Version                        1.0
Private-Package                         aQute.lib.io;version="4.5.0"
                                        aQute.lib.stringrover;version="1.2.0"
                                        aQute.libg.glob;version="1.6.0"
                                        com.github.mustachejava
                                        com.github.mustachejava.codes
                                        com.github.mustachejava.reflect
                                        com.github.mustachejava.reflect.guards
                                        com.github.mustachejava.resolver
                                        com.github.mustachejava.util
                                        com.google.common.base
                                        com.google.common.base.internal
                                        com.google.common.cache
                                        com.google.common.collect
                                        com.google.common.graph
                                        com.google.common.hash
                                        com.google.common.io
                                        com.google.common.math
                                        com.google.common.primitives
                                        com.google.common.util.concurrent
                                        com.google.common.util.concurrent.internal
                                        org.bndtools.templating.engine.mustache
                                        org.bndtools.templating.engine.st
                                        org.stringtemplate.v4
                                        org.stringtemplate.v4.compiler
                                        org.stringtemplate.v4.debug
                                        org.stringtemplate.v4.gui
                                        org.stringtemplate.v4.misc
                                        st4hidden.org.antlr.runtime
                                        st4hidden.org.antlr.runtime.misc
                                        st4hidden.org.antlr.runtime.tree
Provide-Capability                      osgi.service;objectClass:List<String>="org.bndtools.templating.TemplateEngine";uses:="org.bndtools.templating"
Require-Capability                      osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=17))"
                                        osgi.extender;filter:="(&(osgi.extender=osgi.component)(version>=1.3.0)(!(version>=2.0.0)))"
SPDX-License-Identifier                 (Apache-2.0 OR EPL-2.0)
Service-Component                       OSGI-INF/org.bndtools.templating.engine.mustache.xml
                                        OSGI-INF/org.bndtools.templating.engine.st.xml


```

### DIFF
--- a/org.bndtools.templating/bnd.bnd
+++ b/org.bndtools.templating/bnd.bnd
@@ -17,8 +17,8 @@
 	org.antlr:ST4:jar:complete;maven-scope=provided,\
 	com.github.spullara.mustache.java:compiler;maven-scope=provided,\
 	com.google.guava;version="[33.4.8,33.4.9)";maven-scope=provided,\
-	org.jspecify.jspecify;version='[1.0.0,2.0.0)';maven-scope=provided
-
+	com.google.guava.failureaccess;version=latest;maven-scope=provided
+	
 -testpath: \
 	slf4j.api,\
 	slf4j.simple,\
@@ -32,7 +32,7 @@
 	aQute.libg.*,\
 	com.github.mustachejava.*,\
 	com.google.common.*,\
-	org.jspecify.*
+	
 	
 Bundle-ActivationPolicy: lazy
 
@@ -41,6 +41,8 @@ Import-Package: \
  com.google.appengine.*;resolution:=optional,\
  com.google.apphosting.*;resolution:=optional,\
  android.os.*;resolution:=optional,\
+ !org.jspecify.*,\
+ !com.google.errorprone.annotations.*,\
  ${eclipse.importpackage},\
  *
 


chrisrueger:
@laeubi since you were also  part of https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3579

Are you ok with the changes?

org.bndtools.templating/bnd.bnd
```diff
@@ -17,8 +17,10 @@
 	org.antlr:ST4:jar:complete;maven-scope=provided,\
 	com.github.spullara.mustache.java:compiler;maven-scope=provided,\
 	com.google.guava;version="[33.4.8,33.4.9)";maven-scope=provided,\
-	org.jspecify.jspecify;version='[1.0.0,2.0.0)';maven-scope=provided
-
+	org.jspecify.jspecify;version='[1.0.0,2.0.0)';maven-scope=compile,\
+	com.google.errorprone.annotations;version=latest;maven-scope=compile,\
+	com.google.guava.failureaccess;version=latest;maven-scope=compile
```
laeubi: Looking at the other declaration should this not be maven scope provided as well? I don't think this needs to be transitively available?

chrisrueger: Oh you are right, I think I misinterpreted those scopes.
Fixed.

This basically only affects what is written in the pom.xml which is embedded in the final jar under `META-INF/maven/pom.xml`

Thanks

org.bndtools.templating/bnd.bnd
```diff
@@ -17,8 +17,10 @@
 	org.antlr:ST4:jar:complete;maven-scope=provided,\
 	com.github.spullara.mustache.java:compiler;maven-scope=provided,\
 	com.google.guava;version="[33.4.8,33.4.9)";maven-scope=provided,\
-	org.jspecify.jspecify;version='[1.0.0,2.0.0)';maven-scope=provided
-
+	org.jspecify.jspecify;version='[1.0.0,2.0.0)';maven-scope=provided,\
+	com.google.errorprone.annotations;version=latest;maven-scope=provided,\
+	com.google.guava.failureaccess;version=latest;maven-scope=provided
```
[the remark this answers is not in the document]
merks: Sounds like really bad advice to me.  Let's say everyone does this with every dependency.  We (any downstream consumers) would end up with a mess of duplicates that a) likely don't work together because of wiring problems and b) if there were a CVE in an existing dependency (which seems more likely than a CVE in a new one but not in an old one) we'd not be able to update to a fixed higher version until everyone and everything provides new versions of their content with higher upper bounds in that content.  Taken to the extreme, this would appear to be advice to stop using version ranges in bundles.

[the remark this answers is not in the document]
laeubi: I think copilot (and you) just got this wrong. `latest` only means, use the latest from all repositories _during the build_ what seems valid here given one likely manages the actual version in the build config rather than having multiple one there and select specific ones here.

So it does not has an influence on the runtime or how bnd calculates the manifest, it just controls what is selected for the classpath.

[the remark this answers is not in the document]
chrisrueger: Regarding `latest`, I think @laeubi is right.
I used `latest` here because bnd will only look into the repositories it knows and controls. bnd defines its dependencies in 

https://github.com/bndtools/bnd/blob/cdc2c40730523d0f8b124e0f423e45211ba392b9/cnf/ext/central.mvn#L132-L134

Regarding your other points:
But regarding your CVE topic: I am not sure anymore.

**Private-Package**
In order to make the packages appear under `Private-Package` (instead of Import-Package) I added them to the [-conditional-package](https://bnd.bndtools.org/instructions/conditionalpackage.html) which basically embeds the packages and their classes inside the `org.bndtools.templating.jar`

image


I did that because it was done for other guava packages too, so I thought I continue this pattern. That way downstream consumers do not have to care providing the dependency themselves, because `org.bndtools.templating.jar` already includes what it needs and does not export it (keep it private). I thought for those annotation bundles that approach is ok. 

But the side effect is, in case of a CVE those classes are inside `org.bndtools.templating.jar` (similar to shadowed), so they cannot be controlled from outside (downstream consumers). 

**`Import-Package` but with `;resolution:=optional`**

If I should instead put it back to `Import-Package` but with `;resolution:=optional` , this would mean, that downstream consumers would have more control over this and could in theory provide another version fitting the version range. 

Question: To embed or not embed?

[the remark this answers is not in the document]
laeubi: > Question: To embed or not embed?

See https://bnd.bndtools.org/instructions/conditionalpackage.html#objections

As its unlikely someone will ever need these annotations in the junit platform runner at all and the runner is most meant to be used standalone, this is mostly byte-garbage never used so I won't care that much, you can even exclude these package here and it will simply not matter at all.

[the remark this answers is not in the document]
chrisrueger: > need these annotations in the junit platform runner

The embedding question was about `org.bndtools.templating.jar` and not `junit platform runner` (sorry, both topics are part of this PR).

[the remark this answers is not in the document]
chrisrueger: > you can even exclude these package here and it will simply not matter at all.

This was a good point 👍  I push forced again and updated the description above for the **Guava** part and included the updated manifest.

- Removed `org.jspecify.jspecify` and `com.google.errorprone.annotations`
- following packages are now private-package (instead of Import-Package) and embedded into the jar

```
com.google.common.util.concurrent.internal.*
```
The latter was done because guave needs those packages at runtime and since we already embed all guava packages (since 2016) I think it is ok for this too.
image

On that note: bnd does not use guava directly as far as I know, but some of its dependencies need it (com.github.spullara.mustache.java)

biz.aQute.tester.junit-platform/bnd.bnd
```diff
@@ -20,7 +20,7 @@ Import-Package: \
 	aQute.*;resolution:=optional,\
 	junit.*;version="${range;[==,5);${junit3.version}}";resolution:=optional,\
 	org.apache.felix.service.command;resolution:=optional,\
-	org.junit.platform.*;version="[${junit.platform.tester.version},1.13)",\
+	org.junit.platform.*;version="[${junit.platform.tester.version},1.13)";resolution:=optional,\
```
merks: Isn't a hard coded `1.13` just waiting to bite you again in the future?  I mean at some point one expects the version to change and it might be that `org.junit.platform.tester.version >= 1.13` and then this will be completely bogus/broken and that might go unnoticed...

Maybe it should be more like one of these building a range around the current version:
```
${range;[==,+);${junit.platform.tester.version}}
${range;[==,=+);${junit.platform.tester.version}}
```

Have you really verified/tested that 1.14.1 does not work and needed to be excluded?  Excluded something within the major version bounds seems generally questionable.

chrisrueger: > Have you really verified/tested that 1.14.1 does not work and needed to be excluded? Excluded something within the major version bounds seems generally questionable.

Yes, see https://github.com/bndtools/bnd/issues/6651

But good point to discuss this now. 
Yesterday I did some more work on https://github.com/bndtools/bnd/pull/7000 which does exactly that

https://github.com/bndtools/bnd/pull/7000/changes#diff-8baae411a65dbf2547879dec01c93a74b1ded6d32ce242eef118835cb917335fR19 

If #7000 gets approved than the hard range limit `1.13` is not needed anymore. 

What are your thoughts on `;resolution:=optional` ? Should this be added even with #7000 with a larger range?

merks: I can't really know the semantics of what's optional or not.  If it doesn't function without that bundle (which given the bundle name suggests is the case), it probably should not be optional...

chrisrueger: That makes sense. Yes, I think that bundle needs `org.junit.platform.*` to function correctly ( as the name suggest and you point out). 

I guess I will remove `org.junit.platform.*` related changes from this PR. So this PR will then only about `org.bndtools.templating.jar` and `biz.aQute.tester.junit-platform` will be better handled in #7000

merks: BTW, I'm concerned that any upper bound other than 2 on any non-optional junit dependency will cause problems for the platform in the future.  E.g., if 1.14.2 comes out and the Platform cannot update because BND restricts that, it will be probably and we might not noticed until later.  😟

chrisrueger: > I'm concerned that any upper bound other than 2 on any non-optional junit dependency will cause problems for the platform in the future. E.g., if 1.14.2 comes out and the Platform cannot update because BND restricts that, it will be probably and we might not noticed until later.

In #7000 the range is < 2 again

```
org.junit.platform.*;version="${range;[==,+);${junit.platform.tester.version}}",\
```

chrisrueger:
I removed changes for `biz.aQute.tester.junit-platform`, because #7000 seems to be a better fix for that.

So this PR here is mainly about `org.bndtools.templating` now

chrisrueger:
@merks `org.bndtools.templating` 7.3.0 SNAPSHOT is available here https://bndtools.jfrog.io/artifactory/update-snapshot/org/bndtools/org.bndtools.templating/7.3.0-SNAPSHOT/org.bndtools.templating-7.3.0-20251229.122458-34.jar in case you want to have a look at it.

merks:
@chrisrueger 

Looking at the MANIFEST.MF content, everything looks sensible and is structured with sensible version ranges.  Thanks!!

PR #7021, opened 2025-12-27
Remove test-bnd module from in-build pom.xml

chrisrueger:
Closes #7020

The test-bnd module reference was removed from the modules section in the in-build pom.xml

This fixes the failing test in #7020

But not sure this is really the best solution, since it might fix only a symptom, and not the root cause. But maybe it serves as a base for discussion.

Issue #7020, opened 2025-12-27
Master build fails after 7.2.0 release


chrisrueger:
Strangly our build in master branch suddenly starts failing e.g. https://github.com/bndtools/bnd/actions/runs/20529529172/job/58982746219

with: 

```
[INFO] Building in-build 0.0.1                                            [3/3]
[INFO]   from pom.xml
[INFO] --------------------------------[ pom ]---------------------------------
[INFO] 
[INFO] --- bnd-indexer:7.2.0:index (index) @ in-build ---
[INFO] Resolved all Dependencies for biz.aQute.bnd-test:in-build:pom:0.0.1. Checked: 4, Skipped(no need to revisit): 0, Failed: 0, Resulting Dependencies: 4
[ERROR] Failed to determine the artifact URI for artifact biz.aQute:bnd:jar:1.50.0
java.io.FileNotFoundException: Unable to index artifact biz.aQute:bnd:jar:1.50.0. The repository workspace is not known to this resolver
    at aQute.bnd.maven.indexer.plugin.IndexerMojo$RepositoryURLResolver.resolver (IndexerMojo.java:353)
    at aQute.bnd.maven.indexer.plugin.IndexerMojo.execute (IndexerMojo.java:188)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:126)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:328)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:316)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:212)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:174)
    at org.apache.maven.lifecycle.internal.MojoExecutor.access$000 (MojoExecutor.java:75)
    at org.apache.maven.lifecycle.internal.MojoExecutor$1.run (MojoExecutor.java:162)
    at org.apache.maven.plugin.DefaultMojosExecutionStrategy.execute (DefaultMojosExecutionStrategy.java:39)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:159)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:105)
[...]

PR #7025, opened 2025-12-28
Remove test-bnd module from in-build pom.xml

chrisrueger:
This is replacement PR for https://github.com/bndtools/bnd/pull/7021 (which was in the wrong branch)

This is a (maybe a temporary) fix for https://github.com/bndtools/bnd/issues/7020 to make the master build work again. 

## Description

The test-bnd module reference was removed from the modules section in the in-build pom.xml

This fixes the failing test in https://github.com/bndtools/bnd/issues/7020

But not sure this is really the best solution, since it might fix only a symptom, and not the root cause. But maybe it serves as a base for discussion.

## Analysis

The cause seemed to be that maven-plugins/bnd-indexer-maven-plugin/src/it/test-bnd builds a fake bnd v1.50.0 bundle.

```
[INFO] Reactor Summary:
[INFO] 
[INFO] bnd 1.50.0 ......................................... SUCCESS [  0.272 s]
[INFO] org.apache.aries.async 0.0.1-SNAPSHOT .............. SUCCESS [  0.006 s]
[INFO] in-build 0.0.1 ..................................... FAILURE [  0.081 s]

```

And then the test in-build/pom.xml find this bnd 1.50.0 from the reactor instead of the remote maven central, and thus fails.

But I don't know why it suddenly happens and did not happen earlier.
Not sure my fix here is only fighting symptoms.

@timothyjward , since you worked on related PRs (see description of https://github.com/bndtools/bnd/issues/7020), maybe you have an idea.

Or mayb @bjhargrave has thoughts too since you did a larger PR related to the tests https://github.com/bndtools/bnd/pull/5252

### DIFF
--- a/maven-plugins/bnd-indexer-maven-plugin/src/it/in-build/pom.xml
+++ b/maven-plugins/bnd-indexer-maven-plugin/src/it/in-build/pom.xml
@@ -14,7 +14,6 @@
 	<packaging>pom</packaging>
 
 	<modules>
-		<module>../test-bnd</module>
 		<module>../test-snapshot</module>
 	</modules>
 


chrisrueger:
I merged that to make the master build pass again. We keep https://github.com/bndtools/bnd/issues/7020 open for now, so we can discuss.

Issue #7030, opened 2025-12-30
Fix SNAPSHOT deployments for Sonatype


peterkir:
Sonatype SNAPSHOT deployments are not working.
After debugging SNAPSHOT releases process, I detected that it is also required to deployment via artefact creation as required for releases.

PR #7032, opened 2025-12-30
fix issue #7030 - enable sonatype snapshot deployments

peterkir:


chrisrueger:
to make the build pass, just rebase on master, which contains my fix of https://github.com/bndtools/bnd/pull/7025

PR #7033, opened 2026-01-01
biz.aQute.bnd.annotation: compile with java 8

### LINKED ISSUE #7005: `biz.aQute.bnd.annotation` bytecode level causes downstream build issues
Unlike most of the BND toolchain, `biz.aQute.bnd.annotation` types are referenced directly from compiled library bytecode.

In Apache Log4j, we build with **JDK 17** and therefore use the latest BND release, but we must continue publishing artifacts with **Java 8 bytecode compatibility**. Recent `log4j-api` versions (Java 8 bytecode) expose a `compileOnly` dependency on `biz.aQute.bnd.annotation`, which is currently built with a **Java 17 bytecode baseline**.

As a result, users compiling with JDK 8 or 11 and `-Xlint:classfile` (often promoted to errors via `-Werror`) see compiler warnings or build failures due to the class file version mismatch.

While we can work around this by pinning an older annotation version, is there a reason `biz.aQute.bnd.annotation` (a pure annotation artifact) cannot be built with `--release 8`?

For context, other annotation libraries follow this approach:

- `error_prone_annotations` only requires Java 8, while the rest of the suite requires Java 21.
- `spotbugs-annotations` rolled back their baseline change in spotbugs/spotbugs#3497.

chrisrueger:
Closes #7005

remove Java 9+ features: It was simply the attributes of the `@Deprecated` annotation. So we just put that into the comment and use plain `@Deprecated` which is Java8 compatible.

Then enforced compilation with java8

TODO: Not sure if we really want to do this or what the drawbacks are. Maybe with [bnd 8.0.0](https://github.com/bndtools/bnd/wiki/WIP-ideas-%E2%80%90-bnd-8.0.0-breaking-changes) we should remove java 8 support fully.

### DIFF
--- a/biz.aQute.bnd.annotation/bnd.bnd
+++ b/biz.aQute.bnd.annotation/bnd.bnd
@@ -1,5 +1,13 @@
 # Set javac settings from JDT prefs
--include: ${workspace}/cnf/includes/jdt.bnd
+-include: ~${workspace}/cnf/includes/jdt.bnd
+
+# #7005: Keep this bundle compatible with Java 8 as long as possible. All other code is min. Java 17
+# It is because some libraries reference bnd annotations
+# like aQute.bnd.annotation.Cardinality, aQute.bnd.annotation.Resolution, aQute.bnd.annotation.spi.ServiceConsumer
+# directly in source code (e.g. https://github.com/apache/logging-log4j2/blob/752d023795fc518a38ecae73bb6455e2723cecea/log4j-api/src/main/java/org/apache/logging/log4j/util/PropertiesUtil.java)
+# but they need to stay compatible with java 8.
+javac.source = 8
+javac.target = 8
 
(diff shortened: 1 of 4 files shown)


peterkir:
Supporting Java 8 is a good idea from my point for now.
So removing Java 9 features for now, is a an obvious reason.

chrisrueger:
`biz.aQute.bnd.annotation` now has Java 8 class file version 52 (34 Hex)

https://bndtools.jfrog.io/artifactory/update-snapshot/biz/aQute/bnd/biz.aQute.bnd.annotation/7.3.0-SNAPSHOT/biz.aQute.bnd.annotation-7.3.0-20260103.163301-38.jar

image

laeubi:
For annotation I think its fine keep them at java8 I see no benefit in any higher java version.

PR #7077, opened 2026-02-02
DS: Ignore "Activate" annotations on super constructors

### LINKED ISSUE #7075: "Multiple constructors are annotated @Activate" when inherited class has also one constructor with @Activate annotation
Sometimes there is an inheritance among DS components. Consider the following setup:

- ComponentA having constructor with `@Activate` and another constructor (not having `@Activate` annotation).
- ComponentB extending ComponentA having one constructor with `@Activate`
- The instruction `-dsannotations-options: inherit` is set

At compile time the following error is emitted by Bnd:

```
ComponentA.<init>*(...)Multiple constructors are annotated @Activate
```

According to [DS Spec 1.5 112.3.4](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#service.component-constructor.injection) and [112.5.11](https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#service.component-activate.method) there is nothing specifically mentioned regarding inheritance, however for each class there is really only one suitable constructor having `@Activate` annotation, therefore this should be accepted by Bnd.

I tested with Bnd 7.2.1.

kwin:
Otherwise they could lead to compile time errors when "dsannotations-options" is set to "inherit".

This closes #7075

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/component/DSAnnotationTest.java
+++ b/biz.aQute.bndlib.tests/test/test/component/DSAnnotationTest.java
@@ -1537,6 +1537,79 @@ public class DSAnnotationTest {
 
 	}
 
+	@Component(name="top")
+	public static class TopConstructorInjection {
+
+		@Activate
+		public TopConstructorInjection(@SuppressWarnings("unused") BundleContext context) {
+			
+		}
+		@Reference
+		void setLogService(@SuppressWarnings("unused") LogService l) {}
+
+		void updatedLogService(@SuppressWarnings("unused") ServiceReference<?> ref) {
+
+		}
+
+		@Reference
+		protected void setPrivateLogService(@SuppressWarnings("unused") LogService l) {
+
+		}
+
+		@SuppressWarnings("unused")
+		private void updatedPrivateLogService(ServiceReference<?> ref) {
+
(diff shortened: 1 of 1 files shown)


chrisrueger:
Thanks @kwin 
Do you think it is possible to write a test for this? Maybe around  `test.component.DSAnnotationTest.testInheritance(String, String, String)` https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib.tests/test/test/component/DSAnnotationTest.java#L1509C7-L1509C22

kwin:
> Thanks @kwin Do you think it is possible to write a test for this? Maybe around `test.component.DSAnnotationTest.testInheritance(String, String, String)` https://github.com/bndtools/bnd/blob/master/biz.aQute.bndlib.tests/test/test/component/DSAnnotationTest.java#L1509C7-L1509C22

Done in https://github.com/bndtools/bnd/pull/7077/commits/2184eb0d438b64a54a0e412396ed420f4dc72493.

chrisrueger:
FYI I restarted the build. Github Actions had issues [yesterday](https://www.githubstatus.com/incidents/xwn6hjps36ty).

PR #7082, opened 2026-02-03
DS: Error on missing public noArg constructor and no Activate

### LINKED ISSUE #7078: Bail out at compile time if no suitable constructor in DS component found
Currently Bnd doesn't enforce the usage of the `@Activate` on constructors correctly. So the following component doesn't lead to an error from Bnd.

```
package demo;

import org.osgi.framework.BundleContext;

@Component
public class TestComponent {

    public TestComponent(BundleContext bundleContext) {
        
    }
}
```

Instead it generates an incorrect component descriptor:

```
<?xml version="1.0" encoding="UTF-8"?>
<scr:component xmlns:scr="http://www.osgi.org/xmlns/scr/v1.3.0" name="demo.TestComponent">
  <implementation class="demo.TestComponent"/>
</scr:component>
```

Constructor injection was only introduced with DS 1.4.0 (but the descriptor emits 1.3.0) and 1.3.0 only allows public constructors without an argument:

> The class must be public and have a public constructor without arguments (this is normally the default constructor) so component instances may be created by SCR with the newInstance method on Class.

(https://docs.osgi.org/download/r6/osgi.cmpn-6.0.0.pdf, 112.4.5)

This is clearly not the case in this example.

chrisrueger:
Closes #7078

Bnd will now throw an error if a class with a `@Component` annotation 
- does not have a public no-arg constructor 
- and no `@Activate` annotation

Why?

This makes bnd more compliant with the DS Spec: https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#service.component-constructor.injection

> If the constructor has a parameter count that does not match the value of the init attribute in the component element, then the constructor must not be considered. If the value of the init attribute is 0, the default value, then the public no-parameter constructor must be used.

The error looks like this:
```
The DS component class MyClass must declare a public no-arg constructor, or a public constructor annotated with @Activate.
```

e.g. for classes like:

```
@Component
public class ConstructorInjectionMissingDefaultNoArgConstructor {

   public ConstructorInjectionMissingDefaultNoArgConstructor(BundleContext bundleContext) {

   }
}
```

or 

```
@Component
class ConstructorInjectionMissingDefaultNoArgConstructor {

}
```

The latter has an implicit no-arg constructor, but since the class is package private, the constructor is not considered public by java.

This error will make builds fail. 
In bndtools Eclipse plugin this will result in an error and a marker at the Class. 

image

image

In each case no .jar is created and no XML DS Component XML is created.

## Why the  Breaking change label

- Added the `breakingchange` label to make people aware and to remind me to mention it in the next release notes. 
- But this might not be a breaking change per se.
- This PR makes bnd more compliant with the DS Spec https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#service.component-constructor.injection

> If the constructor has a parameter count that does not match the value of the init attribute in the component element, then the constructor must not be considered. If the value of the init attribute is 0, the default value, then the public no-parameter constructor must be used.

So basically bnd will behave more strict so that it does **not** always generate a DS Component XML Descriptor anymore. It will throw an error lke:

```
The DS component class MyClass must declare a public no-arg constructor, or a public constructor annotated with @Activate.
```

The new thing **public** no-arg constructor.
Before this PR bnd just required a **no-arg constructor** and generate a XML component descriptor. 
This would be an invalid component descriptor which would fail at runtime, because e.g. Felix SCR [explicitly looks for public constructors](https://github.com/apache/felix-dev/blob/2282645785948c7acc5d6a5c1c9359cb8eb3a34c/scr/src/main/java/org/apache/felix/scr/impl/inject/internal/ComponentConstructorImpl.java#L98). 

So although this is a behavioral change, it must not be considered breaking, because it basically makes bnd more spec compliant.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/component/DSAnnotationReader.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/component/DSAnnotationReader.java
@@ -117,6 +117,10 @@ public class DSAnnotationReader extends ClassDataCollector {
 	MethodSignature											constructorSig;
 	int														parameter;
 	int														constructorArg;
+
+	private TypeRef											implementationClass;
+	private boolean											inImplementationClass;
+	private boolean											hasPublicNoArgConstructor;
 	TypeRef													className;
 	Analyzer												analyzer;
 	MultiMap<String, MethodDef>								methods						= new MultiMap<>();
@@ -171,6 +175,17 @@ public class DSAnnotationReader extends ClassDataCollector {
 		clazz.parseClassFileWithCollector(this);
 		if (component.implementation == null)
 			return null;
+
+		// Validate constructors for the component implementation class:
+		// - either it has a public no-arg constructor
+		// - or it uses constructor injection (public @Activate constructor)
+		if ((component.init == null || component.init.intValue() == 0) && !hasPublicNoArgConstructor) {
+			DeclarativeServicesAnnotationError details = new DeclarativeServicesAnnotationError(
+				implementationClass.getFQN(), null, null, ErrorType.INVALID_COMPONENT_TYPE);
+			details.addError(analyzer,
+				"[%s] The DS component class %s must declare a public no-arg constructor, or a public constructor annotated with @Activate.",
+				details.location(), implementationClass.getFQN());
+		}
 
 		if (options.contains(Options.inherit)) {
 			baseclass = false;
@@ -1508,6 +1523,14 @@ public class DSAnnotationReader extends ClassDataCollector {
 	@Override
 	public void classBegin(int access, TypeRef name) {
 		className = name;
+
+		// The first class we visit during parsing is the component implementation class.
+		// When scanning superclasses (inherit option), we do not want to change the
+		// constructor validation target.
+		if (implementationClass == null) {
+			implementationClass = name;
+		}
+		inImplementationClass = (implementationClass == name);
 	}
 
 	@Override
@@ -1583,6 +1606,15 @@ public class DSAnnotationReader extends ClassDataCollector {
 
 		methods.add(method.getName(), method);
 		methodSig = getMethodSignature(method);
+
+		// Track whether the *implementation* class has a public no-arg ctor.
+		if (inImplementationClass && method.isConstructor() && method.isPublic()) {
+			// methodSig is available and based on signature/descriptor
+
+			if ((methodSig != null) && (methodSig.parameterTypes.length == 0)) {
+				hasPublicNoArgConstructor = true;
+			}
+		}
 	}
 
 	private MethodSignature getMethodSignature(MethodDef method) {
--- a/biz.aQute.bndlib.tests/test/test/component/DSAnnotationTest.java
+++ b/biz.aQute.bndlib.tests/test/test/component/DSAnnotationTest.java
@@ -3876,7 +3876,7 @@ public class DSAnnotationTest {
 	}
 
 	@Component
-	static class VolatileField {
+	public static class VolatileField {
 		@Reference
 		private volatile LogService	log1;
 		@Reference
@@ -3915,7 +3915,7 @@ public class DSAnnotationTest {
 	}
 
 	@Component
-	static class FinalDynamicCollectionField {
+	public static class FinalDynamicCollectionField {
 		@Reference(policy = ReferencePolicy.DYNAMIC)
 		private final List<LogService>					logs1	= new CopyOnWriteArrayList<>();
 
@@ -3977,7 +3977,7 @@ public class DSAnnotationTest {
 	}
 
 	@Component
-	static class FieldCardinality {
+	public static class FieldCardinality {
 		@Reference
 		private List<LogService>				log1;
 		@Reference
@@ -4043,7 +4043,7 @@ public class DSAnnotationTest {
 	}
 
 	@Component
-	static class MismatchedUnbind {
+	public static class MismatchedUnbind {
 		@Reference
 		void setLogService10(LogService ls) {}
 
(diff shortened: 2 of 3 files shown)


biz.aQute.bndlib/src/aQute/bnd/component/DSAnnotationReader.java
```diff
@@ -172,6 +176,17 @@ private ComponentDef getDef() throws Exception {
 		if (component.implementation == null)
 			return null;
 
+		// Validate constructors for the component implementation class:
+		// - either it has a public no-arg constructor
+		// - or it uses constructor injection (public @Activate constructor)
+		if ((component.init == null) && !hasPublicNoArgConstructor && V1_3.equals(component.version)) {
```
kwin: I think `component.init` being set to `0` explicitly should be treated equally to not set at all (https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#service.component-component.element). That way no need to check the version of DS.

chrisrueger: Ok, I will push something in a minute.

biz.aQute.bndlib/src/aQute/bnd/component/DSAnnotationReader.java
```diff
@@ -1579,6 +1602,14 @@ public void method(MethodDef method) {
 
 		methods.add(method.getName(), method);
 		methodSig = getMethodSignature(method);
+
+		// Track whether the *implementation* class has a public no-arg ctor.
+		if (inImplementationClass && "<init>".equals(method.getName()) && method.isPublic()) {
```
chrisrueger: @kwin I still struggle with a now failing test and I think it is realated to the checks

```
if (inImplementationClass && "<init>".equals(method.getName()) && method.isPublic()) {
```

Test failing: `test.component.DSAnnotationTest.testFinalDynamicCollectionField()` which now errors with 

```
[test.component.DSAnnotationTest$FinalDynamicCollectionField] The DS component class test.component.DSAnnotationTest$FinalDynamicCollectionField must declare a public no-arg constructor, or a public constructor annotated with @Activate.
```

```
@Component
	static class FinalDynamicCollectionField {
		@Reference(policy = ReferencePolicy.DYNAMIC)
		private final List<LogService>					logs1	= new CopyOnWriteArrayList<>();

		@Reference(policy = ReferencePolicy.DYNAMIC)
		private final CopyOnWriteArrayList<LogService>	logs2	= new CopyOnWriteArrayList<>();
	}
```

Theoretically this error is correct. 
Because the class in the test is package private (thus not public), so the implicit constructor is a no-args constructor, but not public, as it inherits the visibility of the class (package private). 


https://docs.osgi.org/specification/osgi.cmpn/8.0.0/service.component.html#service.component-constructor.injection

> If the constructor has a parameter count that does not match the value of the init attribute in the component element, then the constructor must not be considered. If the value of the init attribute is 0, the default value, then the **public no-parameter constructor** must be used.

If the spec really means **public** no-parameter constructor, then the test should be modified, because the class is not public. 

But before my change the test was passing, so I have the feeling that the spec actually means "the default no-args constructor, whatever the inherited class visibility is. 

Not sure how to proceed here.

Feel free to copy my branch and play around with it.

chrisrueger: @kwin (for now) I fixed the failing tests in 0184e9ec9fe635186bfb078acf034073496c0a93 by changing all test classes from `static class` to `public static class` so that their implicit no-arg constructors become public too.

kwin: At least Felix SCR Runtime only considers the public constructors (in https://github.com/apache/felix-dev/blob/2282645785948c7acc5d6a5c1c9359cb8eb3a34c/scr/src/main/java/org/apache/felix/scr/impl/inject/internal/ComponentConstructorImpl.java#L98)

chrisrueger: Thanks. Ok, my interpretation would be: If SCR is spec compliant (explicitly requiring public), then bnd should also be equally strict. Meaning that throwing an error if no-arg constructor is not public is correct.

My worry was about breaking change. But theoretically nobody should have non-public no-arg constructors, because they would fail at runtime, right?

kwin: I am not aware of any other DS impl. Equinox dropped their own some time ago, so being compliant with both Felix SCR and OSGi spec is safe enough.

kwin:
For me this is not exactly a breaking change. Granted it now breaks at compile time instead of at run time as before. But even before this was not working.

chrisrueger:
> For me this is not exactly a breaking change. Granted it now breaks at compile time instead of at run time as before. But even before this was not working.

I agree. Added explanation in the description, why I added this label.

peterkir:
Looks for me like bugfix, which should be incorporated as soon as possible.

chrisrueger:
Just linking https://github.com/osgi/osgi/issues/638 since in OSGi Slack it was suggested to bring this topic up in OSGi issue tracker.

chrisrueger:
This PR seems to be not fully correct: see https://github.com/bndtools/bnd/issues/7078#issuecomment-3848359026

laeubi:
Just FYI this is how it is handled by PDE

grafik

so maybe one can resort to this simpler check that the class itself must be public?

chrisrueger:
> so maybe one can resort to this simpler check that the class itself must be public?

I tested and at least this case is covered by the current check too. I think the current check for the constructor still is useful because it covers also a case like this with non-public constructors:

```
@Component
public class Foo {

	Foo() {
	}
}

```

or 

```
@Component
public class Foo {

	private Foo() {
	}
}

```

kriegfrj:
> I tested and at least this case is covered by the current check too. I think the current check for the constructor still is useful because it covers also a case like this with non-public constructors:
>
:+1: I think both checks are ultimately necessary. Having the error on a non-public class would have saved me some dev cycles recently, as I tried to implement a component as a package-private class and didn't get a compile-time warning. Even the runtime warning was not explicit and I eventually figured out that my component simply wasn't starting, and figured out by trial-and-error that it was because the class was not public.

The problem with relying on the constructor check only might be that the error message is a bit confusing. You might have a constructor that is flagged as public and then struggle to interpret the error message that says it isn't. A more explicit error message indicating that the class must be public would be a more direct (and more helpful) way to pinpoint the exact problem for the developer.

chrisrueger:
@kriegfrj thanks for the suggestion. I have added a rewording proposal here https://github.com/bndtools/bnd/pull/7101/changes

PR #7088, opened 2026-02-05
Fix baseline problems because of missing stuff in About.java to fix Eclipse version issue

### LINKED ISSUE #7087: bnd 7.3.0 introduces a Unresolved requirement problem with org.eclipse.pde.core
When starting 7.3.0 (master) in I get this:

```
Unresolved requirement: Import-Package: org.eclipse.pde.core.target; bundle-symbolic-name="org.eclipse.pde.core"; bundle-version="[3.17.0,4.0.0)"
    -> Export-Package: org.eclipse.pde.core.target; bundle-symbolic-name="org.eclipse.pde.core"; bundle-version="3.17.200.v20231118-2008"; version="0.0.0"
       org.eclipse.pde.core [235]
         Unresolved requirement: Import-Package: aQute.bnd.osgi; version="[5.5.0,8.0.0)"
```

and also the `/bndtools.core/bndtools.cocoa.macosx.aarch64.bndrun` does not resolve anymore.

Image

This happend since e5d4676c5fac889485d676d47921ad4bb24c22ca where I added

```
@Version("8.0.0")
package aQute.bnd.osgi;
```

The reason for this commit was that bnd reported a baseline error, because in this package I changed `About.java`

```
public static final Version					CURRENT		= _7_2;
```

to

```
public static final Version					CURRENT		= _7_3;
```

and then I got something this:

```
Baseline mismatch for package aQute.bnd.osgi, MAJOR change. Current is 7.6.0, repo is 7.5.0, suggest 8.0.0 or -
MAJOR                PACKAGE    aQute.bnd.osgi
  MAJOR              CLASS      aQute.bnd.osgi.About
    REMOVED          FIELD      CHANGES_7_2_1
      REMOVED        ACCESS     final
      REMOVED        ACCESS     static
      REMOVED        RETURN     java.lang.String[]
    ADDED            FIELD      CHANGES_7_3
      ADDED          ACCESS     final
      ADDED          ACCESS     static
      ADDED          RETURN     java.lang.String[]
    REMOVED          FIELD      _7_2_1
      REMOVED        ACCESS     final
      REMOVED        ACCESS     static
      REMOVED        RETURN     aQute.bnd.version.Version
    ADDED            FIELD      _7_3
      ADDED          ACCESS     final
      ADDED          ACCESS     static
      ADDED          RETURN     aQute.bnd.version.Version
  MICRO              ANNOTATED  org.osgi.annotation.versioning.Version
    REMOVED          PROPERTY   value='7.5.0'
    ADDED            PROPERTY   value='7.6.0'
  REMOVED            VERSION    7.5.0
  ADDED              VERSION    7.6.0
But now it seems that a bundle of Eclips has a version range which excludes this 8.0.0

```

But Eclipse PDE seems to have a hard range limit on `Import-Package: aQute.bnd.osgi; version="[5.5.0,8.0.0)` which then suddenly would fail since 8.0.0 is not in this range. 

## Idea

We should exclude `About.java` from baselining because it will increase with every release we make and the changes are not worth bumping the package version - in my opinion.

chrisrueger:
Closes #7087
Import and apply @BaselineIgnore("7.6.0") to aQute.bnd.osgi.About and change the package version in package-info.java from "8.0.0" to "7.6.0". This aligns the package metadata and suppresses baseline checks for the About class

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/About.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/About.java
@@ -47,6 +47,8 @@ import aQute.lib.utf8properties.UTF8Properties;
  * The namespace is maintained by {@link Descriptors}, which here is owned by
  * {@link Analyzer}. A special class, {@link Packages} maintains the attributes
  * that are found in the code. @version $Revision: 1.2 $
+ * <hr />
+ * Ensure you only add Major and minor versions (the 7.2.1 below was a mistake)
  */
 public class About {
 	private final static Logger					logger		= LoggerFactory.getLogger(About.class);
@@ -74,11 +76,15 @@ public class About {
 	public static final Version					_7_0		= new Version(7, 0, 0);
 	public static final Version					_7_1		= new Version(7, 1, 0);
 	public static final Version					_7_2		= new Version(7, 2, 0);
+	public static final Version					_7_2_1			= new Version(7, 2, 1);
 	public static final Version					_7_3		= new Version(7, 3, 0);
 	public static final Version					CURRENT		= _7_3;
 
 	public static final String[]				CHANGES_7_3	= {
 		"See https://github.com/bndtools/bnd/wiki/Changes-in-7.3.0 for a list of changes."
+	};
+	public static final String[]				CHANGES_7_2_1	= {
+		"See https://github.com/bndtools/bnd/wiki/Changes-in-7.2.1 for a list of changes."
 	};
 	public static final String[]				CHANGES_7_2	= {
 		"See https://github.com/bndtools/bnd/wiki/Changes-in-7.2.0 for a list of changes."
--- a/biz.aQute.bndlib/src/aQute/bnd/osgi/package-info.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/osgi/package-info.java
@@ -1,4 +1,4 @@
-@Version("8.0.0")
+@Version("7.6.0")
 package aQute.bnd.osgi;
 
 import org.osgi.annotation.versioning.Version;


chrisrueger:
@peterkir @laeubi Could you have a look if this makes sense?

biz.aQute.bndlib/src/aQute/bnd/osgi/About.java
```diff
@@ -48,6 +49,7 @@
  * {@link Analyzer}. A special class, {@link Packages} maintains the attributes
  * that are found in the code. @version $Revision: 1.2 $
  */
+@BaselineIgnore("7.6.0")
 public class About {
```
laeubi, line 53: Maybe it should be final then I think baseline compare will not mind if you add new things in a minor version.

chrisrueger: Ah, didn't know. I will try.

chrisrueger: Hmm making the class final still resulted in Baseonline error. 
I will try to merge this to master since the current baseline in the repo (I guess from the SNAPSHOT build) is still 8.0.0 . So I first need something in the repo which is reverted to 7.6.0, to see if the error goes away. 

image

laeubi: > since the current baseline in the repo (I guess from the SNAPSHOT build)

This sounds wrong. The baseline should be the previous release (7.2.1), but maybe the baseline check also has an error, what exactly it reasons about what a major change is needed?

In any case I suspect making it final is a breaking change so it will only be effective after the next release.

chrisrueger: Yes I had a messed up local .m2 repo. I had a look together with @peterkir and we fixed it.

chrisrueger: We also discovered the actual problem: 
I made a mistake when I released 7.2.1. I added a version field for 7.2.1 in About.java which I forgot to transfer back to master branch. This was the reason baselining said "REMOVED". I didn't really noticed this at first. 
So my fix is to add it back in master branch.

PR #7097, opened 2026-02-13
Upgrade bndtools to Eclipse 2025-12

chrisrueger:
This allows starting bndtools development environment in Eclipse with a latest Eclipse 2025-12. This makes it easier debugging eclipse issues in more recent eclipse versions. We might now merge this to master any time soon, as we want to stay behind. But in case of debugging, we could just switch to this branch.


chrisrueger:
The build fails probably because Eclipse now brings JUnit 6 and this mixes with JUnit 5 and explodes, does not discover tests and what not. 
I won't bother spending time fixing this. I just needed the Eclipse portion of it. I can start bndtools Eclipse dev instance and that is all I needed from this branch. Just an FYI. This PR will stay broken for a long time.

kriegfrj:
As an alternative, why not have both launch files side-by-side? We could name the launch files `bndtools.<arch>-latest.bndrun` and they can live side-by-side. Similarly, we can have the Eclipse 2025 repo live side-by-side with the baseline repo. This repo could be excluded from the build, and could also be blacklisted in the baseline bndruns so that the baseline bndruns only use the baseline repo for resolving. Then it would be ready and available any time you needed it.

I started down this path once before (it's how I've debugged issues that happened only on later versions of Eclipse) but I never managed to find the time to polish it well enough to make it a PR.

chrisrueger:
Good point , i saw your "-latest" bndrun. I give that a try, but not sure two eclipse repos side by side will not cause issues. But at least i could document the steps somewhere.

kriegfrj:
> Good point , i saw your "-latest" bndrun. I give that a try, but not sure two eclipse repos side by side will not cause issues. But at least i could document the steps somewhere.

Two Eclipse repos side-by-side with no further configuration probably would cause problems. However, you can configure the bndrun files to omit certain repositories, and I believe that we can do the same in the build repositories (though my memory is hazy). So as long you configure the build + base bndruns to use the base Eclipse repo, and the "latest" bndruns to use the latest Eclipse repo and not the base Eclipse repo, it should all work.

chrisrueger:
> I believe that we can do the same in the build repositories (though my memory is hazy)

I don't know about the build-repo. I believe all repositories present are considered for the build. I could imagine there will be problems when using `version=latest` in bundles, because then a newer Eclipse repo might be pulled in. 

Some java Eclipse API changes could also be problematic (they removed methods and constants), but the changes in this PR are minimal, and I could probably backport them to even work with an older Eclipse. 

But let's see when I am one step further.

kriegfrj:
> > I believe that we can do the same in the build repositories (though my memory is hazy)
> 
> I don't know about the build-repo. I believe all repositories present are considered for the build. I could imagine there will be problems when using `version=latest` in bundles, because then a newer Eclipse repo might be pulled in.
> 

This is true. This is why it's generally a good idea to not include the `version` attribute in `-buildpath` - without it, Bnd will choose the lowest available version to compile against, which would be what we want in this situation.

> Some java Eclipse API changes could also be problematic (they removed methods and constants), but the changes in this PR are minimal, and I could probably backport them to even work with an older Eclipse.
> 

These would not be an issue if we were still building against the oldest version of Eclipse.

> But let's see when I am one step further.

:+1: Sorry I'm not more active help these days...

peterkir:
I also wanted to update this branch and incorporate compatible and latest eclipse version into bnd repo.
As soon as the feature support PR is merged we can hopefully more easily create launch configs on the newer eclipse repos. Otherwise we have to mirror the subset of eclipse repo into our JFrog to reference it.
The oomph setups are also always on the latest eclipse version and we should be able to work on the latest version in parallel, when a issue or bug occurs.

kriegfrj:
> I also wanted to update this branch and incorporate compatible and latest eclipse version into bnd repo. As soon as the feature support PR is merged we can hopefully more easily create launch configs on the newer eclipse repos. Otherwise we have to mirror the subset of eclipse repo into our JFrog to reference it. 

Just for the record: You don't actually *have* to mirror the subset of the Eclipse repo into JFrog. Bnd has inbuilt support for p2 repos and you can reference it directly at the Eclipse download repos.

The main reason the JFrog subset mirror was created, historically, was for CI build stability. The Eclipse download mirrors would occasionally transiently fail during CI build and cause the build to fail. Also worth noting: due to the structure of P2, the P2 implementation in Bnd must download every bundle in the P2 repo upon initialisation so that it can create its own index in its own format. I think from memory this is an atomic operation and so a single bundle download failure can cause the initialisation to fail, which makes it more vulnerable to transient errors.

@bjhargrave mirrored it on JFrog as an OSGi-shaped repo and this proved more stable. The main reason he only mirrored a subset was to save space on the mirror servers. Mirroring only a subset may also speed up build-time and run-time performance of Bnd a little bit (though probably not much as Bnd can happily work with very large repos), but it is also a hinderance sometimes if you want to add a new Bndtools feature that relies on an Eclipse bundle that's not mirrored. Admittedly this doesn't happen often, but when it does happen it is such a big barrier to adding the new feature that the result is typically that the new feature simply doesn't see the light of day.

In the case of a "latest" Eclipse repo in the Bndtools workspace, we could disable it in CI and avoid the transient errors, and hence we could probably get away with using the P2 repo directly.

PR #7168, opened 2026-03-20
Docs: fix index.md clash for commands

chrisrueger:
- now 00-overview.md is accessible under /commands/overview.html
- and the generated index.md (from the index CLI command) under /commands/index.html
- navbar works too.

TODO we should think about if we adopt this pattern for _heads and _macros  too


peterkir:
All improvements are welcome 🥳

PR #7221, opened 2026-05-07
add -XX:+UseStringDeduplication

chrisrueger:
Hope that this helps with large XML files during long running testcases which currently may be responsible for the failing fuild with Java Out Of Memory errors


chrisrueger:
Does not seem to have an effect

PR #7247, opened 2026-05-25
Allow overriding HTTPClient request headers

### LINKED ISSUE #7228: Maven Central is blocking our IP address when we use maven via bndtools
Seems like Maven Central modified their protection yesterday. Since yesterday, our builds are often failing because the downloading of maven libraries fails.

I discovered that it's triggered by an invalid User-Agent. When we proxy the requests through an reverse proxy and modify the User-Agent to match the original maven User-Agent in style of "Apache-Maven ...", it works.

Original maven is using a custom user-agent, configured here: https://github.com/apache/maven/blob/master/impl/maven-core/src/main/java/org/apache/maven/internal/aether/DefaultRepositorySystemSessionFactory.java#L172

Is it feasable to also configure the same User-Agent in bndtools?
Is this the correct place to add the configuration parameter? https://github.com/bndtools/bnd/blob/master/biz.aQute.bnd.maven/src/aQute/bnd/maven/lib/resolve/RemotePostProcessor.java#L96

chrisrueger:
Closes #7228

This pull request improves the handling and documentation of HTTP request headers in the `HttpClient` class. The main change ensures that custom headers provided by plugins are not overridden by default headers, and the documentation now explains how to customize headers such as `User-Agent`.

**HTTP Header Handling Improvements:**

* Modified `setHeaders` in `HttpClient.java` to only set a request property if it has not already been set, preventing default headers from overwriting custom ones provided by plugins.

**Documentation Updates:**

* Added a new section to the documentation explaining how to override HTTP request headers (such as `User-Agent`) using a plugin in your build configuration.

### DIFF
--- a/biz.aQute.bndlib.comm.tests/test/aQute/bnd/comm/tests/HttpClientTest.java
+++ b/biz.aQute.bndlib.comm.tests/test/aQute/bnd/comm/tests/HttpClientTest.java
@@ -899,4 +899,53 @@ public class HttpClientTest {
 			}
 		}
 	}
+
+	@Test
+	public void testPluginCannotOverrideExplicitUserAgent() throws Exception {
+
+		// 1. no overriding of User-Agent
+		try (Processor p = new Processor()) {
+
+			// try to customize the user-agent, which should not work
+			p.setProperty("-plugin.my-override-useragent", "aQute.bnd.url.ConnectionSettings;match=\""
+				+ httpServer.getBaseURI() + "/*\";User-Agent=\"My own User Agent\"");
+
+			try (HttpClient client = new HttpClient()) {
+				client.setRegistry(p);
+
+				// this sets explicit User-Agent via .headers()
+				String response = client.build()
+					.headers("User-Agent", "Use me!")
+					.get(String.class)
+					.go(httpServer.getBaseURI("user-agent"));
+
+				assertThat(response).contains("Use me!");
+			}
+
+		}
+	}
+
+	@Test
+	public void testPluginCanSetUserAgent() throws Exception {
+
+		try (Processor p = new Processor()) {
+
+			// now customize the user-agent, which now should work
+			p.setProperty("-plugin.my-override-useragent", "aQute.bnd.url.ConnectionSettings;match=\""
+				+ httpServer.getBaseURI() + "/*\";User-Agent=\"My own User Agent\"");
+
+			try (HttpClient client = new HttpClient()) {
+				client.setRegistry(p);
+
+				// because .headersIfAbsent() only sets the default if not set
+				// otherwise
+				String response = client.build()
+					.headersIfAbsent("User-Agent", "Don't use me!")
+					.get(String.class)
+					.go(httpServer.getBaseURI("user-agent"));
+
+				assertThat(response).contains("My own User Agent");
+			}
+		}
+	}
 }
--- a/biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
@@ -567,6 +567,7 @@ public class HttpClient implements Closeable, URLConnector {
 					Dates.formatMillis(Dates.RFC_7231_DATE_TIME, request.ifUnmodifiedSince));
 			}
 
+			setHeadersIfAbsent(request.headersIfAbsent, con);
 			setHeaders(request.headers, con);
 
 			configureHttpConnection(request.verb, hcon);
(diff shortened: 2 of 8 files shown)


biz.aQute.bndlib.comm.tests/test/aQute/bnd/comm/tests/HttpClientTest.java
```diff
@@ -899,4 +899,42 @@ public boolean isCanceled() {
...
+
+			// now customize the user-agent
+			p.setProperty("-plugin.my-override-useragent", "aQute.bnd.url.ConnectionSettings;match=\""
+				+ httpServer.getBaseURI() + "/*\";User-Agent=\"My own User Agent\"");
+
+			try (HttpClient client = new HttpClient()) {
+				client.setRegistry(p);
+
+				String response = client.build()
+					.headers("User-Agent", "Don't use me!")
```
bjhargrave: Once concern is that there is no way to distinguish between headers that default vs headers which are specifically set. In the case here, I would expect that the "Don't use me!" user agent would be set since it is that most specific in the code, no?

Perhaps what we need is a `headersIfAbsent` method on HttpClient which only sets the header if there is no other known value like one from ConnectionsSettings. Then all code in Bnd calling `headers("User-Agent", "Bnd")` would instead call `headersIfAbsent("User-Agent", "Bnd")` to provide a default value if not otherwise set. 

This would allow, from an HttpClient API use, that calling `headers` will set the header (e.g/. `Map.put`) while `headersIfAbsent` would set the header only if there is no existing value (e.g. `Map.putIfAbsent`).

chrisrueger: I like this idea - makes the intent more expplicit. I'll give that a try.

biz.aQute.bndlib.comm.tests/test/aQute/bnd/comm/tests/HttpClientTest.java
```diff
@@ -899,4 +899,46 @@ public boolean isCanceled() {
 			}
 		}
 	}
+
+	@Test
+	public void testPluginCanOverrideDefaultUserAgent() throws Exception {
```
bjhargrave: Show probably be named something like `testPluginCannotOverrideExplicitUserAgent`.

biz.aQute.bndlib.comm.tests/test/aQute/bnd/comm/tests/HttpClientTest.java
```diff
@@ -899,4 +899,46 @@ public boolean isCanceled() {
...
+					.headers("User-Agent", "Use me!")
+					.get(String.class)
+					.go(httpServer.getBaseURI("user-agent"));
+
+				assertThat(response).contains("Use me!");
+			}
+
+		}
+
+		try (Processor p = new Processor()) {
```
bjhargrave, line 931: I would break this out into its own test method. `testPluginCanSetUserAgent`

bjhargrave (commented):
Looks good with some comments.

biz.aQute.bndlib.comm.tests/test/aQute/bnd/comm/tests/HttpClientTest.java
chrisrueger: Done and squashed

biz.aQute.bndlib.comm.tests/test/aQute/bnd/comm/tests/HttpClientTest.java
chrisrueger: Done and squashed

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -726,6 +727,21 @@ private void setHeaders(Map<String, String> headers, URLConnection con) {
...
+			MapStream<String, String> stream = MapStream.ofNullable(headers);
+			if (logger.isDebugEnabled()) {
+				stream = stream.peek((k, v) -> logger.debug("set header {}={}", k, v));
+			}
+			stream.forEachOrdered((k, v) -> {
+				String existing = con.getRequestProperty(k);
+				// don't override headers which are already set
+				// e.g. via -plugin aQute.bnd.url.ConnectionSettings
+				if (existing == null) {
+					con.setRequestProperty(k, v);
```
bjhargrave, line 740:
```suggestion
                    logger.debug("set header {}={}", k, v)
					con.setRequestProperty(k, v);
```

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -726,6 +727,21 @@ private void setHeaders(Map<String, String> headers, URLConnection con) {
 			stream.forEachOrdered(con::setRequestProperty);
 		}
 
+		private void setHeadersIfAbsent(Map<String, String> headers, URLConnection con) {
+			MapStream<String, String> stream = MapStream.ofNullable(headers);
+			if (logger.isDebugEnabled()) {
+				stream = stream.peek((k, v) -> logger.debug("set header {}={}", k, v));
+			}
```
bjhargrave, line 734:
```suggestion
```

You may want to move the debug into the `existing == null` block so it does not log when it does not set the header.

chrisrueger:
Thanks @bjhargrave for reviewing

biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -567,6 +567,7 @@ private TaggedData connect() throws Exception {
 					Dates.formatMillis(Dates.RFC_7231_DATE_TIME, request.ifUnmodifiedSince));
 			}
 
+			setHeadersIfAbsent(request.headersIfAbsent, con);
 			setHeaders(request.headers, con);
```
bjhargrave, line 571:
```suggestion
			setHeaders(request.headers, con);
			setHeadersIfAbsent(request.headersIfAbsent, con);
```

Set the explicit headers first, them the ifAbsent. With the logging change I suggest in `setHeadersIfAbsent` this will accurately log the headers being set.

bjhargrave (commented):
Some changes for accurate logging.

chrisrueger:
@bjhargrave sorry I was too quick merging. I'll create another PR with your suggestions

chrisrueger:
@bjhargrave created PR https://github.com/bndtools/bnd/pull/7249 for the remaining changes.

PR #7249, opened 2026-05-26
Set explicit headers before defaults and refine logging

chrisrueger:
Contains some suggested improvements of previous PR which I merged too fast.

- Call setHeaders(...) before setHeadersIfAbsent(...) so header defaults are only applied when a header is not already present.
- Improve logging of default headers (only log if set)

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
@@ -567,8 +567,8 @@ public class HttpClient implements Closeable, URLConnector {
 					Dates.formatMillis(Dates.RFC_7231_DATE_TIME, request.ifUnmodifiedSince));
 			}
 
-			setHeadersIfAbsent(request.headersIfAbsent, con);
 			setHeaders(request.headers, con);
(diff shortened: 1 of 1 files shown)


biz.aQute.bndlib/src/aQute/bnd/http/HttpClient.java
```diff
@@ -728,18 +728,18 @@ private void setHeaders(Map<String, String> headers, URLConnection con) {
...
-					con.setRequestProperty(k, v);
-				}
-			});
+			MapStream.ofNullable(headers)
+				.forEachOrdered((k, v) -> {
+					String existing = con.getRequestProperty(k);
+					// don't override headers which are already set
+					// e.g. via -plugin aQute.bnd.url.ConnectionSettings
+					if (existing == null) {
+						if (logger.isDebugEnabled()) {
```
bjhargrave: You don't need the isDebugEnabled test now since there is no work to avoid (e.g. adding a peek to the stream).

chrisrueger: Done

bjhargrave (approved):
Thanks!

PR #7296, opened 2026-06-26
Exclude Eclipse features from P2 Repo .versions()

### LINKED ISSUE #7279: [bndtools 7.3.0] mix up features and bundles in p2 repository
Hi,

we're using bndtools `7.2.1` to build our applications. We also put several commonly used repositories like Eclipse's P2 platform repository into a `bnd workspace library` to conveniently supply it into different applications (as own repository entry to retrieve dependencies from). The used repository provider is `aQute.bnd.repository.p2.provider.P2Repository`.

With bndtools `7.3.0` this isn't working anymore because the content of the P2 repository appears to be mixed up, specifically the versions for `org.eclipse.emf.common`. Also the icon shown is not the usual bnd bundle "brick" but the Eclipse PDE bundle icon.

Image

Cleaning the cache folder, syncing the workspace and restarting Eclipse multiple times does not help. The behavior occurs with bndtools 7.3.0 and Eclipse version 2025-12, 2026-03 and 2026-06.

The used P2 repository is the one from `download.eclipse.org/eclipse/updates/4.36/R-4.36-202505281830`, and we assume that its content doesn't change so the difference appears to be between bndtools `7.2.1` and bndtools `7.3.0`.

chrisrueger:
Closes #7279 
This part 2 of https://github.com/bndtools/bnd/pull/7292

With this PR the compilation error is gone, because only one version of the real bundle remains (the feature does not appear as a bundle version)

image

@peterkir I am not sure if there are side-effects I have not thought of if we filter out Eclipse Features resources like this. But at least it fixes the compilation probem stated in https://github.com/bndtools/bnd/issues/7279#issuecomment-4809326571 because `-buildpath` picks the feature bundle (which should not be available at all because it contains no classes)

## Summary

This pull request improves how Eclipse features are handled in the P2 repository implementation. The main focus is to ensure that Eclipse features are not returned as bundle versions, preventing them from being incorrectly included in build paths, while still allowing them to be accessed via dedicated feature APIs. It also introduces a new method to retrieve detailed version information and updates tests to verify the correct filtering of features.

**Feature filtering and handling:**

* The `versions(String bsn)` method in `P2Indexer` now filters out versions corresponding to Eclipse features (`org.eclipse.update.feature`), so only actual bundle versions are returned. This prevents features from being added to the build path accidentally.
* A static constant `ECLIPSE_FEATURE_CAPABILITY` is introduced to avoid hardcoding the feature type string in multiple places, improving maintainability and reducing errors. [[1]](diffhunk://#diff-d0bef1fb0f92387de1e01c72fb565264dbbfd5d52f5953e4cdcbd893a1747630R75) [[2]](diffhunk://#diff-d0bef1fb0f92387de1e01c72fb565264dbbfd5d52f5953e4cdcbd893a1747630L165-R169) [[3]](diffhunk://#diff-d0bef1fb0f92387de1e01c72fb565264dbbfd5d52f5953e4cdcbd893a1747630L402-R428) [[4]](diffhunk://#diff-d0bef1fb0f92387de1e01c72fb565264dbbfd5d52f5953e4cdcbd893a1747630L467-R493) [[5]](diffhunk://#diff-d0bef1fb0f92387de1e01c72fb565264dbbfd5d52f5953e4cdcbd893a1747630L567-R593)

**API enhancements:**

* The `BridgeRepository` class adds a new `versionInfos(String bsn)` method, which returns a map of versions to their associated `ResourceInfo`, allowing more granular inspection and filtering of resources.

**Testing improvements:**

* The `P2RepositoryTest` is updated with new assertions to verify that feature versions are properly filtered out from the main repository, but remain accessible via the `getFeatures()` API. This ensures the intended behavior is enforced and tested.

**Versioning:**

* The package version for `aQute.bnd.osgi.repository` is incremented from 3.3.0 to 3.4.0 to reflect the new API addition.

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
@@ -13,9 +13,11 @@ import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
 import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.concurrent.TimeUnit;
 
 import org.osgi.framework.namespace.IdentityNamespace;
@@ -29,6 +31,7 @@ import org.slf4j.LoggerFactory;
 
 import aQute.bnd.http.HttpClient;
 import aQute.bnd.osgi.repository.BridgeRepository;
+import aQute.bnd.osgi.repository.BridgeRepository.ResourceInfo;
 import aQute.bnd.osgi.repository.ResourcesRepository;
 import aQute.bnd.osgi.repository.XMLResourceGenerator;
 import aQute.bnd.osgi.repository.XMLResourceParser;
@@ -69,6 +72,7 @@ class P2Indexer implements Closeable {
 	private final PromiseFactory			promiseFactory;
 	private volatile BridgeRepository		bridge;
 	private static final SupportingResource	RECOVERY	= new ResourceBuilder().build();
+	private static final String				ECLIPSE_FEATURE_CAPABILITY	= "org.eclipse.update.feature";
 
 	P2Indexer(Unpack200 processor, Reporter reporter, File location, HttpClient client, URI url, String name)
 		throws Exception {
@@ -152,7 +156,7 @@ class P2Indexer implements Closeable {
 				if (!bsn.equals(id) || !osgiVersion.equals(capVersion)) {
 					continue;
 				}
-				
+
 				// Filter by type: when requestedType is specified, it must match
 				// When requestedType is null (default case), exclude features to prefer bundles
 				if (requestedType != null) {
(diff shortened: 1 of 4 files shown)


chrisrueger:
Found a problem...

chrisrueger:
New approach. Now I filter only in `P2Repo.versions(bsn)`

biz.aQute.repository/src/aQute/bnd/repository/p2/provider/P2Indexer.java
```diff
@@ -179,7 +183,28 @@ List<String> list(String pattern) throws Exception {
...
+
+			boolean isFeature = ResourceUtils.hasType(resource, ECLIPSE_FEATURE_CAPABILITY);
+			if (isFeature) {
+				continue;
+			}
+
+			versions.add(entry.getKey());
+		}
+
+		return versions;
```
chrisrueger, line 210: @peterkir this is the main change of this PR.

peterkir: Let's discuss this topic in dev meeting on monday.

peterkir (approved):
Works like a charm

PR #7297, opened 2026-06-27
Increase gradle 9 compatibility

peterkir:
updates the bnd Gradle plugins so they work cleanly on Gradle 9. It replaces deprecated/internal Gradle API usage, switches build-scan integration to the Develocity plugin, adjusts tests and compatibility checks, and updates the README to document Gradle 9 support and the configuration-cache limitation.


gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BeanProperties.java
```diff
@@ -87,10 +86,12 @@ private static Object unwrap(Object value) {
 
 	private static Object getField(Object target, String fieldName) {
 		try {
-			if (target instanceof DynamicObjectAware dynamicObjectAware) {
-				DynamicObject dynamicObject = dynamicObjectAware.getAsDynamicObject();
-				DynamicInvokeResult result = dynamicObject.tryGetProperty(fieldName);
-				return result.isFound() ? result.getValue() : null;
+			if (target instanceof GroovyObject groovyObject) {
```
bjhargrave: I don't think this is the same result. It does not consider ext properties I think. 

What is the issue here? I don't see any issue in the build scan.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java
```diff
@@ -477,8 +477,10 @@ public void execute(Task tt) {
 			TaskProvider<AbstractArchiveTask> jar = tasks.named(JavaPlugin.JAR_TASK_NAME, AbstractArchiveTask.class,
 				t -> {
 					t.setDescription("Jar this project's bundles.");
-					t.getActions()
-						.clear(); /* Replace the standard task actions */
+					/* Bnd Workspace model objects are not serializable; configuration cache is not supported. */
+					t.notCompatibleWithConfigurationCache(
+						"Bnd workspace plugin does not support configuration cache: bnd Project objects are not serializable");
+					/* Gradle 9: getActions().clear() is not supported. Configuration via inputs/outputs is sufficient. */
```
bjhargrave: This depends upon clearing the previous actions. It is not enough to just remove the clear call. Some other solution is needed. Actions have nothing to do with inputs/outputs.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndUtils.java
```diff
@@ -43,8 +43,21 @@ private BndUtils() {}
 	 * @return Whether the current Gradle is at least the specified version.
 	 */
 	public static boolean isGradleCompatible(String requestedVersion) {
-		return GradleVersion.current()
-			.compareTo(GradleVersion.version(requestedVersion)) >= 0;
+		return MavenVersion.parseMavenString(currentGradleVersion())
```
bjhargrave: Why is this change necessary?

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndWorkspacePlugin.java
```diff
@@ -293,13 +292,19 @@ public static Workspace getBndWorkspace(Project workspace) throws Exception {
...
 	private static void bndWorkspaceConfigure(Workspace workspace, Gradle gradle) {
-		ExtraPropertiesExtension ext = new DslObject(gradle).getExtensions()
-			.getExtraProperties();
+		ExtraPropertiesExtension ext = extraProperties(gradle);
 		if (ext.has("bndWorkspaceConfigure")) {
 			Object bndWorkspaceConfigure = ext.get("bndWorkspaceConfigure");
-			if (bndWorkspaceConfigure instanceof Closure<?> closure) {
-				closure.call(workspace);
-			} else if (bndWorkspaceConfigure instanceof Action) {
+			/* Check for Closure first (for Groovy DSL compatibility when Groovy is available) */
```
bjhargrave: When is groovy not available in a gradle build execution?

gradle-plugins/biz.aQute.bnd.gradle/src/test/groovy/aQute/bnd/gradle/TestBndPlugin.groovy
```diff
@@ -8,8 +8,7 @@ import java.util.jar.Attributes
 import java.util.jar.JarFile
 import java.util.regex.Pattern
 
-import static org.gradle.testkit.runner.TaskOutcome.FAILED
-import static org.gradle.testkit.runner.TaskOutcome.SUCCESS
+import static org.gradle.testkit.runner.TaskOutcome.*
```
bjhargrave: Lets not use wildcard import.

gradle-plugins/biz.aQute.bnd.gradle/src/test/groovy/aQute/bnd/gradle/TestBndPlugin.groovy
```diff
@@ -42,12 +41,12 @@ class TestBndPlugin extends Specification {
 				.build()
 
 		then:
-		result.task(":test.simple:jar").outcome == SUCCESS
-		result.task(":test.simple:test").outcome == SUCCESS
-		result.task(":test.simple:testOSGi").outcome == SUCCESS
-		result.task(":test.simple:check").outcome == SUCCESS
-		result.task(":test.simple:build").outcome == SUCCESS
-		result.task(":test.simple:release").outcome == SUCCESS
+		result.task(":test.simple:jar").outcome in [SUCCESS, UP_TO_DATE]
```
bjhargrave: This seems wrong. Changing tests to re-implement methods smells. The outputs of the test run are wiped before running the tests.

gradle-plugins/biz.aQute.bnd.gradle/src/test/groovy/aQute/bnd/gradle/TestHelper.groovy
```diff
@@ -28,6 +28,26 @@ class TestHelper {
 		return runner(version)
 	}
 
+	public static GradleRunner getCompatibilityGradleRunner(String version) {
```
bjhargrave: Why does this exist? I don't seen it being used in the existing test cases. You wrote a new test case to test this test helper method? I don't understand.

gradle-plugins/README.md
```diff
@@ -6,6 +6,7 @@ A Bnd Workspace build uses the information specified in the Bnd Workspace's `cnf
 
 The [`biz.aQute.bnd.gradle`][2] jar contains the Bnd Gradle Plugins.
 These plugins require Java 17 and at least Gradle 8.0.
+Gradle 9.x is supported and validated (tested against Gradle 9.2.0).
```
bjhargrave: This will quickly go stale. I don't think it is helpful or necessary.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java
```diff
@@ -477,8 +477,10 @@ public void execute(Task tt) {
 			TaskProvider<AbstractArchiveTask> jar = tasks.named(JavaPlugin.JAR_TASK_NAME, AbstractArchiveTask.class,
 				t -> {
 					t.setDescription("Jar this project's bundles.");
-					t.getActions()
-						.clear(); /* Replace the standard task actions */
+					/* Bnd Workspace model objects are not serializable; configuration cache is not supported. */
+					t.notCompatibleWithConfigurationCache(
+						"Bnd workspace plugin does not support configuration cache: bnd Project objects are not serializable");
```
bjhargrave: This should probably be done in the plugin's apply method. Fail fast.

gradle-plugins/README.md
```diff
@@ -764,8 +766,91 @@ When you use the Gradle plugin for another JVM language like Groovy, Scala, or K
 
 ---
 
+# Running builds with local Bnd Gradle Plugins
```
bjhargrave: This is a developer note not useful to the general user. Perhaps it should be in a developer readme?

gradle-plugins/README.md
```diff
@@ -764,8 +766,91 @@ When you use the Gradle plugin for another JVM language like Groovy, Scala, or K
...
+    id "biz.aQute.bnd.builder" version "7.4.0-SNAPSHOT"
+  }
+}
+```
+
+If you update plugin code, re-run `publishToMavenLocal` before rebuilding the consumer.
+
 # Using the latest development SNAPSHOT build of the Bnd Gradle Plugins
 
+
```
bjhargrave: Excess blank line

gradle-plugins/settings.gradle.kts
```diff
@@ -1,21 +1,22 @@
...
 
-if (System.getenv("CI").toBoolean()) {
-	gradleEnterprise {
-		buildScan {
-			publishAlways()
-			termsOfServiceUrl = "https://gradle.com/terms-of-service"
-			termsOfServiceAgree = "yes"
+develocity {
+	buildScan {
+		val isCI = System.getenv("CI").toBoolean()
```
bjhargrave: Why pull this test into the configuration?

build.gradle
```diff
@@ -49,7 +49,7 @@ subprojects {
 		tasks.named("test") {
 			useJUnitPlatform()
 			if (isCI) {
-				retry {
+				develocity.testRetry {
```
bjhargrave: What does this do? It makes us dependent on non-open source function?

bjhargrave (changes requested):
I am not clear what all of this is necessary. I can see making a config cache warning in the workspace plugin's apply method to fail fast for users. But the rest does not seem necessary or useful.

gradle-plugins/biz.aQute.bnd.gradle/src/test/groovy/aQute/bnd/gradle/TestBndPlugin.groovy
peterkir: Reverted to explicit imports.

peterkir:
> I am not clear what all of this is necessary. I can see making a config cache warning in the workspace plugin's apply method to fail fast for users. But the rest does not seem necessary or useful.

@bjhargrave Thx for the review. Blunt as always, but I asked for ;-) You are right. 
I reduced the PR to the minimal set of changes. All unnecessary
API replacements are reverted, and the config cache concern is now handled by a
fail-fast check in the workspace plugin's apply path as you suggested. What
remains is the fail-fast (+ test) and the removal of internal API usages
(DefaultManifest, VersionNumber), and the com.gradle.plugin-publish update.

Created a new clean commit, without the noise.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
```diff
@@ -581,12 +574,14 @@ public void execute(Task t) {
 			}
 		}
 
-		private org.gradle.api.java.archives.Manifest mergeManifest(Manifest builtManifest) {
-			org.gradle.api.java.archives.Manifest mergeManifest = new DefaultManifest(null);
-			mergeManifest.attributes(new AttributesMap(builtManifest.getMainAttributes()));
+		private void updateTaskManifest(org.gradle.api.java.archives.Manifest taskManifest, Manifest builtManifest) {
+			taskManifest.getAttributes()
```
bjhargrave: I am afraid of this mutation of the existing manifest. Perhaps you can  create a new Manifest by using `taskManifest.getEffectiveManifest()` which you can then mutate. The purpose of mergeManifest was to create a new Manifest object rather the mutate the existing Manifest.

gradle-plugins/settings.gradle.kts
```diff
@@ -1,22 +1,7 @@
 pluginManagement {
 	plugins {
-		id("com.gradle.plugin-publish") version("1.2.0")
+		id("com.gradle.plugin-publish") version("2.1.1")
 		id("dev.hargrave.addmavendescriptor") version("1.1.0")
-		id("com.gradle.enterprise") version("3.13.4")
```
bjhargrave, line 5: Why is this removed? It enables build scan.

build.gradle
```diff
@@ -48,12 +46,6 @@ subprojects {
 		}
 		tasks.named("test") {
 			useJUnitPlatform()
-			if (isCI) {
```
bjhargrave, line 51: Why is this removed? It attempt to work around flakiness in testing.

settings.gradle
```diff
@@ -4,12 +4,6 @@
 
 import aQute.bnd.osgi.Constants
 
-pluginManagement {
```
bjhargrave, line 7: Why removed?

gradle-plugins/biz.aQute.bnd.gradle/src/test/groovy/aQute/bnd/gradle/TestBndPlugin.groovy
```diff
@@ -514,4 +514,32 @@ class TestBndPlugin extends Specification {
...
+
+	def "Bnd Workspace Plugin configuration cache fail fast Test"() {
+		given:
+		String testProject = "workspaceplugin1"
+		File testProjectDir = new File(testResources, testProject)
+		assert testProjectDir.isDirectory()
+
+		when: "the configuration cache is requested"
+		// The BuildFeatures service used for the fail fast check requires Gradle 8.5
+		def result = TestHelper.getGradleRunner("8.5")
```
bjhargrave, line 531: Asking for version 8.5 seems wrong since 8.5 will always return false. The floor version for this test should be 9.0, no?

gradle-plugins/biz.aQute.bnd.gradle/src/test/groovy/aQute/bnd/gradle/TestBndPlugin.groovy
```diff
@@ -514,4 +514,32 @@ class TestBndPlugin extends Specification {
...
+				.withProjectDir(testProjectDir)
+				.withArguments("-Pbnd_plugin=${pluginClasspath}", "--configuration-cache", "--stacktrace", ":tasks")
+				.forwardOutput()
+				.buildAndFail()
+
+		then: "the build fails fast with a clear message"
+		result.output.contains("The Bnd Workspace plugin does not support the Gradle configuration cache")
+
+		when: "the configuration cache is not requested"
+		result = TestHelper.getGradleRunner("8.5")
```
bjhargrave, line 541: Same here

bjhargrave (changes requested):

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
peterkir: Agreed the task manifest should not be mutated — I restored the original semantics with a new Manifest object as merge source. However, `getEffectiveManifest()` can't provide that object: when a manifest has no merge specs, `DefaultManifest.getEffectiveManifestInternal()` returns `this` (both Gradle 8.0 and 9.0 sources), so merging it back creates a cycle and a `StackOverflowError`. Since Gradle has no public factory for a detached Manifest, I added a small read-only adapter implementing the public `Manifest` interface over the bnd-generated manifest. Same behavior as before, and no internal Gradle API usage.

gradle-plugins/settings.gradle.kts
peterkir: I put it back and updated it to the latest version of the deprecated/legacy plugin. Internet is full of suggestions to move to Develocity - but I agree on your point of staying with ths OSS version.

build.gradle
peterkir: My bad. Wanted to speed up tests.

settings.gradle
peterkir: re-introduced with updated version ... same as above

gradle-plugins/biz.aQute.bnd.gradle/src/test/groovy/aQute/bnd/gradle/TestBndPlugin.groovy
peterkir: `getGradleRunner("8.5")` is a floor, not a pin, JDK17→8.5, JDK21→8.10.1, JDK25→9.1.0
8.5 is deliberate: it's the version that introduced `BuildFeatures`
I've extended the code comment to make this explicitly visible.

gradle-plugins/biz.aQute.bnd.gradle/src/test/groovy/aQute/bnd/gradle/TestBndPlugin.groovy
peterkir: see above

settings.gradle
```diff
@@ -50,20 +50,6 @@ buildscript {
...
-if (Boolean.parseBoolean(System.getenv("CI"))) {
-	gradleEnterprise {
-		buildScan {
-			publishAlways()
-			termsOfServiceUrl = "https://gradle.com/terms-of-service"
-			termsOfServiceAgree = "yes"
-		}
-	}
-}
-
```
bjhargrave, line 66: This is still deleted. Without this, we do not get a scan for each CI biuld.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
```diff
@@ -636,4 +644,69 @@ public int size() {
...
 		}
 	}
+
+	/**
+	 * A read-only {@link org.gradle.api.java.archives.Manifest} adapter over
+	 * the manifest generated by the bnd builder. It is used as the merge
+	 * source for the task's manifest, so only the query methods used by the
+	 * Gradle manifest merge machinery are supported.
+	 */
+	static final class BuiltManifest implements org.gradle.api.java.archives.Manifest {
```
bjhargrave, line 654: This choice can bite us one day. The gradle Manifest interface is really a ProviderType. Consumers like us are not meant to implement it. So we can be easily broken when gradle adds a new (non-default) method which could happen on a non-major version release.

So the trade-off seems to be implement their ProviderType interface and risk class load errors one day, or use the internal implementation class and risk that it is removed one day (probably on a major version release boundary).

That is why I originally implemented this using the internal implementation class since it seemed the least risky.

Another option would be to implement the interface using a dynamic proxy. See https://github.com/eclipse-osgi-technology/osgi-test/blob/main/org.osgi.test.common/src/main/java/org/osgi/test/common/context/CloseableBundleContext.java for an example of this. It is more robust if the ProviderType is extended in the future.

bjhargrave (changes requested):

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
peterkir: ok. I will have a look at the dyn proxy option.

settings.gradle
peterkir: I first commented, but haven't pushed everything. Please wait for a new "Review request".

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
peterkir: @bjhargrave 
I implemented the dynamic proxy. Can you please review.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
```diff
@@ -600,7 +610,7 @@ private boolean isEmpty(String header) {
 		}
 	}
 
-	static final class AttributesMap extends AbstractMap<String, Object> {
+	static final class AttributesMap extends AbstractMap<String, Object> implements org.gradle.api.java.archives.Attributes {
```
bjhargrave: Implementing this Gradle interface is a risk. While it is API, I would assume it is ProviderType which means Gradle can change at will which could break this code. You did not want to use DefaultManifest which is internal code but this just changes the risk rather than eliminating it. If you need this type to implement Attributes, you will probably also need to proxy this like you do for Manifest.

build.gradle
```diff
@@ -27,11 +32,21 @@ if (JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_1_9)) {
 	]
 }
 
-boolean isCI = Boolean.parseBoolean(System.getenv("CI"))
+// biz.aQute.launchpad is only registered for full builds, not scoped task
+// invocations like ":gradle-plugins:build" - use findProject to avoid failing
+// those with an UnknownProjectException.
+findProject(":biz.aQute.launchpad")?.tasks?.named("test") {
```
bjhargrave: This is a bad idea here. It break project isolation which is a new thing Gradle is pushing and will probably not work in a future gradle version. Any configuration of the project should be in the project's build.gradle.

settings.gradle
bjhargrave: Still not replaced.

build.gradle
```diff
@@ -27,11 +32,21 @@ if (JavaVersion.current().isCompatibleWith(JavaVersion.VERSION_1_9)) {
...
+		reports {
+			html.required = false
+		}
+	}
+}
 
 /* Configure the subprojects */
 subprojects {
 	if (pluginManager.hasPlugin("biz.aQute.bnd")) {
+		pluginManager.apply("org.gradle.test-retry")
```
bjhargrave: Should there be configuration for this plugin? number of retries? etc?

bjhargrave (changes requested):

build.gradle
peterkir: Restored the master configuration - with the re-try configuration inside

build.gradle
peterkir: Moved launchpad-specific test configuration to `build.gradle`, so the root project no longer accesses another project’s task state.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
peterkir: Switching it also to a dynamic proxy.

peterkir:
@bjhargrave Thanks for the review. I addressed the four requested changes:

- Attributes now uses a dynamic proxy, matching the Manifest approach
- The launchpad-specific test configuration moved to build.gradle, so the root project no longer accesses another project’s task state.
- Root CI build-scan application and publishAlways() configuration were restored.
- CI retry configuration was restored with maxRetries = 2 and maxFailures = 20

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
```diff
@@ -636,4 +647,101 @@ public int size() {
 			};
 		}
 	}
+
+	static final class BuiltAttributes implements InvocationHandler {
+		private final Map<String, Object> attributes;
+
+		private BuiltAttributes(java.util.jar.Attributes attributes) {
+			this.attributes = Collections.unmodifiableMap(new AttributesMap(attributes));
```
bjhargrave:
```suggestion
			this.attributes = new AttributesMap(attributes);
```

AttributesMap is already unmodifiable.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
```diff
@@ -636,4 +647,101 @@ public int size() {
 			};
 		}
 	}
+
+	static final class BuiltAttributes implements InvocationHandler {
```
bjhargrave, line 661: It would be good to have javadoc here.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
```diff
@@ -636,4 +647,101 @@ public int size() {
...
+					return sections;
+				case "getEffectiveManifest" :
+					return proxy;
+				case "hashCode" :
+					return System.identityHashCode(proxy);
+				case "equals" :
+					return proxy == args[0];
+				case "toString" :
+					return "BuiltManifest[attributes=" + attributes + ", sections=" + sections + "]";
+				default :
```
bjhargrave, line 775: I think the default here should be more robust like in BuiltAttributes to defend against future changes to the gradle Manifest interface.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
```diff
@@ -636,4 +647,101 @@ public int size() {
...
+	 * consumers, since a future Gradle version can add a new abstract method
+	 * to it in a minor release. A class compiled against an older version of
+	 * the interface would then fail with an {@link AbstractMethodError} once
+	 * Gradle invoked the missing method. A dynamic {@link Proxy} avoids that
+	 * risk: the proxy class is generated at run time against whatever version
+	 * of the interface Gradle loads, so it always has every method: unknown or
+	 * future methods land in the {@code default} case below instead of
+	 * crashing with a linkage error.
+	 */
+	static final class BuiltManifest implements InvocationHandler {
```
bjhargrave, line 735: I don't think these proxy implementations are as robust as https://github.com/eclipse-osgi-technology/osgi-test/blob/main/org.osgi.test.common/src/main/java/org/osgi/test/common/context/CloseableBundleContext.java.

bjhargrave (changes requested):
The fixes are an improvement. A couple of more comments.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
peterkir: revised

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BundleTaskExtension.java
peterkir: revised

peterkir:
@bjhargrave 
Thanks once more - taking the time. 
I’ll added Javadoc for BuiltAttributes and refactor both handlers to exact Method-based dispatch, including support for future interface default methods and explicit rejection of unsupported mutators. This follows the robustness pattern used by CloseableBundleContext without depending on Gradle internals.
Looking forward to your review :-)

bjhargrave (approved):
Looks great!

peterkir:
@bjhargrave Thanks for your patience and feedback. I learned a lot and appreciate it!

PR #7304, opened 2026-07-06
Fix NPE in RepoCollector initialization during Project construction

### LINKED ISSUE #7303: Bug: ${repo} in -include causes NPE in Gradle builds with bnd 7.3.0
## Summary

Using `${repo;...}` inside a `-include` instruction in `bnd.bnd` causes a `NullPointerException` in Gradle builds with bnd 7.3.0. This worked correctly in 7.0.0.

The error message is:
```
error: Cannot invoke "aQute.bnd.build.RepoCollector.repoContainers(java.lang.String[])"
       because "this.repoCollector" is null, for cmd: repo, arguments; [repo, org.ow2.asm:asm, 9.10]
```

## Environment

- **bnd version:** 7.3.0 (regression from 7.0.0)
- **Build tool:** Gradle
- **Eclipse bndtools:** Not affected (see explanation below)

## Minimal Reproducer

5 files:

### Directory structure
```
workspace/
├── cnf/
│   ├── bnd.bnd
│   ├── build.bnd
│   └── deps.maven
├── project/
│   └── bnd.bnd
└── settings.gradle
```

### `settings.gradle`
```groovy
buildscript {
    repositories { maven { url = uri('https://repo1.maven.org/maven2') } }
    dependencies {
        classpath 'biz.aQute.bnd:biz.aQute.bnd.gradle:7.3.0'
    }
}

import aQute.bnd.build.Workspace
import aQute.bnd.osgi.Constants
import aQute.bnd.gradle.BndPlugin

Workspace.setDriver(Constants.BNDDRIVER_GRADLE)
Workspace.addGestalt(Constants.GESTALT_BATCH, null)
Workspace workspace = new Workspace(rootDir, 'cnf').setOffline(startParameter.offline)

workspace.getAllProjects().each { include it.name }

gradle.projectsLoaded { gradle ->
    gradle.rootProject.ext.bndPlugin = BndPlugin.class
    gradle.rootProject.ext.bndWorkspace = workspace
    gradle.rootProject.subprojects.each { subproject ->
        if (subproject.name != 'cnf') {
            subproject.apply plugin: BndPlugin
        }
    }
}
```

### `cnf/bnd.bnd`
```
-nobundles=true
```

### `cnf/build.bnd`
```
-plugin.1.Central: \
    name = Central; \
    releaseUrl = https://repo.maven.apache.org/maven2/; \
    index = ${build}/deps.maven
```

### `cnf/deps.maven`
```
org.ow2.asm:asm:9.10
```

### `project/bnd.bnd`
```
-include= jar:${fileuri;${repo;org.ow2.asm:asm;9.10}}!/META-INF/MANIFEST.MF

-buildpath: org.ow2.asm:asm;version=9.10
```

### Run
```
./gradlew project:jar
```

## Expected

`${repo;org.ow2.asm:asm;9.10}` resolves to the jar path and the build succeeds.

## Actual

```
project/bnd.bnd:0: error: Cannot invoke "aQute.bnd.build.RepoCollector.repoContainers(java.lang.String[])"
    because "this.repoCollector" is null, for cmd: repo, arguments; [repo, org.ow2.asm:asm, 9.10]
project/bnd.bnd:0: error: Included file .../project/$%7Brepo does not exist
```

## Root Cause

The bug is in the `Project` constructor in `biz.aQute.bndlib`. `RepoCollector` was introduced in 7.3.0 and added as a `final` field on `Project`, but it is assigned _after_ `setProperties(buildFile)` is called:

```java
// Project.java — current (broken in 7.3.0):
public Project(Workspace workspace, File unused, File buildFile) {
    super(workspace);
    this.workspace = workspace;
    setFileMustExist(false);
    if (buildFile != null)
        setProperties(buildFile);    // ← evaluates -include, calls _repo() on this
                                     //   but repoCollector is still null here!
    readBuildProperties();
    repoCollector = new RepoCollector(this);  // ← too late
    addClose(repoCollector);
}
```

`setProperties(buildFile)` eagerly processes `-include` instructions via `doIncludes()`, which calls `getReplacer().process()` on the include value. When the value contains `${repo;...}`, bnd dispatches `_repo()` on the `Project` instance. In 7.3.0, `Project._repo()` delegates to `this.repoCollector.repoContainers(args)` — but `repoCollector` is `null` at this point because it hasn't been assigned yet.

The NPE is caught by `Macro.doCommand()`'s `catch (Exception e)` block and reported as a bnd error, causing the build to fail.

## Why Eclipse bndtools is not affected

Eclipse calls `new Workspace(workspaceDirectory)` (single-arg constructor) which calls `open()`. `open()` eventually triggers `Workspace.refresh()` → `Processor.refresh()` → `forceRefresh()` → `setProperties(propertiesFile, base)` _after_ all project constructors have completed and `repoCollector` is valid. This second call to `setProperties()` heals the broken state from the constructor.

Gradle uses `new Workspace(rootDir, cnfDir)` (two-arg constructor) which does **not** call `open()`, so this healing re-load never happens.

## Fix

Move the `repoCollector` initialization before `setProperties()` in the `Project` constructor. `RepoCollector(this)` is safe to construct at this point — its constructor only stores a reference to the `Project` and reads no properties:

```java
// Project.java — fixed:
public Project(Workspace workspace, File unused, File buildFile) {
    super(workspace);
    this.workspace = workspace;
    setFileMustExist(false);
    repoCollector = new RepoCollector(this);  // ← moved before setProperties()
    addClose(repoCollector);
    if (buildFile != null)
        setProperties(buildFile);    // ← now safe: repoCollector is valid
    readBuildProperties();
}
```

chrisrueger:
Closes #7303 

Move RepoCollector initialization before setProperties() in Project constructor to ensure it's available when properties are being processed. This fixes issue #7303 where an NPE could occur when using the 2-arg Workspace constructor (which doesn't call open()). Adds test case with repocollector workspace to verify the fix.

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/ProjectTest.java
+++ b/biz.aQute.bndlib.tests/test/test/ProjectTest.java
@@ -870,6 +870,25 @@ public class ProjectTest {
 		}
 	}
 
+	/**
+	 * Test for https://github.com/bndtools/bnd/issues/7303 that no NPE is
+	 * thrown.
+	 */
+	@Test
+	public void testRepoCollectorNoNPEWorkspaceOpenNotCalled() throws Exception {
+
+		try (Workspace ws = getWorkspace(IO.getFile("testresources/ws-repocollector"), "cnf").setOffline(false);
+			Project project = ws.getProject("p1");
+		) {
+			List<String> errors = project.getErrors();
+			List<String> warnings = project.getWarnings();
+
+			assertThat(errors).hasSize(0);
+			assertThat(warnings).hasSize(0);
+		}
+
+	}
+
 	@Test
 	public void testRepoCollectorNonJar() throws Exception {
 		try (Workspace ws = getWorkspace(IO.getFile("testresources/ws-repononjar"));
@@ -1308,6 +1327,16 @@ public class ProjectTest {
 	private Workspace getWorkspace(File file) throws Exception {
 		IO.copy(file, tmp);
 		return new Workspace(tmp);
+	}
+
+	/**
+	 * This method calles the 2-arg constructor of Workspace, which was
+	 * important for a bugfix, because this constructor does not call
+	 * Workspace.open()
+	 */
+	private Workspace getWorkspace(File file, String bndDir) throws Exception {
+		IO.copy(file, tmp);
+		return new Workspace(tmp, bndDir);
 	}
 
 	private Workspace getWorkspace(String dir) throws Exception {
--- a/biz.aQute.bndlib.tests/testresources/ws-repocollector/cnf/bnd.bnd
+++ b/biz.aQute.bndlib.tests/testresources/ws-repocollector/cnf/bnd.bnd
@@ -1,0 +1,1 @@
+-nobundles=true
--- a/biz.aQute.bndlib.tests/testresources/ws-repocollector/cnf/build.bnd
+++ b/biz.aQute.bndlib.tests/testresources/ws-repocollector/cnf/build.bnd
@@ -1,0 +1,5 @@
+-plugin.1.Central: \
+  aQute.bnd.repository.maven.provider.MavenBndRepository; \
+    name = Central; \
+    releaseUrl = https://repo.maven.apache.org/maven2/; \
+    index = ./cnf/deps.maven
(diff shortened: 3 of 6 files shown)


jhanders34 (approved):
The code changes looks good and I ran the test with and without the change and it passes with the updated code and fails without the updated code.

chrisrueger:
> The code changes looks good and I ran the test with and without the change and it passes with the updated code and fails without the updated code.

Thanks a lot. I wait for the build to succeed and then I merge. 
The change then will be available in the next [SNAPSHOT build](https://github.com/bndtools/bnd#installation-development-snapshot-builds--master-branch) about 30-45 minutes after merging. 

The next bnd release is planned for end of july due to vacation.

PR #7307, opened 2026-07-07
Updates to avoid memory visibility problem

### LINKED ISSUE #7305: Bug: MavenBndRepository race condition in parallel builds
## Summary

`MavenBndRepository` has a Java Memory Model visibility bug that causes intermittent `NullPointerException` during Gradle parallel builds. The error manifests as:

```
error  : null, for cmd: repo, arguments; [repo, group:artifact, version]
```

---

## Environment

- **Bnd version:** 7.0.0
- **Build tool:** Gradle with `--parallel`
- **Java:** 17+
- **OS:** Linux (also reproducible on other platforms with parallel builds)

---

## Steps to Reproduce

1. Configure a Bnd workspace with `MavenBndRepository` plugins
2. Run a parallel Gradle build: `./gradlew build --parallel`
3. Observe intermittent errors of the form:
   ```
   error  : null, for cmd: repo, arguments; [repo, group:artifact, version]
   ```
4. The error does not occur every build — only when thread scheduling causes the race window to be hit

---

### Non-volatile fields accessed across threads

In [`MavenBndRepository.java`](https://github.com/bndtools/bnd/blob/master/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java), lines 125–126:

```java
private boolean  inited;   // line 125 — NOT volatile
IndexFile        index;    // line 126 — NOT volatile
```

The `init()` method (line 641) is `synchronized`, which correctly prevents two threads from initializing simultaneously:

```java
synchronized boolean init() {
    if (inited)
        return true;

    inited = true;           // written inside synchronized block

    // ... long initialization ...

    this.index = new IndexFile(...);   // line 749 — written inside synchronized
    this.index.open();
    // ...
    return true;
}
```

However, `versions()` (line 629) and `list()` (line 620) call `init()` **without holding a lock**, then immediately access `index` also **without a lock**:

```java
public SortedSet<Version> versions(String bsn) throws Exception {
    if (!init())            // calls synchronized init() — acquires/releases lock
        return new TreeSet<>();

    return index.versions(bsn);   // accesses index WITHOUT lock — may see null!
}

public List<String> list(String pattern) throws Exception {
    if (!init())
        return Collections.emptyList();

    return index.getBridge().list(pattern);   // same problem
}
```

### The Race Window

The Java Memory Model guarantees that writes made inside a `synchronized` block are visible to **other threads that subsequently enter the same synchronized block**. However, once `init()` returns and the lock is released, a different thread calling `versions()` will:

1. Call `init()` — enters and exits the `synchronized` block, sees `inited == true`, returns `true`
2. Exit `init()` — the lock is released
3. Read `index` — **outside any synchronized block**, with no `volatile` guarantee

Because `index` is neither `volatile` nor accessed under a lock at the read site, the JVM/CPU is free to serve the read from a stale cached value. If the write to `index` (line 749) has not been flushed to main memory yet — or if the reading thread's cache has not been invalidated — `index` reads as `null`.

This leads to a `NullPointerException` in `index.versions(bsn)` or `index.getBridge().list(pattern)`. Because `NullPointerException.getMessage()` returns `null`, Bnd's error reporter prints it as:

```
error  : null, for cmd: repo, arguments; [repo, group:artifact, version]
```

---

## Suggested Fix

The simplest fix is to declare both fields `volatile`:

```java
private volatile boolean  inited;   // ensures visibility across threads
volatile IndexFile        index;    // ensures visibility across threads
```

Because `volatile` provides a happens-before guarantee, any thread that reads `inited == true` is guaranteed to also see the fully initialized value of `index`, since the write to `inited` happened after the write to `index` within the same thread.

---

## Impact

- **Severity:** Medium — intermittent build failures that are hard to reproduce consistently
- **Frequency:** Higher with more parallel workers and larger workspaces
- **Workaround:** Disable parallel builds with `org.gradle.parallel=false`

jhanders34:
Fixes #7305

### DIFF
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
@@ -122,8 +122,8 @@ public class MavenBndRepository extends BaseRepository implements RepositoryPlug
 	private File								localRepo;
 	Reporter									reporter;
 	IMavenRepo									storage;
-	private boolean								inited;
-	IndexFile									index;
+	private volatile boolean					inited;
+	volatile IndexFile							index;
 	private ScheduledFuture<?>					indexPoller;
 	private RepoActions							actions							= new RepoActions(this);
 	private String								name;
--- a/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
+++ b/biz.aQute.repository/src/aQute/bnd/repository/maven/provider/MavenBndRepository.java
@@ -746,8 +746,10 @@ public class MavenBndRepository extends BaseRepository implements RepositoryPlug
 			}
 			Set<String> multi = Strings.splitAsStream(configuration.multi())
 				.collect(Sets.toSet());
-			this.index = new IndexFile(domain, reporter, indexFile, source, storage, client.promiseFactory(), multi);
-			this.index.open();
+			// Set the index only after it completes opening
+			IndexFile newIndex = new IndexFile(domain, reporter, indexFile, source, storage, client.promiseFactory(), multi);
+			newIndex.open();
+			this.index = newIndex;
 
 			try (Formatter f = new Formatter()) {
 				validateUris(release, f);


chrisrueger:
@jhanders34 build passed. Should I merge or do you want to add another commit with your idea of https://github.com/bndtools/bnd/issues/7305#issuecomment-4905549389

jhanders34:
> @jhanders34 build passed. Should I merge or do you want to add another commit with your idea of [#7305 (comment)](https://github.com/bndtools/bnd/issues/7305#issuecomment-4905549389)

@chrisrueger I have added a second commit with the second fix

chrisrueger (approved):
Let's give it a try. And thanks :)

Issue #6896, opened 2025-10-28
Bnd sometimes leaves classes out of bundles


ejjcase:
This is a very odd one where I can see the symptoms but can't reproduce it consistently (thank goodness). I thought about reporting it earlier but decided to try investigating it myself first. However, that hasn't led anywhere.

I have a Bnd workspace containing many bundle projects. Their sources are laid out in the traditional Eclipse way with `java` classes and other resources mixed together under `src/`. I'm not sure if that's important.

They are all set up with these properties:
```properties
tool-modifier:          ${if;${is;${driver};gradle};/gradle}
target-dir:             target
src:                    src
testsrc:                unitTests/src
bin:                    ${target-dir}${tool-modifier}/classes
testbin:                ${target-dir}${tool-modifier}/testClasses
```

I sometimes use Eclipse, and sometimes Gradle. I've seen this bug in both.

Sometimes when I'm rebuilding the workspace, the build will stop with compile errors. When I investigate, I see that one of the earlier bundles has resources in it but no classes. If it has an activator, there's also a warning that the activator was not found in the bundle.

`target/classes/` or `target/gradle/classes/` (depending on the driver) only contains resources, but I can see that compilation took place, because there are usually some compiler warnings.

If I delete `target/classes/` (in Eclipse) or simply try the build again (in Gradle) it will usually build the bundle properly, but then I'll get the same problem with a different bundle.

This usually continues until I perform the right combination of cleaning the workspace, stopping the Gradle daemon, restarting the IDE … I'm not sure. Then I might not see the problem again for a few days, but it has always come back.

CI builds use Gradle but I don't think I've seen this bug in one of those; I thought I had, then I found [another problem with similar symptoms](https://bnd.discourse.group/t/confusing-behaviour-as-bnd-runs-tests-against-empty-bundles/538) and worked around it. I suppose that since every CI build starts with a clean checkout, there isn't time for the problem to occur.

I can also consistently make the bug go away in Gradle by disabling all the `processResources` tasks. Then there are no resources in the bundles unless they are placed there by an `-includeresource`, but the classes are all present.

My hypothesis is that the classes are indeed being written into the correct folder, but then they are somehow overwritten when the resources are copied in. I understand that Gradle or Eclipse, not Bnd, is handling the resources in this case. But I haven't seen a bug like this when I wasn't using Bnd, so I think Bnd must have something to do with it.

[...]

Issue #7306, opened 2026-07-06
Bug:  Gradle Plugin: compileTestJava fails when using overlay file for a class found in a bundle on the buildpath


jhanders34:
## Summary

When moving to the bnd gradle plugin (biz.aQute.bnd:biz.aQute.bnd.gradle) 7.0.0 and higher, unit tests do not compile cleanly when the source file in a project overlays a file in a buildpath dependency.

For example, if the overlay file adds a new method that a unit test references, it fails to find it and gets a compile failure

---

## Environment

- **Bnd version:** 7.0.0+
- **Build tool:** Gradle
- **Java:** 17+

---

## Steps to Reproduce

1. Reference a group:artifact:version in your -buildpath
2. Overlay one of the classes from that dependency and add a new method
3. Include contents of the dependency in the bundle and add the overlay class on top of it using `-includeresource` or `Include-Resource`
4. Have a unit test that references the new method
5. Run `./gradlew build` or `./gradlew compileTestJava`
6. Observe java compile failures when building the unit test

---

## Root Cause

I don't know the cause for this particular problem.  It is as if the project's output directory or the built bundle for the project is not prepending to the build path.  Maybe it is appended or not added at all when doing unit test compilation.  Note that the Eclipse bndtools plugin does not have this problem.  It is appears to only be a problem with the gradle plugin.

---

### Workaround

To workaround the problem, I add the following to my gradle configuration.

```
compileTestJava {
  // With the 7.x versions of the bnd gradle plugin the test classpath for compiling
  // does not have the output directory of the source of the project so if there are
  // overlays that add new methods, the new methods are not found during test compile time.
  classpath = files(sourceSets.main.output.classesDirs) + classpath
}
```

chrisrueger:
@bjhargrave @peterkir any ideas?

bjhargrave:
> any ideas?

Not from this description. I would need a simple repo example to debug.

jhanders34:
Using similar set up to what I did in #7303, here are some steps to reproduce the problem:

### Directory structure
```
workspace/
├── cnf/
│   ├── bnd.bnd
│   ├── build.bnd
│   └── deps.maven
├── project/
│   ├── bnd.bnd
│   ├── build.gradle
│   ├── src/
│   │   └── org/objectweb/asm/
│   │       └── Attribute.java
│   └── test/
│       └── org/objectweb/asm/
│           └── AttributeTest.java
└── settings.gradle
```

### `settings.gradle`
[...]

PR #7323, opened 2026-07-20
Fix task order between compileJava and processResources.

ejjcase:
Fix for https://github.com/bndtools/bnd/issues/6896 in the Gradle case only.

Both these tasks have the same output directory. If `compileJava` runs first, which can sometimes happen when cached state is reused, then `processResources` [clears out the output directory when it runs](https://github.com/gradle/gradle/blob/cbe914f9a3dbe91a5d83fbe4d5493ed1751544e1/platforms/jvm/language-jvm/src/main/java/org/gradle/language/jvm/tasks/ProcessResources.java#L34), deleting all the classes. But if `processResources` runs first, `compileJava` happily writes classes into the same directory tree without deleting the resources.

I've done the same thing for the `test` source set.

I noticed that when the source sets have extensions for other languages, `BndPlugin` [adds the outputs of `processResources` as inputs](https://github.com/bndtools/bnd/blob/cdd3fe3b81817c9b393905e8ea27e8b9b9656ca9/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java#L339) to the extra `AbstractCompile` tasks. This was added recently in https://github.com/bndtools/bnd/pull/7313. I'm not sure whether this was done to enforce a task ordering dependency or if the inputs are actually needed for some reason. I know that they're not needed for Java compilation, so I opted for a simple ordering between the tasks.

I tried to write a test:

* Start with a Bnd workspace containing a bundle that has a Java class and a resource
* In the bundle's Gradle configuration, add a `doLast` block to the `jar` task to verify that the jar contained a class
* Repeatedly run the tasks `clean` and `jar` to see if the check ever failed

Unfortunately I couldn't get the build to ever fail. I think it's because the Gradle Test Kit doesn't use a Gradle daemon in the usual way, so there's no opportunity for the cache to get into a state where it starts changing the task order.

### DIFF
--- a/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java
+++ b/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java
@@ -262,9 +262,12 @@ public class BndPlugin implements Plugin<Project> {
 				sourceSet.setCompileClasspath(jarLibraryElements(project, sourceSet.getCompileClasspathConfigurationName()));
 				sourceSet.setRuntimeClasspath(objects.fileCollection()
 					.from(sourceSet.getOutput(), jarLibraryElements(project, sourceSet.getRuntimeClasspathConfigurationName())));
+				TaskProvider<Task> processResourcesTask = tasks.named(sourceSet.getProcessResourcesTaskName());
+				TaskProvider<AbstractCompile> compileTask = tasks.named(sourceSet.getCompileJavaTaskName(),
+					AbstractCompile.class, t -> t.mustRunAfter(processResourcesTask));
 				generateInputAction.ifPresent(action -> {
-					tasks.named(sourceSet.getCompileJavaTaskName()).configure(action);
-					tasks.named(sourceSet.getProcessResourcesTaskName()).configure(action);
+					compileTask.configure(action);
+					processResourcesTask.configure(action);
 				});
 			});
 			sourceSets.getByName(SourceSet.TEST_SOURCE_SET_NAME, sourceSet -> {
@@ -287,6 +290,8 @@ public class BndPlugin implements Plugin<Project> {
 					.from(mainSourceSet.getOutput(), jarLibraryElements(project, sourceSet.getCompileClasspathConfigurationName())));
 				sourceSet.setRuntimeClasspath(objects.fileCollection()
 					.from(sourceSet.getOutput(), mainSourceSet.getOutput(), jarLibraryElements(project, sourceSet.getRuntimeClasspathConfigurationName())));
+				tasks.named(sourceSet.getCompileJavaTaskName(),
+					AbstractCompile.class,t -> t.mustRunAfter(tasks.named(sourceSet.getProcessResourcesTaskName())));
 			});
 			/* Configure srcDirs for any additional languages */
 			project.afterEvaluate(p -> {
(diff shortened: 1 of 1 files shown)


chrisrueger:
Thanks for the PR. 

> I noticed that when the source sets have extensions for other languages, `BndPlugin` [adds the outputs of `processResources` as inputs](https://github.com/bndtools/bnd/blob/cdd3fe3b81817c9b393905e8ea27e8b9b9656ca9/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java#L339) to the extra `AbstractCompile` tasks. This was added recently in #7313. I'm not sure whether this was done to enforce a task ordering dependency or if the inputs are actually needed for some reason. 

I think the answer is in the Example of comment (see the "Expected" headline): https://github.com/bndtools/bnd/issues/7306#issuecomment-4907196892

> I know that they're not needed for Java compilation

I am not so sure, given the problem of #7306 (and if I understand your sentence correctly). 
If I understand #7306 correctly `compileJava` needs result of `processResource` (because `procesResource` adds `.class` files which should be visible for compilation.

Maybe @bjhargrave and @jhanders34 have an opinion.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java
```diff
@@ -266,6 +266,8 @@ public void execute(Task tt) {
 					tasks.named(sourceSet.getCompileJavaTaskName()).configure(action);
 					tasks.named(sourceSet.getProcessResourcesTaskName()).configure(action);
 				});
+				tasks.named(sourceSet.getCompileJavaTaskName(), compileJavaTask ->
+					compileJavaTask.mustRunAfter(tasks.named(sourceSet.getProcessResourcesTaskName())));
```
bjhargrave: This is wrong. `mustRunAfter` is a last resort as it hides dependencies. It is better to make the process resources task be an input to the compiler task to enforce the ordering in a transparent way.

bjhargrave (changes requested):

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java
ejjcase: But the resources are not really inputs to `compileJava`, in the sense that the compiler doesn't need to see them. Even in a case like https://github.com/bndtools/bnd/issues/7306 where some of the resources were class files, then if sources in the main source set needed to compile against them, you'd deal with that by putting them on the buildpath in the normal way.

The only reason there's a "dependency" between the two tasks is that they share an output directory, in violation of Gradle best practice. This actually causes Gradle cache misses for `compileJava` whenever a bundle has any resources (with my `mustRunAfter` relationship) or always (with a file input relationship).

For this reason, I think `mustRunAfter` is more honest. It's still an explicit relationship between the two tasks.

It would be better if we could allow `processResources` to write to a different directory instead.

ejjcase:
> Thanks for the PR.
> 
> > I noticed that when the source sets have extensions for other languages, `BndPlugin` [adds the outputs of `processResources` as inputs](https://github.com/bndtools/bnd/blob/cdd3fe3b81817c9b393905e8ea27e8b9b9656ca9/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java#L339) to the extra `AbstractCompile` tasks. This was added recently in #7313. I'm not sure whether this was done to enforce a task ordering dependency or if the inputs are actually needed for some reason.
> 
> I think the answer is in the Example of comment (see the "Expected" headline): [#7306 (comment)](https://github.com/bndtools/bnd/issues/7306#issuecomment-4907196892)
> 
> > I know that they're not needed for Java compilation
> 
> I am not so sure, given the problem of #7306 (and if I understand your sentence correctly). If I understand #7306 correctly `compileJava` needs result of `processResource` (because `procesResource` adds `.class` files which should be visible for compilation.

My understanding of the example you linked is that the bundle mostly contains classes included from an external jar, but wants to replace one class with a locally compiled version. The test needs the local version to appear before the included one on the classpath, but it wouldn't normally get that, because Gradle normally runs tests against the output of `classes` and `processResources` (i.e. two separate trees of files) instead of against the built jar. This is why prepending the jar to the test compile classpath is a valid workaround.

There aren't any compile tasks in the example where the compiler needs to see any resources belonging to the _same_ source set. Even if there were, you could just put them on the buildpath/testpath.

chrisrueger:
> the bundle mostly contains classes included from an external jar, but wants to replace one class with a locally compiled version. The test needs the local version to appear before the included one on the classpath

You are right. Forget my remark. I misinterpreted.

gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java
bjhargrave: Upon further reading, I agree that `mustRunAfter` is a better option here. It is better in that it does not require the process resource task to be run before the compile task, but if both tasks are to be run, they have a defined order.

We should also change the 2 occurrences for the other language extensions to also use mustRunAfter instead of inputs.

> It would be better if we could allow `processResources` to write to a different directory instead.

That would be a larger change to Bnd Workspace model builds :-)

bjhargrave: I will push up a fix

bjhargrave (approved):

PR #7324, opened 2026-07-20
Hold read lock to avoid refresh closing repository

jhanders34:
The problem in #7305 was still seen after the fix in PR #7307.  With IBM Bob's help, was able to track it down to needing to have a read lock when calling getBundle() where the repository could be closed as part of workspace refresh.  

By holding the workspace read lock, it is able to make sure that the repository is not closed in the middle of processing of a bundle because there is a discovery and use window where the repositories are discovered with getRepositories(), but used later where the repositories could be changed.  When Workspace.refresh() is called, it gets a write lock which is how this change fixes the problem because the refresh does not proceed until we are done using the discovered repositories.

Co-authored-by-AI: IBM Bob 2.0.1

For: #7305

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/Project.java
@@ -770,66 +770,72 @@ public class Project extends Processor {
 			if (bsnPattern.length() == 0 || bsnPattern.equals("*"))
 				bsnPattern = null;
 		}
+		String useBsnPattern = bsnPattern;
 
-		SortedMap<String, Pair<Version, RepositoryPlugin>> providerMap = new TreeMap<>();
+		// Hold the workspace read lock for the entire resolution. This prevents
+		// a concurrent workspace refresh (write lock) from closing repository
+		// instances between discovery (getRepositories()) and use.
+		return workspace.readLocked(() -> {
+			SortedMap<String, Pair<Version, RepositoryPlugin>> providerMap = new TreeMap<>();
 
-		List<RepositoryPlugin> plugins = getRepositories();
-		for (RepositoryPlugin plugin : plugins) {
+			List<RepositoryPlugin> plugins = getRepositories();
+			for (RepositoryPlugin plugin : plugins) {
 
-			if (repoFilter != null && !repoFilter.match(plugin))
-				continue;
+				if (repoFilter != null && !repoFilter.match(plugin))
+					continue;
 
-			List<String> bsns = plugin.list(bsnPattern);
-			if (bsns != null)
-				for (String bsn : bsns) {
-					SortedSet<Version> versions = plugin.versions(bsn);
-					if (versions != null && !versions.isEmpty()) {
-						Pair<Version, RepositoryPlugin> currentProvider = providerMap.get(bsn);
+				List<String> bsns = plugin.list(useBsnPattern);
+				if (bsns != null)
+					for (String bsn : bsns) {
+						SortedSet<Version> versions = plugin.versions(bsn);
+						if (versions != null && !versions.isEmpty()) {
+							Pair<Version, RepositoryPlugin> currentProvider = providerMap.get(bsn);
 
-						Version candidate;
-						switch (strategyx) {
-							case HIGHEST :
-								candidate = versions.last();
-								if (currentProvider == null || candidate.compareTo(currentProvider.getFirst()) > 0) {
-									providerMap.put(bsn, new Pair<>(candidate, plugin));
-								}
-								break;
+							Version candidate;
+							switch (strategyx) {
+								case HIGHEST :
+									candidate = versions.last();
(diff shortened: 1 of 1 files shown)


jhanders34:
@chrisrueger and @bjhargrave here is another one to address the issues in #7305

chrisrueger:
I don't have a strong opinion here, since I have not discovered those concurrency issues myself, which you are dealing with - although I do recognize that the issues exist.
  
The thought I had when looking at the change is, whether it is the only place where something like this should be done.

For example `getBundleByHash()`

https://github.com/jhanders34/bnd/blob/38a481f55067389afc600bd7ed5a63899c98bb5a/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L1558

called by 

`getBundle()` in 

https://github.com/jhanders34/bnd/blob/38a481f55067389afc600bd7ed5a63899c98bb5a/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L1302

is also iterating over repos 

https://github.com/jhanders34/bnd/blob/38a481f55067389afc600bd7ed5a63899c98bb5a/biz.aQute.bndlib/src/aQute/bnd/build/Project.java#L1570

and doing stuff. 

So maybe the whole `getBundle()` method should use that lock? Or would that be too coarse? 
Maybe locking should be done at the callers of `Project.getBundle()` which is currently the following places if I do a reference search:

image

But on the other hand I also think that the scope of locking should be as small as possible.
Not sure what the right thing to do here is.

@bjhargrave @pkriens ?

jhanders34:
@chrisrueger You were right to point out the additional places that I missed.  I updated the PR with those places addressed as well.

chrisrueger:
@jhanders34 thanks. One minor subjective / cosmetic thing: Could you maybe try refactoring the code between `workspace.readLocked(() -> {` into its own method right below the method where you used `workspace.readLocked(() -> {`? This would make the diff a bit simpler and easier to reason about, since it keeps the original indentation mostly and the diff shows mainly the wrapping of `readLocked(() -> {})`

for example:

```
return workspace.readLocked(() -> {
    return getBundlesWildcardInternal(range, strategyx, attrs, repoFilter, useBsnPattern, providerMap);
});
```

and 

```
return workspace.readLocked(() -> {
    return getBundleWithReadLockInternal(bsn, range, useStrategy, attrs, repoFilter, useBundleByHash);
});
```

I just tried that locally and the diff looks simpler. But I don't want to mess in your PR. 

image

UPDATE: But I notice that "Hide whitepsace" in Github also makes the PR diff easier to digest. So do as you like. 


Another thing I noticed and just want to point out:  `readLocked` vs `writeLocked`
In your PR description you mention 

> where the repository could be closed as part of refresh

Do you mean a "refresh" of a "Workspace"?

If yes, I just noticed that `Workspace.refresh()` uses a `writeLocked`

https://github.com/bndtools/bnd/blob/e1b5f4fa9fbb15f500d69d2b49b83ecfcd7fbc8f/biz.aQute.bndlib/src/aQute/bnd/build/Workspace.java#L469-L472

Just want to point out as something which should / could be explained in case the current approach is correct.
But since I am not familiar with concurrency handling in bnd I would prefer another opinion.

jhanders34:
> @jhanders34 thanks. One minor subjective / cosmetic thing: Could you maybe try refactoring the code between `workspace.readLocked(() -> {` into its own method right below the method where you used `workspace.readLocked(() -> {`? This would make the diff a bit simpler and easier to reason about, since it keeps the original indentation mostly and the diff shows mainly the wrapping of `readLocked(() -> {})`

You can easily remove all of the white space comparison in the diff with Eclipse or in the github.com comparison using https://github.com/bndtools/bnd/pull/7324/changes?w=1.  The `?w=1` removes the white space comparison. With using that option, it is very easy to see the minimal changes.  That is how I am doing it easily without having to deal with the extra indentation.

> Another thing I noticed and just want to point out: `readLocked` vs `writeLocked` In your PR description you mention
> 
> > where the repository could be closed as part of refresh
> 
> Do you mean a "refresh" of a "Workspace"?

I have updated the PR description to include additional details.  I hope that is what you were looking for.  The comments in the code already referred to workspace refresh.

chrisrueger:
> The `?w=1` removes the white space comparison.

You are absolutely right. I noticed after submitting my response :)

Regarding readLocked and writeLocked : Thanks for the clarification. I got myself some tutoring about `ReentrantReadWriteLock` and it seems your approach is fine - I just didn't know :)

chrisrueger:
Thanks @jhanders34  I'm traveling today and if nobody objects in the meantime , i will merge tomorrow.

chrisrueger:
@jhanders34 ready to merge?

jhanders34:
> @jhanders34 ready to merge?

Yes

chrisrueger:
@jhanders34 thanks for your PR. Looking forward for feedback if it solves all your issues.

PR #7355, opened 2026-08-02
Bugfix pr builds

peterkir:


chrisrueger:
I think this won't fix the problem with forks (PRs where the branch lives outside the bndtools organization). 
Because of `if: ${{ github.repository == 'bndtools/bnd' }}`

Copilot suggests:

**To fix this for external contributor PRs**, the condition should allow runs from forks:

```yaml
build-canonical:
  if: ${{ (github.repository == 'bndtools/bnd') || (github.event_name == 'pull_request') }}
```

Or more explicitly:
```yaml
build-canonical:
  if: ${{ !cancelled() }}  # Always run, no repo check
```

Then keep the **publish** steps restricted to the main repo/branches only.

PR #7356, opened 2026-08-03
[FEA] Display and Navigation of  (merged) properties for includes

peterkir:
- AbstractRequirementListPart: detect when requirements are inherited from an included file and show them greyed out; double-click opens the source file
- AbstractRequirementListPart: add abstract getPrimaryPropertyKey() used to check the correct stem for local vs merged property detection
- BndEditModel: add isLocalProperty, hasLocalMergeProperty, getMergedRequirements and getPropertyProvenance to support the above
- RunRequirementsPart: implement getPrimaryPropertyKey and fall back to getMergedRequirements(-runrequires) when no local -runrequires is present
- RunBlacklistPart: same treatment for -runblacklist
- RepositoryBundleSelectionPart: remove unused ResourceUtils import

### DIFF
--- a/bndtools.m2e/bnd.bnd
+++ b/bndtools.m2e/bnd.bnd
@@ -41,6 +41,8 @@
 	org.eclipse.osgi,\
 	org.codehaus.plexus:plexus-build-api
 
+-testpath: ${junit}
+
 Bundle-SymbolicName: ${p};singleton:=true
 Bundle-ActivationPolicy: lazy
 
--- a/bndtools.m2e/src/bndtools/m2e/MavenBndrunContainer.java
+++ b/bndtools.m2e/src/bndtools/m2e/MavenBndrunContainer.java
@@ -61,8 +61,11 @@ public class MavenBndrunContainer {
 					bundles = new Bundles();
 				}
 
-				useMavenDependencies = maven.getMojoParameterValue(mavenProject, mojoExecution, "useMavenDependencies",
-					Boolean.class, monitor);
+				Boolean configuredUseMavenDependencies = maven.getMojoParameterValue(mavenProject, mojoExecution,
+					"useMavenDependencies", Boolean.class, monitor);
+				if (configuredUseMavenDependencies != null) {
+					useMavenDependencies = configuredUseMavenDependencies;
+				}
 
 				// Comparing OSGi versions makes '4.2.0.SNAPSHOT' > '4.2.0'.
 				// Comparing Maven versions fails because '4.2.0-SNAPSHOT' is <
@@ -70,8 +73,11 @@ public class MavenBndrunContainer {
 				Version mojoVersion = new MavenVersion(mojoExecution.getVersion()).getOSGiVersion();
 
 				if (mojoVersion.compareTo(scopesVersion) >= 0) {
-					includeDependencyManagement = maven.getMojoParameterValue(mavenProject, mojoExecution,
-						"includeDependencyManagement", Boolean.class, monitor);
+					Boolean configuredIncludeDependencyManagement = maven.getMojoParameterValue(mavenProject,
+						mojoExecution, "includeDependencyManagement", Boolean.class, monitor);
+					if (configuredIncludeDependencyManagement != null) {
+						includeDependencyManagement = configuredIncludeDependencyManagement;
+					}
 
 					@SuppressWarnings({
 						"rawtypes", "unchecked"
--- a/bndtools.m2e/src/bndtools/m2e/MavenImplicitProjectRepository.java
+++ b/bndtools.m2e/src/bndtools/m2e/MavenImplicitProjectRepository.java
@@ -9,6 +9,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.concurrent.CompletableFuture;
 
 import org.apache.maven.project.MavenProject;
 import org.eclipse.core.resources.IResourceChangeEvent;
@@ -26,7 +27,6 @@ import org.slf4j.LoggerFactory;
 import aQute.bnd.build.Run;
 import aQute.bnd.exceptions.Exceptions;
 import aQute.bnd.maven.lib.resolve.BndrunContainer;
-import aQute.bnd.osgi.resource.ResourceUtils;
 import aQute.bnd.repository.fileset.FileSetRepository;
 import aQute.bnd.service.Refreshable;
 import aQute.bnd.version.Version;
@@ -37,10 +37,10 @@ public class MavenImplicitProjectRepository extends AbstractMavenRepository
 
 	private static final org.slf4j.Logger	logger	= LoggerFactory.getLogger(MavenImplicitProjectRepository.class);
 
-	private volatile FileSetRepository		fileSetRepository;
-	private final IMavenProjectFacade		projectFacade;
-	private final Run						run;
-	private final IPath						bndrunFilePath;
+	private final RepositoryHolder<FileSetRepository>	repositoryHolder	= new RepositoryHolder<>();
+	private final IMavenProjectFacade					projectFacade;
+	private final Run									run;
+	private final IPath									bndrunFilePath;
 
 	public MavenImplicitProjectRepository(IMavenProjectRegistry mavenProjectRegistry,
 		IMavenProjectFacade projectFacade, Run run) {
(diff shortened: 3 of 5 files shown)


biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
```diff
@@ -1204,6 +1204,43 @@ public void setRunRequires(List<Requirement> requires) {
 		doSetObject(Constants.RUNREQUIRES, oldValue, requires, requirementListFormatter);
 	}
 
+	/** Returns true if the given property key is defined locally, not inherited from an included file. */
+	public boolean isLocalProperty(String key) {
+		return documentProperties.containsKey(key);
+	}
```
chrisrueger: I suggest using the existing `private isLocalPropertyKey()` 

```suggestion
	public boolean isLocalProperty(String key) {
		return isLocalPropertyKey(key);
	}
```

Or we make the private method `public`.

biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
```diff
@@ -1204,6 +1204,43 @@ public void setRunRequires(List<Requirement> requires) {
...
+		if (merged == null || merged.isBlank())
+			return null;
+		return requirementListConverter.convert(merged);
+	}
+
+	/** Returns the provenance (absolute file path) of the given property key, or empty if local or unknown. */
+	public Optional<String> getPropertyProvenance(String key) {
+		Processor p = getOwner();
+		if (p == null)
+			return Optional.empty();
```
chrisrueger: Let's discuss which of the new `public` convenience shortcuts we really need.
I am a bit hesitant since they increase the public API-surface of `BndEditModel` we have to maintain and I am not yet sure if they add enough value (yes they add value, but not sure if it justifies a new public method). 

Maybe we could start with a helper util class in `bndtools.core` to make access more convenient and then see if it works in `BndSourceEffectivePage.java` and the `RunRequirements` lists.

```
// Instead of: model.getMergedRequirements(Constants.RUNREQUIRES)
Processor p = model.getOwner();
String merged = p.mergeProperties(Constants.RUNREQUIRES);
List<Requirement> reqs = requirementListConverter.convert(merged);
```

```
// Instead of: model.getPropertyProvenance(primaryKey)
PropertyKey.findVisible(model.getOwner().getMergePropertyKeys(primaryKey))
    .stream()
    .findFirst()
    .flatMap(PropertyKey::getProvenance);
```


Quick Example of a convenience accessor Util, living in `bndtools.core` alongside the UI code that needs it:

```
package bndtools.editor.project;

import java.util.List;
import java.util.Optional;

import org.osgi.resource.Requirement;

import aQute.bnd.build.model.BndEditModel;
import aQute.bnd.build.model.conversions.RequirementListConverter;
import aQute.bnd.osgi.Processor.PropertyKey;

/**
 * Utility accessor for BndEditModel that provides convenience methods
 * for checking property provenance and merged properties.
 * Used by UI components that need to distinguish between local and inherited properties.
 */
public class BndEditModelAccessor {
    
    private final BndEditModel model;
    private final RequirementListConverter requirementListConverter = new RequirementListConverter();
    
    public BndEditModelAccessor(BndEditModel model) {
        this.model = model;
    }
    
    /** Returns true if the given property key is defined locally, not inherited from an included file. */
    public boolean isLocalProperty(String key) {
        return model.getDocumentProperties().containsKey(key);
    }
    
    /** Returns true if any local property key matches the stem or a stem.* variant. */
    public boolean hasLocalMergeProperty(String stem) {
        if (model.getDocumentProperties().containsKey(stem))
            return true;
        String prefix = stem + ".";
        return model.getDocumentProperties().stringPropertyNames()
            .stream()
            .anyMatch(k -> k.startsWith(prefix));
    }
    
    /** Returns requirements merged across all stem.* variants from the owner processor. */
    public List<Requirement> getMergedRequirements(String stem) {
        Processor p = model.getOwner();
        if (p == null)
            return null;
        String merged = p.mergeProperties(stem);
        if (merged == null || merged.isBlank())
            return null;
        return requirementListConverter.convert(merged);
    }
    
    /** Returns the provenance (absolute file path) of the given property key, or empty if local or unknown. */
    public Optional<String> getPropertyProvenance(String key) {
        Processor p = model.getOwner();
        if (p == null)
            return Optional.empty();
        return PropertyKey.findVisible(p.getMergePropertyKeys(key))
            .stream()
            .findFirst()
            .flatMap(PropertyKey::getProvenance);
    }
}
```

chrisrueger (commented):
Added some comments for our next discussion about BndEditModel.

chrisrueger:
@peterkir  Feedback: Did a quick test with this PR in our project.
Display works great (I can see the greyed out inherited entries and click on it and go to the parent (provenance) file directly 👍 

One issue i discovered:
- I added some an entry to Run Blacklist (via the `+` button) in a shared.bnd (the one I include) 
- entry was added
- but when I save I get this error:

image

```
Save Failed
class java.lang.String cannot be cast to class java.util.List (java.lang.String and java.util.List are in module java.base of loader 'bootstrap')
```

Unfortunatelly a stacktrace does not get logged but I could obtain one via debugger manually:

```
java.lang.ClassCastException: class java.lang.String cannot be cast to class java.util.List (java.lang.String and java.util.List are in module java.base of loader 'bootstrap')
	at aQute.bnd.build.model.BndEditModel.getRunRequires(BndEditModel.java:1199)
	at aQute.bnd.build.model.BndEditModel.setRunRequires(BndEditModel.java:1203)
	at bndtools.editor.project.RunRequirementsPart.doCommitToModel(RunRequirementsPart.java:188)
	at bndtools.editor.project.AbstractRequirementListPart.commitToModel(AbstractRequirementListPart.java:284)
	at bndtools.editor.common.BndEditorPart.commit(BndEditorPart.java:82)
	at org.eclipse.ui.forms.ManagedForm.commit(ManagedForm.java:193)
	at org.eclipse.ui.forms.editor.FormEditor.commitPages(FormEditor.java:299)
	at bndtools.editor.BndEditor.commitDirtyPages(BndEditor.java:275)
	at bndtools.editor.BndEditor.doSave(BndEditor.java:251)
```

The ClassCastException is not that obvious, but it happens here:

image

To me it seems that the caller expects a `List<Requirement>` but `aQute.bnd.build.model.BndEditModel.doGetObject(String, Converter<? extends R, ? super String>, boolean)` returns a plain `String` in the first `if` branch , since it is **not** using the `converter`. 

I could fix it by changing from:
```
if (objectProperties.containsKey(name)) {
				R temp = (R) objectProperties.get(name);
				result = temp;
			}
```

to 

```
if (objectProperties.containsKey(name)) {
	return converter.convert((String) objectProperties.get(name));
}
```

image

But let's not do this as fix. I think the actual bug is in:

`aQute.bnd.build.model.BndEditModel.doSetObject(String, T, T, Converter<String, ? super T>)`
which puts the un-formatted value into the `objectProperties` cache which is the root cause for the ClassCastException. 

image

When I change this to `objectProperties.put(name, v);` the problem with the ClassCastException above is also fix and saving works.

...something like that

UPDATE: I think the problem may have been introduced by 

https://github.com/bndtools/bnd/pull/7356/changes#diff-af324c809dd4117f47e1a35dbb0318547a295dce78737367d08a57d612eee73dR92

```
/** Returns requirements to write to an arbitrary merge key using the standard requirements format. */
	static void setRequirementListByKey(BndEditModel model, String key, List<Requirement> requires) {
		model.setGenericString(key, requirementListFormatter.convert(requires));
	}
```

I think we should **not** use `setGenericString()`. Instead we should use some combination of:

```
BndEditModel.getConverter(Map, String)
```
to get a known converter 
and then find a way to call 

```
BndEditModel.doSetObject(String, T, T, Converter<String, ? super T>)
```

with that converter, in case we don't know exactly with what keys we are dealing with. 
If you do a reference search fo`setGenericString()` you notice that this not called often in existing code (only for `setResolveMode()`)

So I think we should revist the new methods of this PR:
- `bndtools.editor.project.BndEditModelAccessor.setPropertiesByKey(BndEditModel, String, Map<String, String>)`
- `bndtools.editor.project.BndEditModelAccessor.setRequirementListByKey(BndEditModel, String, List<Requirement>)`
- `bndtools.editor.project.BndEditModelAccessor.setVersionedClausesByKey(BndEditModel, String, List<VersionedClause>)`

and the change this PR made to `bndtools.editor.project.RunPropertiesPart.commitToModel(boolean)`

chrisrueger:
Another issue: Scrollbars are missing and scrolling is not working in the RunrequiresPart and BlacklistPart 
(I noticed in the Eclipse debug instance. ).

image

In my current main Eclipse I see scroll bars:

image

PR #7371, opened 2026-08-12
Artifactory deploy of PR and branches from bndtools/bnd repo

peterkir:


maven-plugins/.project
```diff
@@ -14,4 +14,15 @@
 	<natures>
 		<nature>org.eclipse.m2e.core.maven2Nature</nature>
 	</natures>
+	<filteredResources>
+		<filter>
```
chrisrueger: I think those files were changed by accident?

PR #7378, opened 2026-08-20
tester: Clean up launcher after failed preparation

### LINKED ISSUE #7372: Temporary launch properties file is left behind when test preparation fails
### Description

When running openHAB integration tests with `bnd-testing-maven-plugin` 7.3.0, a temporary `launch*.properties` file can be left behind when the test fails while preparing the OSGi runtime, for example because the required bundles cannot be resolved.

We ran into this in openHAB here:

[https://github.com/openhab/openhab-addons/actions/runs/31677377310/job/94374812202](https://github.com/openhab/openhab-addons/actions/runs/31677377310/job/94374812202)

openHAB currently uses bnd 7.3.0:

[https://github.com/openhab/openhab-addons/blob/main/pom.xml](https://github.com/openhab/openhab-addons/blob/main/pom.xml)

After the failed integration test, a generated `launch*.properties` file remains in the test module instead of being cleaned up. Since it is a temporary launcher artifact, it would be preferable for it to be removed on failure as well as on a normal test run.

### Expected behavior

Temporary launcher files are cleaned up even if test preparation or bundle resolution fails before the OSGi framework is launched.

### Actual behavior

If preparation fails, the temporary `launch*.properties` file remains on disk.

### Possible cause

`ProjectLauncherImpl` creates the temporary launch properties file in its constructor and removes it from `cleanup()`.

The testing path appears to have failure paths before the cleanup protected by `ProjectLauncher.launch()` is reached. In particular, `ProjectTesterImpl.test()` calls `prepare()` before calling `getProjectLauncher().launch()`, and `ProjectLauncherImpl.launch()` also calls `prepare()` before delegating to `super.launch()`.

So if `prepare()` throws, for example due to a bundle resolution/preparation problem, `cleanup()` may never be called.

Looking at current `master`, this control flow appears to still be present, so this may not be specific to bnd 7.3.0.

### Related issue

This looks similar to #326, where temporary `launch*.properties` files were left behind. That issue was fixed by calling `cleanup()` for the export path, so this may be an equivalent cleanup gap in the testing/preparation path.

wborn:
- Always clean up the project launcher after a test run attempt.
- Cover preparation exceptions and errors that prevent the launcher from starting.
- Prevent temporary launch*.properties files from being left behind.
- Add regression tests for both preparation failure paths.

Fixes #7372

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/ProjectTest.java
+++ b/biz.aQute.bndlib.tests/test/test/ProjectTest.java
@@ -5,7 +5,16 @@ import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.doAnswer;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


chrisrueger:
Thanks a lot for the PR @wborn 
We will merge it later today. Just waiting for the current build snapshot build.

chrisrueger (approved):
Looks good. Seems to be a minimal change of mainly adding the finally block with the call to cleanup.

wborn:
Yes the fix is smaller than the annoyance. 😉

Maybe I'll also find some time to fix the annoyance of the resolver not always ending bndrun files with a newline. These bndrun files then show up with a 🛑 marker in the changed files view on GitHub.

Issue #6380, opened 2024-11-30
Bnd Run File Editor: “Resolve” button stopped working in Maven workspaces since Bndtools 7.1.0


sad-poet:
Starting with Bndtools 7.1.0, opening a *.bndrun file in a Maven-only workspace and clicking the *Resolve* button results in a NullPointerException (see below).

The problem is not reproduced when using Bndtools 7.0.0.

# Environment

* Bndtools 7.1.0.202411251545 (installed from https://bndtools.org/bndtools.p2.repo/7.1.0/)
  * `Git-Descriptor: 7.0.0.DEV-1660-g47e504d78`
  * `Git-SHA: 47e504d7881ba466703c55a8dca7b0578561582d`
* Eclipse Temurin JDK 21.0.5
* Eclipse IDE for Java Developers 2024-09
* Eclipse Java Development Tools 3.19.600.v20240903-0240
* Eclipse PDE is not installed
* Apache Maven 3.9.9
* bnd-maven-plugin 7.1.0
* Devuan GNU+Linux codename Ceres (corresponding to Debian sid/unstable)

It was also possible to reproduce the error using Eclipse 2024-12 preview version, and on Windows 11.

# Steps to reproduce

1. Install Eclipse IDE (from Eclipse Installer) and latest Bndtools (from Eclipse Marketplace Client), including Bndtools m2e integration feature.
2. Clone the example project and open it in Eclipse IDE: https://github.com/sad-poet/bndtools-resolve-npe-reproducer
4. Open app/app.bndrun file in Bnd Run File Editor.
5. Click Resolve button.

# Expected result

There should be no error. A *Resolution Results* window should open, displaying the required resources used to create the Run Bundles list.

# Actual result

The following error message is displayed in the *Resolution Results* window:

```
Exception during resolution.
java.lang.NullPointerException: Cannot invoke "aQute.bnd.build.Project.get(String)" because "project" is null
	at biz.aQute.resolve.RunResolution.resolve(RunResolution.java:101)
	at org.bndtools.core.resolve.ResolveOperation.lambda$run$0(ResolveOperation.java:62)
	at org.bndtools.core.resolve.ResolveOperation.coordinate(ResolveOperation.java:102)
	at org.bndtools.core.resolve.ResolveOperation.run(ResolveOperation.java:52)
	at org.bndtools.core.resolve.ResolveJob.run(ResolveJob.java:89)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

[...]

PR #7386, opened 2026-08-23
Fix Maven bndrun resolution with current m2e

wborn:
## Summary

Fix Maven-backed `.bndrun` resolution when m2e does not expose optional Maven plugin parameters and when the implicit Maven repository is still being initialized asynchronously.

This has affected the openHAB Eclipse setup since Bndtools 7.1.0. openHAB had to [pin Bndtools to 7.0.x](https://github.com/openhab/openhab-distro/pull/1803) in its [`openHAB2.setup`](https://github.com/openhab/openhab-distro/blob/main/launch/openHAB2.setup) because resolving the Maven-backed [`app.bndrun`](https://github.com/openhab/openhab-distro/blob/main/launch/app/app.bndrun) stopped working with newer versions. This is also related to the Maven workspace resolution problem reported in #6380.

With current m2e versions, `getMojoParameterValue(...)` can return `null` for parameters that rely on Maven plugin defaults. Bndtools assumed these values were always populated. This caused an NPE for `bndruns` and could also result in null values being unboxed for `useMavenDependencies` and `includeDependencyManagement`.

After fixing those parameter handling issues, another problem became visible: in editor mode, Bndtools creates the implicit Maven repository asynchronously, but resolution can start before that initialization has finished. In that case the resolver sees an empty repository and reports bundles as unavailable even though they are present in the Maven dependency graph.

This race is particularly easy to reproduce with openHAB because its [`launch/app` Maven project](https://github.com/openhab/openhab-distro/blob/main/launch/app/pom.xml) has a relatively large Maven dependency graph and resulting implicit repository. During testing, the implicit Maven repository contained 387 bundles, so creating and indexing it takes long enough for resolution to overtake the background initialization more easily than in smaller Maven projects.

## Changes

* Apply the resolver Maven plugin defaults when m2e returns `null` for:

  * `bndruns`
  * `bndrunDir`
  * `useMavenDependencies`
  * `includeDependencyManagement`
* Keep implicit Maven repository initialization asynchronous in editor mode.
* Wait for the initial repository to become available when it is accessed before initialization has completed.
* Continue using refreshed repository instances after Maven/project changes.
* Add regression tests for waiting on initial repository creation and using updated repository instances.

## Testing

Tested with the openHAB [`launch/app` project](https://github.com/openhab/openhab-distro/tree/main/launch/app) using Eclipse 2026-06 and Bndtools 7.4.

Before these changes, opening/resolving [`app.bndrun`](https://github.com/openhab/openhab-distro/blob/main/launch/app/app.bndrun) could first fail with an NPE caused by the missing `bndruns` Maven parameter value. After that was addressed, immediate resolution could still fail with bundles such as `org.eclipse.osgi` and `org.eclipse.equinox.metatype` reported as unavailable.

During diagnosis, the initialized implicit Maven repository contained 387 bundles and included both of those bundles, confirming that the remaining failure was caused by resolution accessing the repository before its asynchronous initialization completed.

With these changes, `app.bndrun` resolves successfully while repository initialization remains asynchronous.

## Backport

Since this fixes Maven `.bndrun` resolution with current Eclipse/m2e versions and restores functionality that has been broken for openHAB since Bndtools 7.1.0, it would be great if this could also be cherry-picked to the 7.4.x branch.

### DIFF
--- a/bndtools.m2e/bnd.bnd
+++ b/bndtools.m2e/bnd.bnd
@@ -41,6 +41,8 @@
 	org.eclipse.osgi,\
 	org.codehaus.plexus:plexus-build-api
 
+-testpath: ${junit}
+
 Bundle-SymbolicName: ${p};singleton:=true
 Bundle-ActivationPolicy: lazy
 
--- a/bndtools.m2e/src/bndtools/m2e/MavenBndrunContainer.java
+++ b/bndtools.m2e/src/bndtools/m2e/MavenBndrunContainer.java
@@ -61,8 +61,11 @@ public class MavenBndrunContainer {
 					bundles = new Bundles();
 				}
 
-				useMavenDependencies = maven.getMojoParameterValue(mavenProject, mojoExecution, "useMavenDependencies",
-					Boolean.class, monitor);
+				Boolean configuredUseMavenDependencies = maven.getMojoParameterValue(mavenProject, mojoExecution,
+					"useMavenDependencies", Boolean.class, monitor);
+				if (configuredUseMavenDependencies != null) {
+					useMavenDependencies = configuredUseMavenDependencies;
+				}
 
 				// Comparing OSGi versions makes '4.2.0.SNAPSHOT' > '4.2.0'.
 				// Comparing Maven versions fails because '4.2.0-SNAPSHOT' is <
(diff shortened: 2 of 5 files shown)


chrisrueger (approved):

chrisrueger:
Thanks @wborn changes look good to me. Since you have a large test code base and I don't have any m2e related things I trust you to check the next the SNAPSHOT build for any problems. Thus I merge now.

PR #7388, opened 2026-08-24
Show inherited entries from -include files in Run editor

peterkir:
The bnd/bndrun editor previously displayed only properties defined in the edited file itself. Entries contributed via -include files were invisible, silently shadowed, or lost on save.

The Run page parts (-runrequires, -runblacklist, -runbundles, -runproperties, -runvm, -runprogramargs) now:

* show inherited entries gray and read-only (merged view via the owner Processor), local entries editable as before
* show per-entry provenance tooltips and open the defining file on double-click; macro-reference clauses (e.g. ${feature.xyz}) stay collapsed and show their one-level definition in the tooltip
* write local additions to the correct merge key: an existing local key is reused, otherwise "<stem>.local" is created when inherited entries exist, so bnd merges instead of shadows
* raise error markers with rename quickfixes for two conflict kinds (marker type bndtools.core.includeconflict): plain merge-stem keys defined in multiple files of the include tree (values shadow each other), and keys defined twice within one file (later definition silently overwrites the earlier one)
* reload the editor model and recompute markers when an included file is saved

API changes (aQute.bnd.build.model 4.5.0 -> 4.6.0, minor bump):

* BndEditModel.getTypedProperty/setTypedProperty added. The UI must read and write suffixed merge keys (e.g. -runrequires.local) with the same conversion as their stem. No existing public API supports this: setGenericString would store a raw String in the typed object cache and cause ClassCastExceptions in the typed getters. The new methods resolve converter/formatter via the key stem (Constants.MERGED_HEADERS) and fall back to strings for unknown keys.
* RUNBLACKLIST registered in the converter/formatter maps so -runblacklist merge keys convert like -runrequires.

No other bndlib API is touched; all remaining logic builds on existing API (Processor.getMergePropertyKeys, PropertyKey provenance, mergeProperties) and lives in the package-private BndEditModelAccessor inside bndtools.core, which exports no affected package. The only new externally visible Eclipse surface is the marker type and its resolution generator in plugin.xml, where a public class is required by the extension registry. New protected hooks (getPrimaryPropertyKey, loadMergedFromModel) are internal; loadMergedFromModel defaults to null so unrelated subclasses keep their previous behavior.

### DIFF
--- a/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
+++ b/biz.aQute.bndlib/src/aQute/bnd/build/model/BndEditModel.java
@@ -307,6 +307,7 @@ public class BndEditModel {
 		// converters.put(BndConstants.RUNVMARGS, stringConverter);
 		converters.put(Constants.TESTCASES, listConverter);
 		converters.put(Constants.RUNREQUIRES, requirementListConverter);
+		converters.put(Constants.RUNBLACKLIST, requirementListConverter);
 		converters.put(Constants.RUNEE, eeConverter);
 		converters.put(Constants.RUNREPOS, listConverter);
 		// converters.put(BndConstants.RESOLVE_MODE, resolveModeConverter);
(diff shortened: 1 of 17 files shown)


synesty-cr (approved):

chrisrueger (approved):
Works great. Tested with our real life project. I can see inherited entries greyed out and double click jumps to the provenance file. Finally no empty Run Requirements anymore -> now it is WYSIWYG
Also tested the Quickfix. 

👍

PR #7389, opened 2026-08-24
Fix for baseliner DiffPluginImpl Require-Capability & Provide-Capability when attribute order is different

### LINKED ISSUE #7377: Baselining - undefined order of Require-Capability parameters
When switching from bnd 7.1.0 to 7.2.3 we recognized that the parameters of the Require-Capability headers are generated differently between our CI pipeline (Linux) and our IDEs (Windows). This was not the case in 7.1.0. The behaviour is still present using 7.4.0.
This might not be an issue to the runtime artifacts, however it forces us to unneccessarily bump the bundle versions after every release when using baseling. Baselining correctly reports a true negative.

Here is an example of the baseline report:

The bundle version (1.0.3/1.0.3) is too low, must be at least 1.0.4
CHANGED              BUNDLE     de.tw.lib.servlet.tokenhandler
  CHANGED            MANIFEST   <manifest>
    REMOVED          HEADER     Require-Capability:osgi.service;effective:=active;filter:="(objectClass=de.tw.lib.servlet.tokenhandler.TokenRequestHandlerRegistry)",osgi.service;effective:=active;filter:="(objectClass=org.osgi.service.log.LoggerFactory)",osgi.extender;filter:="(&(osgi.extender=osgi.component)(version>=1.5.0)(!(version>=2.0.0)))",osgi.extender;filter:="(&(osgi.extender=osgi.configurator)(version>=1.0.0)(!(version>=2.0.0)))",osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=21))"
    ADDED            HEADER     Require-Capability:osgi.service;filter:="(objectClass=de.tw.lib.servlet.tokenhandler.TokenRequestHandlerRegistry)";effective:=active,osgi.service;filter:="(objectClass=org.osgi.service.log.LoggerFactory)";effective:=active,osgi.extender;filter:="(&(osgi.extender=osgi.component)(version>=1.5.0)(!(version>=2.0.0)))",osgi.extender;filter:="(&(osgi.extender=osgi.configurator)(version>=1.0.0)(!(version>=2.0.0)))",osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=21))"

Expected behaviour: the build must be reproducable with same toolchain over CI pipeline and IDE so that baselining does not report changes on an untouched codebase.

chrisrueger:
Closes #7377 

Fixes the Differ Problem when two headers had the same structural content but different string representation.

**What it does**  
Ensures that `REQUIRE_CAPABILITY` and `PROVIDE_CAPABILITY` header is parsed into clauses + parameters and compares the structured tree (order becomes irrelevant).

This fixes the problem reported in the issue where the MANIFEST in the physical baseline .jar file has a different attribute order than the in-memory MANIFEST passed to `DiffPluginImpl`. Their values are equal but their string representation differs. 

The change adds the two headers to `MAJOR_HEADERS` which causes the DiffPlugin to do a Tree-comparison instead only a naive String comparison. And this comparison can show than both headers are equal although their string representation is not.

**Ecosystem impact**
- **low risk of breaking builds.**  
Existing baselines that only differ in attribute order will suddenly start being treated as equal -> fewer false positives, not more.
- People who intentionally relied on the textual order (unlikely?) would no longer see a change -> isn't that a good thing?
- No change to the written manifests
- This is the same treatment already given to `Export-Package`, `Import-Package`, etc., so it is consistent with the rest of the Differ.

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/diff/DiffRequireCapabilityOrderTest.java
+++ b/biz.aQute.bndlib.tests/test/test/diff/DiffRequireCapabilityOrderTest.java
@@ -114,6 +114,85 @@ class DiffRequireCapabilityOrderTest {
                 "After clean()/reorderClause() the written headers must be identical");
     }
 
+	/**
+	 * #7377: the fix must equally cover Provide-Capability
+	 */
+	@Test
+	void provideCapabilityDifferentInsertionOrder() throws Exception {
+		Parameters capA = new Parameters();
+		Attrs a = new Attrs();
+		a.put("objectClass", "com.example.Foo");
+		a.put("uses:", "com.example");
+		capA.put("osgi.service", a);
+
+		Parameters capB = new Parameters();
+		Attrs b = new Attrs();
+		b.put("uses:", "com.example");
+		b.put("objectClass", "com.example.Foo");
+		capB.put("osgi.service", b);
+
+		assertNotEquals(capA.toString(), capB.toString(), "insertion order must produce different strings");
+
+		Diff manifestDiff = diffManifests(Constants.PROVIDE_CAPABILITY, capA.toString(), capB.toString());
+		assertEquals(Delta.UNCHANGED, manifestDiff.getDelta(),
+			"Provide-Capability with different attribute order must be treated as equal");
+	}
+
+	/**
+	 * #7377: repeated capability namespaces (duplicate clause keys, e.g. two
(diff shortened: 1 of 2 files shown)


peterkir:
I could reproduce the bug and this is fixing the issue for me. 
Suggest a slightly different testcase via https://github.com/chrisrueger/bnd/pull/2

PR #7398, opened 2026-08-28
Allow explicit index filenames for P2 repositories

peterkir:
Treat P2 repository locations ending in a slash as directories containing index.xml.gz. Treat other locations as complete index filenames and store downloaded bundles in the parent directory.

Add tests for both location forms and update the documentation.


chrisrueger (approved):
Let's give it a try in next SNAPSHOT and see how it goes.

Issue #6465, opened 2025-02-13
Support a `resolution:=conditional` directive for a more flexible and automated way of handling of package-imports based on the classpath


laeubi:
Currently one can use for [Import-Package](https://bnd.bndtools.org/heads/import_package.html) a `resolution:=dynamic` so an import becomes an `DynamicImport-Package`, beside that one can add `resolution:=optional` to make it an optional import.
And then there is also [conditionalpackage](https://bnd.bndtools.org/instructions/conditionalpackage.html) to embed packages (and not import them at all).

What I'm sometimes missing when wrapping third party code (but it can be useful in other cases as well) is something I will call here `resolution:=conditional` (while maybe `computed` would also be suitable as a name) that can be used like this:

 `Import-Package: *;resolution:=conditional` 

and then behaves in the following way:

1. Assume we have a package to be imported `p`
2. If `p` **is found** on the classpath and the jar **is an OSGi** bundle then `p` should be imported as usual possibly with a version range
3. If `p` **is found** on the classpath but the jar is **not an OSGI** bundle then `p` is considered an conditional package and embedded into the jar
4. If `p` is **not found** on the classpath it is imported with `resolution:=optional` (e.g. might be some `com.sun.*` class from the jdk or whatever e.g. a prebuild jar is analyzed and some dependencies are not available at that time)

This would be especially useful for cases where one needs to wrap an existing jar but can't add metadata in the first place, or if metadata should be added but the authors of that jar do not care much about OSGi and are afraid of having to know/manage that stuff individually.

chrisrueger:
Also I tried googling and found this, which sounds a bit related.

- https://groups.google.com/g/bndtools-users/c/94Wb2uQghf8
- https://github.com/bndtools/bnd/pull/2039

chrisrueger:
@laeubi Let me know if we want to work on that.

laeubi:
I'm currently not plan to work on this as it goes deep into the bnd-internals (and last time I did something similar the contribution experience was very bad ...). Still I think that's an important and valuable feature.

chrisrueger:
Let's talk about this the next days. There won't be anybody else than you and probably me. So the contribution experience will probably be between you and me 😄

chrisrueger:
@laeubi do you want to drive this forward? It's sounds useful and i don't see a reason to reject that. Maybe create a PR from the branch you created and we continue from there.

laeubi:
[...]

Issue #7274, opened 2026-06-17
Add SBOM support to bnd


maho7791:
Refering to [https://github.com/DataInMotion/OSGi_STF/issues/16](https://github.com/DataInMotion/OSGi_STF/issues/16), we want to introduce SBOM support for bnd. I'll start the contribution based on this issue.

We want to start with a simple *bndrun* based sbom generation first. 

This is an application based sbom generation using dependencies from the following instructions:

* -runbundles
* -runpath
* -runfw

PR #7403, opened 2026-09-04
Add resolution:=conditional directive implementation and tests

laeubi:
As requested by @chrisrueger in 

- https://github.com/laeubi/bnd/pull/2#issuecomment-5444678652

See also
- https://github.com/bndtools/bnd/issues/6465

### DIFF
--- a/biz.aQute.bndlib.tests/test/test/AnalyzerTest.java
+++ b/biz.aQute.bndlib.tests/test/test/AnalyzerTest.java
@@ -1578,4 +1578,6 @@ public class AnalyzerTest {
 			a.close();
 		}
 	}
+
+
 }
--- a/biz.aQute.bndlib.tests/test/test/AnalyzerTest.java
+++ b/biz.aQute.bndlib.tests/test/test/AnalyzerTest.java
@@ -1504,4 +1504,78 @@ public class AnalyzerTest {
 			return null;
 		return clauses.get(attr);
 	}
+
+	/**
+	 * Test Import-Packages marked with resolution:=conditional.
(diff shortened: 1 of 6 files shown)


chrisrueger:
Just linking https://github.com/bndtools/bnd/issues/7274 (SBOM topic)
This is also something which could  be considered in SBOM generation, because in the SBOM it is useful to know from which dependency something was embedded. We should think about how to preserve the origin information.

This is not something new this PR created. It just came to mind just now.

peterkir:
The feature itself would be really useful.
Current implementation has a flaw...
Case 2: transient/transitive dependencies

It embeds only files in the directly referred package. embedConditionalPackages copies the package directory to the output bundle and records it as contained, but does not analyze the copied classes or iterate over their referred packages.

So this does not emulate existing Conditional-Package behavior. Conditional-Package uses getExtra() followed by analyzeJar(...), allowing newly embedded classes to introduce further package references and repeat the conditional-package selection. Current PR copies after normal analyzer work is complete, during manifest generation.

Practical consequence: if application code refers to com.vendor.a, and com.vendor.a refers to com.vendor.b:

com.vendor.a is embedded.
com.vendor.b is not automatically embedded.
No import for com.vendor.b is generated from the embedded a classes either.
Runtime can fail with NoClassDefFoundError unless b is separately embedded, imported/provided by framework, or otherwise available.
That is a major mismatch with documentation’s “automatic handling of dependencies” claim.
Could you re-use the existing embedConditionalPackages implementation?

chrisrueger:
Development continues in https://github.com/bndtools/bnd/pull/7407

